PR #44, opened 2023-11-23
feat: signers

DaniPopes:
## Motivation

Add `signer[-*]` crates.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


refcell:
huge

crates/signer-ledger/src/app.rs
```diff
@@ -0,0 +1,347 @@
...
+                one_byte_chain_id - signature.v
+            };
+
+            signature.v = match tx {
+                TypedTransaction::Eip2930(_) | TypedTransaction::Eip1559(_) => {
+                    (ecc_parity % 2 != 1) as u64
+                }
+                TypedTransaction::Legacy(_) => eip155_chain_id + ecc_parity,
+                #[cfg(feature = "optimism")]
+                TypedTransaction::DepositTransaction(_) => 0,
```
tynes, line 161: Deposit transactions never contain signatures since they are side effects of events emitted from L1. It should never hit this branch so perhaps should `panic` or error if a user tries to sign a `DepositTransaction`?

DaniPopes: 1. ported 1-1 from ethers
2. we don't even have this type so it's configured out for now

README.md
```diff
@@ -42,7 +42,7 @@ When updating this, also update:
 
 Alloy will keep a rolling MSRV (minimum supported rust version) policy of **at
 least** 6 months. When increasing the MSRV, the new Rust version must have been
-released at least six months ago. The current MSRV is 1.65.0.
+released at least six months ago. The current MSRV is 1.68.
```
DaniPopes, line 45: aws-sdk-kms 1.68, with msrv at [1.70 in readme](https://github.com/awslabs/aws-sdk-rust#supported-rust-versions-msrv). I think we'll want to bump this to latest stable for AFIT anyway

crates/signer/src/signature.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
+    #[inline]
+    pub fn apply_eip155(&mut self, chain_id: u64) {
+        self.set_v(to_eip155_v(self.recid.to_byte(), chain_id));
+    }
+
+    /// Recovers a [`VerifyingKey`] from this signature and the given message by first hashing the
+    /// message with Keccak-256.
+    #[inline]
+    pub fn recover_address_from_msg<T: AsRef<[u8]>>(
```
prestwich, line 215: doc nit: match the more full docs in `recover_from_msg` below

crates/signer/src/signer.rs
```diff
@@ -0,0 +1,202 @@
...
+            })
+            .await
+    }
+
+    /// Encodes and signs the typed data according to [EIP-712].
+    ///
+    /// [EIP-712]: https://eips.ethereum.org/EIPS/eip-712
+    #[cfg(feature = "eip712")]
+    #[inline]
+    fn sign_typed_data<T: SolStruct + Send + Sync>(
```
prestwich, line 99: should `Send` and `Sync` and `'static` be supertraits of `SolStruct`?  is there any situation in which we allow/want `SolStructs` to have anything other than dumb data in them?

crates/signer/src/wallet/private_key.rs
```diff
@@ -0,0 +1,427 @@
...
+use rand::{CryptoRng, Rng};
+use std::str::FromStr;
+use thiserror::Error;
+
+#[cfg(feature = "keystore")]
+use {elliptic_curve::rand_core, eth_keystore::KeystoreError, std::path::Path};
+
+/// Error thrown by the Wallet module
+#[derive(Debug, Error)]
+pub enum WalletError {
```
prestwich, line 19: should be in its own module file

crates/signer/src/wallet/mod.rs
```diff
@@ -0,0 +1,87 @@
+use crate::{Result, Signature, Signer};
```
prestwich, line 1: Should wallet be a separate crate?

prestwich (changes requested):
nothing major. straightforward port. I think the arch questions outstanding are

- what do we do about txns?
- Wallet
    - does Wallet provide anything meaningful besides address caching? 
    - should the ledger/trezor/aws signers use the Wallet scaffold?
    - should wallet be a separate crate?
- Can we smoothly provide signers/wallets with access to multiple addresses by revisiting the API design?
    - e.g. all of ledger/trezor/aws/mnemonic may have a key inventory instead of just a key

crates/signer/src/wallet/mod.rs
DaniPopes: I thought about this, but I don't believe it makes much sense since it uses the same dependencies

prestwich: yeah, i don't love that `Wallet` is special wrt other implementors of `Signer` 🤔  should all of them be `Wallet<T>` ?

crates/signer/src/signature.rs
```diff
@@ -0,0 +1,326 @@
...
+    Secp256k1,
+};
+use std::str::FromStr;
+
+/// An Ethereum ECDSA signature.
+///
+/// This is a wrapper around [`ecdsa::Signature`] and a [`RecoveryId`] to provide public key
+/// recovery functionality.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[allow(missing_copy_implementations)]
```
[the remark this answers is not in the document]
DaniPopes: No reason, if ecdsa sig is copy I guess this should be as well

crates/signer/src/signer.rs
DaniPopes: I generally dislike slapping "send sync static" combo everywhere

crates/signer/src/wallet/mod.rs
DaniPopes: Don't understand this question, wallet is generic over the underlying private key type basically

crates/signer-ledger/src/app.rs
prestwich: nothing in this workspace should have optimism support yet either. please remove

crates/signer/src/wallet/private_key.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    /// Creates a new random keypair seeded with [`rand::thread_rng()`].
+    #[inline]
+    pub fn random() -> Self {
+        Self::random_with(&mut rand::thread_rng())
+    }
+
+    /// Creates a new random keypair seeded with the provided RNG.
+    #[inline]
+    pub fn random_with<R: Rng + CryptoRng>(rng: &mut R) -> Self {
```
prestwich, line 37: can these be feature-gated? Generally we don't want people to be doing this 😅

DaniPopes: Why not? Rand is brought in from deps anyway, and it's nice in tests

prestwich: Creating a random raw key is essentially never best practice, and shouldn't be in the standard API. We want to recommend users not use this API and feature gating it is a good way to do that

crates/signer/src/signature.rs
```diff
@@ -0,0 +1,325 @@
...
+/// An Ethereum ECDSA signature.
+///
+/// This is a wrapper around [`ecdsa::Signature`] and a [`RecoveryId`] to provide public key
+/// recovery functionality.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Signature {
+    /// The inner ECDSA signature.
+    inner: ecdsa::Signature,
+    /// The recovery ID.
+    recid: RecoveryId,
```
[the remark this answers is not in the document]
prestwich: https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v

pnnkr:
This is amazing, thanks. 

Is there any plan to have a more basic signing in the core crates so we can have no_std support please?

crates/signer/src/signer.rs
```diff
@@ -0,0 +1,214 @@
...
+///
+/// All provided implementations rely on [`sign_hash`](Signer::sign_hash). A signer may not always
+/// be able to implement this method, in which case it should return
+/// [`UnsupportedOperation`](crate::Error::UnsupportedOperation), and implement all the signing
+/// methods directly.
+///
+/// Synchronous signers should implement both this trait and [`SignerSync`].
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait Signer: Send + Sync {
```
mattsse, line 18: Finally object safe Signers

mattsse, line 18: I think we want to impl this for `Box` as well?
but can do seprately

mattsse (approved):

crates/signer/src/signer.rs
DaniPopes: can do `&mut` and `Box` on Signer only due to `set_chain_id`

crates/signer/src/signer.rs
```diff
@@ -0,0 +1,218 @@
...
+    async fn sign_message(&self, message: &[u8]) -> Result<Signature> {
+        self.sign_hash(&eip191_hash_message(message)).await
+    }
+
+    /// Signs the transaction.
+    #[cfg(TODO)] // TODO: TypedTransaction
+    #[inline]
+    async fn sign_transaction(&self, message: &TypedTransaction) -> Result<Signature> {
+        self.sign_hash(&message.sighash()).await
+    }
```
gakonst, line 37: Confirming we need to derive RLP here, and that's the main blocker? https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types/src/eth/transaction/typed.rs#L18

gakonst (approved):
Didn't go through every part of the impls, but signer trait design looks good, and I like that we're using the new AWS sdk. Let's track adding https://github.com/georgewhewell/ethers-gcp-kms-signer upstream in an issue?

If there's no other blocker, let's do the typed tx derive, and get this in! And I think then we need:
1. a signer-aware provider, with some path for general middlewares ( we should discuss design in person)
2. wiring of the provider/middleware with `sol!` and with dyn-abi
3. ..? done?

crates/signer/src/signer.rs
prestwich: no, a transaction request is not a transaction. transaction requests may contain incomplete data, and are never RLP serialized

prestwich: > builder style Request types (Optional fields) that have a corresponding Transaction type (valid Transaction).

yes, love this plan :)

> going from Request -> Transaction is what we called fill in ethers.

This is a little more complex, I think. The `TransactionRequest` generally allows a `from` field, which the `Transaction` that implements RLP should not

So we may want something like:
- `Transaction` - complete, no from field, implements RLP, no serde
- `SignedTransaction` - complete, no from field, contains Transaction + Signature, no serde
- TransactionRequest - builder for the JSON request object
- `Provider::fill(&mut TransactionRequest)` 
    - populates all possible fields from RPC node
    - should ensure that `TransactionRequest::try_into_transaction` definitely succeeds
- `Signer::sign_transaction(&TransactionRequest) -> Result<SignedTransaction>` 
    - errors when the request is missing info, or if `from ` is populated with address signer can't sign from 
- `TransactionResponse` - JSON object returned by `eth_getTransactionByXXXX` methods 
    - should ALWAYS contain a `SignedTransaction`

gakonst: OK. Supportive of this. Let's do in follow-up @DaniPopes so we can merge this in?

PR #45, opened 2023-11-23
feat(`TempProvider`): `raw_request`

Evalir:
## Motivation

We need a way to issue raw requests for the `rpc` cheatcode in foundry.

## Solution

Add a `raw_request` function to `TempProvider` that allows sending arbitrary requests, with arbitrary params, expecting an arbitrary result. Of course, this function isn't object safe though.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/providers/src/provider.rs
+++ b/crates/providers/src/provider.rs
@@ -13,7 +13,7 @@ use alloy_transport::{BoxTransport, Transport, TransportErrorKind, TransportResu
 use alloy_transport_http::Http;
 use auto_impl::auto_impl;
 use reqwest::Client;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


crates/providers/src/provider.rs
```diff
@@ -663,6 +673,26 @@ impl<T: Transport + Clone + Send + Sync> TempProvider for Provider<T> {
...
+        P: Serialize + Send + Sync + Clone,
+        R: Serialize + DeserializeOwned + Send + Sync + Unpin + 'static,
+        Self: Sync
+    {
+        let res = async move {
+            let res: R = self.inner.prepare(method, &params).await?;
+            Ok(res)
+        }.await;
+
+        res
```
DaniPopes, line 693: Why does this need a separate async block?

Evalir: yea no need, was experimenting with bounds for this. wil lremove

DaniPopes (approved):

PR #50, opened 2023-11-29
Impl TryFrom<alloy_rpc_types::Log> for alloy_primitives::Log

leruaa:
Came across the following use case:

- Retrieve logs with `Provider::get_logs()` that returns a `Vec` of `alloy_rpc_types::Log`
- Then decode logs using `SolEvent::decode_log_object()` that accept a `alloy_primitives::Log`

So I needed to convert `alloy_rpc_types::Log` to `alloy_primitives::Log`, hence this PR.

### DIFF
--- a/crates/rpc-types/src/eth/log.rs
+++ b/crates/rpc-types/src/eth/log.rs
@@ -26,6 +26,22 @@ pub struct Log {
     pub removed: bool,
 }
 
+impl TryFrom<Log> for alloy_primitives::Log {
+    type Error = LogError;
+
+    fn try_from(value: Log) -> Result<Self, Self::Error> {
+        alloy_primitives::Log::new(value.topics, value.data).ok_or(LogError::InvalidTopics)
+    }
+}
+
+/// Error that can occur when converting other types to logs
+#[derive(Debug, thiserror::Error)]
+pub enum LogError {
+    /// The topics are invalid
+    #[error("invalid topics")]
+    InvalidTopics,
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/rpc-types/src/eth/log.rs
+++ b/crates/rpc-types/src/eth/log.rs
@@ -30,16 +30,16 @@ impl TryFrom<Log> for alloy_primitives::Log {
     type Error = LogError;
 
     fn try_from(value: Log) -> Result<Self, Self::Error> {
-        alloy_primitives::Log::new(value.topics, value.data).ok_or(LogError::InvalidTopics)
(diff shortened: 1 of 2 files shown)


crates/rpc-types/src/eth/log.rs
```diff
@@ -26,6 +26,22 @@ pub struct Log {
...
+
+    fn try_from(value: Log) -> Result<Self, Self::Error> {
+        alloy_primitives::Log::new(value.topics, value.data).ok_or(LogError::InvalidTopics)
+    }
+}
+
+/// Error that can occur when converting other types to logs
+#[derive(Debug, thiserror::Error)]
+pub enum LogError {
+    /// The topics are invalid
```
Evalir, line 40: I would prefer something a bit more specific, as the invalid case here is having _too many_ topics. Maybe `LogError::TooManyTopics`?

Evalir (changes requested):
looks good! i have a nit:

prestwich:
@Evalir this ties back to the conversation we were having in telegram the other day. Should RPC types instead embed a primitive Log?

Evalir:
@prestwich I think it should—this rpc type is a fuller version of the log object, and the additional APIs from the primitive log add additional safety. However, there's some work to do if we want to embed it here:
- `alloy_primitives::Log` should be `LogData`, as `Log` omits the emitter (as you said on tg)
- We then need to add a `Log` type with the emitter. This can then also remove the `raw_log.rs` type which is essentially the primitive log with the emitter.
- Finally, we can embed `Log` into this rpc type.

leruaa:
@Evalir Updated following your comment.

 Would you like I create an issue for the bigger refactoring you mentioned? I would be happy to work on this.

prestwich:
> * `alloy_primitives::Log` should be `LogData`, as `Log` omits the emitter (as you said on tg)
> * We then need to add a `Log` type with the emitter. This can then also remove the `raw_log.rs` type which is essentially the primitive log with the emitter.
> * Finally, we can embed `Log` into this rpc type.

Is there a good way to make it generic over encoded and decoded logs? E.g. can we achieve something like this

```
struct Log<T = LogData> {
    emitter: Address,
    inner: T,
}

impl Log<LogData> { 
    fn try_decode<T: SolEvent>(self) -> Result<Log<T>, Self>;
} 
impl Log<T> where T: sol_types::SolEvent { 
    fn encode(self) -> Log<LogData>;
}

impl RlpEncodable for Log<LogData> { ... }
impl RlpEncodable for Log<T> where T: sol_types::SolEvent { ... }
```

gakonst (approved):

PR #59, opened 2023-12-02
feature: interprocess-based IPC

prestwich:
Adds interprocess-based IPC pubsub backend.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Evalir (approved):
lgtm—will add at least a basic test for this and see if we can get Ci on multiple OSs. I think we need to stop CI from running wasm support on this crate as well

PR #83, opened 2023-12-15
alloy-consensus crate

prestwich:
## Motivation

Have signature type support the JSON v and yParity formats 

## Solution

Signature keeps vrs AND an ecdsa::Signature in memory.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


Cargo.toml
```diff
@@ -74,3 +75,7 @@ semver = "1.0"
 serial_test = "2.0"
 thiserror = "1.0"
 url = "2.4"
+
+
+[patch.crates-io]
```
prestwich, line 80: TODO: remove this after `Address::from_public_key_bytes`

DaniPopes: Released 0.5.3

prestwich:
ty ty i will update here soon

crates/consensus/src/header.rs
```diff
@@ -0,0 +1,467 @@
...
+        U256::from(self.gas_limit).encode(out);
+        U256::from(self.gas_used).encode(out);
+        self.timestamp.encode(out);
+        self.extra_data.encode(out);
+        self.mix_hash.encode(out);
+        B64::new(self.nonce.to_be_bytes()).encode(out);
+
+        // Encode base fee. Put empty list if base fee is missing,
+        // but withdrawals root is present.
+        if let Some(ref base_fee) = self.base_fee_per_gas {
```
gakonst, line 326: Cc @rjected

crates/consensus/src/pure/eip4844.rs
```diff
@@ -0,0 +1,150 @@
+use crate::constants::eip4844::{
```
gakonst, line 1: Nit I'd maybe group by EIP instead of by constants as it seems nicer to import everything EIP related in one call

crates/consensus/src/receipt/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+/// This enum distinguishes between tagged and untagged legacy receipts, as the
+/// in-protocol merkle tree may commit to EITHER 0-prefixed or raw. Therefore
+/// we must ensure that encoding returns the precise byte-array that was
+/// decoded, preserving the presence or absence of the `TransactionType` flag.
+///
+/// Transaction receipt payloads are specified in their respective EIPs.
+///
+/// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ReceiptEnvelope {
```
gakonst, line 18: Nice

crates/consensus/src/receipt/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        Self::decode_receipt(buf)
+    }
+}
+
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a> arbitrary::Arbitrary<'a> for Receipt {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        let success = bool::arbitrary(u)?;
+        let cumulative_gas_used = u64::arbitrary(u)?;
+        let logs = Vec::<Log>::arbitrary(u)?;
```
gakonst, line 321: This should be bound cc @DaniPopes

crates/consensus/src/transaction/envelope.rs
```diff
@@ -0,0 +1,182 @@
...
+pub enum TxEnvelope {
+    /// An untagged [`TxLegacy`].
+    Legacy(Signed<TxLegacy>),
+    /// A [`TxLegacy`] tagged with type 0.
+    TaggedLegacy(Signed<TxLegacy>),
+    /// A [`TxEip2930`].
+    Eip2930(Signed<TxEip2930>),
+    /// A [`TxEip1559`].
+    Eip1559(Signed<TxEip1559>),
+}
```
gakonst, line 68: Love it

crates/consensus/src/transaction/legacy.rs
```diff
@@ -0,0 +1,329 @@
...
+        let sig = Signature::from_scalars_and_parity(
+            b256!("2a378831cf81d99a3f06a18ae1b6ca366817ab4d88a70053c41d7a8f0368e031"),
+            b256!("450d831a05b6e418724436c05c155e0a1b7b921015d0fbc2f667aed709ac4fb5"),
+            37,
+        )
+        .unwrap();
+
+        let signed_tx = tx.into_signed(sig);
+
+        dbg!(signed_tx.signature_hash());
```
gakonst, line 324: .

crates/network/src/eip2718.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Some other error occurred.
+    #[error(transparent)]
+    Custom(#[from] Box<dyn std::error::Error>),
+}
+
+/// Decoding trait for [EIP-2718] envelopes. These envelopes wrap a transaction
+/// or a receipt with a type flag.
+///
+/// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+pub trait Decodable2718: Sized {
```
gakonst, line 29: V nice

crates/network/src/lib.rs
```diff
@@ -0,0 +1,76 @@
...
+/// A block response
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
+pub struct BlockResponse<N: Network> {
+    #[serde(flatten)]
+    header: N::HeaderResponse,
+    transactions: TransactionList<N::TransactionResponse>,
+}
+
+/// Captures type info for network-specific RPC requests/responses.
+pub trait Network: Sized + Send + Sync + 'static {
```
gakonst, line 52: I forget - can we remove the static?

gakonst (commented):
slowly reading through

crates/consensus/src/header.rs
prestwich: this section is copy+pasted from reth

crates/network/src/lib.rs
prestwich: because we assert the type is zero-sized, adding all the relevant bounds here makes other bounds downstream shorter. there's really no purpose to removing the `'static` bound, as in this context it means "ownership can be held indefinitely", which is trivially true of  any ZST

crates/consensus/src/pure/eip4844.rs
prestwich: do you have a better name for this class of things, btw? the pure functions and constants?

crates/consensus/src/pure/eip4844.rs
```diff
@@ -0,0 +1,150 @@
...
+    BLOB_GASPRICE_UPDATE_FRACTION, BLOB_TX_MIN_BLOB_GASPRICE, TARGET_DATA_GAS_PER_BLOCK,
+};
+
+/// Calculates the `excess_blob_gas` from the parent header's `blob_gas_used` and `excess_blob_gas`.
+///
+/// See also [the EIP-4844 helpers](https://eips.ethereum.org/EIPS/eip-4844#helpers
+/// (`calc_excess_blob_gas`).
+#[inline]
+pub const fn calc_excess_blob_gas(parent_excess_blob_gas: u64, parent_blob_gas_used: u64) -> u64 {
```
prestwich, line 10: also cc @mattsse  I inhoused these 4844 functions which reth imports from revm in `reth/crates/primitives/src/eip4844.rs`. They were using identical constants under different names, so i've changed the names here. mind double-checking my work?

mattsse: this looks fine, but it makes me think that we need a crate for constants?
because maintaining this stuff twice is probably a bad idea?

we could add a lean alloy-eip-constants or something and then upstream to revm?

prestwich: it might make sense to do alloy-eips or something to house the pure logic, condtants, and helpers for each significant eip?

mattsse: sg!
should be very small, just constants+helper functions so we can use it in revm-primitives

crates/signer/src/signer.rs
```diff
@@ -13,17 +13,56 @@ use alloy_sol_types::{Eip712Domain, SolStruct};
...
+                    return Err(crate::Error::TransactionChainIdMismatch {
+                        signer: chain_id,
+                        tx: tx_chain_id,
+                    });
+                }
+            }
+            None => {
+                self.set_chain_id(chain_id);
+            }
+        }
```
gakonst, line 43: can simplify with

```rust
let Some(tx_chain_id)  = self.chain_id()  { else self.set_chain_Id(chain_id); return Ok(()) };
```

and pull the error check outside

crates/signer/src/signer.rs
```diff
@@ -51,8 +99,16 @@ pub trait Signer: Send + Sync {
...
-        self.sign_hash(message.signature_hash()).await
+    async fn sign_transaction(&self, tx: &mut dyn SignableTx) -> Result<Signature> {
+        let chain_id = self.chain_id();
+        if let Some(chain_id) = chain_id {
+            tx.set_chain_id_checked(chain_id)?;
+        }
+        let mut sig = self.sign_hash(tx.signature_hash()).await?;
+        if let Some(chain_id) = chain_id.or_else(|| tx.chain_id()) {
+            sig = sig.with_chain_id(chain_id);
+        }
```
gakonst, line 110: great!

crates/signer/src/wallet/private_key.rs
```diff
@@ -224,97 +225,78 @@ mod tests {
...
+        assert_eq!(sig_none_1, sig_1);
+
+        tx.chain_id = None;
+        let sig_none_2 = sign_tx_test(&mut tx, Some(2)).await.unwrap();
+        assert_eq!(sig_none_2, sig_2);
+
+        // Errors on mismatch.
+        tx.chain_id = Some(2);
+        let error = sign_tx_test(&mut tx, Some(1)).await.unwrap_err();
+        let expected_error = crate::Error::TransactionChainIdMismatch { signer: 1, tx: 2 };
```
gakonst, line 298: great!

crates/signer-aws/src/signer.rs
```diff
@@ -96,20 +96,9 @@ pub enum AwsSignerError {
...
-    #[cfg(TODO)] // TODO: TypedTransaction
-    #[instrument(err)]
-    async fn sign_transaction(&self, tx: &TypedTransaction) -> Result<Signature> {
-        let mut tx_with_chain = tx.clone();
-        let chain_id = tx_with_chain.chain_id().map(|id| id.as_u64()).unwrap_or(self.chain_id);
-        tx_with_chain.set_chain_id(chain_id);
-
-        let sighash = tx_with_chain.sighash();
-        self.sign_digest_with_eip155(sighash, chain_id).await
+        self.sign_digest_inner(hash).await.map_err(alloy_signer::Error::other)
```
gakonst, line 101: did we lose tx signing on aws signer?

crates/signer-trezor/src/signer.rs
```diff
@@ -46,9 +46,9 @@ impl Signer for TrezorSigner {
         self.sign_message_(message).await.map_err(alloy_signer::Error::other)
     }
 
-    #[cfg(TODO)] // TODO: TypedTransaction
     #[inline]
-    async fn sign_transaction(&self, tx: &TypedTransaction) -> Result<Signature> {
+    #[cfg(TODO)]
```
gakonst, line 50: remove?

crates/signer-aws/src/signer.rs
DaniPopes: uses default impl

crates/signer-trezor/src/signer.rs
DaniPopes: work in progress

DaniPopes: See #100 as a possible solution

crates/signer/src/signer.rs
DaniPopes: don't think it makes it any simple

gakonst:
Would love if we can use this to remove a bunch of the Reth storage types. Might mean either manual impls of the Compact trait, or derives on newtypes. Either way would simplify things on Reth side.

crates/eips/src/eip4844.rs
mattsse, line 1: as discussed with @prestwich we want to move a lot of these into a alloy-eip-constants crate or something similar, so revm can easily import those and we don't have them in two places

crates/network/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Special case for uncle response
+    Uncled,
+}
+
+/// A block response
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
+pub struct BlockResponse<N: Network> {
+    #[serde(flatten)]
+    header: N::HeaderResponse,
+    transactions: TransactionList<N::TransactionResponse>,
```
mattsse, line 50: this probably doesn't work for Uncle blocks, so we either need a type for the uncle requests which is just N::HeaderResponse or do some serde magic here, using a dedicated UncleResponse is probably better

mattsse (approved):
all of this looks reasonable to me.

it's a lot of code, most of it comes from reth directly. imo it's fine (for now) that we have them duplicated, but want to upstream this to reth soonish.

didn't review the trait abstractions too closely, because I think those are still highly unstable, but the approach is fine.

we have a lot of room for follow ups, for example, extracting constants

crates/eips/src/eip2718.rs
```diff
@@ -0,0 +1,162 @@
...
+pub trait Encodable2718: Sized + Send + Sync + 'static {
+    /// Return the type flag (if any).
+    ///
+    /// This should return `None` for the default (legacy) variant of the
+    /// envelope.
+    fn type_flag(&self) -> Option<u8>;
+
+    /// True if the envelope is the legacy variant.
+    fn is_legacy(&self) -> bool {
+        self.type_flag().is_none()
```
gakonst, line 105: shouldn't this also be for Some(0)?

crates/eips/src/eip2930.rs
```diff
@@ -0,0 +1,93 @@
...
+pub struct AccessListItem {
+    /// Account addresses that would be loaded at the start of execution
+    pub address: Address,
+    /// Keys of storage that would be loaded at the start of execution
+    // #[cfg_attr(
+    //     any(test, feature = "arbitrary"),
+    //     proptest(
+    //         strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
+    //     )
+    // )]
```
gakonst, line 27: bring back?

crates/network/src/transaction/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+#[doc(hidden)]
+impl<S: 'static> dyn Transaction<Signature = S> {
+    pub fn __downcast_ref<T: std::any::Any>(&self) -> Option<&T> {
+        if std::any::Any::type_id(self) == std::any::TypeId::of::<T>() {
+            unsafe { Some(&*(self as *const _ as *const T)) }
+        } else {
+            None
+        }
+    }
+}
```
gakonst, line 88: FYI @prestwich this was needed for the trezor signer

crates/network/src/transaction/signed.rs
```diff
@@ -0,0 +1,80 @@
...
+        &self.signature
+    }
+
+    /// Get the transaction hash.
+    pub const fn hash(&self) -> B256 {
+        self.hash
+    }
+}
+
+impl<T: Transaction> Signed<T> {
```
gakonst, line 38: prob also want sighash method?

gakonst (approved):
good w me - let's get this in so we can finish the anvil core transition

crates/eips/src/eip2718.rs
DaniPopes: I guess so, added in https://github.com/alloy-rs/alloy/commit/1a240d018b77b958290c47fcb4fc2d6d91c2141b

Cargo.toml
```diff
@@ -37,11 +39,12 @@ alloy-sol-types = { version = "0.5.1", default-features = false, features = ["st
 alloy-rlp = "0.3"
 
 # crypto
-elliptic-curve = { version = "0.13.5", default-features = false, features = ["std"] }
-generic-array = { version = "0.14.7", default-features = false, features = ["std"] }
-k256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std"] }
-sha2 = { version = "0.10.8", default-features = false, features = ["std"] }
-spki = { version = "0.7.2", default-features = false, features = ["std"] }
+c-kzg = "0.4.0" # for eip-4844
```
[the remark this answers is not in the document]
DaniPopes: This is the workspace dependencies, but anyway it's unused

Cargo.toml
```diff
@@ -78,7 +81,13 @@ serde_json = "1.0"
...
 assert_matches = "1.5"
-similar-asserts = "1.5"
 proptest = "1.4"
 proptest-derive = "0.4"
 serial_test = "2.0"
+similar-asserts = "1.5"
+tempfile = "3.8"
+
+[patch.crates-io]
+alloy-primitives = { git = "https://github.com/alloy-rs/core" }
```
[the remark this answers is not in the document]
DaniPopes: No, this will need a 0.6 release

PR #89, opened 2023-12-18
feat: use reth-rpc-types

mattsse:
~~move all of reth-rpc-types to alloy-rpc-types.~~

This moves commonly used types, tracer and eth namespace types from reth.
The transaction types are removed here because they will be added via #83 

This likely needs a bunch of followup PRs but should unblock https://github.com/foundry-rs/foundry/pull/6219

this is still a bit messy and I think we want to divide this up further,
perhaps we even want:
rpc-eth-core-types (this is for things like BlockId BlockNumber) etc.
rpc-eth-types (with common namespaces)
rpc-tracer-types
rpc-common (like serde_helpers)
rpc-engine-types
rpc-relay/beacon-types


crates/rpc-types/Cargo.toml
```diff
@@ -12,32 +12,36 @@ repository.workspace = true
...
 # jsonrpsee
 jsonrpsee-types = { version = "0.20", optional = true }
 
+
 [features]
-default = ["jsonrpsee"]
+default = ["jsonrpsee-types"]
 arbitrary = ["dep:arbitrary", "dep:proptest-derive", "dep:proptest", "alloy-primitives/arbitrary"]
-jsonrpsee = ["dep:jsonrpsee-types"]
+ssz = ["dep:ethereum_ssz" ,"dep:ethereum_ssz_derive", "alloy-primitives/ssz"]
```
DaniPopes, line 38: We should keep the `jsonrpsee` feature instead of using implicit features.

crates/rpc-types/src/eth/fee.rs
```diff
@@ -51,6 +50,6 @@ pub struct FeeHistory {
     pub oldest_block: U256,
     /// An (optional) array of effective priority fee per gas data points from a single
     /// block. All zeroes are returned if the block is empty.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
DaniPopes, line 53: Is this intentional?

DaniPopes (changes requested):

Evalir (approved):

Issue #71, opened 2023-12-12
[Feature] Add GCP KMS signer


DaniPopes:
### Component

providers

### Describe the feature you would like

https://github.com/georgewhewell/ethers-gcp-kms-signer

### Additional context

https://github.com/alloy-rs/alloy/pull/44#pullrequestreview-1776205022

gakonst:
George @georgewhewell said he'll take it, so let's let him do it when he has time, no rush for it

PR #94, opened 2023-12-27
feat(signers): adds alloy-signer-gcp

georgewhewell:
## Motivation
Closes https://github.com/alloy-rs/alloy/issues/71

## Solution

upstreams https://github.com/georgewhewell/ethers-gcp-kms-signer

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


Cargo.toml
```diff
@@ -38,7 +39,7 @@ alloy-rlp = "0.3"
 # crypto
 elliptic-curve = { version = "0.13.5", default-features = false, features = ["std"] }
 generic-array = { version = "0.14.7", default-features = false, features = ["std"] }
-k256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std"] }
+k256 = { version = "0.13.2", default-features = false, features = ["ecdsa", "std", "pem"] }
```
prestwich, line 42: can avoid adding the feature to the workspace by specifying in the relevant crate's `Cargo.toml` as `k256 = { workspace = true, features = ["pem"] }`

crates/signer-gcp/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+[dependencies]
+alloy-primitives.workspace = true
+alloy-signer.workspace = true
+
+async-trait.workspace = true
+gcloud-sdk = { version = "0.23.1", features = ["google-cloud-kms-v1"] }
+k256.workspace = true
+spki.workspace = true
+thiserror.workspace = true
+tonic = { version = "0.10.2", default-features = false }
```
prestwich, line 23: can be re-imported from `gcloud-sdk` instead of explicitly added as a dep?

crates/signer-gcp/src/signer.rs
```diff
@@ -0,0 +1,348 @@
...
+use spki::DecodePublicKey;
+use std::{fmt, fmt::Debug};
+use thiserror::Error;
+use tonic::Request;
+
+type Client = GoogleApi<KeyManagementServiceClient<GoogleAuthMiddleware>>;
+
+/// Reference to a GCP KeyRing.
+#[derive(Clone, Debug)]
+pub struct GcpKeyRingRef {
```
prestwich, line 24: check my understanding: 
- the `GcpKeyRingRef` specifies a keyring by project, location, and name
- it is extended with a `key_id` and a `version` to form a full key specifier
- that key specifier is passed in the requests

we may want to clean it up with something like this? 

```
struct KeySpecifier { keyring: GcpKeyRingRef, key_id: String, version: u64 }

impl KeySpecifier { 
    fn to_key_version_ref(&self) -> String {
        // current behavior of `GcpKeyRingRef::to_key_version_ref`
    }
}
```

crates/signer-gcp/src/signer.rs
```diff
@@ -0,0 +1,348 @@
...
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl Signer for GcpSigner {
+    #[instrument(err)]
+    #[allow(clippy::blocks_in_conditions)]
+    async fn sign_hash(&self, hash: &B256) -> Result<Signature> {
+        self.sign_digest_with_eip155(hash, self.chain_id).await.map_err(alloy_signer::Error::other)
+    }
+
+    #[cfg(TODO)] // TODO: TypedTransaction
+    #[instrument(err)]
```
prestwich, line 153: #83 has the latest work on this. We should merge this PR in advance of #83, and then update this code to match that code

prestwich (changes requested):

crates/signer-gcp/src/signer.rs
georgewhewell: yeah, that matches my understanding- eventually all these parameters are just used to construct a single string so have implemented `KeySpecifier(String)` since the inputs are never used individually

gakonst:
@DaniPopes whenever you have time let's try to get this over the line!

DaniPopes (approved):
Thanks! 
FYI I've edited the description to use [GitHub keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) when referencing issues.

PR #96, opened 2024-01-04
feat: split rpc types into trace types and rpc types

Evalir:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/trace-rpc-types/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "alloy-trace-rpc-types"
```
mattsse, line 2: this should be called alloy-rpc-trace-types

and the folder name rpc-trace-types

PR #98, opened 2024-01-05
fix(`rpc-types`): set Uncle as default for BlockTransactions

Evalir:
## Motivation

When fetching uncle blocks, these blocks never contain a `transactions` field. This means that normal serialization into a `Block` can fail as `BlockTransactions` doesn't have a default and therefore will just not be added to the `Block` struct at all, causing a serde failure.

## Solution

In the case the field is missing, simply include the field as default with an `BlockTransactions::Uncle` instance.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/block.rs
+++ b/crates/rpc-types/src/eth/block.rs
@@ -27,8 +27,10 @@ pub struct Block {
     pub uncles: Vec<B256>,
     /// Block Transactions. In the case of an uncle block, this field is not included in RPC
     /// responses, and when deserialized, it will be set to [BlockTransactions::Uncle].
-    #[serde(skip_serializing_if = "BlockTransactions::is_uncle")]
-    #[serde(default = "BlockTransactions::uncle")]
+    #[serde(
+        skip_serializing_if = "BlockTransactions::is_uncle",
+        default = "BlockTransactions::uncle"
+    )]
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/block.rs
```diff
@@ -25,8 +25,10 @@ pub struct Block {
     pub total_difficulty: Option<U256>,
     /// Uncles' hashes.
     pub uncles: Vec<B256>,
-    /// Transactions.
+    /// Block Transactions. In the case of an uncle block, this field is not included in RPC
+    /// responses, and when deserialized, it will be set to [BlockTransactions::Uncle].
     #[serde(skip_serializing_if = "BlockTransactions::is_uncle")]
+    #[serde(default = "BlockTransactions::uncle")]
```
mattsse, line 31: nit merge those two

mattsse (approved):
smol style nit

crates/rpc-types/src/eth/block.rs
Evalir: done in https://github.com/alloy-rs/alloy/pull/98/commits/0e1eff8a633fa675265cca80e7a56397ad3ee025

PR #100, opened 2024-01-05
feat: implement sign_transaction on Trezor

DaniPopes:
## Motivation

Merge onto #83

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/transaction/mod.rs
+++ b/crates/network/src/transaction/mod.rs
@@ -75,6 +75,18 @@ pub trait Transaction: std::any::Any + Encodable + Send + Sync + 'static {
     fn set_gas_price(&mut self, price: U256);
 }
 
+// TODO: Remove in favor of dyn trait upcasting (1.76+)
+#[doc(hidden)]
+impl<S: 'static> dyn Transaction<Signature = S> {
+    pub fn __downcast_ref<T: std::any::Any>(&self) -> Option<&T> {
+        if std::any::Any::type_id(self) == std::any::TypeId::of::<T>() {
+            unsafe { Some(&*(self as *const _ as *const T)) }
+        } else {
+            None
+        }
+    }
+}
+
 /// Captures getters and setters common across EIP-1559 transactions across all networks
 pub trait Eip1559Transaction: Transaction {
     /// Get `max_priority_fee_per_gas`.
--- a/crates/network/src/transaction/mod.rs
+++ b/crates/network/src/transaction/mod.rs
@@ -1,21 +1,20 @@
-use crate::Receipt;
 use alloy_primitives::{Bytes, ChainId, Signature, B256, U256};
-use alloy_rlp::{BufMut, Decodable, Encodable};
+use alloy_rlp::{BufMut, Encodable};
+
+mod common;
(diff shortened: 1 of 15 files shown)


crates/signer-trezor/src/signer.rs
```diff
@@ -143,59 +156,81 @@ impl TrezorSigner {
...
+        let value = tx.value();
+        let value = u256_to_trezor(value);
+
+        let data = tx.input().to_vec();
+
+        // TODO: Uncomment in 1.76
+        /*
+        let signature = if let Some(tx) = (tx as &dyn std::any::Any).downcast_ref::<TxEip1559>() {
+        */
+        let signature = if let Some(tx) = tx.__downcast_ref::<TxEip1559>() {
```
gakonst, line 193: good with me :+1:

crates/signer/src/signer.rs
```diff
@@ -7,23 +7,20 @@ use auto_impl::auto_impl;
...
+
+/// Extension trait for utilities for signable transactions.
 ///
-/// This trait allows us to generically sign transactions without knowing the
-/// receipt type, and is blanket implemented for transactions which our signer
-/// trait can produce signatures for.
-pub trait SignableTx: Send + Sync + 'static {
+/// This trait is implemented for all types that implement [`Transaction`] with [`Signature`] as the
+/// signature associated type.
+pub trait TransactionExt: Transaction<Signature = Signature> {
```
gakonst, line 17: good w me also

crates/network/src/transaction/mod.rs
```diff
@@ -1,22 +1,21 @@
...
     /// The signature type for this transaction.
     ///
     /// This is usually [`alloy_primitives::Signature`], however, it may be different for future
     /// EIP-2718 transaction types, or in other networks. For example, in Optimism, the deposit
     /// transaction signature is the unit type `()`.
     type Signature;
 
-    /// The receipt type for this transaction.
-    type Receipt: Receipt;
-
```
gakonst, line 19: we'll bring this back per Prestwich thinking right? in that case let's leave this commented out instead of removing

crates/network/src/transaction/mod.rs
```diff
@@ -1,22 +1,21 @@
...
+
+mod common;
+pub use common::TxKind;
 
 mod signed;
 pub use signed::Signed;
 
 /// Represents a minimal EVM transaction.
-pub trait Transaction: Encodable + Decodable + Send + Sync + 'static {
+pub trait Transaction: std::any::Any + Encodable + Send + Sync + 'static {
```
gakonst, line 11: bleh do we need to explicitly declare the Any here? I guess it'll be always there for most types we impl transaction for, would add a note in the code on why it's needed

gakonst (approved):

crates/network/src/transaction/mod.rs
DaniPopes: I don't know honestly, should receipts be tied to a specific transaction? Keeping this will make it impossible to have signers use dyn transaction because know that also has to be specified

PR #104, opened 2024-01-12
fix(providers): receipts of unmined blocks should be null

kallydev:
## Motivation

The provider cannot distinguish between non-existent blocks and blocks without transactions when using the `eth_getBlockReceipts` method.

### RPC responses

- The block exists, but there are no transactions.

```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": []
}
```

 - The block does not exist.

```json
{
  "jsonrpc": "2.0",
  "id": 0,
  "result": null
}
```

## Solution

Use `Option` to distinguish between `[]` and `null`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/providers/src/provider.rs
+++ b/crates/providers/src/provider.rs
@@ -124,7 +124,7 @@ pub trait TempProvider: Send + Sync {
     async fn get_block_receipts(
         &self,
         block: BlockNumberOrTag,
-    ) -> TransportResult<Vec<TransactionReceipt>>;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think this is reasonable and the right fix

pending @Evalir

Evalir (approved):
yep this looks good, agreed this is a good fix.

PR #107, opened 2024-01-12
fix(pubsub): handle subscription response on reconnects (#105)

### LINKED ISSUE #105: [Bug] Pubsub backend reconnect doesn't work
### Component

pubsub

v0.1.0 (commit 50e38e845afb27fc311a74d2260dd087f8e972f3)

### Operating System

Linux

### Describe the bug

I'm using pubsub to subscribe to `neHeads` and `logs` via my Alchemy provider, and a get this error a few times a day:

```
ERROR WS connection error while receiving a response err=WebSocket protocol error: Connection reset without closing handshake
```

immediately followed by these logs:

```
 INFO Reconnecting pubsub service backend.
DEBUG Draining old backend to_handle
DEBUG Reissuing pending requests count=0
DEBUG Re-starting active subscriptions count=2
```

And that's it, I don't receive notifications anymore after that.

After adding more traces I noticed that new subscriptions are sent to the server, but the responses are not handled, because the reconnect logic doesn't add back in-flight request, and it's needed:

https://github.com/alloy-rs/alloy/blob/4db40b051a1f2443b9c18fa0cf62d0ac24d07012/crates/pubsub/src/service.rs#L165-L176

Another issue: Looking at `PubSubService::reconnect()`, I noticed that  [all server ids are dropped](https://github.com/alloy-rs/alloy/blob/4db40b051a1f2443b9c18fa0cf62d0ac24d07012/crates/pubsub/src/service.rs#L93), leading to close the connection between the backend and the frontend.

leruaa:
## Motivation

Closes #105 

## Solution

* Add a in-flight request for each subscription requests to be dispached on reconnect
* Don't drop server ids to be able to update them when handling subscription responses

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/pubsub/src/service.rs
```diff
@@ -89,15 +91,13 @@ where
...
-        self.subs.drop_server_ids();
         // Dispatch all subscription requests
         self.subs
             .iter()
-            .map(|(_, sub)| sub.request().serialized().to_owned())
+            .map(|(_, sub)| sub.request().to_owned())
             .collect::<Vec<_>>()
             .into_iter()
-            .try_for_each(|brv| self.dispatch_request(brv))?;
+            .try_for_each(|req| self.dispatch_subscription(req))?;
```
DaniPopes, line 100: What's the reason for collecting here? cc @prestwich

leruaa: I think that without collecting closure requires unique access to `*self` but it is already borrowed

prestwich: yes, self is borrowed by the `Map<Iter<..>>` and collecting ends that borrow so that the `try_for_each` can get `&mut self`

crates/pubsub/src/managers/in_flight.rs
```diff
@@ -13,26 +13,31 @@ pub(crate) struct InFlight {
     pub(crate) request: SerializedRequest,
 
     /// The channel to send the response on.
-    pub(crate) tx: oneshot::Sender<Result<Response, TransportError>>,
+    pub(crate) tx: Option<oneshot::Sender<Result<Response, TransportError>>>,
```
prestwich, line 16: instead of changing the type here, you can just make a broken sender with `let (tx, _) = oneshot::channel();`

prestwich: or use `Inflight::new` and then drop the receiver/waiter

crates/pubsub/src/service.rs
prestwich: could probably be simplified into a single `try_for_each`

crates/pubsub/src/service.rs
```diff
@@ -89,15 +91,13 @@ where
         // Re-subscribe to all active subscriptions
         tracing::debug!(count = self.subs.len(), "Re-starting active subscriptions");
 
-        // Drop all server IDs. We'll re-insert them as we get responses.
-        self.subs.drop_server_ids();
```
prestwich, line 93: removing this causes memory leak on each reconnect

prestwich (commented):
can you explain the reasoning behind this fix? i.e. why do we believe it addresses the issue?

It looks like the main difference is the Inflight added [here](https://github.com/alloy-rs/alloy/pull/107/files#diff-46c2aec33612b9f33b878dfd28165670efb45a552e06e4bf2fac8f55e51b7a11R111)

crates/pubsub/src/service.rs
prestwich: server IDs are dropped as they are no longer valid, btw. the server will issue new IDs on reconnect and all old IDs are now dead

crates/pubsub/src/service.rs
DaniPopes: If `dispatch_subscription` is inlined it can be because the borrows get split to the individual fields

leruaa:
@prestwich Indeed. It seems that on reconnects, if there are no in-flight request, new subscription responses are not handled, see below:
https://github.com/alloy-rs/alloy/blob/4db40b051a1f2443b9c18fa0cf62d0ac24d07012/crates/pubsub/src/service.rs#L167

crates/pubsub/src/service.rs
leruaa: I remember that new subscription messages were not sent to the frontend if server ids weren't dropped... I'll do more tests

leruaa: Ok, so here's what's happening:

On reconnect server ID are dropped (i.e. `local_to_server` is cleared). When the new subscription response is handled (in [handle_sub_response](https://github.com/alloy-rs/alloy/blob/d68e1103bda0a0b95e1928c56091dc48e64c95de/crates/pubsub/src/service.rs#L179)), we call `upsert` to insert or update the server_id. 

But, as `local_to_server` is cleared, instead of just changing the server ID, we insert a new `ActiveSubscription`, dropping the old one *and* the associated `Sender`, leading to close the corresponding `Receiver` on the frontend side.

leruaa:
@prestwich Yes I just did, see my comment [above ](https://github.com/alloy-rs/alloy/pull/107#discussion_r1466648031)

prestwich:
ah simultaneous comments. seems the check in upsert `self.local_to_server.contains_left(&local_id)` should be `self.local_to_sub.contains(&local_id)` ?

leruaa:
That make sense, I will test.

prestwich (approved):
lgtm

PR #118, opened 2024-01-16
Addition of engine rpc-types from reth

1010adigupta:
## Motivation
This pr aims to add engine related rpc-types to alloy.
closes paradigmxyz/reth#5989
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -62,6 +64,17 @@ impl From<OptimismTransactionFields> for OtherFields {
...
+#[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[repr(C)]
+pub struct BlobTransactionSidecar {
+    /// The blob data.
+    pub blobs: Vec<Blob>,
+    /// The blob commitments.
+    pub commitments: Vec<Bytes48>,
+    /// The blob proofs.
+    pub proofs: Vec<Bytes48>,
+}
```
1010adigupta, line 77: I have added BlobTransactionSidecar here which is the type for sidecar field for EIP-4844 transactions, in TransactionRequest, is this the correct approach or should I do this using other::OtherFields?

mattsse (changes requested):
great draft!

I'd all of this in a new crate rpc-engine-types

mattsse (approved):
nice job!

now we can replace this in reth-rpc-types and simply reexport everything in the engine module there

crates/rpc-engine-types/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+# ethereum
+alloy-rlp = { workspace = true, features = ["arrayvec", "derive"] }
+alloy-primitives = { workspace = true, features = ["rlp", "serde"] }
+alloy-rpc-types.workspace = true
+
+serde = { workspace = true, features = ["derive"] }
+thiserror.workspace = true
+
+# arbitrary
+arbitrary = { version = "1.3", features = ["derive"], optional = true }
```
Evalir, line 24: Do we need arbitrary?

Evalir (approved):
lgtm! thanks! just wanna make sure we're not copy extra unused deps

1010adigupta:
> nice job!
> 
> now we can replace this in reth-rpc-types and simply reexport everything in the engine module there

on it!
should I create a issue in reth for this, and attach a new pr there now?

PR #120, opened 2024-01-18
feat: move reth genesis to alloy-genesis

supernovahs:
closes https://github.com/paradigmxyz/reth/issues/6085


crates/genesis/src/account.rs
mattsse, line 1: I don't think we need this type,

this is the same as 

https://github.com/paradigmxyz/reth/blob/7c9949962b97a6e3cb230e82a44c91ce8d0b1c98/crates/primitives/src/account.rs#L14-L21

right?

crates/genesis/src/genesis.rs
mattsse, line 1: this should be moved to lib.rs

mattsse (changes requested):
great!

a few more nits, then this is good

crates/genesis/src/genesis.rs
```diff
@@ -0,0 +1,1479 @@
...
+impl From<GenesisAccount> for Account {
+    fn from(value: GenesisAccount) -> Self {
+        Account {
+            // nonce must exist, so we default to zero when converting a genesis account
+            nonce: value.nonce.unwrap_or_default(),
+            balance: value.balance,
+            bytecode_hash: value.code.map(keccak256),
+        }
+    }
+}
```
supernovahs, line 189: @mattsse Deleting Account type would mean , we would have to import Account from reth -primitives? That's not an option i think right?

mattsse: we don't need this conversion anymore, instead we can move this to reth-primitives when we replace the genesis type there

mattsse (approved):
nice, as next step we can also replace the genesis types in 

https://github.com/alloy-rs/alloy/blob/main/crates/node-bindings/src/genesis.rs

with this one

PR #130, opened 2024-01-18
feat: helper function to check pending block filter

allnil:
helper function to check if the filter configured to match pending block
originally from: https://github.com/paradigmxyz/reth/pull/6108

### DIFF
--- a/crates/rpc-types/src/eth/filter.rs
+++ b/crates/rpc-types/src/eth/filter.rs
@@ -356,6 +356,14 @@ impl Filter {
         self
     }
 
+    /// Return `true` if filter configured to match pending block
+    pub fn is_pending_block_filter(&self) -> bool {
+        let is_pending_block_filter =
+            self.block_option.get_from_block().map_or(false, BlockNumberOrTag::is_pending)
+                && self.block_option.get_to_block().map_or(false, BlockNumberOrTag::is_pending);
+        is_pending_block_filter
+    }
+
     /// Pins the block hash for the filter
     #[must_use]
     pub fn at_block_hash<T: Into<B256>>(mut self, hash: T) -> Self {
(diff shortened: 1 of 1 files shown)


mattsse:
sweet

will bump this via https://github.com/paradigmxyz/reth/pull/6123

crates/rpc-types/src/eth/filter.rs
```diff
@@ -798,6 +806,13 @@ impl FilteredParams {
         true
     }
 
+    /// Return `true` if the filter configured to match pending block
+    pub fn is_pending_block_filter(&self) -> bool {
+        let is_pending_block_filter =
+            self.filter.as_ref().map_or(false, |f| f.is_pending_block_filter());
+        is_pending_block_filter
```
mattsse, line 813: style: no need for the tmp var

crates/rpc-types/src/eth/filter.rs
```diff
@@ -356,6 +356,14 @@ impl Filter {
         self
     }
 
+    /// Return `true` if filter configured to match pending block
+    pub fn is_pending_block_filter(&self) -> bool {
+        let is_pending_block_filter =
+            self.block_option.get_from_block().map_or(false, BlockNumberOrTag::is_pending)
+                && self.block_option.get_to_block().map_or(false, BlockNumberOrTag::is_pending);
+        is_pending_block_filter
```
mattsse, line 364: same

crates/rpc-types/src/eth/filter.rs
```diff
@@ -356,6 +356,14 @@ impl Filter {
         self
     }
 
+    /// Return `true` if filter configured to match pending block
```
mattsse, line 359: this should mention that this is the case if both from and to block tags are set to pending

crates/rpc-types/src/eth/filter.rs
```diff
@@ -798,6 +806,13 @@ impl FilteredParams {
         true
     }
 
+    /// Return `true` if the filter configured to match pending block
+    pub fn is_pending_block_filter(&self) -> bool {
```
mattsse, line 810: this should reference [Filter::is_pending_block_filter]

mattsse (changes requested):
smol nits

mattsse (approved):

DaniPopes (approved):

PR #133, opened 2024-01-19
fix: use U256 for eth_getStorageAt

DaniPopes:
## Motivation

fix (maybe) https://github.com/foundry-rs/foundry/issues/6855

See links in comments

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/providers/src/provider.rs
+++ b/crates/providers/src/provider.rs
@@ -85,7 +85,7 @@ pub trait TempProvider: Send + Sync {
     async fn get_storage_at(
         &self,
         address: Address,
-        key: StorageKey,
+        key: U256,
         tag: Option<BlockId>,
     ) -> TransportResult<StorageValue>;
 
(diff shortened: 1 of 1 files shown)


crates/providers/src/provider.rs
```diff
@@ -280,12 +275,10 @@ impl<T: Transport + Clone + Send + Sync> TempProvider for Provider<T> {
...
     ) -> TransportResult<StorageValue> {
-        self.inner
-            .prepare(
-                "eth_getStorageAt",
-                (address, key, tag.unwrap_or(BlockNumberOrTag::Latest.into())),
-            )
-            .await
+        // Key is a QUANTITY according to the [spec](https://eth.wiki/json-rpc/API#eth_getstorageat).
+        // See: https://github.com/gakonst/ethers-rs/blob/b60c7fe1fea41a902e3b62dc716d45b326520fd0/ethers-providers/src/rpc/provider.rs#L674
+        let key = U256::from_be_bytes(key.into());
```
DaniPopes, line 280: Should this be changed in the parameter?

mattsse: I think so, the user should do that, the storageKey type is intended for the server so the off spec format is also supported

we should use spec compliance here

mattsse (changes requested):

mattsse (approved):

PR #137, opened 2024-01-19
fix(`consensus`): populate chain id when decoding signed legacy txs

### LINKED ISSUE #125: [Bug] `alloy_consensus`: Incorrect legacy signer being recovered
### Component

signers

latest

### Operating System

macOS (Apple Silicon)

### Describe the bug

Decoding a legacy transaction from its raw RLP seems to not work. 

Using https://etherscan.io/tx/0x280cde7cdefe4b188750e76c888f13bd05ce9a4d7767730feefe8a0e50ca6fc4 as an example, we can access the raw RLP in Etherscan: https://etherscan.io/getRawTx?tx=0x280cde7cdefe4b188750e76c888f13bd05ce9a4d7767730feefe8a0e50ca6fc4 . As we can see, the expected signer is `0xa12e1462d0ced572f396f58b6e2d03894cd7c8a4`. Although, when running a test to recover the signer using alloy, it gives `0x133e78b108f8ffb7b1e07790d60e2babbcc9b3d8`, which is not the expected signer.

Test vector:
```
    #[test]
    #[cfg(feature = "k256")]
    fn decode_legacy_and_recover_signer() {
        use crate::TxLegacy;
        use alloy_network::Signed;
        use alloy_primitives::address;

        let raw_tx = "f9015482078b8505d21dba0083022ef1947a250d5630b4cf539739df2c5dacb4c659f2488d880c46549a521b13d8b8e47ff36ab50000000000000000000000000000000000000000000066ab5a608bd00a23f2fe000000000000000000000000000000000000000000000000000000000000008000000000000000000000000048c04ed5691981c42154c6167398f95e8f38a7ff00000000000000000000000000000000000000000000000000000000632ceac70000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006c6ee5e31d828de241282b9606c8e98ea48526e225a0c9077369501641a92ef7399ff81c21639ed4fd8fc69cb793cfa1dbfab342e10aa0615facb2f1bcf3274a354cfe384a38d0cc008a11c2dd23a69111bc6930ba27a8";

        let tx = <Signed<TxLegacy> as Decodable>::decode(&mut alloy_primitives::hex::decode(raw_tx).unwrap().as_slice()).unwrap();
        let recovered = tx.recover_signer().unwrap();
        let expected = address!("a12e1462d0ceD572f396F58B6E2D03894cD7C8a4");
        assert_eq!(expected, recovered, "Expected same signer");
    }
```

Test vector obtained from https://github.com/gakonst/ethers-rs/issues/1732 — both issues seem to be very similar.

Evalir:
## Motivation

The chain id was not being populated when decoding a signed legacy tx, leading to decoding issues.

Closes #125

## Solution

Populate the chain id param when decoding.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/legacy.rs
+++ b/crates/consensus/src/transaction/legacy.rs
@@ -232,10 +232,9 @@ impl Transaction for TxLegacy {
 
         let v = signature.v();
 
-        match v {
-            Parity::Eip155(_) => tx.chain_id = v.chain_id(),
-            _ => {}
-        };
+        if let Parity::Eip155(_) = v {
+            tx.chain_id = v.chain_id();
+        }
 
         Ok(tx.into_signed(signature))
     }
--- a/crates/consensus/src/transaction/legacy.rs
+++ b/crates/consensus/src/transaction/legacy.rs
@@ -226,7 +226,16 @@ impl Transaction for TxLegacy {
         if !header.list {
             return Err(alloy_rlp::Error::UnexpectedString);
         }
-        let tx = Self::decode_fields(buf)?;
+        let mut tx = Self::decode_fields(buf)?;
+
+        // If the buffer is not empty, it should be an eip-155 encoded chain_id.
+        if !buf.is_empty() {
+            let mut chain_buf = *buf;
+            let chain_id = ChainId::decode(&mut chain_buf)?;
+            // EIP-155: v = {0, 1} + CHAIN_ID * 2 + 35
+            tx.chain_id = ((chain_id - 35) >> 1).into();
+        }
+
         let signature = Signature::decode_rlp_vrs(buf)?;
 
         Ok(tx.into_signed(signature))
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/legacy.rs
```diff
@@ -226,7 +226,16 @@ impl Transaction for TxLegacy {
...
         }
-        let tx = Self::decode_fields(buf)?;
+        let mut tx = Self::decode_fields(buf)?;
+
+        // If the buffer is not empty, it should be an eip-155 encoded chain_id.
+        if !buf.is_empty() {
+            let mut chain_buf = *buf;
+            let chain_id = ChainId::decode(&mut chain_buf)?;
+            // EIP-155: v = {0, 1} + CHAIN_ID * 2 + 35
+            tx.chain_id = ((chain_id - 35) >> 1).into();
```
DaniPopes, line 236: this can overflow

crates/consensus/src/transaction/legacy.rs
```diff
@@ -226,7 +226,16 @@ impl Transaction for TxLegacy {
...
+        }
+
         let signature = Signature::decode_rlp_vrs(buf)?;
```
DaniPopes, line 239: Shouldn't this be decoded from here? <https://github.com/gakonst/ethers-rs/blob/88095ba47eb6a3507f0db1767353b387b27a6e98/ethers-core/src/types/transaction/request.rs#L267>

Evalir: done in https://github.com/alloy-rs/alloy/pull/137/commits/a840f94dc372c4ce57837a260ee565879bdf333e

mattsse (changes requested):
I'm a bit confused why TxLegacy is also Encodeable and Decodeable,

this can be misused easily, because that's most likely never what you want?

re signature and eip155:

https://github.com/paradigmxyz/reth/blob/77f38574e20d2871c7e38cf99fae2fc5cd1cee02/crates/primitives/src/transaction/signature.rs#L98-L98

Evalir:
@mattsse agreed it feels a bit odd to have `TxLegacy` as encodable/decodable, but I think we should not include removing that in this PR as this is a bugfix for the signed decoding case

crates/consensus/src/transaction/legacy.rs
```diff
@@ -226,9 +226,16 @@ impl Transaction for TxLegacy {
...
-        let tx = Self::decode_fields(buf)?;
+        let mut tx = Self::decode_fields(buf)?;
+
         let signature = Signature::decode_rlp_vrs(buf)?;
 
+        let v = signature.v();
+
+        if let Parity::Eip155(_) = v {
+            tx.chain_id = v.chain_id();
+        }
```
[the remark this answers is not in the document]
Evalir: not sure what you mean? this chain id is coming from the signature parity?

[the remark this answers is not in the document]
DaniPopes: https://docs.rs/alloy-primitives/latest/src/alloy_primitives/signature/parity.rs.html#59-70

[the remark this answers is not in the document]
Evalir: https://github.com/alloy-rs/alloy/pull/137/commits/d175c77c02edfec2cc94b4fbd93241039f330c3f

crates/consensus/src/transaction/legacy.rs
```diff
@@ -336,4 +341,26 @@ mod tests {
         assert_eq!(*signed_tx.hash(), hash, "Expected same hash");
         assert_eq!(signed_tx.recover_signer().unwrap(), signer, "Recovering signer should pass.");
     }
+
+    #[test]
+    #[cfg(feature = "k256")]
+    fn decode_legacy_and_recover_signer() {
+        use crate::TxLegacy;
```
mattsse, line 348: this should include the etherscan url

mattsse (approved):
last nit

DaniPopes (approved):

crates/consensus/src/transaction/legacy.rs
Evalir: Will add the alloy issue which contains all the details

PR #148, opened 2024-01-23
fix(`eips`/`consensus`): correctly decode txs on `TxEnvelope`

### LINKED ISSUE #141: [Bug] TxEnvelope rlp decoding returns unexpected length
### Component

consensus, eips, genesis

https://github.com/alloy-rs/alloy#42778ba4

### Operating System

macOS (Apple Silicon)

### Describe the bug

```rust
    let raw_tx = hex::decode("02f86f0102843b9aca0085029e7822d68298f094d9e1459a7a482635700cbc20bbaf52d495ab9c9680841b55ba3ac080a0c199674fcb29f353693dd779c017823b954b3c69dffa3cd6b2a6ff7888798039a028ca912de909e7e6cdef9cdcaf24c54dd8c1032946dfa1d85c206b32a9064fe8").unwrap();
    let res = alloy_consensus::TxEnvelope::decode(&mut raw_tx.as_slice());
```
Returns `alloy_rlp::Error::UnexpectedLength`. Seems it's expecting a non-empty buffer here: https://github.com/alloy-rs/alloy/blob/42778ba44fbf9fdcfd44de81436f65d73e70ede2/crates/eips/src/eip2718.rs#L84
but it's empty at that point, as expected. Should this be checking for `buf.len() != 0` instead?

Evalir:
## Motivation

Closes #141.

TxEnvelope wasn't able to properly decode txs due to a few issues with the eip-2718 `network_decode` impl. Namely, as the issue says:
- The original buffer was getting modified when decoding the header instead of being copied, leading to issues decoding legacy txs.
- The last check was wrong—we should only perform the check if we have a header.

## Solution

Correct the checks.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -64,9 +64,11 @@ pub trait Decodable2718: Sized {
     /// RLP for all current Ethereum transaction types, but may not be in future
     /// versions of the protocol.
     fn network_decode(buf: &mut &[u8]) -> Result<Self, Eip2718Error> {
-        let h_decode = &mut *buf;
-        let h = Header::decode(h_decode)?;
+        // Keep the original buffer around by copying it.
+        let mut h_decode = *buf;
+        let h = Header::decode(&mut h_decode)?;
 
+        // If it's a list, we need to fallback to the legacy decoding.
         if h.list {
             return Self::fallback_decode(buf);
         } else {
@@ -81,7 +83,7 @@ pub trait Decodable2718: Sized {
         let buf = &mut &buf[1..];
         let tx = Self::typed_decode(ty, buf)?;
 
-        if buf.len() != pre_len - h.payload_length {
+        if !buf.is_empty() {
             return Err(alloy_rlp::Error::UnexpectedLength.into());
         }
 
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -3,7 +3,7 @@
 //! [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
 
 use alloy_primitives::{keccak256, Sealed, B256};
-use alloy_rlp::{BufMut, Header};
+use alloy_rlp::{BufMut, Header, EMPTY_STRING_CODE};
 
 // https://eips.ethereum.org/EIPS/eip-2718#transactiontype-only-goes-up-to-0x7f
 const TX_TYPE_BYTE_MAX: u8 = 0x7f;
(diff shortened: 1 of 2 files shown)


crates/eips/src/eip2718.rs
```diff
@@ -81,7 +83,7 @@ pub trait Decodable2718: Sized {
         let buf = &mut &buf[1..];
         let tx = Self::typed_decode(ty, buf)?;
 
-        if buf.len() != pre_len - h.payload_length {
+        if !buf.is_empty() {
```
Rjected, line 86: as far as I understand, this error is meant for when the RLP string has a payload length `n`, but the actual encoding ends up taking some `length != n`, so I'm not sure that this is the right way to fix. It might be worth checking what the `payload_length` actually is in the failing test.

Evalir: Interestingly, i checked the failing test (eip1559) and the payload length is 1—this seems incorrect, no?

Rjected: > Interestingly, i checked the failing test (eip1559) and the payload length is 1—this seems incorrect, no?

The `Header::decode` method is decoding the tx type byte then. IIRC if you try to `Header::decode` a single byte (less than `0x7f`), then `Header::decode` will:
* _NOT_ advance the buffer, and
* return successfully with `Header { payload_length: 1, list: false }`

Rjected: Not saying that `Header::decode` _should_ do this for the record, maybe `Header` should be an enum, or trying to decode a single byte should fail, but this is out of scope w.r.t. bug fixing:
```rust
enum Header {
    Byte{
        value: u8
    },
    ListHeader {
        payload_length: u64,
    },
    StringHeader {
        payload_length: u64,
    },
}
```

Evalir: Ah, that's right! this checks out with how we used to decode typed txs in anvil. I think, in this case, fallibly decoding the tx should be enough and this length check should not be required at all?

Rjected: This is how we do it in reth at least (see comments about quirks):
https://github.com/paradigmxyz/reth/blob/fd683e5a5c38e27aee2b7256f5cd270d17954019/crates/primitives/src/transaction/mod.rs#L1373-L1400

which also contains another check in `decode_enveloped_typed_transaction` for the inner typed tx rlp header payload length

this check should be performed IMO

Evalir: Awesome, sounds good. Corrected in https://github.com/alloy-rs/alloy/pull/148/commits/055a21192931348fc4ae6d658396089a2bff55bf

Rjected (approved):
LGTM!

mattsse:
nice!

PR #149, opened 2024-01-24
feat: alloy-dyn-contract

onbjerg:
## Motivation

We want something like `BaseContract`/`ContractInstance` in ethers-rs for dynamic ABI in alloy.

## Solution

Port over what we can of those existing APIs to alloy (i.e. really only `call`) and simplify the `CallBuilder` a bit

In ethers there are 2 call builders, a main one, `FunctionCall`, and a secondary one you can get from `FunctionCall` (called `RawCall`) which you can use to set state overrides. I inlined the state override API into `CallBuilder` instead, and opted to only add the 3rd parameter for `eth_call` if state overrides are set, as this is not supported by all nodes.

I'm not sure if there is any risk in sending a third "bogus" parameter in `eth_call`, but if so, maybe the API in this PR could be updated to split out the `eth_call` + state override to a different function in the provider.

This also does not include sending transactions (that part is a `todo!()`) because that flow is not really done yet in the provider or the signer.

A separate pr for "static" contracts should be opened

This PR does also not differentiate between legacy/EIP-1559 tx's as we don't have a "typed tx" construct in Alloy yet

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub const fn gas(mut self, gas: U256) -> Self {
+        self.request.gas = Some(gas);
+        self
+    }
+
+    /// Sets the `gas_price` field in the transaction to the provided value
+    /// If the internal transaction is an EIP-1559 one, then it sets both
+    /// `max_fee_per_gas` and `max_priority_fee_per_gas` to the same value
+    pub const fn gas_price(mut self, gas_price: U256) -> Self {
+        // todo: differentiate between eip-1559 and legacy once we have typedtx
```
Evalir, line 62: I'd say let's default to legacy txs while we don't have typedtx—it's what we use in tests IIRC

crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+///
+/// The builder can be `.await`ed directly which is equivalent to invoking [`CallBuilder::call`].
+///
+/// # Note
+///
+/// Sets the [state overrides](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set) for `eth_call`, but this is not supported by all clients.
+pub struct CallBuilder<P> {
+    // todo: this will not work with `send_transaction` and does not differentiate between EIP-1559
+    // and legacy tx
+    request: CallRequest,
```
Evalir, line 23: I think having access to the CallRequest would be ideal. This way we can get around not having `send_transaction` on the provider by transferring the fields manually to a signable/rlp-able tx type from alloy consensus e.g `TxLegacy`

crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+
+impl<P> Clone for CallBuilder<P>
+where
+    P: Clone,
+{
+    fn clone(&self) -> Self {
+        CallBuilder {
+            request: self.request.clone(),
+            function: self.function.clone(),
+            block: self.block,
```
Evalir, line 150: why are we not copying the block? can't we derive this automatically and put a `P: Clone` boundary?

crates/alloy-dyn-contract/src/interface.rs
```diff
@@ -0,0 +1,123 @@
...
+use crate::{
+    error::{Error, Result},
+    instance::ContractInstance,
+};
+use alloy_dyn_abi::{DynSolValue, FunctionExt, JsonAbiExt};
+use alloy_json_abi::{Function, JsonAbi};
+use alloy_primitives::{Address, Selector};
+
+/// A smart contract interface.
+#[derive(Debug, Clone)]
```
Evalir, line 12: Could we add a bit more docs on why it's necessary/more ergonomic to have this type vs extending `JsonAbi`?

crates/alloy-dyn-contract/src/interface.rs
```diff
@@ -0,0 +1,123 @@
...
+    functions: HashMap<Selector, (String, usize)>,
+}
+
+impl From<JsonAbi> for Interface {
+    fn from(abi: JsonAbi) -> Self {
+        Self { abi, functions: Default::default() }
+    }
+}
+
+// TODO: events/errors
```
Evalir, line 24: just flagging :D

crates/providers/src/provider.rs
```diff
@@ -359,8 +368,24 @@ impl<T: Transport + Clone + Send + Sync> TempProvider for Provider<T> {
...
+    /// Not all client implementations support state overrides.
+    async fn call(
+        &self,
+        tx: CallRequest,
+        block: Option<BlockId>,
+        state: Option<StateOverride>,
+    ) -> TransportResult<Bytes> {
+        // todo: if we just set `None` for the 3rd parameter, will this be ok for clients who do not
+        // expect a third parameter? if not, should we split this into `call` and
+        // `call_with_overrides`?
```
Evalir, line 383: I think separating both is okay. Frankly, not many people use call with overrides and having to set none becomes a chore

Evalir (commented):
looks great, some comments

crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Sets the `gas_price` field in the transaction to the provided value
+    /// If the internal transaction is an EIP-1559 one, then it sets both
+    /// `max_fee_per_gas` and `max_priority_fee_per_gas` to the same value
+    pub const fn gas_price(mut self, gas_price: U256) -> Self {
+        // todo: differentiate between eip-1559 and legacy once we have typedtx
+        self.request.gas_price = Some(gas_price);
+        self.request.max_fee_per_gas = Some(gas_price);
+        self.request.max_priority_fee_per_gas = Some(gas_price);
```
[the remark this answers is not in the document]
onbjerg: hmm, I guess eip-1559 for now is okay then

crates/providers/src/provider.rs
```diff
@@ -133,7 +133,16 @@ pub trait TempProvider: Send + Sync {
...
+    ///
+    /// # Note
+    ///
+    /// Not all client implementations support state overrides.
+    async fn call(
+        &self,
+        tx: CallRequest,
+        block: Option<BlockId>,
+        state: Option<StateOverride>,
+    ) -> TransportResult<Bytes>;
```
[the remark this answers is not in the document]
onbjerg: sg to me - was not sure what the best approach was so open to either

crates/alloy-dyn-contract/src/call.rs
onbjerg: i don't want to force it to be `P: Clone`, so `Clone` is only impld for `CallBuilder` if `P: Clone`. i think that's ok

the block is `Copy`, so we don't `.clone()` or clippy will shoot us

crates/alloy-dyn-contract/src/interface.rs
onbjerg: yea we don't have the facilities rn for events so it's more of a follow up to do

Evalir: sweet np, let's open an issue for this and roll

crates/providers/src/provider.rs
gakonst: Separate IMO yep

crates/alloy-dyn-contract/src/instance.rs
```diff
@@ -0,0 +1,108 @@
...
+        this
+    }
+}
+
+impl<P: TempProvider + Clone> ContractInstance<P> {
+    /// Returns a transaction builder for the provided function name.
+    /// If there are  multiple functions with the same name due to overloading, consider using
+    /// the [`ContractInstance::function_from_selector`] method instead, since this will use the
+    /// first match.
+    pub fn function(&self, name: &str, args: &[DynSolValue]) -> Result<CallBuilder<P>> {
```
gakonst, line 65: I would love if we had a great UX about this. Right now it'd require the user to manually do `DynSolValue::X` for each argument. The way we did it with the generics `Tokenize` in Ethers-rs was nice, but understand if it's a PITA to work with.

cc @DaniPopes

crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// If executed on a mutating smart contract function, it will do a "dry run" of the call
+    /// and return the return type of the transaction without mutating the state
+    ///
+    /// # Note
+    ///
+    /// This function _does not_ send a transaction from your account
+    pub async fn call(&self) -> Result<Vec<DynSolValue>> {
+        let bytes = self
+            .provider
+            .call(self.request.clone(), self.block, self.state.clone())
```
gakonst, line 126: @DaniPopes  @mattsse is there a world where we can get rid of these clones and use refs under the hood

crates/alloy-dyn-contract/src/call.rs
gakonst: Let's default to 1559, no?

gakonst (commented):
nice

crates/alloy-dyn-contract/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "alloy-dyn-contract"
```
DaniPopes, line 2: needs description to be the same as readme

crates/alloy-dyn-contract/src/instance.rs
```diff
@@ -0,0 +1,108 @@
...
+    P: Clone,
+{
+    fn clone(&self) -> Self {
+        ContractInstance {
+            address: self.address,
+            provider: self.provider.clone(),
+            interface: self.interface.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
onbjerg: we can't unless we require that `P: Clone` which we talked about last sync that we do not want to enforce

crates/alloy-dyn-contract/src/instance.rs
onbjerg: I tried to convey that at last sync but there did not seem to be a lot of agreement

DaniPopes (changes requested):

crates/alloy-dyn-contract/src/call.rs
```diff
@@ -0,0 +1,187 @@
...
+    Self: 'static,
+    P: TempProvider,
+{
+    type Output = Result<Vec<DynSolValue>>;
+
+    #[cfg(target_arch = "wasm32")]
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output>>>;
+
+    #[cfg(not(target_arch = "wasm32"))]
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send>>;
```
DaniPopes, line 171: Could `impl Future` work here?

crates/alloy-dyn-contract/src/error.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::UnknownFunction(name) => {
+                write!(f, "unknown function: function {name} does not exist",)
+            }
+            Self::UnknownSelector(selector) => {
+                write!(
+                    f,
+                    "unknown function: function with selector {} does not exist",
+                    hex::encode(selector)
```
DaniPopes, line 55:
```suggestion
                    "unknown function: function with selector {selector} does not exist"
```

crates/alloy-dyn-contract/src/call.rs
DaniPopes: Sorry I don't really understand the reasoning here and below of not deriving? It's the exact same as the compiler derived implementation.

crates/alloy-dyn-contract/src/interface.rs
```diff
@@ -0,0 +1,123 @@
...
+        data: &[u8],
+        validate: bool,
+    ) -> Result<Vec<DynSolValue>> {
+        self.get_from_selector(selector)?.abi_decode_output(data, validate).map_err(Into::into)
+    }
+
+    /// Returns a reference to the contract's ABI
+    pub const fn abi(&self) -> &JsonAbi {
+        &self.abi
+    }
```
DaniPopes, line 103: + `fn into_abi(self)`

DaniPopes (changes requested):

crates/alloy-dyn-contract/src/call.rs
onbjerg: I see it works now, earlier I tried to derive w/o requiring `P: Clone` but it complained

crates/alloy-dyn-contract/src/call.rs
onbjerg: that feature is still unstable unfortunately

crates/dyn-contract/src/call.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Sets the [state overrides](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set) for `eth_call`, but this is not supported by all clients.
+#[derive(Clone)]
+pub struct CallBuilder<P> {
+    // todo: this will not work with `send_transaction` and does not differentiate between EIP-1559
+    // and legacy tx
+    request: CallRequest,
+    block: Option<BlockId>,
+    state: Option<StateOverride>,
+    provider: P,
+    // todo: only used to decode - should it be some type D to dedupe with `sol!` contracts?
```
DaniPopes, line 28: Will need a follow-up for this so we can also use it in sol!

DaniPopes (approved):
lgtm

onbjerg:
will merge as is and we can make improvements on top, we're prob bound to have missed something from the old api that someone has a use case for

PR #150, opened 2024-01-24
fix(`consensus`): ensure into_signed forces correct format for eip1559/2930 txs

### LINKED ISSUE #144: [Bug] Signers apply EIP-155 even for EIP-{1559,2718} transactions
### Component

signers

_No response_

### Operating System

None

### Describe the bug

Signers::sign_transaction and SignerSync::sign_transaction_sync should not apply EIP-155 when the signer's chain_id is None. Indeed, this is the expected behavior according to [Signer's docstring](https://github.com/alloy-rs/alloy/blob/main/crates/signer/src/signer.rs#L59), but EIP-155 is applied anyway if the chain ID is found on the transaction instead.

Ref: https://github.com/alloy-rs/alloy/blob/2ee7b8b2821294f0661b2588e1fd7fcd2640adc0/crates/signer/src/signer.rs#L87

EIP-{1559,2718} always embed a chain ID value but they're not compatible with EIP-155. Thus, they get an EIP-155 signature when they shouldn't.

Evalir:
## Motivation

Closes #144. 

## Solution

This ensures that, at the time of combining the signature with the tx, the correct format is applied for 2930/1559 txs, removing any eip-155 application.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -231,7 +231,6 @@ impl Decodable for TxEip1559 {
 
 impl Transaction for TxEip1559 {
     type Signature = Signature;
-    // type Receipt = ReceiptWithBloom;
 
     fn into_signed(self, signature: Signature) -> Signed<Self> {
         let mut buf = vec![];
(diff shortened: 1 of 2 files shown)


prestwich (approved):
lgtm

PR #151, opened 2024-01-24
chore(`genesis`): do not make genesis fill out its fields by default

Evalir:
## Motivation

#120 ported over the genesis utilities from ethers, but also made it `#[serde(default)]`. I don't think this is desirable as this can lead to footguns—the provider of the genesis file must ensure it's filled out correctly.

## Solution

Revert to the way we used to mark the fields as default.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -257,7 +257,7 @@ impl GenesisAccount {
 /// struct](https://github.com/ethereum/go-ethereum/blob/64dccf7aa411c5c7cd36090c3d9b9892945ae813/params/config.go#L349)
 /// for the source of each field.
 #[derive(Clone, Debug, Deserialize, Serialize, Default, PartialEq, Eq)]
-#[serde(default, rename_all = "camelCase")]
+#[serde(rename_all = "camelCase")]
 pub struct ChainConfig {
     /// The network's chain ID.
     #[serde(default = "mainnet_id")]
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -28,7 +28,7 @@ use std::collections::HashMap;
 
 /// The genesis block specification.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "camelCase", default)]
+#[serde(rename_all = "camelCase")]
 pub struct Genesis {
     /// The fork configuration for this network.
     #[serde(default)]
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -34,22 +34,25 @@ pub struct Genesis {
     #[serde(default)]
     pub config: ChainConfig,
     /// The genesis header nonce.
-    #[serde(with = "u64_hex_or_decimal")]
+    #[serde(default, with = "u64_hex_or_decimal")]
     pub nonce: u64,
     /// The genesis header timestamp.
-    #[serde(with = "u64_hex_or_decimal")]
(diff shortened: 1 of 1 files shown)


mattsse (commented):
> I don't think this is desirable as this can lead to footguns—the provider of the genesis file must ensure it's filled out correctly.

sorry can you clarify? looks like this PR does the opposite

mattsse:
wdyt @Rjected

crates/genesis/src/lib.rs
```diff
@@ -60,13 +63,13 @@ pub struct Genesis {
...
-    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt")]
+    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt", default)]
     pub base_fee_per_gas: Option<u64>,
     /// The genesis header excess blob gas
-    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt")]
+    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt", default)]
     pub excess_blob_gas: Option<u64>,
     /// The genesis header blob gas used
-    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt")]
+    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt", default)]
```
mattsse, line 72: these are redundant, no?

mattsse (changes requested):

Evalir:
So, before the genesis utility on ethers didn't mark the `Genesis` struct as default, but rather _most_ fields it had. When this was ported over to alloy, the whole `Genesis` struct was marked as default which actually caused the way this deserialized to change.

mattsse:
tbh idk if we want this, defer to @Rjected here, but sounds more like a loadAllocs issue

Evalir:
it's aight if we want to fix it directly on load allocs, but it does feel weird to me to let `Genesis` be instantiated by default as we usually read it from a file.

crates/genesis/src/lib.rs
mattsse: doesn't sound right

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=768fa5e9e41d6a85af59b97b824077fb

crates/genesis/src/lib.rs
```diff
@@ -28,28 +28,31 @@ use std::collections::HashMap;
 
 /// The genesis block specification.
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "camelCase", default)]
+#[serde(rename_all = "camelCase")]
```
mattsse, line 31: I only notice this now,

I really don't understand why it is necessary to remove the struct's default and put it on every field

Evalir:
Let's fix on `loadAllocs` directly

mattsse:
what needs fixing exactly?
that's what's super unclear here to me

Evalir:
If you diff ethers-rs's `Genesis` vs alloy's `Genesis`, you'll see that alloy's genesis is marked as `#[serde(default)]`. This causes then deserialization to not fail even if fields are missing, e.g `alloc`. This is what causes the `loadAllocs` cheatcode to fail, as it always thinks it successfully read a `Genesis` file even when it might've just default-deserialized it.

The fix proposed here was to revert back to how ethers-rs was deserializing the `Genesis` struct. However, we can also just not change this, and fix over on the cheatcode.

Issue #123, opened 2024-01-18
[Feature] Move alloy-rpc-types serde_helpers mod to standalone crate


mattsse:
### Component

rpc-types

### Describe the feature you would like

having the commonly used helpers separately is useful.

for example

alloy-genesis needs them but doesn't need any of the rpc types itself

PR #155, opened 2024-01-25
add serde helpers crates 

DoTheBestToGetTheBest:
related : https://github.com/alloy-rs/alloy/issues/123

### DIFF
--- a/crates/rpc-types/Cargo.toml
+++ b/crates/rpc-types/Cargo.toml
@@ -32,13 +32,27 @@ proptest-derive = { version = "0.4", optional = true }
 # jsonrpsee
 jsonrpsee-types = { version = "0.20", optional = true }
 
+#serde_helper
+alloy-serde.workspace = true
+
+
 [features]
-arbitrary = ["dep:arbitrary", "dep:proptest-derive", "dep:proptest", "alloy-primitives/arbitrary"]
+arbitrary = [
+    "dep:arbitrary",
+    "dep:proptest-derive",
+    "dep:proptest",
+    "alloy-primitives/arbitrary",
+]
 jsonrpsee-types = ["dep:jsonrpsee-types"]
-ssz = ["dep:ethereum_ssz" ,"dep:ethereum_ssz_derive", "alloy-primitives/ssz"]
+ssz = ["dep:ethereum_ssz", "dep:ethereum_ssz_derive", "alloy-primitives/ssz"]
 
 [dev-dependencies]
-alloy-primitives = { workspace = true, features = ["rand", "rlp", "serde", "arbitrary"] }
+alloy-primitives = { workspace = true, features = [
+    "rand",
+    "rlp",
+    "serde",
+    "arbitrary",
+] }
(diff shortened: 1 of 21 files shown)


crates/serders/src/lib.rs
mattsse, line 1: this needs the same header as other crates

mattsse (changes requested):
good start, 
we now want this in all the other rpc crates so we can get rid of the serde_helpers.mod file

DoTheBestToGetTheBest:
> we now want this in all the other rpc crates so we can get rid of the serde_helpers.mod file

do you agree if this can be in separate pr ?

mattsse:
hey @DoTheBestToGetTheBest sorry for the late response

we want this as separate crate yes and then replace the current serde_helpers file in the rpc types crate

currently this duplicates the stuff, let's do it in one go

DoTheBestToGetTheBest:
> hey @DoTheBestToGetTheBest sorry for the late response


no problem (:
> we want this as separate crate yes and then replace the current serde_helpers file in the rpc types crate
> 
> currently this duplicates the stuff, let's do it in one go

okey will finish this

DoTheBestToGetTheBest:
@mattsse does it look great to you ? i don't know how can i run test here

DoTheBestToGetTheBest:
hey boss @mattsse does this looks great to you?

DoTheBestToGetTheBest:
@mattsse hey can you take look here pls

crates/rpc-types/Cargo.toml
```diff
@@ -32,13 +32,27 @@ proptest-derive = { version = "0.4", optional = true }
 # jsonrpsee
 jsonrpsee-types = { version = "0.20", optional = true }
 
+#serde_helper
+alloy-serde.workspace = true
```
mattsse, line 36: this should be moved to other alloy deps

crates/serders/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+//! Serde helpers for primitive types.
```
mattsse, line 1: this needs the header, as used in other crates

crates/rpc-types/src/lib.rs
```diff
@@ -20,7 +20,5 @@
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
 mod eth;
-pub mod serde_helpers;
```
mattsse, line 23: we want to keep the reexport for convenience, for now

mattsse (changes requested):

crates/rpc-types/src/lib.rs
DoTheBestToGetTheBest: > we want to keep the reexport for convenience, for now

sorry but i don't understand really what you means here

mattsse: `pub use alloy_serde as serde_helpers`

this way we don't break anything

PR #171, opened 2024-02-01
feat(`alloy-genesis`): pk support

Evalir:
## Motivation

missing pk support.

## Solution

Adds missing private key support to alloy genesis, see https://github.com/foundry-rs/foundry/pull/6970#pullrequestreview-1856796777.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/genesis/Cargo.toml
+++ b/crates/genesis/Cargo.toml
@@ -14,6 +14,8 @@ exclude.workspace = true
 [dependencies]
 alloy-primitives.workspace = true
 alloy-rpc-types.workspace = true
+alloy-signer.workspace = true
+k256.workspace = true
 
 # serde
 serde.workspace = true
--- a/crates/genesis/Cargo.toml
+++ b/crates/genesis/Cargo.toml
@@ -14,8 +14,6 @@ exclude.workspace = true
 [dependencies]
 alloy-primitives.workspace = true
 alloy-rpc-types.workspace = true
-alloy-signer.workspace = true
-k256.workspace = true
 
 # serde
 serde.workspace = true
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -27,7 +27,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 /// The genesis block specification.
-#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase", default)]
 pub struct Genesis {
     /// The fork configuration for this network.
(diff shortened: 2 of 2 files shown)


crates/genesis/Cargo.toml
```diff
@@ -14,6 +14,8 @@ exclude.workspace = true
 [dependencies]
 alloy-primitives.workspace = true
 alloy-rpc-types.workspace = true
+alloy-signer.workspace = true
+k256.workspace = true
```
DaniPopes, line 18: we should re-export this from alloy-signer actually

crates/genesis/src/lib.rs
```diff
@@ -217,6 +224,55 @@ pub struct GenesisAccount {
...
     )]
     pub storage: Option<HashMap<B256, B256>>,
+    /// The account's private key.
+    #[serde(
+        rename = "secretKey",
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "secret_key"
+    )]
+    pub private_key: Option<LocalWallet>,
```
DaniPopes, line 234: Can we just do B256?

mattsse, line 234: this should mention that this is only for testing

DaniPopes (changes requested):

crates/genesis/src/lib.rs
```diff
@@ -217,6 +224,55 @@ pub struct GenesisAccount {
...
+}
+
+/// serde support for `secretKey` in genesis
+pub mod secret_key {
+    use alloy_primitives::Bytes;
+    use alloy_signer::LocalWallet;
+    use k256::{ecdsa::SigningKey, SecretKey};
```
mattsse, line 241: this should not be required here,
the LocalWallet type should have all the necessary functions, if not we should add

mattsse (changes requested):

crates/genesis/src/lib.rs
Evalir: yeah would prefer that actually, much easier to simplify. wdyt @mattsse

mattsse: @Evalir I'd actually prefer just B256 as well.

point in my review still stands, we should have functions on LocalWallet that take care of that

crates/genesis/Cargo.toml
Evalir: k256 u mean right?

crates/genesis/src/lib.rs
DaniPopes: Not a great idea to serialize pk by default as raw bytes

mattsse (approved):
lgtm, lets add an issue for the b256 -> LocalWallet and LocalWallet -> B256 calls?

Issue #172, opened 2024-02-01
[Feature] Add `B256` to `LocalWallet` conversions for `alloy-signer`


Evalir:
### Component

signers

### Describe the feature you would like

It'd be nice to have direct conversions from `B256` to `LocalWallet`.

PR #173, opened 2024-02-01
feat(signer): re-export k256, add `Wallet::from_bytes(B256)`

DaniPopes:
Closes https://github.com/alloy-rs/alloy/issues/172


Evalir (approved):

mattsse (approved):
can we get -> B256 as well?

PR #175, opened 2024-02-01
feat: subscription type

prestwich:
Adds the following types:
- `RawSubscription`
- `SubscriptionItem<T>`
- `Subscription<T>`

## Motivation

Lay groundwork for RAII subscriptions and subscription configuration.

## Solution

Add `Subscription` type around the raw `tokio::broadcast::Receiver`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/pubsub/src/sub.rs
+++ b/crates/pubsub/src/sub.rs
@@ -143,8 +143,8 @@ impl<T> Subscription<T> {
         self.inner.len()
     }
 
-    /// Wrapper for [`resubscribe`]. Create a new [`SubInner`], starting from
-    /// the current tail element.
+    /// Wrapper for [`resubscribe`]. Create a new [`RawSubscription`], starting
+    /// from the current tail element.
     ///
     /// [`resubscribe`]: broadcast::Receiver::resubscribe
     pub fn resubscribe_inner(&self) -> RawSubscription {
@@ -197,6 +197,8 @@ impl<T: DeserializeOwned> Subscription<T> {
 
     /// Wrapper for [`blocking_recv`]. Block the current thread until a message
     /// of the expected type is available.
+    ///
+    /// [`blocking_recv`]: broadcast::Receiver::blocking_recv
     pub fn blocking_recv(&mut self) -> Result<T, broadcast::error::RecvError> {
         loop {
             match self.blocking_recv_any()? {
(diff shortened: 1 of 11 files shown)


crates/pubsub/src/sub.rs
```diff
@@ -0,0 +1,236 @@
...
+impl<T> Subscription<T> {
+    /// Get the local ID of the subscription.
+    pub const fn local_id(&self) -> B256 {
+        self.inner.local_id()
+    }
+
+    /// Get a reference to the inner subscription.
+    pub const fn inner(&self) -> &RawSubscription {
+        &self.inner
+    }
```
mattsse, line 124: add into_inner?

crates/pubsub/src/sub.rs
```diff
@@ -0,0 +1,236 @@
...
+
+impl<T: DeserializeOwned> From<Box<RawValue>> for SubscriptionItem<T> {
+    fn from(value: Box<RawValue>) -> Self {
+        if let Ok(item) = serde_json::from_str(value.get()) {
+            SubscriptionItem::Item(item)
+        } else {
+            SubscriptionItem::Other(value)
+        }
+    }
+}
```
mattsse, line 95: this will silently drop deserialization errors in the Subscription

can we add `recv_item` to RawSubscription ?

prestwich: > this will silently drop deserialization errors in the Subscription

this is intentional, as a deser error implies the subscription is of an unexpected type

prestwich: > can we add recv_item to RawSubscription ?

what's the expected behavior?  something like `async fn recv_item<T>(&mut self) -> Result<SubscriptionItem<T>, RecvError>` ?

mattsse: yeah,

sometimes the json types are not quite right, so it'd be beneficial to also receive failed deserde items.

but technically this would then behave like a result, so maybe we can add a subscription type that returns results
not sure what the best ux here is

but for debugging, I think I'd want to quickly swap out the `subscribe` call to a different function that returns `Result<T, >` or subscription

prestwich:
> separately, sooner or later we want a stream type

yeah, I was considering implementing these as streams, but there's an upcoming conflict between Unpin/projection and Drop impl for RAII guards I think?

prestwich:
basically need to figure out https://docs.rs/pin-project/latest/pin_project/attr.pinned_drop.html

crates/pubsub/src/sub.rs
prestwich: how about this api

`recv` - get next item, returning `T`, discarding all errors
`recv_result` - get next item, returning `Result<T. serde_json::Error>`
`recv_any` - get next item, returning `SubscriptionItem<T>`

mattsse (approved):
all of these make sense to me.

having a new type is better than dealing with receiver, so we can add functions as we like

PR #177, opened 2024-02-01
Allow connector to configure subscription buffer size

### LINKED ISSUE #170: [Feature] Subscription broadcast buffer size is fixed
### Component

providers, pubsub

The buffer size for subscriptions is currently a constant.

https://github.com/alloy-rs/alloy/blob/fcbf995278007c19b75e40d4fda7fd477e2d1598/crates/pubsub/src/managers/active_sub.rs#L56-L60

This should be configurable.

we should add settings for the 

https://github.com/alloy-rs/alloy/blob/fcbf995278007c19b75e40d4fda7fd477e2d1598/crates/pubsub/src/service.rs#L20-L20

so you can call connect(connector, settings)

wdyt @prestwich 

### Additional context

_No response_

prestwich:
## Motivation

Closes #170 

## Solution

Add a fn to the `PubSubConnect` trait that allows the connector to override the default subscription channel size

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/pubsub/src/connect.rs
```diff
@@ -9,6 +9,16 @@ pub trait PubSubConnect: Sized + Send + Sync + 'static {
...
 
+    /// Select the buffer size for subscription channels. This value will
+    /// be used to determine the capacity of the broadcast channels used to send
+    /// notifications to subscribers. See [`broadcast::channel`] for more
+    /// information about channel behavior.
+    ///
+    /// [`broadcast::channel`]: tokio::sync::broadcast::channel
+    fn subscription_buffer_size(&self) -> usize {
+        16
+    }
```
mattsse, line 20: is only used by the `PubSubService`?

would it be better to move this to the PubSubFrontEnd? so it can be modified for future subscriptions?

for example:
you notice a lot of lagging responses and want to reconnect with larger buffer

prestwich: right now pubsub frontend and the service communicate only over the channel

we could do one of the following easily
- have pubsub frontend pass updates over the channel to the service, instructing it to change its conf value
- introduce shared memory between all frontends and the service

we can do the following messily but effectively:
- attach the conf value to each request, and have the backend store that conf value on a per-request basis

thoughts on which you like?

mattsse: per request would be best imo

prestwich: would you prefer that this is a single configuration value shared by all front ends and shared memory or a per front end configuration value?

crates/json-rpc/src/request.rs
```diff
@@ -10,6 +10,21 @@ pub struct RequestMeta {
...
     /// The request ID.
     pub id: Id,
+    /// Whether the request is a subscription, other than `eth_subscribe`.
+    pub is_non_standard_sub: bool,
+}
+
+impl RequestMeta {
+    /// Returns `true` if the request is a subscription.
+    pub const fn is_subscription(&self) -> bool {
+        self.is_non_standard_sub || matches!(self.method.as_bytes(), b"eth_subscribe")
```
DaniPopes, line 20:
```suggestion
        self.is_non_standard_sub || self.method == "eth_subscribe"
```

crates/json-rpc/src/request.rs
```diff
@@ -161,6 +189,17 @@ impl SerializedRequest {
...
 
+    /// Mark the request as a non-standard subscription (i.e. not
+    /// `eth_subscribe`)
+    pub fn set_non_standard_sub(&mut self) {
+        self.meta.is_non_standard_sub = true;
+    }
+
+    /// Returns `true` if the request is a subscription.
+    pub const fn is_subscription(&self) -> bool {
+        self.meta.is_non_standard_sub || matches!(self.method().as_bytes(), b"eth_subscribe")
```
DaniPopes, line 200: self.meta.method()

crates/pubsub/src/connect.rs
prestwich: i want to clean up a bit by taking it off the inflight and making a `PubSubInstruction::Subscribe(Inflight, channel_size)`

crates/json-rpc/src/request.rs
prestwich: this is non-const

crates/pubsub/src/connect.rs
prestwich: > i want to clean up a bit by taking it off the inflight and making a PubSubInstruction::Subscribe(Inflight, channel_size)

nvm putting it on the IX would require storing it int he backend separately. so i'll leave it on the inflight

crates/json-rpc/src/request.rs
```diff
@@ -161,6 +199,17 @@ impl SerializedRequest {
         self.meta.method
     }
 
+    /// Mark the request as a non-standard subscription (i.e. not
+    /// `eth_subscribe`)
+    pub fn set_non_standard_sub(&mut self) {
+        self.meta.is_subscription = true;
```
onbjerg, line 205:
```suggestion
        self.meta.set_is_subscription();
```

nit

onbjerg (changes requested):

crates/json-rpc/src/request.rs
```diff
@@ -10,6 +10,26 @@ pub struct RequestMeta {
...
+    /// Returns `true` if the request is a subscription.
+    pub const fn is_subscription(&self) -> bool {
+        self.is_subscription || matches!(self.method.as_bytes(), b"eth_subscribe")
+    }
+
+    /// Indicates that the request is a non-standard subscription (i.e. not
+    /// "eth_subscribe").
+    pub fn set_is_subscription(&mut self) {
+        self.is_subscription = true;
+    }
```
mattsse, line 32: this could use some helper functions to:
set sub state with bool arg

crates/json-rpc/src/request.rs
```diff
@@ -29,6 +49,24 @@ pub struct Request<Params> {
...
+    /// Returns `true` if the request is a subscription.
+    pub const fn is_subscription(&self) -> bool {
+        self.meta.is_subscription()
+    }
+
+    /// Indicates that the request is a non-standard subscription (i.e. not
+    /// "eth_subscribe").
+    pub fn set_is_subscription(&mut self) {
+        self.meta.set_is_subscription()
+    }
```
mattsse, line 67: same

mattsse (approved):
this solution gives the user the most control and we're no longer limited to eth_subscribe

DaniPopes (approved):
LG to the ME

PR #178, opened 2024-02-02
refactor: dedupe `CallRequest`/`TransactionRequest`

onbjerg:
## Motivation

`CallRequest` and `TransactionRequest` are basically the same type, with very few differences:

- `TransactionRequest` was missing an optional chain ID
- `TransactionRequest` was missing `max_fee_per_blob_gas`
- `CallRequest` and `TransactionRequest` had different blob fields

Since they also serve the same purpose, this PR dedupes them, to make sure there is only one tx request type, which is important for the network abstraction.

## Solution

Dedupe them, porting over fields missing from either.

I removed the blob transaction sidecar, but only because I didn't really know if it is actually a part of the spec. As far as I could tell, only `blob_versioned_hashes` and the gas field is. I asked @mattsse for direction in case he knew, but ultimately we didn't land on anything.

This is a breaking change for obvious reasons, and it not simply a "import new type" migration path, as I also ported over `CallInput` for `TransactionRequest`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/request.rs
+++ b/crates/rpc-types/src/eth/transaction/request.rs
@@ -3,48 +3,48 @@ use crate::{eth::transaction::AccessList, other::OtherFields, BlobTransactionSid
 use alloy_primitives::{Address, Bytes, B256, U128, U256, U64, U8};
 use serde::{Deserialize, Serialize};
 
-/// Represents _all_ transaction requests to/from RPC
+/// Represents _all_ transaction requests to/from RPC.
 #[derive(Clone, Debug, PartialEq, Eq, Default, Serialize, Deserialize)]
 #[serde(deny_unknown_fields)]
 #[serde(rename_all = "camelCase")]
 pub struct TransactionRequest {
-    /// from address
+    /// The address of the transaction author.
     pub from: Option<Address>,
-    /// to address
+    /// The destination address of the transaction.
     pub to: Option<Address>,
-    /// legacy, gas Price
+    /// The legacy gas price.
     #[serde(default)]
     pub gas_price: Option<U128>,
-    /// max base fee per gas sender is willing to pay
+    /// The max base fee per gas the sender is willing to pay.
     #[serde(default)]
     pub max_fee_per_gas: Option<U128>,
-    /// miner tip
+    /// The max priority fee per gas the sender is willing to pay, also called the miner tip.
     #[serde(default)]
     pub max_priority_fee_per_gas: Option<U128>,
-    /// Max Fee per Blob gas for EIP-4844 transactions
+    /// The max fee per blob gas for EIP-4844 blob transactions.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub max_fee_per_blob_gas: Option<U256>,
-    /// gas
+    /// The gas limit for the transaction.
     pub gas: Option<U256>,
-    /// value of th tx in wei
+    /// The value transferred in the transaction, in wei.
     pub value: Option<U256>,
-    /// Any additional data sent
+    /// Transaction data.
     #[serde(default, flatten)]
     pub input: TransactionInput,
-    /// Transaction nonce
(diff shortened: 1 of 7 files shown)


onbjerg:
@DaniPopes where do I note breaking changes for the checklist?

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -36,8 +39,11 @@ pub struct TransactionRequest {
     /// EIP-2718 type
     #[serde(rename = "type")]
     pub transaction_type: Option<U8>,
-    /// sidecar for EIP-4844 transactions
-    pub sidecar: Option<BlobTransactionSidecar>,
```
mattsse, line 40: we still need this if we want to use this to convert into a full eip4844 transaction.

it's not needed for anything rpc besides eth_sendTransaction, but if we wan to use this type as the builder for constructing sendable transactions we need this

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -36,8 +39,11 @@ pub struct TransactionRequest {
...
+    /// Blob Versioned Hashes for EIP-4844 transactions
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub blob_versioned_hashes: Option<Vec<B256>>,
+    /// chain id
+    pub chain_id: Option<U64>,
```
mattsse, line 46: this should use order as CallRequest before, should come after nonce

mattsse (changes requested):
I'm for unifying them, not worth having duplicate versions of the same thing. especially if we want to use this type as the foundation for building transactions.

before we merge this here, I'd like companion prs on foundry and reth so we can merge them in one go.

crates/rpc-types/src/eth/transaction/request.rs
onbjerg: thanks, that's what i was asking about in alloy maintainers, will add it back.

Evalir (approved):
This is fine w/ me. I'll take care of the foundry companion PR

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -1,51 +1,106 @@
...
-    /// EIP-2718 type
+    /// The EIP-2718 transaction type. See [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718) for more information.
     #[serde(rename = "type")]
     pub transaction_type: Option<U8>,
-    /// sidecar for EIP-4844 transactions
+    /// Blob versioned hashes for EIP-4844 transactions.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub blob_versioned_hashes: Option<Vec<B256>>,
+    /// Blob sidecar for EIP-4844 transactions.
     pub sidecar: Option<BlobTransactionSidecar>,
```
mattsse, line 50: this should also be #[serde(skip_serializing_if = "Option::is_none")]

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -1,51 +1,106 @@
...
     /// Support for arbitrary additional fields.
     #[serde(flatten)]
     pub other: OtherFields,
 }
 
+impl Hash for TransactionRequest {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.from.hash(state);
```
mattsse, line 58: why can't we derive this?

mattsse (changes requested):
last nit

crates/rpc-types/src/eth/transaction/request.rs
onbjerg: `OtherFields` contains `serde_json::Value` which is not `Hash`

Evalir:
gm @onbjerg can this be merged?

PR #179, opened 2024-02-02
ClientRefs, Poller, and Streams

prestwich:
Separate `RpcClient` from `RpcClientInner` to allow weak and lifetime-borrowed use of RpcClient without holding a reference beyond

## Motivation

- Make it so that long-lived tasks do not permanently hold RpcClients open
- lay groundwork for block heartbeat and other tasks that live as long as an Arc<Client>
- make it clear which to use in any given moment 

## Solution

- Make RpcClient<Arc<RpcClientInner>>

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/providers/src/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+                Some(provider) => provider,
+                None => {
+                    tracing::debug!("provider dropped");
+                    break 'task;
+                },
+            };
+
+            // Then try to fill as many blocks as possible
+            while !known_blocks.contains(&block_number) {
+                let block = provider.get_block_by_number(block_number, false).await;
```
gakonst, line 77: would consider doing a tokio::future::join_all if the gap between `block_number` and `next_yield` is >N blocks

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// Reap any timeout
+    fn reap_timeouts(&mut self) {
+        let now = std::time::Instant::now();
+        let to_keep = self.reap_at.split_off(&now);
+        let to_reap = std::mem::replace(&mut self.reap_at, to_keep);
+
+        for tx_hash in to_reap.values() {
+            self.unconfirmed.remove(tx_hash);
+        }
+    }
```
gakonst, line 161: Reap is basically "unsubscribe from transactions that have been pending for longer than T"?

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,255 @@
...
+        if let Some(timeout) = to_watch.timeout {
+            self.reap_at.insert(std::time::Instant::now() + timeout, to_watch.tx_hash);
+        }
+        self.unconfirmed.insert(to_watch.tx_hash, to_watch);
+    }
+
+    /// Handle a new block by checking if any of the transactions we're
+    /// watching are in it, and if so, notifying the watcher. Also updates
+    /// the latest block.
+    fn handle_new_block(&mut self, block: Block, latest: &watch::Sender<Block>) {
```
gakonst, line 176: 👍

gakonst (changes requested):
cool! i think we can disentangle this from @onbjerg's #190 work? any blockers / next steps?

crates/pubsub/src/sub.rs
```diff
@@ -274,4 +301,116 @@ impl<T: DeserializeOwned> Subscription<T> {
...
+        mut self: Pin<&mut Self>,
+        cx: &mut task::Context<'_>,
+    ) -> task::Poll<Option<Self::Item>> {
+        match ready!(self.inner.poll_next_unpin(cx)) {
+            Some(Ok(value)) => task::Poll::Ready(Some(Ok(serde_json::from_str(value.get())))),
+            Some(Err(e)) => task::Poll::Ready(Some(Err(e))),
+            None => task::Poll::Ready(None),
+        }
+    }
 }
```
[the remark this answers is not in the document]
DaniPopes: Don't think new types are necessary because we don't use the internal ID and can be just expressed as a stream map combinator

crates/rpc-client/src/client.rs
```diff
@@ -1,10 +1,95 @@
...
 
+/// An [`RpcClient`] in a [`Weak`] reference.
+pub type WeakClient<T> = Weak<RpcClientInner<T>>;
+
+/// A borrowed [`RpcClient`].
+pub type ClientRef<'a, T> = &'a RpcClientInner<T>;
+
+/// A JSON-RPC client.
+#[derive(Debug)]
+pub struct RpcClient<T>(Arc<RpcClientInner<T>>);
```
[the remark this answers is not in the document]
DaniPopes: Allows breaking out of task loops when the provider is dropped in addition to the channel dropping, makes sense with broadcast channels ig

crates/rpc-client/src/poller.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    /// Spawn the poller task, producing a stream of responses.
+    pub fn spawn(self) -> PollChannel<Resp> {
+        let (tx, rx) = broadcast::channel(self.channel_size);
+
+        let fut = async move {
+            let limit = self.limit.unwrap_or(usize::MAX);
+            for _ in 0..limit {
+                let client = match self.client.upgrade() {
```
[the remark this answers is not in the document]
DaniPopes: I believe we want to break if this loop is the only provider left

crates/providers/src/new.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_send_tx() {
+        init_tracing();
+
+        let anvil = alloy_node_bindings::Anvil::new().block_time(1u64).spawn();
+        let url = anvil.endpoint().parse().unwrap();
+        let http = Http::<Client>::new(url);
+        let provider = RootProvider::<TmpNetwork, _>::new(RpcClient::new(http, true));
```
gakonst, line 349: we should sugar this to `Provider::new(url)` and have it work for all HTTP/WS/IPC etc.

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Require a number of confirmations.
+    confirmations: u64,
+
+    /// Optional timeout for the transaction.
+    timeout: Option<Duration>,
+}
+
+impl WatchConfig {
+    /// Create a new watch for a transaction.
+    pub fn new(tx_hash: B256) -> Self {
```
mattsse, line 33: we can do `impl From<B256>` so the other functions can accept `impl Into<WatchConfig>`

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl PendingTransaction {
+    /// Returns this transaction's hash.
+    pub const fn tx_hash(&self) -> &B256 {
+        &self.tx_hash
+    }
+}
+
+impl Future for PendingTransaction {
+    type Output = TransportResult<()>;
```
mattsse, line 95: this should probably return the hash

so you can do:

```
let hash = pending.await?;
```

because future consumes the value

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,276 @@
...
+pub(crate) struct HeartbeatHandle {
+    tx: mpsc::Sender<TxWatcher>,
+    latest: watch::Receiver<Option<Block>>,
+}
+
+impl HeartbeatHandle {
+    /// Watch for a transaction to be confirmed with the given config.
+    pub(crate) async fn watch_tx(
+        &self,
+        config: WatchConfig,
```
mattsse, line 116:
```suggestion
        config: impl Into<WatchConfig>,
```

mattsse (changes requested):
this should work,

some nits

crates/providers/src/heart.rs
DaniPopes: It should return the receipt

crates/providers/src/heart.rs
DaniPopes: It's not a public function, doesn't really matter

gakonst:
@DaniPopes let's review & get this in if feature complete and proceed with events?

crates/providers/src/heart.rs
```diff
@@ -0,0 +1,283 @@
...
+                        },
+
+                        // Wake up to handle new blocks.
+                        block = self.stream.select_next_some() => {
+                            self.handle_new_block(block, &latest);
+                        },
+
+                        // This arm ensures we always wake up to reap timeouts,
+                        // even if there are no other events.
+                        _ = tokio::time::sleep_until(next_reap.into()) => {},
```
mattsse, line 271: hmm, I don't think this is correct because Sleep is not Unpin.
 
I believe this should be pinned:

https://docs.rs/tokio/latest/tokio/time/struct.Sleep.html#examples

mattsse (changes requested):
I think I can see how it all fits together.

I have one concern re usage of sleep in select!

otherwise lgtm

mattsse (approved):

PR #181, opened 2024-02-04
feat: support no_std for `alloy-eips`

yjhmelody:
Also, fix `alloy-signer-gcp` compile problems


mattsse (changes requested):
I think we can do that, so supportive, doesn't hurt.

though I'd like to not use thiserror-no-std and instead do all of that manually
can do expand and copy the generated code+cleanup

crates/signer-aws/src/signer.rs
```diff
@@ -230,7 +230,7 @@ fn sig_from_digest_bytes_trial_recovery(
 
 /// Makes a trial recovery to check whether an RSig corresponds to a known `VerifyingKey`.
 fn check_candidate(signature: &Signature, hash: B256, pubkey: &VerifyingKey) -> bool {
-    signature.recover_from_prehash(&hash).map(|key| key == *pubkey).unwrap_or(false)
+    signature.recover_from_prehash(hash).map(|key| key == *pubkey).unwrap_or(false)
```
DaniPopes, line 233: This breaks the crates, run `cargo update` to use what CI uses

crates/eips/src/eip2718.rs
```diff
@@ -2,26 +2,25 @@
 //!
 //! [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
 
+use alloc::{vec, vec::Vec};
 use alloy_primitives::{keccak256, Sealed, B256};
 use alloy_rlp::{BufMut, Header, EMPTY_STRING_CODE};
+use thiserror_no_std::Error;
```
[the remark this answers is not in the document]
yjhmelody: ok

Cargo.toml
```diff
@@ -83,10 +81,11 @@ rand = "0.8.5"
 reqwest = { version = "0.11.18", default-features = false }
 semver = "1.0"
 thiserror = "1.0"
+thiserror-no-std = { version = "2.0", default-features = false }
```
mattsse, line 84: we don't need this

crates/eips/src/eip2718.rs
```diff
@@ -2,28 +2,46 @@
...
     /// Rlp error from [`alloy_rlp`].
-    #[error(transparent)]
-    RlpError(#[from] alloy_rlp::Error),
+    RlpError(alloy_rlp::Error),
     /// Got an unexpected type flag while decoding.
-    #[error("Unexpected type flag. Got {0}.")]
     UnexpectedType(u8),
-    /// Some other error occurred.
-    #[error(transparent)]
-    Custom(#[from] Box<dyn std::error::Error>),
```
mattsse, line 24: I think eventually we'd need this, so perhaps the better question for this PR is, does this type even belong in here? 

perhaps we should rather move this to the consensus crate instead, I believe it's better suited in there

mattsse (changes requested):
lg on the no-std changes

though about the error type. I'm skeptical that this even belongs in there if it's only used in the consensus crate

https://github.com/search?q=repo%3Aalloy-rs%2Falloy%20Eip2718Error&type=code

we should move it there imo

wdyt @DaniPopes

crates/eips/src/eip2718.rs
yjhmelody: What do you mean?
Move all eip error type to consensus crate? and add Error type to trait such as `Decodable2718`?

yjhmelody: But that might be a break change, which the current one is not?

mattsse: removing a variant is a breaking change as well, I'd really prefer moving the error. imo it does not belong in the eip crate

crates/eips/src/eip2718.rs
```diff
@@ -2,28 +2,46 @@
...
 
 // https://eips.ethereum.org/EIPS/eip-2718#transactiontype-only-goes-up-to-0x7f
 const TX_TYPE_BYTE_MAX: u8 = 0x7f;
 
 /// [EIP-2718] decoding errors.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
-#[derive(thiserror::Error, Debug)]
+#[derive(Debug, Copy, Clone)]
 pub enum Eip2718Error {
```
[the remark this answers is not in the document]
yjhmelody: > I still think we should move this error type and keep the custom variant

i want to how tohandle the related trait

prestwich:
> though about the error type. I'm skeptical that this even belongs in there if it's only used in the consensus crate

not opposed to removing it, the purpose of it is to allow for non-RLP encoding schemes in EIP-2718 in non-Ethereum networks. E.g. if Optimism decides to have a fallible SSZ encoding for some transaction type, the existing API can still capture that behavior via the `Custom` error

mattsse (approved):

prestwich:
cleaning up CI before merge 👍

prestwich:
nightly clippy strikes again

https://github.com/rust-lang/rust/issues/121708

prestwich:
ty for you contribution @yjhmelody, sorry for leaving it open so long 🙇‍♀️

PR #182, opened 2024-02-05
feat: alloy-contract

DaniPopes:
Upgrade `alloy-dyn-contract` to `alloy-contract` by making `CallBuilder` generic over the output decoder, which can now be `sol!`-generated (see https://github.com/alloy-rs/core/pull/510) as well "noop" decoder which makes `call` behave the same as `call_raw`.

Bumps MSRV to 1.75, but this is not strictly necessary for the core functionality.


crates/contract/src/instance.rs
```diff
@@ -16,79 +17,70 @@ pub struct ContractInstance<P> {
...
-    #[must_use]
-    pub fn at(&self, address: Address) -> ContractInstance<P> {
-        let mut this = self.clone();
-        this.address = address;
-        this
+impl<P: Clone> ContractInstance<&P> {
+    /// Clones the provider and returns a new contract instance with the cloned provider.
+    #[inline]
+    pub fn with_cloned_provider(self, provider: P) -> ContractInstance<P> {
+        ContractInstance { address: self.address, provider, interface: self.interface }
```
mattsse, line 61: This does not clone the provider but swaps it from a ref to owned

looks like this doesn't need the P: Clone bound and the function could accept any Provider type?

crates/contract/src/call.rs
```diff
@@ -0,0 +1,404 @@
...
+/// # }
+/// ```
+///
+/// [sol]: alloy_sol_types::sol
+#[derive(Clone)]
+#[must_use = "call builders do nothing unless you `.call`, `.send`, or `.await` them"]
+pub struct CallBuilder<P, D> {
+    // TODO: this will not work with `send_transaction` and does not differentiate between EIP-1559
+    // and legacy tx
+    request: CallRequest,
```
mattsse, line 191: right for signing we need another step where the request is converted into a TypedTransaction first

crates/contract/src/call.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// Decodes the output of a contract function using the provided decoder.
+    #[inline]
+    pub fn decode_output(&self, data: Bytes, validate: bool) -> Result<D::Output> {
+        self.decoder.abi_decode_output(data, validate)
+    }
+
+    /// Signs and broadcasts this call as a transaction.
+    pub async fn send(&self) -> Result<()> {
+        todo!()
+    }
```
mattsse, line 356: Perhaps we can move this behind a Provider trait bound, requiring that the provider supports signing

crates/contract/src/call.rs
Evalir: Related `TypedTransaction` work: https://github.com/alloy-rs/alloy/pull/183

crates/contract/src/call.rs
gakonst: we want it to work with eth_sendTransaction too, so i think just need to modify the comment

crates/contract/src/call.rs
```diff
@@ -0,0 +1,424 @@
...
+
+/// [`CallBuilder`] using a [`SolCall`] type as the call decoder.
+// NOTE: please avoid changing this type due to its use in the `sol!` macro.
+pub type SolCallBuilder<P, C> = CallBuilder<P, PhantomData<C>>;
+
+/// [`CallBuilder`] using a [`Function`] as the call decoder.
+pub type DynCallBuilder<P> = CallBuilder<P, Function>;
+
+/// [`CallBuilder`] that does not have a call decoder.
+pub type RawCallBuilder<P> = CallBuilder<P, ()>;
```
gakonst, line 22: amazing -- exactly how i had it in mind

onbjerg (approved):
lgtm (minus doc nits from other reviewers) and also noting that `SolCallBuilder` will end up changing pretty soon

DaniPopes:
CallBuilder changing is fine, as long as the APIs marked with my NOTE comment stay the same, so as to not break `sol!`-generated code. Otherwise it would need to be updated in alloy-rs/core too.

onbjerg:
`SolCallBuilder` will end up taking an additional generic for `N: Network`, so this would break the API.

onbjerg (approved):
pog personally, this looks good to me. excited to see this macro in the wild

need to fix deny check, otherwise can merge imo

PR #188, opened 2024-02-06
feat: extend FeeHistory type with eip-4844 fields

allnil:
Related to reth task: https://github.com/paradigmxyz/reth/issues/6330
according to: https://github.com/ethereum/execution-apis/pull/486

### DIFF
--- a/crates/rpc-types/src/eth/fee.rs
+++ b/crates/rpc-types/src/eth/fee.rs
@@ -46,6 +46,13 @@ pub struct FeeHistory {
     ///
     /// The `Option` is only for compatability with Erigon and Geth.
     pub gas_used_ratio: Vec<f64>,
+    /// An array of block base fees per blob gas. This includes the next block after the newest
+    /// of  the returned range, because this value can be derived from the newest block. Zeroes
+    /// are returned for pre-EIP-4844 blocks.
+    pub base_fee_per_blob_gas: Vec<U256>,
+    /// An array of block blob gas used ratios. These are calculated as the ratio of gasUsed and
+    /// gasLimit.
+    pub blob_gas_used_ratio: Vec<f64>,
     /// Lowest number block of the returned range.
     pub oldest_block: U256,
     /// An (optional) array of effective priority fee per gas data points from a single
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/fee.rs
```diff
@@ -46,6 +46,13 @@ pub struct FeeHistory {
     ///
     /// The `Option` is only for compatability with Erigon and Geth.
     pub gas_used_ratio: Vec<f64>,
+    /// An array of block base fees per blob gas. This includes the next block after the newest
+    /// of  the returned range, because this value can be derived from the newest block. Zeroes
+    /// are returned for pre-EIP-4844 blocks.
+    pub base_fee_per_blob_gas: Vec<U256>,
+    /// An array of block blob gas used ratios. These are calculated as the ratio of gasUsed and
+    /// gasLimit.
+    pub blob_gas_used_ratio: Vec<f64>,
```
mattsse, line 55: these need to be option for now, because only supported after cancun
and skip if none

mattsse (changes requested):

crates/rpc-types/src/eth/fee.rs
DaniPopes: Should be enough: `#[serde(default, skip_serializing_if = "Vec::is_empty")]`

onbjerg: >these need to be option for now, because only supported after cancun
and skip if none

The spec says just to fill with 0s for pre-cancun blocks, would this not be ok even with cancun not activated?

DaniPopes: That's if the network and RPC provider implements the EIP

mattsse: right let's do

> #[serde(default, skip_serializing_if = "Vec::is_empty")]

then @allnil

allnil: that's both for base_fee_per_blob_gas and blob_gas_used_ratio or only for blob_gas_used_ratio?

onbjerg: both!

mattsse (approved):
lgtm

onbjerg (approved):
lgtm

PR #190, opened 2024-02-06
feat: network abstraction and transaction builder

prestwich:
## Motivation

We want a transaction builder that can build one of several strongly-typed consensus transactions for signing to implement `send_transaction`. The intended flow is:

- User specifies fields they want to add to their transaction
- They pass it off to a signer, where the concrete transaction type is resolved from the builder (e.g. legacy, EIP-1559, ..)
- The transaction is signed and sent

This is made complicated due to our network abstraction:

- Different networks have a different set of transactions
- Different networks can have different signature schemes, and some are valid for the same transaction types, i.e. our current associated type `Signature` for transactions is not a good abstraction

This flow should also work server side, i.e. if I get a `TransactionRequest` over the wire from a client, I should be able to resolve it to a consensus type transaction. I should also be able to take an already-signed RLP-encoded transaction and decode it into one of the valid transaction types.

This PR intends to:

- Add a builder
- Complete the request → signer flow, including adding `send_transaction` in the provider and the `alloy-contract` call builder
- Remove the temporary provider (and related code) and fully transition to the network-parameterized provider type

## Solution

- Adds a `TransactionBuilder` trait and implements this for `TransactionRequest`
- Adds an `Ethereum` network with all types defined
- Removes the temporary provider, replacing it with a new, network-parameterized provider
- Adjusts `alloy-contract` to work with the new provider
- Moves a lot of types around, because a lot of the abstractions we made had not been put together, and could not without circular dependencies
- Splits `Transaction` into `Transaction` and `SignableTransaction`, since not all transactions are signable (e.g. Optimism deposit tx's)
- Moves the associated `Signature` type in `Transaction` into a parameter instead
- Adds 3 new signer traits: `TxSigner`, `TxSignerSync` and `NetworkSigner`
  - `TxSigner` is essentially ripped out of the `Signer` trait and should be implemented for all signers, even if the signer does not care about transactions. 
  - It is needed for e.g. Ledger/Trezor where tx info is needed. A blanket implementation is not provided because it isn't allowed.
  - These two traits essentially state that a signer can sign *any* transaction with the given `Signature`. Signer implementors need to implement this and `Signer`.
  - `NetworkSigner` instead is implemented by network crate authors. This trait is implemented for a wrapper type and essentially states that the signer can sign *any* transaction for the given network. (See `EthereumSigner`)
- Finish up the `ProviderBuilder` abstraction and add a `SignerLayer` to add signing on top of an existing provider

Companion PRs:

- https://github.com/alloy-rs/core/pull/557
- https://github.com/alloy-rs/core/pull/542

Things left:

- Possibly improve the ergonomics of `ProviderBuilder` a bit

## Follow ups

- Block responses in the provider are not network parameterized yet, so the transactions and header are the same for all networks. This should be simple to adjust later.  (https://github.com/alloy-rs/alloy/issues/267)
- Add an `AnyNetwork` that is essentially `Ethereum` + `OtherFields` on all types https://github.com/alloy-rs/alloy/issues/272
- Maybe move `Wallet` etc. out of `alloy-signers` to make `alloy-network` wasm compatible again, see https://github.com/alloy-rs/alloy/pull/190#discussion_r1516749547 (https://github.com/alloy-rs/alloy/issues/262)
- Conversions/impls making it easier to instantiate a `RootProvider` (https://github.com/alloy-rs/alloy/issues/266)

## Breaking changes

Anything that uses `alloy-contract`, `alloy-provider`, or `alloy-consensus` likely broke with this PR as the API changed.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/signer/src/signer.rs
```diff
@@ -58,32 +27,18 @@ impl<T: ?Sized + Transaction<Signature = Signature>> TransactionExt for T {}
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 #[auto_impl(&mut, Box)]
-pub trait Signer: Send + Sync {
+pub trait Signer<Sig: 'static = Signature>: Send + Sync {
```
gakonst, line 30: why do we need it to be `Static`?

crates/signer/src/signer.rs
```diff
@@ -1,43 +1,12 @@
...
-pub type SignableTx = dyn Transaction<Signature = Signature>;
-
-/// Extension trait for utilities for signable transactions.
-///
-/// This trait is implemented for all types that implement [`Transaction`] with [`Signature`] as the
-/// signature associated type.
-pub trait TransactionExt: Transaction<Signature = Signature> {
-    /// Set `chain_id` if it is not already set. Checks that the provided `chain_id` matches the
-    /// existing `chain_id` if it is already set.
-    fn set_chain_id_checked(&mut self, chain_id: ChainId) -> Result<()> {
```
gakonst, line 21: assuming we'll be doing this somewhere in the provider now maybe?

crates/consensus/src/transaction/typed.rs
```diff
@@ -0,0 +1,147 @@
...
+
+/// The TypedTransaction enum represents all Ethereum transaction request types.
+///
+/// Its variants correspond to specific allowed transactions:
+/// 1. Legacy (pre-EIP2718) [`TxLegacy`]
+/// 2. EIP2930 (state access lists) [`TxEip2930`]
+/// 3. EIP1559 [`TxEip1559`]
+/// 4. EIP4844 [`TxEip4844`]
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum TypedTransaction {
```
gakonst, line 13: Should this impl RLP?

crates/network/src/lib.rs
```diff
@@ -51,6 +54,8 @@ pub struct BlockResponse<N: Network> {
 }
 
 /// Captures type info for network-specific RPC requests/responses.
+// todo: block responses are ethereum only, so we need to include this in here too, or make `Block`
+// generic over tx/header type
```
gakonst, line 58: let's do as follow-up IMO when we're adding the 2nd network?

crates/network/src/transaction/builder.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn gas_price(&self) -> Option<u128>;
+
+    /// Set the gas price for the transaction.
+    fn set_gas_price(&mut self, gas_price: u128);
+
+    /// Builder-pattern method for setting the gas price.
+    fn with_gas_price(mut self, gas_price: u128) -> Self {
+        self.set_gas_price(gas_price);
+        self
+    }
```
gakonst, line 117: same

crates/network/src/transaction/signer.rs
```diff
@@ -0,0 +1,91 @@
...
+
+// todo: move
+/// An async signer capable of signing any [SignableTransaction] for the given [Signature] type.
+#[async_trait]
+pub trait TxSigner<Signature> {
+    /// Asynchronously sign an unsigned transaction.
+    async fn sign_transaction(
+        &self,
+        tx: &mut dyn SignableTransaction<Signature>,
+    ) -> alloy_signer::Result<Signature>;
```
gakonst, line 80: what's the difference between these 2?

crates/providers/src/lib.rs
```diff
@@ -116,11 +323,25 @@ pub trait Provider<N: Network, T: Transport = BoxTransport>: Send + Sync {
...
     }
+
+    // todo: this breaks object safety, but we need something like it for foundry
+    /*
+    async fn raw_request<P, R>(&self, method: &'static str, params: P) -> TransportResult<R>
+    where
+        P: Serialize + Send + Sync + Clone,
+        R: Serialize + DeserializeOwned + Send + Sync + Unpin + 'static,
+        Self: Sync;
+        */
```
gakonst, line 344: @prestwich @mattsse have you guys thought about this?

We could pull this to separate trait, perhaps? Or not even have it on the trait and assume that for raw RPC requests in Foundry you'll need to have the concrete client type?

crates/providers/src/lib.rs
```diff
@@ -134,22 +355,201 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for NetworkRpcClient<N, T>
...
-    ) -> TransportResult<alloy_primitives::U256> {
-        self.prepare(
-            "eth_getTransactionCount",
-            Cow::<(Address, String)>::Owned((address, "latest".to_string())),
-        )
-        .await
+        key: U256,
+        tag: Option<BlockId>,
+    ) -> TransportResult<StorageValue> {
+        self.client.prepare("eth_getStorageAt", (address, key, tag.unwrap_or_default())).await
```
gakonst, line 413: generally find it weird we're calling this `self.client.prepare` when it's sending the request, shouldnt it be `self.client.prepare(...).send().await` or `self.client.request(...)`?

crates/providers/src/lib.rs
```diff
@@ -134,22 +355,201 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for NetworkRpcClient<N, T>
...
+        &self,
+        tx: &N::TransactionRequest,
+        block: Option<BlockId>,
+    ) -> TransportResult<U256> {
+        todo!()
+        // if let Some(block_id) = block {
+        //     self.client.prepare("eth_estimateGas", (tx, block_id)).await
+        // } else {
+        //    self.client.prepare("eth_estimateGas", (tx,)).await
+        // }
```
gakonst, line 504: what is our todo here?

crates/contract/src/call.rs
```diff
@@ -190,25 +187,25 @@ impl CallDecoder for () {
...
+    request: N::TransactionRequest,
     block: Option<BlockId>,
     state: Option<StateOverride>,
     provider: P,
     decoder: D,
 }
 
 // See [`ContractInstance`].
-impl<P: TempProvider> DynCallBuilder<P> {
+impl<N: Network, P: Provider<N>> DynCallBuilder<N, P> {
```
DaniPopes, line 201: This will only accept BoxTransport because of `Provider<N, T = BoxTransport>`, I think we need to make this generic over T: Transport as well

crates/consensus/src/receipt/receipts.rs
```diff
@@ -28,30 +28,12 @@ impl Receipt {
     }
 }
 
-impl alloy_network::Receipt for Receipt {
```
onbjerg, line 31: These impls have been moved to the network crate because of a circular dep.

Note to self: we can move this trait to `alloy-consensus` too, so the impl is colocated

crates/contract/src/call.rs
```diff
@@ -308,6 +308,8 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
         self
     }
 
+    // todo map function fn(req) -> req
```
onbjerg, line 311: note to self

crates/contract/src/call.rs
```diff
@@ -456,10 +459,10 @@ mod tests {
     use super::*;
     use alloy_node_bindings::{Anvil, AnvilInstance};
     use alloy_primitives::{address, b256, bytes, hex};
-    use alloy_providers::tmp::{HttpProvider, Provider};
+    use alloy_providers::{HttpProvider, Provider};
     use alloy_sol_types::sol;
 
-    #[test]
+    /*#[test]
```
onbjerg, line 465: note to self: re-enable tests

crates/contract/src/lib.rs
```diff
@@ -15,8 +15,9 @@
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
-#[cfg(test)]
-extern crate self as alloy_contract;
+// todo: re-enable when alloy-sol-macro is adjusted
```
onbjerg, line 18: note to self: re-enable

crates/network/src/ethereum/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+    type TransactionRequest = alloy_rpc_types::transaction::TransactionRequest;
+
+    type TransactionResponse = alloy_rpc_types::Transaction;
+
+    type ReceiptResponse = alloy_rpc_types::TransactionReceipt;
+
+    type HeaderResponse = alloy_rpc_types::Header;
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
onbjerg, line 33: note to self: move to other module

crates/network/src/ethereum/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+            }
+            _ => build_legacy(self).map(Into::into),
+        }
+    }
+
+    async fn build<S: NetworkSigner<Ethereum>>(
+        self,
+        signer: &S,
+    ) -> BuilderResult<<Ethereum as Network>::TxEnvelope> {
+        // todo: BuilderResult + SignerResult
```
onbjerg, line 157: note to self: impl

crates/network/src/ethereum/mod.rs
```diff
@@ -0,0 +1,267 @@
...
+        Ok(signer.sign(self.build_unsigned()?).await.unwrap())
+    }
+}
+
+/// Build a legacy transaction.
+fn build_legacy(request: TransactionRequest) -> Result<TxLegacy, TransactionBuilderError> {
```
onbjerg, line 163: Ideally I'd have these on `TransactionRequest`, but that would cause a circular dep (`alloy-network` depends on `rpc-types` -> `rpc-types` would need to depend on `alloy-network` for `TransactionBuilderError`)

crates/providers/src/chain.rs
```diff
@@ -92,17 +92,26 @@ impl<P, T: Transport + Clone> ChainStreamPoller<P, T> {
...
-                let block = match provider.get_block_by_number(number, false).await {
-                    Ok(block) => block,
+                let block = match provider.get_block_by_number(number.into(), false).await {
+                    Ok(Some(block)) => block,
                     Err(RpcError::Transport(err)) if retries > 0 && err.recoverable() => {
                         debug!(number, %err, "failed to fetch block, retrying");
                         retries -= 1;
                         continue;
                     }
+                    Ok(None) if retries > 0 => {
```
onbjerg, line 102: the rpc can respond with the block not existing, which can be the case if we are querying the latest block but the node doesn't have it yet. in that case i think we retry imo

crates/providers/src/new.rs
```diff
@@ -84,41 +98,41 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
-        Ok(())
+    // todo eip-1559 and blobs as well
+    async fn populate_gas(
+        &self,
+        tx: &mut N::TransactionRequest,
+        block: Option<BlockId>,
+    ) -> TransportResult<()> {
+        let _ = self.estimate_gas(&*tx, block).await;
+
+        todo!()
```
onbjerg, line 134: note to self

crates/providers/src/new.rs
```diff
@@ -138,8 +152,281 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+                base_fee_per_gas,
+                fee_history.reward.unwrap_or_default(),
+            )
+        };
+
+        Ok((max_fee_per_gas, max_priority_fee_per_gas))
+    }
+
+    // todo: move to extension trait
+    #[cfg(feature = "anvil")]
```
onbjerg, line 363: note to self

crates/network/src/transaction/signer.rs
onbjerg: `NetworkSigner` is a signer capable of signing any transaction with any signature type for a given network, `TxSigner` is a signer capable of signing any transaction regardless of network for a given signature type.

The idea is to have fewer generics for signers

crates/consensus/src/transaction/builder.rs
```diff
@@ -0,0 +1,258 @@
...
+    fn build_unsigned(self) -> Result<TypedTransaction, BuilderError> {
+        todo!()
+    }
+
+    fn build<S: alloy_network::NetworkSigner<Ethereum>>(
+        self,
+        _signer: &S,
+    ) -> Result<TxEnvelope, BuilderError> {
+        todo!()
+    }
```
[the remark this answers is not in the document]
onbjerg: copied over the conversion from reth

crates/providers/src/lib.rs
onbjerg: pulled into a separate trait

crates/providers/src/lib.rs
onbjerg: I think it's called `prepare` because it doesn't do any serialization or really anything at all until `await`ed. The `send` method in your example is awaiting it

crates/signer-ledger/src/signer.rs
```diff
@@ -27,10 +28,37 @@ pub struct LedgerSigner {
     pub(crate) address: Address,
 }
 
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+impl alloy_network::TxSigner<Signature> for LedgerSigner {
```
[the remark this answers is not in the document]
onbjerg: nvm, this signer is async only

crates/consensus/src/transaction/typed.rs
onbjerg: unsure, do we want it? cc @mattsse perhaps

crates/network/src/lib.rs
```diff
@@ -50,8 +50,20 @@ pub struct BlockResponse<N: Network> {
     transactions: TransactionList<N::TransactionResponse>,
 }
 
+/// A receipt response.
+///
+/// This is distinct from [`TxReceipt`], since this is for JSON-RPC receipts.
+///
+/// [`TxReceipt`]: alloy_consensus::TxReceipt
+pub trait ReceiptResponse {
```
onbjerg, line 58: I added this because consensus receipts do not have the contract address field, and they do not have the means to calculate it either. We cannot recalculate it ourselves where needed, since the receipt might be the result of a transaction where we do not know the nonce (e.g. a partially filled `send_transaction` signed by the node instead of us)

crates/contract/src/call.rs
```diff
@@ -190,36 +189,38 @@ impl CallDecoder for () {
 /// [sol]: alloy_sol_types::sol
 #[derive(Clone)]
 #[must_use = "call builders do nothing unless you `.call`, `.send`, or `.await` them"]
-pub struct CallBuilder<P, D> {
+pub struct CallBuilder<N: Network, T, P, D> {
     // TODO: this will not work with `send_transaction` and does not differentiate between EIP-1559
     // and legacy tx
```
DaniPopes, line 194: Is this TODO outdated?

crates/contract/src/call.rs
```diff
@@ -368,30 +383,25 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
...
-                _cx: &mut std::task::Context<'_>,
-            ) -> std::task::Poll<Self::Output> {
-                todo!()
-            }
-        }
-        Ok(Tmp(PhantomData))
+    pub async fn send(
+        &self,
+    ) -> Result<impl IntoFuture<Output = Result<Option<N::ReceiptResponse>>> + '_> {
+        let pending = self.provider.send_transaction(self.request.clone()).await?;
```
DaniPopes, line 402: todo in followup, return pending transaction

crates/contract/src/call.rs
```diff
@@ -190,36 +189,38 @@ impl CallDecoder for () {
 /// [sol]: alloy_sol_types::sol
 #[derive(Clone)]
 #[must_use = "call builders do nothing unless you `.call`, `.send`, or `.await` them"]
-pub struct CallBuilder<P, D> {
+pub struct CallBuilder<N: Network, T, P, D> {
```
DaniPopes, line 192: I think we want N to default to Ethereum everywhere? @gakonst @prestwich

crates/network/src/ethereum/builder.rs
```diff
@@ -0,0 +1,242 @@
...
+            .ok_or_else(|| TransactionBuilderError::MissingKey("blob_versioned_hashes"))?,
+        max_fee_per_blob_gas: request
+            .max_fee_per_blob_gas
+            .ok_or_else(|| TransactionBuilderError::MissingKey("max_fee_per_blob_gas"))?
+            .to(),
+        input: request.input.into_input().unwrap_or_default(),
+    })
+}
+
+// todo: these types are almost 1:1, minus rlp decoding and ser/de, should dedupe
```
DaniPopes, line 231: https://github.com/alloy-rs/alloy/issues/255

DaniPopes (changes requested):

crates/network/src/transaction/builder.rs
```diff
@@ -0,0 +1,200 @@
...
+/// A Transaction builder for a network.
+///
+/// Transaction builders are primarily used to construct typed transactions that can be signed with
+/// [`TransactionBuilder::build`], or unsigned typed transactions with
+/// [`TransactionBuilder::build_unsigned`].
+///
+/// Transaction builders should be able to construct all available transaction types on a given
+/// network.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
```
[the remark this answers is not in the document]
onbjerg: we need it, otherwise we have no way to set input/value etc. in e.g. `CallBuilder`, since all of that is just `N::TransactionRequest`

crates/contract/src/call.rs
onbjerg: yes, good catch

crates/network/src/transaction/builder.rs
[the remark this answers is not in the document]
onbjerg: i misread, i thought u said we don't need the trait (missed async) lmao

we need async trait for the build fn that signs the tx since the signers are async, are u suggesting we just do sync signing?

crates/network/src/transaction/signer.rs
```diff
@@ -0,0 +1,132 @@
...
+    Signature, Signer, SignerSync, Wallet,
+};
+use async_trait::async_trait;
+
+/// A signer capable of signing any transaction for the given network.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait NetworkSigner<N: Network>: Send + Sync {
+    /// Asynchronously sign an unsigned transaction.
+    async fn sign(&self, tx: N::UnsignedTx) -> alloy_signer::Result<N::TxEnvelope>;
```
onbjerg, line 14: nit: sign_transaction

crates/providers/src/signer.rs
```diff
@@ -0,0 +1,152 @@
...
+
+        let wallet = alloy_signer::Wallet::from(anvil.keys()[0].clone());
+
+        // can we somehow remove the need for <_, Ethereum>? we NEED to call .network<Ethereum>
+        // note: we need to 1) add <_, Ethereum> 2) layer things, and then 3) call .network before
+        // we can call provider
+        let provider = ProviderBuilder::<_, Ethereum>::new()
+            .layer(SignerLayer::new(EthereumSigner::from(wallet)))
+            .network::<Ethereum>()
+            .provider(RootProvider::new(RpcClient::new(http, true)));
```
onbjerg, line 132: .connect sugar

crates/providers/src/signer.rs
```diff
@@ -0,0 +1,152 @@
...
+        let anvil = Anvil::new().spawn();
+        let url = anvil.endpoint().parse().unwrap();
+        let http = Http::<Client>::new(url);
+
+        let wallet = alloy_signer::Wallet::from(anvil.keys()[0].clone());
+
+        // can we somehow remove the need for <_, Ethereum>? we NEED to call .network<Ethereum>
+        // note: we need to 1) add <_, Ethereum> 2) layer things, and then 3) call .network before
+        // we can call provider
+        let provider = ProviderBuilder::<_, Ethereum>::new()
```
onbjerg, line 129: Ethereum::provider -> builder

crates/providers/src/builder.rs
```diff
@@ -54,6 +70,18 @@ pub struct ProviderBuilder<L, N = ()> {
...
+        ProviderBuilder { layer: Identity, network: PhantomData }
+    }
+}
+
+impl<N> Default for ProviderBuilder<Identity, N> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
onbjerg, line 84: add .signer sugar

crates/network/src/lib.rs
```diff
@@ -50,18 +50,28 @@ pub struct BlockResponse<N: Network> {
...
+    /// Address of the created contract, or `None` if the transaction was not a deployment.
+    fn contract_address(&self) -> Option<Address>;
+}
 
+/// Captures type info for network-specific RPC requests/responses.
+// todo: block responses are ethereum only, so we need to include this in here too, or make `Block`
+// generic over tx/header type
+pub trait Network: Clone + Copy + Sized + Send + Sync + 'static {
```
onbjerg, line 66: note about zst

crates/network/src/transaction/builder.rs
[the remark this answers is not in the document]
onbjerg: tried rm'ing, we still need it cus of some bounds in `Provider` (i assume added by async trait)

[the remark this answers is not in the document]
DaniPopes: you have to remove it in impls too

crates/providers/src/signer.rs
```diff
@@ -0,0 +1,152 @@
...
+    async fn poc() {
+        let anvil = Anvil::new().spawn();
+        let url = anvil.endpoint().parse().unwrap();
+        let http = Http::<Client>::new(url);
+
+        let wallet = alloy_signer::Wallet::from(anvil.keys()[0].clone());
+
+        // can we somehow remove the need for <_, Ethereum>? we NEED to call .network<Ethereum>
+        // note: we need to 1) add <_, Ethereum> 2) layer things, and then 3) call .network before
+        // we can call provider
```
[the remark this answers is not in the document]
onbjerg: didn't solve it, leave for follow up?

crates/providers/src/signer.rs
onbjerg: this would cause a circular dependency, so not possible currently. will leave for follow up

crates/providers/src/signer.rs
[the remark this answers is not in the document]
DaniPopes: sure

crates/contract/src/call.rs
onbjerg: this requires moving N to the end, so would need another PR to alloy-sol-macro. up to you

DaniPopes: that's fine

crates/providers/src/signer.rs
```diff
@@ -127,7 +126,7 @@ mod tests {
         // note: we need to 1) add <_, Ethereum> 2) layer things, and then 3) call .network before
         // we can call provider
         let provider = ProviderBuilder::<_, Ethereum>::new()
-            .layer(SignerLayer::new(EthereumSigner::from(wallet)))
+            .signer(EthereumSigner::from(wallet))
```
gakonst, line 129: would love if this could be `.signer(wallet)` via `impl From<Wallet> for EthereumSigner` and `impl From<Wallet> for SignerLayer` perhaps and making `fn signer<S: Into<SignerLayer>>` or something similar

crates/providers/src/signer.rs
gakonst: OK

.github/workflows/ci.yml
```diff
@@ -68,6 +68,7 @@ jobs:
         run: |
           cargo hack check --workspace --target wasm32-unknown-unknown \
             --exclude alloy-contract \
+            --exclude alloy-network \
```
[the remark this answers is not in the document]
gakonst: Yup -- let's make alloy-signer wasm compatilbe in follow up, should be easy

Cargo.toml
```diff
@@ -102,5 +115,5 @@ tempfile = "3.8"
 
 # TODO: Keep around until alloy-contract is stable.
 # This should only be used in `alloy-contract` tests.
-# [patch.crates-io]
-# alloy-sol-macro = { git = "https://github.com/alloy-rs/core", rev = "18b0509950c90d9ec38f25913b692ae4cdd6f227" }
+[patch.crates-io]
+alloy-sol-macro = { git = "https://github.com/alloy-rs/core", rev = "ab0cab1047a088be6cffa4e7a2fcde7cf77aa460" }
```
gakonst, line 119: This is now on main so can remove I think? Or still needed

crates/consensus/src/signed.rs
```diff
@@ -59,22 +59,22 @@ impl<T: Transaction> Signed<T> {
     }
 }
 
-impl<T: Transaction> alloy_rlp::Encodable for Signed<T> {
+impl<T: SignableTransaction<Sig>, Sig> alloy_rlp::Encodable for Signed<T, Sig> {
```
gakonst, line 62: i forget why we did Sig as a generic instead of associated type, maybe a follow-up breaking change to the trait could remove the double mention of `Sig` to make the `Signed` trait easier to impl

crates/contract/src/call.rs
gakonst: let's do as follow-up, smol breakage not blocking us from merging & can pass on to contributor

crates/contract/src/call.rs
```diff
@@ -190,36 +189,36 @@ impl CallDecoder for () {
...
     block: Option<BlockId>,
     state: Option<StateOverride>,
     provider: P,
     decoder: D,
+    transport: PhantomData<T>,
 }
 
 // See [`ContractInstance`].
-impl<P: TempProvider> DynCallBuilder<P> {
+impl<N: Network, T: Transport + Clone, P: Provider<N, T>> DynCallBuilder<N, T, P> {
```
gakonst, line 202: Confirming the `Clone` bound is OK for Transport?

crates/contract/src/call.rs
gakonst: +1 i think not a blocker for the foundry integration given we await the receipt in most areas w anvil

crates/contract/src/call.rs
```diff
@@ -403,25 +411,28 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
     }
 }
 
-impl<P: Clone, D> CallBuilder<&P, D> {
+impl<N: Network, T: Transport, P: Clone, D> CallBuilder<N, T, &P, D> {
```
gakonst, line 414: Confirming P: Clone bound OK

crates/contract/src/call.rs
```diff
@@ -454,10 +465,21 @@ impl<P, D: CallDecoder> std::fmt::Debug for CallBuilder<P, D> {
...
+    use alloy_rpc_client::RpcClient;
     use alloy_sol_types::sol;
+    use alloy_transport_http::Http;
+    use reqwest::Client;
+
+    fn spawn_anvil() -> (HttpProvider<Ethereum>, AnvilInstance) {
+        let anvil = Anvil::new().spawn();
+        let url = anvil.endpoint().parse().unwrap();
+        let http = Http::<Client>::new(url);
+        (RootProvider::<Ethereum, _>::new(RpcClient::new(http, true)), anvil)
```
gakonst, line 481: follow-on we need `Provider::<Ethereum>::connect(url)` which figures out which BoxTransport to use based on URL prefix ("HTTP://" vs "ws://" vs "ipc path")

crates/contract/src/instance.rs
```diff
@@ -54,25 +59,31 @@ impl<P> ContractInstance<P> {
...
     /// Returns a transaction builder for the provided function name.
     ///
     /// If there are multiple functions with the same name due to overloading, consider using
     /// the [`ContractInstance::function_from_selector`] method instead, since this will use the
     /// first match.
-    pub fn function(&self, name: &str, args: &[DynSolValue]) -> Result<CallBuilder<&P, Function>> {
+    pub fn function(
+        &self,
+        name: &str,
+        args: &[DynSolValue],
```
gakonst, line 85: Need follow-up to make this work ideally without the user having to do `DynSolValue::Address(addr)` etc. for each argument, ideally they can do `contract.function("foo", (a, b, c))` or similar instead of `contract.function("foo", &[DynSolValue::(...), DynSolValue::(...), ...)]`

crates/network/src/ethereum/builder.rs
```diff
@@ -0,0 +1,242 @@
...
+            (None, _, _, None, None, None) => build_1559(self).map(Into::into),
+            // EIP-4844
+            // All blob fields required
+            (None, _, _, Some(_), Some(_), Some(_)) => {
+                build_4844(self).map(TxEip4844Variant::from).map(Into::into)
+            }
+            _ => build_legacy(self).map(Into::into),
+        }
+    }
+
```
gakonst, line 128: art

crates/network/src/ethereum/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+mod signer;
+pub use signer::EthereumSigner;
+
+/// Types for a mainnet-like Ethereum network.
+#[derive(Debug, Clone, Copy)]
+pub struct Ethereum {
+    _private: (),
+}
+
+impl Network for Ethereum {
```
gakonst, line 14: art

crates/network/src/ethereum/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+    type ReceiptResponse = alloy_rpc_types::TransactionReceipt;
+
+    type HeaderResponse = alloy_rpc_types::Header;
+}
+
+impl ReceiptResponse for alloy_rpc_types::TransactionReceipt {
+    fn contract_address(&self) -> Option<alloy_primitives::Address> {
+        self.contract_address
+    }
+}
```
gakonst, line 36: i forget why we cannot put this on the main `TxReceipt` trait

crates/network/src/ethereum/signer.rs
```diff
@@ -0,0 +1,152 @@
...
+            }
+            assert_eq!(*tx, before);
+            Ok(sig)
+        }
+
+        async fn sign_dyn_tx_test(
+            tx: &mut dyn SignableTransaction<Signature>,
+            chain_id: Option<ChainId>,
+        ) -> Result<Signature> {
+            let mut wallet: alloy_signer::Wallet<k256::ecdsa::SigningKey> =
```
gakonst, line 93: This generic qualification required? Or can we default to it?

crates/network/src/lib.rs
gakonst: Ah I see, ignore my previous comment. Let's add a note about this in the docs

crates/network/src/transaction/builder.rs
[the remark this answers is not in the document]
gakonst: let's do in followup

crates/network/src/transaction/signer.rs
```diff
@@ -0,0 +1,132 @@
+use crate::Network;
+use alloy_consensus::SignableTransaction;
+use alloy_signer::{
+    k256::ecdsa::{self, signature::hazmat::PrehashSigner, RecoveryId},
+    Signature, Signer, SignerSync, Wallet,
+};
+use async_trait::async_trait;
+
+/// A signer capable of signing any transaction for the given network.
```
gakonst, line 9: Let's add a note on the difference between this and TxSigner below

crates/providers/src/builder.rs
```diff
@@ -67,11 +98,20 @@ impl<L, N> ProviderBuilder<L, N> {
...
     /// [`tower::ServiceBuilder`]: https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html
-
     pub fn layer<Inner>(self, layer: Inner) -> ProviderBuilder<Stack<Inner, L>> {
         ProviderBuilder { layer: Stack::new(layer, self.layer), network: PhantomData }
     }
 
+    /// Add a signer layer to the stack being built.
+    ///
+    /// See [`SignerLayer`].
+    pub fn signer<S>(self, signer: S) -> ProviderBuilder<Stack<SignerLayer<S>, L>> {
```
gakonst, line 108: maybe we can do `S: Into<SignerLayer>` here instead of `SignerLayer::new`

crates/providers/src/chain.rs
gakonst: agreed

crates/providers/src/lib.rs
```diff
@@ -16,21 +16,29 @@
...
 
 pub mod new;
-pub use new::{Provider, ProviderRef, RootProvider, WeakProvider};
 
-pub mod utils;
+#[doc(inline)]
+pub use new::{AnvilProvider, Provider, ProviderRef, RawProvider, RootProvider, WeakProvider};
 
-// TODO: remove
-pub mod tmp;
```
gakonst, line 36: yay

crates/providers/src/new.rs
```diff
@@ -70,6 +82,8 @@ impl<N, T> RootProviderInner<N, T> {
     }
 }
 
+// todo: adjust docs
+// todo: reorder
```
gakonst, line 86: reminder to open issue -- 

really like remco's thing here https://github.com/recmo/uint/blob/main/.github/scripts/issue_tracker.py & https://github.com/recmo/uint/blob/main/.github/workflows/issue_tracker.yml

crates/providers/src/new.rs
```diff
@@ -84,48 +98,63 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
-    /// TODO: block number/hash/tag
-    async fn get_transaction_count(&self, address: Address) -> TransportResult<U256> {
-        self.client().prepare("eth_getTransactionCount", (address, "latest")).await
+    /// Gets the transaction count of the corresponding address.
+    async fn get_transaction_count(
+        &self,
+        address: Address,
+        tag: Option<BlockId>,
+    ) -> TransportResult<alloy_primitives::U256> {
+        self.client().prepare("eth_getTransactionCount", (address, tag.unwrap_or_default())).await
```
gakonst, line 112: personally dont like `prepare` and would like to call it `request`, can do in follow-up

crates/providers/src/new.rs
```diff
@@ -138,8 +167,288 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+}
+
+/// Extension trait for Anvil specific JSON-RPC methods.
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+pub trait AnvilProvider<N: Network, T: Transport + Clone = BoxTransport>: Provider<N, T> {
+    /// Set the bytecode of a given account.
+    async fn set_code(&self, address: Address, code: &'static str) -> TransportResult<()> {
+        self.client().prepare("anvil_setCode", (address, code)).await
+    }
```
gakonst, line 429: todo add other anvil methods and prob could also go to other crate in the future, for set_storage, mine etc.

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -155,19 +155,6 @@ impl TransactionRequest {
         self.transaction_type = Some(U8::from(transaction_type));
         self
     }
-
-    /// Calculates the address that will be created by the transaction, if any.
-    ///
-    /// Returns `None` if the transaction is not a contract creation (the `to` field is set), or if
-    /// the `from` or `nonce` fields are not set.
-    pub fn calculate_create_address(&self) -> Option<Address> {
```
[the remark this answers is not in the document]
gakonst: builder trait should cover us here

crates/signer-aws/src/signer.rs
```diff
@@ -92,12 +93,39 @@ pub enum AwsSignerError {
...
+        };
+
+        let mut sig =
+            self.sign_hash(&tx.signature_hash()).await.map_err(alloy_signer::Error::other)?;
+
+        if let Some(chain_id) = chain_id.or_else(|| tx.chain_id()) {
+            sig = sig.with_chain_id(chain_id);
+        }
+        Ok(sig)
+    }
```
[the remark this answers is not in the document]
gakonst: agree follow on

crates/signer-ledger/src/signer.rs
```diff
@@ -27,10 +28,37 @@ pub struct LedgerSigner {
...
+    #[inline]
+    async fn sign_transaction(
+        &self,
+        tx: &mut dyn SignableTransaction<Signature>,
+    ) -> Result<Signature> {
+        let chain_id = match (self.chain_id(), tx.chain_id()) {
```
gakonst, line 39: agree w dani comment on macro to reduce code dup, or function

crates/signer-ledger/src/signer.rs
```diff
@@ -376,7 +392,7 @@ mod tests {
         test_sign_tx_generic(&mut tx).await;
     }
 
-    async fn test_sign_tx_generic(tx: &mut SignableTx) {
+    async fn test_sign_tx_generic(tx: &mut dyn SignableTransaction<Signature>) {
```
gakonst, line 395: do we have alias that doesnt require us to specify the generic?

crates/signer/src/signer.rs
```diff
@@ -129,58 +76,33 @@ pub trait Signer: Send + Sync {
...
-/// [`sign_transaction_sync`](Self::sign_transaction_sync), and to the resulting signature in all
-/// the methods. If `chain_id` is None, [EIP-155] should not be applied.
-///
 /// Synchronous signers should also implement [`Signer`], as they are always able to by delegating
 /// the asynchronous methods to the synchronous ones.
 ///
 /// [EIP-155]: https://eips.ethereum.org/EIPS/eip-155
 #[auto_impl(&, &mut, Box, Rc, Arc)]
-pub trait SignerSync {
+pub trait SignerSync<Sig: 'static = Signature> {
```
gakonst, line 84: confirming '`static` is fine?

crates/signer/src/wallet/private_key.rs
```diff
@@ -14,6 +14,19 @@ use std::str::FromStr;
...
 impl Wallet<SigningKey> {
+    /// Creates a new Wallet instance from a [`SigningKey`].
+    ///
+    /// This can also be used to create a Wallet from a [`SecretKey`](K256SecretKey).
+    /// See also the `From` implementations.
+    #[doc(alias = "from_private_key")]
+    #[doc(alias = "new_private_key")]
+    #[doc(alias = "new_pk")]
+    #[inline]
+    pub fn from_signing_key(signer: SigningKey) -> Self {
```
gakonst, line 25: Replace with `From<SigningKey>` and remove?

gakonst (approved):
LGTM. Excellent work @onbjerg & ty @prestwich for getting this off the ground, left a few follow-up PR requests. 

I also think we should add tests/examples for `tokio::spawn`'ing various things, in case we screwed up lifetimes somewhere and we cannot satisfy static requirements for task spawning.

I also think we should add a diagram with the types/traits flow across all our builders and show the snowball flow which eventually builds the final payload.

Bullish!!!!

crates/network/src/ethereum/mod.rs
onbjerg: Standard ethereum receipts do not have any tx sender field + nonce, so we can't recompute the created address. They also don't include this info, they only have logs/bloom/status/root

crates/providers/src/signer.rs
onbjerg: done

crates/consensus/src/signed.rs
onbjerg: because transactions can be signed with multiple types of signatures on some networks

crates/providers/src/new.rs
onbjerg: follow up

crates/signer-ledger/src/signer.rs
onbjerg: u cant forward declare them on traits

crates/signer/src/wallet/private_key.rs
onbjerg: follow up imo

onbjerg:
rebased ptal @gakonst @DaniPopes, after merge I will create all the follow up issues

onbjerg:
rebased ptal @gakonst @DaniPopes, after merge I will create all the follow up issues

crates/contract/src/call.rs
DaniPopes: it's necessary for the current implementation, and it can't be a supertrait because it would break object safety

crates/network/src/ethereum/signer.rs
DaniPopes: it's LocalWallet

crates/providers/src/new.rs
DaniPopes: And we don't :D https://discord.com/channels/900503503162724452/1049944297799168080/1212754890338865224

crates/signer-ledger/src/signer.rs
DaniPopes: `type SignableTx = dyn SignableTransaction<Signature>`
but it's a test

DaniPopes:
I realized we're not running doctests in CI because when I ran `cargo t --all-features` it failed on a doctest, @onbjerg can you please look into it? I couldn't figure it out

.github/workflows/ci.yml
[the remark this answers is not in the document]
onbjerg: https://github.com/alloy-rs/alloy/issues/262

Cargo.toml
onbjerg: will keep it until release

crates/contract/src/call.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/263

crates/contract/src/call.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/264

crates/network/src/lib.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/267

crates/network/src/transaction/builder.rs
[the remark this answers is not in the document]
onbjerg: https://github.com/alloy-rs/alloy/issues/268

crates/providers/src/new.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/269

crates/providers/src/new.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/270

crates/providers/src/new.rs
onbjerg: https://github.com/alloy-rs/alloy/issues/271

onbjerg:
created all the follow up issues, will look into doctest

onbjerg:
@DaniPopes we are, it just showed up later. i ran `cargo t --all-features` and got the same error as ci (which is fixed now)

crates/signer-ledger/src/signer.rs
DaniPopes: done with a macro

DaniPopes:
Yes, because I added it in https://github.com/alloy-rs/alloy/pull/190/commits/81c98a572c037eb438c3a328d5d5a3aa8901bf48 :smile:

PR #195, opened 2024-02-09
fix(`rpc-types`): do not deny additional fields

Evalir:
## Motivation

We were explicitly denying unknown fields on `TransactionRequest`.

## Solution

Support them, as we already use `OtherFields`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
yeah, this is not helpful

PR #209, opened 2024-02-13
fix: make l1_fee_scalar f64

0xZerohero:
## Motivation
Related to https://github.com/paradigmxyz/reth/issues/6547
In op-geth l1FeeScalar is represented as a float and is serialized to string
```go
FeeScalar  *big.Float `json:"l1FeeScalar,omitempty"` // always nil after Ecotone hardfork
```
```go
if receipt.FeeScalar != nil { // removed in Ecotone
	fields["l1FeeScalar"] = receipt.FeeScalar.String()
}
```

This introduces a breaking change in reth and need to be synchronized with https://github.com/paradigmxyz/reth/pull/6575

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/optimism.rs
+++ b/crates/rpc-types/src/eth/transaction/optimism.rs
@@ -86,6 +86,14 @@ mod tests {
     use super::*;
 
     #[test]
+    fn serialize_empty_optimism_transaction_receipt_fields_struct() {
+        let op_fields = OptimismTransactionReceiptFields::default();
+
+        let json = serde_json::to_value(op_fields).unwrap();
+        assert_eq!(json, json!({}));
+    }
+
+    #[test]
     fn serialize_l1_fee_scalar() {
         let op_fields = OptimismTransactionReceiptFields {
             l1_fee_scalar: Some(0.678),
--- a/crates/rpc-types/src/eth/transaction/optimism.rs
+++ b/crates/rpc-types/src/eth/transaction/optimism.rs
@@ -20,7 +20,7 @@ pub struct OptimismTransactionFields {
 }
 
 /// Additional fields for Optimism transaction receipts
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/transaction/optimism.rs
```diff
@@ -30,8 +30,8 @@ pub struct OptimismTransactionReceiptFields {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub l1_fee: Option<U256>,
     /// L1 fee scalar for the transaction
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub l1_fee_scalar: Option<U256>,
+    #[serde(default, skip_serializing_if = "Option::is_none", with = "l1_fee_scalar_serde")]
```
mattsse, line 33:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none", with = "l1_fee_scalar_serde")]
```

mattsse (approved):
lgtm, 1 nit

crates/rpc-types/src/eth/transaction/optimism.rs
0xZerohero: Without the serde(default) the deserialization works when the field is Value::Null, but fails when it's not set.
https://stackoverflow.com/a/44303505

```
thread 'eth::transaction::optimism::tests::deserialize_l1_fee_scalar' panicked at crates/rpc-types/src/eth/transaction/optimism.rs:126:88:
called `Result::unwrap()` on an `Err` value: Error("missing field `l1FeeScalar`", line: 0, column: 0)

```

Should I remove it ?

mattsse: TIL

PR #212, opened 2024-02-14
Add debug trace call

DoTheBestToGetTheBest:
added debug_traceCall method as part of the TempProvider trait

### DIFF
--- a/crates/providers/src/builder.rs
+++ b/crates/providers/src/builder.rs
@@ -1,7 +1,9 @@
-use crate::{NetworkRpcClient, Provider};
+use crate::{provider::TempProvider, NetworkRpcClient, Provider};
 use alloy_network::Network;
 use alloy_rpc_client::RpcClient;
-use alloy_transport::Transport;
+use alloy_rpc_trace_types::geth::{GethDebugTracingCallOptions, GethTrace};
+use alloy_rpc_types::{request::TransactionRequest, BlockId};
+use alloy_transport::{Transport, TransportResult};
 use std::marker::PhantomData;
 
 /// A layering abstraction in the vein of [`tower::Layer`]
@@ -115,6 +117,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
         N: Network,
     {
         self.layer.layer(provider)
+    }
+}
+
+/// Builder for constructing a debug trace call request.
+#[derive(Clone)]
+pub struct DebugTraceCallBuilder {
+    pub req: Option<TransactionRequest>,
+    pub block: Option<BlockId>,
+    pub trace_options: Option<GethDebugTracingCallOptions>,
+}
+
+impl DebugTraceCallBuilder {
+    /// Creates a new `DebugTraceCallBuilder`.
+    pub fn new() -> Self {
+        DebugTraceCallBuilder { req: None, block: None, trace_options: None }
+    }
+
+    /// Sets the transaction request for the builder.
+    pub fn request(mut self, req: TransactionRequest) -> Self {
+        self.req = Some(req);
+        self
+    }
+
+    /// Sets the block ID for the builder.
+    pub fn block(mut self, block: BlockId) -> Self {
+        self.block = Some(block);
+        self
+    }
+
+    /// Sets the tracing options for the builder.
+    pub fn trace_options(mut self, options: GethDebugTracingCallOptions) -> Self {
+        self.trace_options = Some(options);
+        self
+    }
+
+    /// Executes the debug trace call using the provided provider.
+    /// Panics if required fields (req and trace_options) are not set.
+    pub async fn execute<T: TempProvider>(&self, provider: &T) -> TransportResult<GethTrace> {
+        provider.debug_trace_call(self.clone()).await
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/providers/src/provider.rs
```diff
@@ -184,6 +184,13 @@ pub trait TempProvider: Send + Sync {
         keys: Vec<StorageKey>,
         block: Option<BlockId>,
     ) -> TransportResult<EIP1186AccountProofResponse>;
+    /// Inspect and debug transaction
+    async fn debug_trace_call(
+        &self,
+        req: TransactionRequest,
+        block: Option<BlockId>,
+        trace_options: GethDebugTracingCallOptions,
```
mattsse, line 192: this should be encapsulated in a helper, builder style struct, which would make calling this a bit easier

mattsse (changes requested):

crates/providers/src/provider.rs
DoTheBestToGetTheBest: > this should be encapsulated in a helper, builder style struct, which would make calling this a bit easier

interesting, so i should not follow how it is implemented in ethers-rs ?

 https://github.com/gakonst/ethers-rs/blob/6136ddeb7227df1ffaf792a7518d7e3a580eb573/ethers-providers/src/middleware.rs#L746

mattsse: I'm fine with this but it would be more convenient to a builder-style request object for this

DoTheBestToGetTheBest: > I'm fine with this but it would be more convenient to a builder-style request object for this

based, i work on builder pattern for this then ! thanks

DoTheBestToGetTheBest: > I'm fine with this but it would be more convenient to a builder-style request object for this

made changes

crates/providers/src/builder.rs
```diff
@@ -118,6 +120,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
     }
 }
 
+/// Builder for constructing a debug trace call request.
+#[derive(Clone)]
+pub struct DebugTraceCallBuilder {
+    pub req: Option<TransactionRequest>,
+    pub block: Option<BlockId>,
+    pub trace_options: Option<GethDebugTracingCallOptions>,
```
mattsse, line 128: this option is redundant

crates/providers/src/builder.rs
```diff
@@ -118,6 +120,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
     }
 }
 
+/// Builder for constructing a debug trace call request.
+#[derive(Clone)]
+pub struct DebugTraceCallBuilder {
+    pub req: Option<TransactionRequest>,
```
mattsse, line 126: I think this option is also redundant

crates/providers/src/builder.rs
```diff
@@ -118,6 +120,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
...
+    pub block: Option<BlockId>,
+    pub trace_options: Option<GethDebugTracingCallOptions>,
+}
+
+impl DebugTraceCallBuilder {
+    /// Creates a new `DebugTraceCallBuilder`.
+    pub fn new() -> Self {
+        DebugTraceCallBuilder { req: None, block: None, trace_options: None }
+    }
```
mattsse, line 135: can be removed, because there's already default

crates/providers/src/builder.rs
```diff
@@ -118,6 +120,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
...
+    pub fn trace_options(mut self, options: GethDebugTracingCallOptions) -> Self {
+        self.trace_options = Some(options);
+        self
+    }
+
+    /// Executes the debug trace call using the provided provider.
+    /// Panics if required fields (req and trace_options) are not set.
+    pub async fn execute<T: TempProvider>(&self, provider: &T) -> TransportResult<GethTrace> {
+        provider.debug_trace_call(self.clone()).await
+    }
```
mattsse, line 159: we don't want this here, this should stay in tmpprovider

crates/providers/src/builder.rs
```diff
@@ -118,6 +120,45 @@ impl<L, N, T> ProviderBuilder<L, N, T> {
...
+    pub fn request(mut self, req: TransactionRequest) -> Self {
+        self.req = Some(req);
+        self
+    }
+
+    /// Sets the block ID for the builder.
+    pub fn block(mut self, block: BlockId) -> Self {
+        self.block = Some(block);
+        self
+    }
```
mattsse, line 147: this should be impl Into<BlockId>

crates/providers/src/builder.rs
mattsse, line 1: this does not belong here, this is more like a rpc helper type

mattsse (changes requested):

crates/providers/src/builder.rs
DoTheBestToGetTheBest: > this does not belong here, this is more like a rpc helper type

hmm we dont have a  crate  helper type for rpc, what can i do instead ?

mattsse: I think we want to move this to the rpc types crate

PR #213, opened 2024-02-14
chore(`provider`): make `BlockId` opt on get_storage_at

Evalir:
## Motivation

This should be optional and default to latest.

## Solution

Make it optional.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/providers/src/provider.rs
+++ b/crates/providers/src/provider.rs
@@ -99,7 +99,7 @@ pub trait TempProvider: Send + Sync {
     ) -> TransportResult<StorageValue>;
 
     /// Gets the bytecode located at the corresponding [Address].
-    async fn get_code_at(&self, address: Address, tag: BlockId) -> TransportResult<Bytes>;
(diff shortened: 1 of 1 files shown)


crates/providers/src/provider.rs
```diff
@@ -673,13 +673,7 @@ mod tests {
...
         let addr = alloy_primitives::Address::with_last_byte(16);
         provider.set_code(addr, "0xbeef").await.unwrap();
-        let _code = provider
-            .get_code_at(
-                addr,
-                crate::provider::BlockId::Number(alloy_rpc_types::BlockNumberOrTag::Latest),
-            )
-            .await
-            .unwrap();
+        let _code = provider.get_code_at(addr, None).await.unwrap();
```
DaniPopes, line 676: can we add an assertion on the returned code here?

DaniPopes (approved):

PR #214, opened 2024-02-14
chore(`rpc-types`): add FromStr impl for BlockId

Evalir:
## Motivation

This was missing, and there were no tests for block id.

## Solution

Adds a `FromStr` impl for `BlockId` and some tests.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/block.rs
+++ b/crates/rpc-types/src/eth/block.rs
@@ -740,31 +740,32 @@ impl<'de> Deserialize<'de> for BlockId {
 
 /// Error thrown when parsing a [BlockId] from a string.
 #[derive(Debug, thiserror::Error)]
-#[error("failed to parse {input:?} as a number: {parse_int_error} or hash: {hex_error}")]
-pub struct ParseBlockIdError {
-    input: String,
-    parse_int_error: ParseIntError,
-    hex_error: alloy_primitives::hex::FromHexError,
+pub enum ParseBlockIdError {
+    /// Failed to parse a block id from a number.
+    #[error(transparent)]
+    ParseIntError(#[from] ParseIntError),
+    /// Failed to parse a block id as a hex string.
+    #[error(transparent)]
+    FromHexError(#[from] alloy_primitives::hex::FromHexError),
 }
 
 impl FromStr for BlockId {
     type Err = ParseBlockIdError;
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/block.rs
```diff
@@ -738,6 +738,30 @@ impl<'de> Deserialize<'de> for BlockId {
...
+    type Err = ParseBlockIdError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match u64::from_str(s) {
+            Ok(val) => Ok(BlockId::Number(BlockNumberOrTag::Number(val))),
+            Err(parse_int_error) => match B256::from_str(s) {
+                Ok(val) => Ok(BlockId::Hash(val.into())),
+                Err(hex_error) => {
+                    Err(ParseBlockIdError { input: s.to_string(), parse_int_error, hex_error })
+                }
+            },
```
mattsse, line 760: am I just missing it or does this not cover the tags? only hash and number

mattsse (changes requested):

crates/rpc-types/src/eth/block.rs
Evalir: yeah this is how ethers handles it—we can cover the tags too.

crates/rpc-types/src/eth/block.rs
```diff
@@ -738,6 +738,37 @@ impl<'de> Deserialize<'de> for BlockId {
...
+impl FromStr for BlockId {
+    type Err = ParseBlockIdError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "latest" => Ok(BlockId::Number(BlockNumberOrTag::Latest)),
+            "finalized" => Ok(BlockId::Number(BlockNumberOrTag::Finalized)),
+            "safe" => Ok(BlockId::Number(BlockNumberOrTag::Safe)),
+            "earliest" => Ok(BlockId::Number(BlockNumberOrTag::Earliest)),
+            "pending" => Ok(BlockId::Number(BlockNumberOrTag::Pending)),
+            _ => match u64::from_str(s) {
```
DaniPopes, line 759: can be simplified to `"0x... "=> u64, _ => B256`

DaniPopes (approved):

PR #217, opened 2024-02-15
feat: add some test to fee

DoTheBestToGetTheBest:


mattsse (changes requested):
I wanna keep the json tests, the others aren't very useful

PR #219, opened 2024-02-17
Move total_difficulty to Header

rodrigoherrerai:
Related to: [#6571](https://github.com/paradigmxyz/reth/issues/6571#issuecomment-1949224821) 

The total_difficulty value goes in the Header type in Geth. 

To keep compatibility in certain RPC requests (getHeaderByHash, getHeaderByNumber) we need to move the total_difficulty to the header instead of the block.


mattsse (approved):

mattsse:
moved to #220 because I can't push to your main branch

PR #221, opened 2024-02-18
suppport legacy implementation in CallBuilder

DoTheBestToGetTheBest:
implemented the `legacy` method in `CallBuilder` to allow switching from EIP-1559 tx to legacy tx

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -271,8 +271,9 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
     }
 
     /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
-    pub fn legacy(self) -> Self {
-        todo!()
+    pub fn to(mut self, to: Option<Address>) -> Self {
+        self.request = self.request.to(to);
+        self
     }
 
     /// Sets the `gas` field in the transaction to the provided value
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -271,8 +271,9 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
     }
 
     /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
-    pub fn to(mut self, to: Option<Address>) -> Self {
-        self.request = self.request.to(to);
+    pub fn legacy(mut self) -> Self {
+        self.request = self.request.max_fee_per_gas(U256::ZERO);
+        self.request = self.request.max_priority_fee_per_gas(U256::ZERO);
         self
     }
 


crates/contract/src/call.rs
```diff
@@ -271,8 +271,10 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
     }
 
     /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
-    pub fn legacy(self) -> Self {
-        todo!()
+    pub fn legacy(mut self) -> Self {
+        self.request = self.request.max_fee_per_gas(U256::ZERO);
+        self.request = self.request.max_priority_fee_per_gas(U256::ZERO);
```
mattsse, line 276: these should be set to None instead
this also needs to remove any blob fields

mattsse (changes requested):

crates/contract/src/call.rs
DoTheBestToGetTheBest: > these should be set to None instead this also needs to remove any blob fields

thanks!

crates/contract/src/call.rs
```diff
@@ -271,8 +271,13 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
     }
 
     /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
-    pub fn legacy(self) -> Self {
-        todo!()
+    pub fn legacy(mut self) -> Self {
+        self.request.max_fee_per_blob_gas = None;
```
onbjerg, line 275: unfortunately this does not add support for legacy transactions, `gas_price` still only sets EIP-1559 gas fields

onbjerg (changes requested):

crates/contract/src/call.rs
onbjerg: sorry - maybe I wasn't precise. the field `gas_price` is for legacy transactions, but the *method* `gas_price` only sets the fields `max_fee_per_gas` and `max_priority_fee_per_gas`. if the tx is a legacy tx *only* the `gas_price` field should be set, in all other cases `max_fee_per_gas` and `max_priority_fee_per_gas` should be set.

currently, if i do `tx.gas_price(..)` it sets the fields `max_fee_per_gas` and `max_priority_fee_per_gas`. the *field* `gas_price` is not set. if the tx is a legacy transaction, it should be. if it is not a legacy transaction, instead the `max_fee_*` fields should be set.

DoTheBestToGetTheBest: > sorry - maybe I wasn't precise. the field `gas_price` is for legacy transactions, but the _method_ `gas_price` only sets the fields `max_fee_per_gas` and `max_priority_fee_per_gas`. if the tx is a legacy tx _only_ the `gas_price` field should be set, in all other cases `max_fee_per_gas` and `max_priority_fee_per_gas` should be set.
> 
> currently, if i do `tx.gas_price(..)` it sets the fields `max_fee_per_gas` and `max_priority_fee_per_gas`. the _field_ `gas_price` is not set. if the tx is a legacy transaction, it should be. if it is not a legacy transaction, instead the `max_fee_*` fields should be set.

thank you so much for this explain! Does this changes looks better to you ?

crates/contract/src/call.rs
```diff
@@ -282,21 +281,20 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
         self
     }
 
-    /// Sets the `gas` field in the transaction to the provided value
-    pub fn gas(mut self, gas: U256) -> Self {
-        self.request = self.request.gas_limit(gas);
-        self
-    }
```
onbjerg: this should not be removed

crates/contract/src/call.rs
```diff
@@ -282,21 +281,20 @@ impl<P: TempProvider, D: CallDecoder> CallBuilder<P, D> {
...
-        self.request = self.request.max_fee_per_gas(gas_price);
-        self.request = self.request.max_priority_fee_per_gas(gas_price);
+        match self.request.max_fee_per_gas {
+            None => self.request.gas_price = Some(gas_price),
+
+            Some(_) => {
+                self.request.max_fee_per_gas = Some(gas_price);
+                self.request.max_priority_fee_per_gas = Some(gas_price);
+            }
+        }
```
onbjerg, line 301: with this change, we will never be able to set the EIP-1559 gas prices with this function, since `max_fee_per_gas` always starts out as `None`

onbjerg (changes requested):

crates/contract/src/call.rs
DoTheBestToGetTheBest: > ld not be removed

thanks for review; but actually it's not removed

crates/contract/src/call.rs
DoTheBestToGetTheBest: > with this change, we will never be able to set the EIP-1559 gas prices with this function, since `max_fee_per_gas` always starts out as `None`

thanks for your review, it should be max_fee_per_gas && max_priority_fee_per_gas ?

or i'm totally out of the context ?

onbjerg: I'm not sure I follow, can you try to rephrase it?

PR #227, opened 2024-02-21
feat: Add parent beacon block root into `ExecutionPayloadEnvelopeV3`

clabby:
## Overview

Adds the `parentBeaconBlockRoot` field to the `ExecutionPayloadEnvelopeV3`, required in the `Ecotone` hardfork. 

Reference: https://github.com/ethereum-optimism/op-geth/blob/70103aa6866c3d965f1f71154958cd9c1a7beb5b/beacon/engine/types.go#L102-L103


crates/rpc-engine-types/src/payload.rs
```diff
@@ -117,6 +117,9 @@ pub struct ExecutionPayloadEnvelopeV3 {
     /// Introduced in V3, this represents a suggestion from the execution layer if the payload
     /// should be used instead of an externally provided one.
     pub should_override_builder: bool,
+    /// Optimism-specific - Ecotone parent beacon block root
+    #[serde(skip_serializing_if = "Option::is_none")]
```
DaniPopes, line 121:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

DaniPopes (changes requested):

DaniPopes (approved):

Evalir (approved):

crates/rpc-engine-types/src/payload.rs
```diff
@@ -117,6 +117,9 @@ pub struct ExecutionPayloadEnvelopeV3 {
     /// Introduced in V3, this represents a suggestion from the execution layer if the payload
     /// should be used instead of an externally provided one.
     pub should_override_builder: bool,
+    /// Optimism-specific - Ecotone parent beacon block root
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub parent_beacon_block_root: Option<B256>,
```
mattsse, line 122: I'm fine with that for now, if you create a new reth issue for better rpc execution payload customization

cc @rjected

mattsse (approved):

mattsse:
@clabby to update reth

please bump pinned alloy commit in paradigm/emv-inspectors first, then bump both on reth

crates/rpc-engine-types/src/payload.rs
gakonst: is this fixed with the customized engine types in node builder?

mattsse: not yet https://github.com/paradigmxyz/reth/pull/6871

PR #228, opened 2024-02-22
 refactor: rename "blob_gas_price" to "blob_base_fee"

AbnerZheng:
## Motivation
According to https://github.com/paradigmxyz/reth/issues/6330, `blob_gas_price` should renamed to `blob_base_fee`. 

## Solution
A minor update.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


AbnerZheng:
@mattsse PTAL

mattsse (commented):
thanks, I need think about this a bit because

https://github.com/ethereum/go-ethereum/blob/b590cae89232299d54aac8aada88c66d00c5b34c/internal/ethapi/api.go#L1722-L1722

prestwich:
@mattsse is this still being considered?

mattsse:
the spec is blobGasPrice, which is in line with other `*Price` fields of the receipt, so we can close this

PR #229, opened 2024-02-22
[WIP] feat(`consensus`): add extra EIP-4844 types needed

Evalir:
## Motivation

We still need to add a few `EIP-4844` types to be able to properly send and decode full EIP-4844 txs with a sidecar. Right now we can't do that, and the `TxEip4844` contained in the envelope is incomplete, as it misses the sidecar.

## Solution

Add the needed types. Mostly ported from reth. WIP, needs tests added and proper encoding/decoding fns to be implemented yet.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -16,6 +16,9 @@ alloy-network.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-eips.workspace = true
+thiserror.workspace = true
+c-kzg = { version = "0.4.2", features = ["serde"] }
+sha2 = "0.10.7"
 
 # arbitrary
 arbitrary = { workspace = true, features = ["derive"], optional = true }
--- a/crates/consensus/src/transaction/eip4844.rs
+++ b/crates/consensus/src/transaction/eip4844.rs
@@ -3,11 +3,35 @@ use alloy_eips::{eip2930::AccessList, eip4844::DATA_GAS_PER_BLOB};
 use alloy_network::{Signed, Transaction};
 use alloy_primitives::{keccak256, Bytes, ChainId, Signature, B256, U256};
 use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable, Header};
-use std::mem;
+use c_kzg::{Blob, Bytes48, KzgCommitment, KzgProof, KzgSettings};
+use sha2::Digest;
+use std::{mem, ops::Deref};
+
+/// An error that can occur when validating a [BlobTransaction].
+#[derive(Debug, thiserror::Error)]
+pub enum BlobTransactionValidationError {
+    /// Proof validation failed.
+    #[error("invalid KZG proof")]
+    InvalidProof,
+    /// An error returned by [`kzg`].
+    #[error("KZG error: {0:?}")]
+    KZGError(#[from] c_kzg::Error),
+    /// The inner transaction is not a blob transaction.
+    #[error("unable to verify proof for non blob transaction: {0}")]
+    NotBlobTransaction(u8),
+    /// The versioned hash is incorrect.
+    #[error("wrong versioned hash: have {have}, expected {expected}")]
+    WrongVersionedHash {
+        /// The versioned hash we got
+        have: B256,
+        /// The versioned hash we expected
+        expected: B256,
+    },
+}
 
 /// [EIP-4844 Blob Transaction](https://eips.ethereum.org/EIPS/eip-4844#blob-transaction)
 ///
-/// A transaction with blob hashes and max blob fee
+/// A transaction with blob hashes and max blob fee. It does not have the Blob sidecar.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Default)]
 pub struct TxEip4844 {
     /// Added as EIP-pub 155: Simple replay attack protection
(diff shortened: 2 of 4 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -65,12 +65,15 @@ pub enum TxEnvelope {
...
     Eip2930(Signed<TxEip2930>),
-    /// A [`TxEip1559`].
+    /// A [`TxEip1559`] tagged with type 2.
     Eip1559(Signed<TxEip1559>),
-    /// A [`TxEip4844`].
+    /// A [`TxEip4844`] tagged with type 3.
     Eip4844(Signed<TxEip4844>),
+    /// A [`BlobTransaction`], which is a [`crate::TxEip4844`] with a
+    /// [`crate::BlobTransactionSidecar`].
+    Eip4844WithSidecar(Signed<BlobTransaction>),
```
mattsse, line 76: hmm, not sure we want two variants for 4844 

this 4844 type is kinda hard to handle, for sending txs we need the sidecar but for historical txs we don't have it.

but I assume we want both formats here
maybe we instead have an enum for 4844?

wdyt @prestwich

Evalir: See https://github.com/alloy-rs/alloy/pull/229/commits/9f433222fa2226cb722b064385cad6fd8ef269e6 for a enum implementation—feels cleaner to me.

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -3,13 +3,303 @@ use alloy_eips::{eip2930::AccessList, eip4844::DATA_GAS_PER_BLOB};
...
+/// A transaction with blob hashes and max blob fee.
+/// It can either be a standalone transaction, mainly seen when retrieving historical transactions,
+/// or a transaction with a sidecar, which is used when submitting a transaction to the network and
+/// when receiving and sending transactions during the gossip stage.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TxEip4844 {
+    /// A standalone transaction with blob hashes and max blob fee.
+    TxEip4844(BlobTx),
+    /// A transaction with a sidecar, which contains the blob data, commitments, and proofs.
+    TxEip4844WithSidecar((BlobTx, BlobTransactionSidecar)),
```
mattsse, line 43:
```suggestion
    TxEip4844WithSidecar((BlobTx, BlobTransactionSidecar)),
```
```suggestion
    TxEip4844WithSidecar(BlobTx, BlobTransactionSidecar),
```

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -3,13 +3,303 @@ use alloy_eips::{eip2930::AccessList, eip4844::DATA_GAS_PER_BLOB};
...
+        }
+    }
+}
+
+/// [EIP-4844 Blob Transaction](https://eips.ethereum.org/EIPS/eip-4844#blob-transaction)
+///
+/// A transaction with blob hashes and max blob fee. It does not have the Blob sidecar.
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Default)]
-pub struct TxEip4844 {
+pub struct BlobTx {
```
mattsse, line 302: this type should be called TxEip4844

crates/consensus/src/transaction/envelope.rs
```diff
@@ -65,11 +65,11 @@ pub enum TxEnvelope {
...
     /// A [`TxLegacy`] tagged with type 0.
     TaggedLegacy(Signed<TxLegacy>),
-    /// A [`TxEip2930`].
+    /// A [`TxEip2930`] tagged with type 1.
     Eip2930(Signed<TxEip2930>),
-    /// A [`TxEip1559`].
+    /// A [`TxEip1559`] tagged with type 2.
     Eip1559(Signed<TxEip1559>),
-    /// A [`TxEip4844`].
+    /// A [`TxEip4844`] tagged with type 3.
```
mattsse, line 72: this needs more docs about the format

crates/consensus/src/transaction/eip4844.rs
mattsse, line 43: I'd prefer a standalone type that has the tx + blob, the tuple here makes it harder to work with

mattsse (changes requested):
the enum approach is the most straight forward one imo.

though we still want an actual type for the 4844+blob variant

crates/consensus/src/transaction/eip4844.rs
Evalir: Yep sg. Not married to names personally—what should we call the enum?

crates/consensus/Cargo.toml
```diff
@@ -26,4 +29,5 @@ arbitrary = { workspace = true, features = ["derive"] }
 
 [features]
 k256 = ["alloy-primitives/k256", "alloy-network/k256"]
+kzg = ["c-kzg", "sha2", "thiserror"]
```
mattsse, line 32: this should be default

mattsse (changes requested):
the cfg on the type itself is not pretty.

I think we should try to limit kzg feature to validation.
we should be able to use primitive types for the blob fields instead and not kzg types here

crates/consensus/Cargo.toml
DaniPopes: Please also use `dep:` syntax

Evalir: note that making kzg default breaks the wasm check, and don't think we want to skip the check for consensus. should we have a diff cargo hack check for consensus only with no default features?

crates/eips/src/eip4844.rs
```diff
@@ -32,6 +34,21 @@ pub const BLOB_TX_MIN_BLOB_GASPRICE: u128 = 1u128;
 /// Commitment version of a KZG commitment
 pub const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
 
+/// How many bytes are in a blob
+pub const BYTES_PER_BLOB: usize = 131072;
```
mattsse, line 38: this is `DATA_GAS_PER_BLOB`

crates/consensus/Cargo.toml
mattsse: if we limit kzg to validation only then we don't need this as default

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -97,7 +404,67 @@ impl TxEip4844 {
         self.blob_versioned_hashes.len() as u64 * DATA_GAS_PER_BLOB
     }
 
-    /// Decodes the inner [TxEip4844] fields from RLP bytes.
+    #[cfg(feature = "kzg")]
```
mattsse, line 407: cfgs should be above the `fn`

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -1,13 +1,320 @@
...
+    /// Verifies that the transaction's blob data, commitments, and proofs are all valid.
+    ///
+    /// See also [TxEip4844::validate_blob]
+    pub fn validate(
+        &self,
+        proof_settings: &KzgSettings,
+    ) -> Result<(), BlobTransactionValidationError> {
+        match self {
+            TxEip4844Wrapper::TxEip4844(tx) => {
+                tx.validate_blob(&BlobTransactionSidecar::default(), proof_settings)
```
mattsse, line 82: we should have an error variant for this case instead

mattsse (changes requested):
last nits

now that we're limiting c-kzg to validation, we can make this no default :)

crates/eips/src/eip4844.rs
Evalir: I know—added an extra comment. I'll keep it as `DATA_GAS_PER_BLOB` is an u64, and i'd rather avoid casting everywhere we need this to be an usize.

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -30,6 +30,10 @@ pub enum BlobTransactionValidationError {
     /// The inner transaction is not a blob transaction.
     #[error("unable to verify proof for non blob transaction: {0}")]
     NotBlobTransaction(u8),
+    /// Using a standalone [TxEip4844] instead of the [TxEip4844WithSidecar] variant, which
+    /// includes the sidecar for validation.
+    #[error("eip4844 tx variant without sidecar being used for verification. Please use the TxEip4844WithSidecar variant, which includes the sidecar")]
+    NoSidecarIncluded,
```
mattsse, line 36:
```suggestion
    MissingSidecar,
```

mattsse (approved):
lgtm,

though maybe we can find a better name than `Wrapper`, maybe `Variant`?

pending @DaniPopes

mattsse:
@Evalir let's rename to `Variant` then send it

Evalir:
clippy failure unrelated.

PR #230, opened 2024-02-22
chore(alloy): re-export `alloy-core` items individually

DaniPopes:
Avoid glob re-export for better docs and source code

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,11 +40,11 @@ alloy-transport-http = { version = "0.1.0", default-features = false, path = "cr
 alloy-transport-ipc = { version = "0.1.0", default-features = false, path = "crates/transport-ipc" }
 alloy-transport-ws = { version = "0.1.0", default-features = false, path = "crates/transport-ws" }
 
-alloy-core = { version = "0.6", default-features = false, features = ["std"] }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -99,6 +99,7 @@ serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.8"
 
-# TODO: Remove once alloy-contract is stable. This is only used in tests for `sol!`.
-[patch.crates-io]
-alloy-sol-macro = { git = "https://github.com/alloy-rs/core", rev = "18b0509950c90d9ec38f25913b692ae4cdd6f227" }
+# TODO: Keep around until alloy-contract is stable.
+# This should only be used in `alloy-contract` tests.
+# [patch.crates-io]
+# alloy-sol-macro = { git = "https://github.com/alloy-rs/core", rev = "18b0509950c90d9ec38f25913b692ae4cdd6f227" }
```
Evalir, line 105: Should this be commented?

Evalir (approved):

Cargo.toml
DaniPopes: Yes, currently there are no changes but it should be re enabled for testing with network changes

PR #232, opened 2024-02-23
fix(consensus): correct TxEip1559.encode_with_signature

vbrvk:
## Motivation

Fixing TxEip1559 encoding, before fix node reply with "error code -32602: Failed to decode transaction" error

## Solution

The tx type byte was missing in `encode_with_signature.encode_with_signature` method

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -142,7 +142,10 @@ impl TxEip1559 {
         signature: &Signature,
         out: &mut dyn alloy_rlp::BufMut,
     ) {
-        self.encode_for_signing(out);
+        let payload_length = self.fields_len() + signature.rlp_vrs_len();
+        let header = Header { list: true, payload_length };
+        header.encode(out);
+        self.encode_fields(out);
         signature.write_rlp_vrs(out);
     }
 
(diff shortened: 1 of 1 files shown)


Evalir (changes requested):
To produce a valid tx you need to encode the tx using a `TxEnvelope`, which will produce an `EIP-2718` compliant tx and add the type byte which was missing.

vbrvk:
How can it be used over abstract Transaction trait?

I have next method, how can i use `TxEnvelope` there?

```rust
    pub async fn sign_tx<T: Transaction<Signature = Signature>>(&self, mut tx: T) -> Result<Bytes> {
        let signature = self.local.sign_transaction(&mut tx).await?;

        dbg!(hex::encode(&signature.as_bytes()));
        let signed = TxEnvelope::from(tx.into_signed(signature));
        Ok(signed.into())
    }
```

Evalir:
This is a good question—we're working on streamlining the flow from a `TransactionRequest` -> signing and converting to a `TxEnvelope` so it's easy to send transactions over the network with a provider, here: https://github.com/alloy-rs/alloy/pull/190 . Right now this flow is not very well polished and it's probably why you found this issue.

The method you posted looks fine—as long as you encode the tx using a tx envelope then this should be working.

PR #236, opened 2024-02-28
fix: map deserde error to ErrorResp if it is an error payload

mattsse:
currently all deserde failures are returned as `deser_error`

this adds a check for error payload if deserialization failed


DaniPopes (approved):

prestwich:
check if this logic error is duplicated in the other transports? Maybe make a convenience function in the jsonrpc crate for doing this conversion?

mattsse:
we could add it to `fn deser_err` because this already gets the text?

DaniPopes (approved):

PR #239, opened 2024-02-28
fix: use enveloped encoding for typed transactions

Rjected:
## Motivation
The encoding for `eth_sendRawTransaction` for typed transactions was not correct, because it was encoding a RLP header for typed transactions.

## Solution
This completely overhauls transaction encoding and decoding in alloy.

There is no longer `encode_signed` and `decode_signed` in `Transaction`. `Signed<T>` transactions cannot be encoded or decoded directly. Now, the main way to encode and decode transactions is through `Encodable2718` and `Decodable2718`, specifically through `TxEnvelope`.

The `TaggedLegacy` variant is removed.
`impl<T> std::ops::Deref for Signed<T>` is removed.

`impl From<Signed<TxEip4844WithSidecar>> for TxEnvelope` is added.
`impl From<Signed<TxEip4844>> for TxEnvelope` is added.

The 2718 trait methods now use `alloy_rlp::Result` instead of `Result<_, Eip2718Error>`

Raw RLP of a blob transaction with sidecar is added to `consensus/testdata`, and is used in a roundtrip test.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


Rjected:
fixing conflicts and adding blob tx with sidecar impl...

Evalir (commented):
Awesome! Agreed we need to iron these out—but I believe enveloped encoding is already handled by `TxEnvelope` & `Decodable2718`/`Encodable2718`? (see `envelope.rs` & `alloy_eips::eip2718`)

prestwich:
encode_signed and decode_signed probably should be documented as "dont call this, you want to call TxEnvelope methods instead"

crates/consensus/src/transaction/legacy.rs
```diff
@@ -91,11 +91,11 @@ impl TxLegacy {
         signature.write_rlp_vrs(out);
     }
 
-    /// Output the length of the RLP signed transaction encoding.
-    pub fn payload_len_with_signature(&self, signature: &Signature) -> usize {
+    /// Returns what the encoded length should be, if the transaction were RLP encoded with the
+    /// given signature.
+    pub(crate) fn encoded_len_with_signature(&self, signature: &Signature) -> usize {
```
prestwich, line 96: side note: i consider these all tech debt that is paid down by a refactor of rlp derives

Rjected: yes, rlp(flatten) etc will usher in a new era of derived and correct encodings

gakonst:
@Rjected Plans here? Flagging that it seems like https://github.com/foundry-rs/foundry/pull/7242#issuecomment-1974253356 is blocked on this

Rjected:
> @Rjected Plans here? Flagging that it seems like [foundry-rs/foundry#7242 (comment)](https://github.com/foundry-rs/foundry/pull/7242#issuecomment-1974253356) is blocked on this

I can get this rfr / done today to unblock

Rjected:
Marked this as RFR, the code is ready to review, i plan on adding more tests today

crates/consensus/src/transaction/mod.rs
```diff
@@ -100,18 +99,6 @@ pub trait SignableTransaction<Signature>: Transaction {
...
-
-    /// Encode with a signature. This encoding is usually RLP, but may be
-    /// different for future EIP-2718 transaction types.
-    fn encode_signed(&self, signature: &Signature, out: &mut dyn BufMut);
-
-    /// Decode a signed transaction. This decoding is usually RLP, but may be
-    /// different for future EIP-2718 transaction types.
-    ///
-    /// This MUST be the inverse of [`SignableTransaction::encode_signed`].
-    fn decode_signed(buf: &mut &[u8]) -> alloy_rlp::Result<Signed<Self, Signature>>
```
onbjerg, line 112: why was this removed? it's useful for converting from mempool rlp -> typed tx

crates/consensus/src/transaction/mod.rs
```diff
@@ -100,18 +99,6 @@ pub trait SignableTransaction<Signature>: Transaction {
     fn into_signed(self, signature: Signature) -> Signed<Self, Signature>
     where
         Self: Sized;
-
-    /// Encode with a signature. This encoding is usually RLP, but may be
-    /// different for future EIP-2718 transaction types.
-    fn encode_signed(&self, signature: &Signature, out: &mut dyn BufMut);
```
onbjerg, line 106: do we not need this?

crates/consensus/src/signed.rs
```diff
@@ -45,32 +41,6 @@ impl<T: SignableTransaction<Sig>, Sig> Signed<T, Sig> {
...
-
-    /// Produce the RLP encoded signed transaction.
-    pub fn rlp_signed(&self) -> Vec<u8> {
-        let mut buf = vec![];
-        self.encode_signed(&mut buf);
-        buf
-    }
-}
-
-impl<T: SignableTransaction<Sig>, Sig> alloy_rlp::Encodable for Signed<T, Sig> {
```
onbjerg, line 62: i assume these were replaced by manual impls? if so, we should also note to implementers of `SignedTransaction` that they need to provide these by themselves, since it is needed to convert mempool tx -> typed tx

another q here: is it even possible to impl this in an external crate because of orphan rule? if not, we need to keep this

crates/consensus/src/transaction/mod.rs
onbjerg: i see in the pr description that this is possible through TxEnvelope, could you give me a small example? we need to note this down for network crate authors, since it is likely a very important feature

crates/consensus/src/transaction/mod.rs
Rjected: `encode_signed` and `decode_signed` are not specific enough for eth transactions, it's ambiguous whether or not you want the network or RPC RLP format for transactions. Encodable2718 provides this distinction

onbjerg: ok, makes sense to me. need to document this, but can be a follow up, would create an issue

crates/consensus/src/signed.rs
onbjerg: i see we can't blanket impl this anymore

onbjerg (approved):
lgtm

prestwich (approved):

mattsse (approved):
rlp 👍

PR #242, opened 2024-03-01
feat: add OptimismExecutionPayloadV3

fgimenez:
## Motivation

Towards https://github.com/paradigmxyz/reth/pull/6871

## Solution

Adds specific type for `OptimismExecutionPayloadV3` to be used in `BuiltPayload`. Very similar to `ExecutionPayloadV3`, adds optimism-specific field `parent_beacon_block_root`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-engine-types/src/optimism.rs
+++ b/crates/rpc-engine-types/src/optimism.rs
@@ -1,5 +1,5 @@
-use crate::PayloadAttributes;
-use alloy_primitives::Bytes;
+use crate::{ExecutionPayloadV2, PayloadAttributes};
+use alloy_primitives::{Bytes, B256};
 use serde::{Deserialize, Serialize};
 
 /// Optimism Payload Attributes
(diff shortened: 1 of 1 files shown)


prestwich:
alloy project does not intend to house any non-Ethereum structs. this probably belongs in a separate repo

mattsse:
this should be in there for ease of use, at least for the time being

crates/rpc-engine-types/src/optimism.rs
```diff
@@ -24,6 +24,29 @@ pub struct OptimismPayloadAttributes {
     pub gas_limit: Option<u64>,
 }
 
+/// This structure maps on the ExecutionPayloadV3 structure of the beacon chain spec.
+///
+/// See also: <https://github.com/ethereum/execution-apis/blob/6709c2a795b707202e93c4f2867fa0bf2640a84f/src/engine/shanghai.md#executionpayloadv2>
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
```
mattsse, line 30: this should also reference the op spec:

https://github.com/ethereum-optimism/specs/blob/main/specs/protocol/exec-engine.md#engine_getpayloadv3

Evalir (approved):
lgtm codewise. as prestwich said, this should only be temporary—I think we should make an issue to track moving OP types to another repo once we have stability

prestwich:
@DaniPopes is the clippy relevant? otherwise good to merge

Evalir:
@prestwich nope, clippy is unrelated—see proptest fix: https://github.com/proptest-rs/proptest/pull/427

Evalir:
Tracking wrt OP types https://github.com/alloy-rs/alloy/issues/243

crates/rpc-engine-types/src/optimism.rs
fgimenez: > this should also reference the op spec:

sure, proposed here https://github.com/alloy-rs/alloy/pull/244

PR #245, opened 2024-03-04
feat: add Optimism execution payload envelope v3

fgimenez:
## Motivation

Towards https://github.com/paradigmxyz/reth/pull/6871

## Solution

Replaces the existing `OptimismExecutionPayloadV3` with `OptimismExecutionPayloadEnvelopeV3`, the latter is what will be used in `BuiltPayload`. 

Also removes the `parent_beacon_block_root` field from `ExecutionPayloadEnvelopeV3`, not needed anymore with the specific envelope types and the changes in https://github.com/paradigmxyz/reth/pull/6871 to use them.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-engine-types/src/payload.rs
```diff
@@ -13,7 +13,7 @@ use std::fmt;
 pub type ExecutionPayloadBodiesV1 = Vec<Option<ExecutionPayloadBodyV1>>;
 
 /// And 8-byte identifier for an execution payload.
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize, Hash, Default)]
```
mattsse, line 16: I think we want to prevent default here

crates/rpc-engine-types/src/payload.rs
```diff
@@ -987,7 +986,7 @@ mod tests {
...
-        let envelope: ExecutionPayloadEnvelopeV3 = serde_json::from_str(response).unwrap();
+        let envelope: OptimismExecutionPayloadEnvelopeV3 = serde_json::from_str(response).unwrap();
         assert_eq!(serde_json::to_string(&envelope).unwrap(), response);
     }
```
mattsse, line 991: this test case can be duplicated in optimims.rs and updated here (removing parentBeaconBlockRoot)

mattsse (changes requested):
smol nit

crates/rpc-engine-types/src/payload.rs
fgimenez: sure done

mattsse (approved):

PR #246, opened 2024-03-05
ci: bump rust version to 1.76 in ci

fenghaojiang:
## Motivation
Bump rust version to 1.76 to pass the cargo clippy ci problems

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


DaniPopes:
This is not right, the clippy job is failing because of a rustc lint that has to be manually allowed. Clippy runs on nightly anyway.

fenghaojiang:
> This is not right, the clippy job is failing because of a rustc lint that has to be manually allowed. Clippy runs on nightly anyway.

Sry for my mistake.

PR #250, opened 2024-03-05
Feat: 4844 SidecarBuilder

prestwich:
## Motivation

There is currently no convenient way to construct valid sidecars from data

## Solution

Add a sidecar builder that ingests arbitrary data, blobs it, and builds into a sidecar

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip4844/builder.rs
+++ b/crates/consensus/src/transaction/eip4844/builder.rs
@@ -62,7 +62,7 @@ impl PartialSidecar {
         self.fe == 0
     }
 
-    /// Push an empty blob to the builder, and reset the unused counter.
+    /// Push an empty blob to the builder.
     fn push_empty_blob(&mut self) {
         self.blobs.push(Blob::new([0u8; BYTES_PER_BLOB]));
     }
@@ -80,12 +80,8 @@ impl PartialSidecar {
     }
 
     /// Get the index of the first unused field element in the current blob.
-    const fn first_unused_fe_index_in_current_blob(&self) -> Option<usize> {
-        if self.fe_in_current_blob() as u64 == FIELD_ELEMENTS_PER_BLOB {
-            None
-        } else {
-            Some(self.fe_in_current_blob())
-        }
+    const fn first_unused_fe_index_in_current_blob(&self) -> usize {
+        self.fe_in_current_blob()
     }
 
     /// Get a mutable reference to the current blob.
@@ -101,10 +97,7 @@ impl PartialSidecar {
 
     /// Get a mutable reference to the next unused field element.
     fn next_unused_fe_mut(&mut self) -> &mut [u8] {
-        if self.first_unused_fe_index_in_current_blob().is_none() {
-            self.push_empty_blob();
-        }
-        self.fe_at_mut(self.first_unused_fe_index_in_current_blob().expect(""))
+        self.fe_at_mut(self.first_unused_fe_index_in_current_blob())
     }
 
     /// Ingest a field element into the current blobs.
@@ -113,6 +106,7 @@ impl PartialSidecar {
     ///
     /// If there are not enough free FEs to encode the data.
     pub fn ingest_valid_fe(&mut self, data: WholeFe<'_>) {
+        self.alloc_fes(1);
         self.next_unused_fe_mut().copy_from_slice(data.as_ref());
         self.fe += 1;
     }
--- a/crates/consensus/src/transaction/eip4844/builder.rs
+++ b/crates/consensus/src/transaction/eip4844/builder.rs
@@ -322,22 +322,14 @@ impl<T: SidecarCoder> SidecarBuilder<T> {
         self,
         settings: &KzgSettings,
     ) -> Result<crate::BlobTransactionSidecar, c_kzg::Error> {
-        let commitments = self
-            .inner
-            .blobs
-            .iter()
-            .map(|blob| KzgCommitment::blob_to_kzg_commitment(blob, settings).map(|c| c.to_bytes()))
-            .collect::<Result<Vec<_>, _>>()?;
-
-        let proofs = self
-            .inner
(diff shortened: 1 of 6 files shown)


crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,408 @@
...
+
+impl SimpleCoder {
+    /// Decode an some bytes from an iterator of valid FEs.
+    ///
+    /// Returns `Ok(Some(data))` if there is some data.
+    /// Returns `Ok(None)` if there is no data (length prefix is 0).
+    /// Returns `Err(())` if there is an error.
+    fn decode_one<'a>(mut fes: impl Iterator<Item = WholeFe<'a>>) -> Result<Option<Vec<u8>>, ()> {
+        let first = fes.next().ok_or(())?;
+        let mut num_bytes = u64::from_be_bytes(first.as_ref()[1..9].try_into().unwrap()) as usize;
```
Rjected, line 197: would be nice to have a method on `WholeFe` that does the array ref conversion, documented with assumptions about how `WholeFe` is constructed

crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,408 @@
...
+    }
+
+    /// Get the number of used field elements in the current blob.
+    const fn fe_in_current_blob(&self) -> usize {
+        self.fe % FIELD_ELEMENTS_PER_BLOB as usize
+    }
+
+    /// Get the index of the first unused field element in the current blob.
+    const fn first_unused_fe_index_in_current_blob(&self) -> Option<usize> {
+        if self.fe_in_current_blob() as u64 == FIELD_ELEMENTS_PER_BLOB {
```
Rjected, line 84: is this ever the case? since the method is `self.fe % FIELD_ELEMENTS_PER_BLOB`

Rjected (changes requested):
looks generally good, the API makes sense to me, but need a few questions answered before I can lgtm

crates/consensus/src/transaction/eip4844/builder.rs
prestwich: correct. this was leftover from an earlier design. removing

crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,390 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PartialSidecar {
+    /// Create a new builder.
+    pub fn new() -> Self {
+        // NB: vecs default to 100 capacity. Blobs are large. We don't want
+        // to allocate 100 blobs if we don't need them.
```
mattsse, line 31: This is not the case, `Vec::new()` does not allocate capacity

crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,390 @@
...
+        Self::with_capacity(2)
+    }
+
+    /// Create a new builder with a given capacity.
+    pub fn with_capacity(capacity: usize) -> Self {
+        let mut blobs = Vec::with_capacity(capacity);
+        blobs.push(Blob::new([0u8; BYTES_PER_BLOB]));
```
mattsse, line 38: this also adds a new blob, 
looking at the other function, I think this is intended

crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,390 @@
...
+/// - If the data is not a multiple of 31 bytes, the last word is right-padded with zeros.
+///
+/// This means that the following regions cannot be used to store data, and are
+/// considered "wasted":
+///
+/// - The first byte of every 32-byte word.
+/// - The right padding on the header word containing the data length.
+/// - Any right padding on the last word for each piece of data.
+#[derive(Debug, Copy, Clone, Default)]
+pub struct SimpleCoder;
```
mattsse, line 177:
```suggestion
#[non_exhaustive]
pub struct SimpleCoder;
```

mattsse (commented):
blob math looks kinda cursed -.-

did not look at the math

crates/consensus/src/transaction/eip4844/builder.rs
DaniPopes: and on the first push it will allocate for 1: https://github.com/rust-lang/rust/blob/3cbb93223f33024db464a4df27a13c7cce870173/library/alloc/src/raw_vec.rs#L130-L141

crates/consensus/src/transaction/eip4844/builder.rs
prestwich: yeah, the intention is to never allow the blobs vec to be empty

prestwich: i've improved instantiation quite a bit now :)

crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -0,0 +1,406 @@
...
+impl Default for PartialSidecar {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PartialSidecar {
+    /// Create a new builder, and push an empty blob to it. This is the default
+    /// constructor, and allocates space for 2 blobs (256 KiB). If you need more
+    /// space, use [`PartialSidecar::with_capacity`].
```
Rjected, line 30: > ```
> If you need more space, use [`PartialSidecar::with_capacity`].
> ```
I would word this as `if you want to pre-allocate capacity`, since `ingest_valid_fe` and other pushing ops would increase the capacity anyways

Rjected (approved):
LGTM, left a note on with_capacity docs

PR #252, opened 2024-03-06
feat(rpc-trace-types): add support for mux tracer

ArtificialPB:
## Motivation
Part of mux tracer implementation, see: https://github.com/paradigmxyz/evm-inspectors/issues/49

## Solution
Add new Geth built-in tracer type: `muxTracer`. Allows executing multiple tracers in a single call.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes (new enum variants)

### DIFF
--- a/crates/rpc-trace-types/src/geth/mod.rs
+++ b/crates/rpc-trace-types/src/geth/mod.rs
@@ -1,6 +1,7 @@
 #![allow(missing_docs)]
 //! Geth tracing types
 
+use crate::geth::mux::{MuxConfig, MuxFrame};
 use alloy_primitives::{Bytes, B256, U256};
 use alloy_rpc_types::{state::StateOverride, BlockOverrides};
 use serde::{de::DeserializeOwned, ser::SerializeMap, Deserialize, Serialize, Serializer};
@@ -19,6 +20,7 @@ pub use self::{
 
 pub mod call;
 pub mod four_byte;
+pub mod mux;
 pub mod noop;
 pub mod pre_state;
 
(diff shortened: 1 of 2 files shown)


crates/rpc-trace-types/src/geth/mux.rs
```diff
@@ -0,0 +1,120 @@
+use crate::geth::{GethDebugBuiltInTracerType, GethDebugTracerConfig, GethTrace};
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+/// A `muxTracer` config that contains the configuration for running multiple tracers in one go.
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct MuxConfig {
+    #[serde(flatten)]
+    pub tracers: HashMap<GethDebugBuiltInTracerType, Option<GethDebugTracerConfig>>,
```
mattsse, line 10: can we also use tuple new type here like MuxFrame

crates/rpc-trace-types/src/geth/mod.rs
```diff
@@ -240,6 +253,13 @@ impl GethDebugTracerConfig {
         }
         self.from_value()
     }
+
```
mattsse, line 256: +one line doc

mattsse (changes requested):
ty, smol nits

crates/rpc-trace-types/src/geth/mux.rs
ArtificialPB: yea, good idea

crates/rpc-trace-types/src/geth/mod.rs
ArtificialPB: done

mattsse (approved):

PR #254, opened 2024-03-07
Convert non-200 http responses into errors

### LINKED ISSUE #253: [Bug] Trying to decode non 200 status code payloads as result object.
### Component

network, json-rpc, rpc

rev 76c70fb9

### Operating System

Linux

### Describe the bug

When hitting a rate limit on a reth node. Was getting the error: Err(DeserError { err: Error("expected value", line: 1, column: 1). This was very missleading as the message was `text: "Too many connections. Please try again later."` that comes with a 429 status code. This should be a ErrorResp afaik.

```
 err=Err(DeserError { err: Error("expected value", line: 1, column: 1), text: "Too many connections. Please try again later." })
````

```rust
#[derive(Debug, Clone)]
pub struct LocalProvider {
    provider: Arc<Provider<Http<reqwest::Client>>>,
    retries:  u8,
}

impl LocalProvider {
    pub fn new(url: String, retries: u8) -> Self {
        let http = Http::new(url.parse().unwrap());

        Self { provider: Arc::new(Provider::new(http)), retries }
    }
}

#[async_trait::async_trait]
impl TracingProvider for LocalProvider {
    async fn eth_call(
        &self,
        request: TransactionRequest,
        block_number: Option<BlockId>,
        state_overrides: Option<StateOverride>,
        block_overrides: Option<Box<BlockOverrides>>,
    ) -> eyre::Result<Bytes> {
        if state_overrides.is_some() || block_overrides.is_some() {
            panic!("local provider doesn't support block or state overrides");
        }
        // for tests, shit can get beefy 
        let mut attempts = 0;
        loop {
            let inner = self.provider.inner();
            let res_str: String = inner
                .prepare("eth_call", (request.clone(), block_number))
                .await
                .unwrap();
            let res = self.provider.call(request.clone(), block_number).await;
            if res.is_ok() || attempts > self.retries {
                return res.map_err(Into::into)
            } else {
                // error message was generated here
                tracing::error!(eth_call_error=%res_str, ?request, err=?res); 
            }
            attempts += 1
        }
    }
}
```

prestwich:
Currently non-200 responses proceed to JSON deser. This causes them to be treated as `TransportError::custom` variants.

Closes #253 

## Motivation

Prevent server errors from being treated as JSON

## Solution

Capture the errors

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/src/hyper.rs
+++ b/crates/transport-http/src/hyper.rs
@@ -30,6 +30,8 @@ where
 
             let resp = this.client.request(req).await.map_err(TransportErrorKind::custom)?;
 
+            let status = resp.status();
+
             // unpack data from the response body. We do this regardless of
             // the status code, as we want to return the error in the body if
             // there is one.
@@ -37,11 +39,10 @@ where
                 .await
                 .map_err(TransportErrorKind::custom)?;
 
-            if resp.status() != hyper::StatusCode::OK {
+            if status != hyper::StatusCode::OK {
                 return Err(TransportErrorKind::custom_str(&format!(
                     "HTTP error: {} with body: {:?}",
-                    resp.status(),
-                    body
+                    status, body
                 )));
             }
 
--- a/crates/transport-http/src/hyper.rs
+++ b/crates/transport-http/src/hyper.rs
@@ -30,7 +30,7 @@ where
 
             let resp = this.client.request(req).await.map_err(TransportErrorKind::custom)?;
 
-            // unpack json from the response body. We do this regardless of
+            // unpack data from the response body. We do this regardless of
             // the status code, as we want to return the error in the body if
             // there is one.
             let body = hyper::body::to_bytes(resp.into_body())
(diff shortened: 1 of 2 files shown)


crates/transport-http/src/hyper.rs
```diff
@@ -30,11 +30,22 @@ where
...
+            // the status code, as we want to return the error in the body if
+            // there is one.
             let body = hyper::body::to_bytes(resp.into_body())
                 .await
                 .map_err(TransportErrorKind::custom)?;
 
+            if status != hyper::StatusCode::OK {
+                return Err(TransportErrorKind::custom_str(&format!(
+                    "HTTP error: {} with body: {:?}",
+                    status, body
```
mattsse, line 45: perhaps we get the respons as text if not Ok:

if !=Ok use body.text() which sanitizes non utf8 iirc

else we get the raw body

mattsse (changes requested):

crates/transport-http/src/hyper.rs
mattsse: oh I guess this is raw hyper not reqwest
I believe reqwest does utf8 decoding on `.text()`

we could convert it to utf8 lossy perhaps, not sure how debug formats here

PR #259, opened 2024-03-10
Move `alloy-rpc-types` `serde_helpers` mod to standalone crate `alloy-serde`

moricho:
## Motivation
Closes: https://github.com/alloy-rs/alloy/issues/123

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/serde/src/lib.rs
```diff
@@ -1,4 +1,4 @@
-//! Serde helpers for primitive types.
+//! Alloy serde helpers for primitive types.
```
mattsse, line 2: this needs the regular alloy lib.rs header:

https://github.com/alloy-rs/alloy/blob/b370cf390e88f893711b1dedd90dd7654846e73a/crates/rpc-types/src/lib.rs#L6-L20

mattsse (changes requested):
I'd like this, the rpc-types crate is already quite big and if we add network specific types, a standalone this crate will be useful

wdyt @DaniPopes

crates/serde/src/lib.rs
moricho: Fixed at 👉  https://github.com/alloy-rs/alloy/pull/259/commits/3c784c4440361c1c871cdd2e5e680389c8fafb9c

mattsse (approved):

PR #274, opened 2024-03-11
feat(providers): event, polling and streaming methods

DaniPopes:
TODO: streams

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -393,13 +393,13 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, D: CallDecoder> CallBu
     }
 
     /// Broadcasts the underlying transaction to the network.
-    // TODO: more docs referring to customizing PendingTransaction
+    ///
+    /// See [`Provider::send_transaction`] for more information.
     pub async fn send(
         &self,
-    ) -> Result<impl IntoFuture<Output = Result<Option<N::ReceiptResponse>>> + '_> {
-        let pending = self.provider.send_transaction(self.request.clone()).await?;
-
-        Ok(pending.and_then(|hash| self.provider.get_transaction_receipt(hash)).map_err(Into::into))
+    ) -> Result<impl Future<Output = Result<Option<N::ReceiptResponse>>> + '_> {
+        let config = self.provider.send_transaction(self.request.clone()).await?;
+        Ok(config.with_provider(&self.provider).get_receipt().map_err(Into::into))
     }
 
     /// Calculates the address that will be created by the transaction, if any.
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -193,7 +193,8 @@ pub struct CallBuilder<N: Network, T, P, D> {
     request: N::TransactionRequest,
     block: Option<BlockId>,
     state: Option<StateOverride>,
-    provider: P,
+    /// The provider.
+    pub provider: P,
     decoder: D,
     transport: PhantomData<T>,
 }
@@ -428,6 +429,13 @@ impl<N: Network, T: Transport, P: Clone, D> CallBuilder<N, T, &P, D> {
 /// [`CallBuilder`] can be turned into a [`Future`] automatically with `.await`.
 ///
 /// Defaults to calling [`CallBuilder::call`].
+///
+/// # Note
+///
+/// This requires `Self: 'static` due to a current limitation in the Rust type system, namely that
+/// the associated future type, the returned future, must be a concrete type (`Box<dyn Future ...>`)
+/// and cannot be an opaque type (`impl Future ...`) because `impl Trait` in this position is not
+/// stable yet. See [rust-lang/rust#63063](https://github.com/rust-lang/rust/issues/63063).
 impl<N, T, P, D> IntoFuture for CallBuilder<N, T, P, D>
 where
     N: Network,
@@ -554,7 +562,6 @@ mod tests {
         );
     }
 
-    // TODO: send_transaction, PendingTransaction
     #[tokio::test(flavor = "multi_thread")]
     async fn deploy_and_call() {
         let (provider, anvil) = spawn_anvil();
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -1,11 +1,11 @@
+use crate::Error;
 use alloy_providers::{new::FilterPollerBuilder, Network, Provider};
 use alloy_rpc_types::{Filter, Log};
 use alloy_sol_types::SolEvent;
 use alloy_transport::{Transport, TransportResult};
+use futures::Stream;
 use futures_util::StreamExt;
 use std::{fmt, marker::PhantomData};
-
-use crate::Error;
 
 /// Helper for managing the event filter before querying or streaming its logs
 #[must_use = "event filters do nothing unless you `query`, `watch`, or `stream` them"]
@@ -49,19 +49,17 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, E: SolEvent> Event<N, 
     /// Watches for events that match the filter.
     ///
     /// Returns a stream of decoded events and raw logs.
-    pub async fn watch(&self) -> TransportResult<EventPoller<(), E>> {
-        // let poller = self.provider.watch_logs(&self.filter).await?;
-        // Ok(EventPoller::new(poller))
-        todo!()
+    pub async fn watch(&self) -> TransportResult<EventPoller<T, E>> {
+        let poller = self.provider.watch_logs(&self.filter).await?;
+        Ok(EventPoller::new(poller))
     }
 
     /// Watches for events that match the filter.
     ///
     /// Returns a stream of decoded events and raw logs.
-    pub async fn subscribe(&self) -> TransportResult<()> {
-        // let poller = self.provider.watch_logs(&self.filter).await?;
-        // Ok(EventPoller::new(poller))
-        todo!()
+    pub async fn subscribe(&self) -> TransportResult<EventPoller<T, E>> {
+        let poller = self.provider.watch_logs(&self.filter).await?;
+        Ok(EventPoller::new(poller))
     }
 }
 
@@ -81,6 +79,22 @@ pub struct EventPoller<T, E> {
     _phantom: PhantomData<E>,
 }
 
+impl<T, E> std::ops::Deref for EventPoller<T, E> {
+    type Target = FilterPollerBuilder<T, Log>;
+
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        &self.poller
+    }
+}
+
+impl<T, E> std::ops::DerefMut for EventPoller<T, E> {
+    #[inline]
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.poller
+    }
+}
+
 impl<T: fmt::Debug, E> fmt::Debug for EventPoller<T, E> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("EventPoller")
@@ -93,14 +107,16 @@ impl<T: fmt::Debug, E> fmt::Debug for EventPoller<T, E> {
 impl<T: Transport + Clone, E: SolEvent> EventPoller<T, E> {
     /// Creates a new event poller with the provided filter poller.
     #[allow(clippy::missing_const_for_fn)]
+    #[inline]
     pub fn new(poller: FilterPollerBuilder<T, Log>) -> Self {
         Self { poller, _phantom: PhantomData }
     }
 
-    /// Converts the event poller into a stream that yields the decoded event and the raw log.
-    pub fn into_stream(self) -> impl futures::Stream<Item = alloy_sol_types::Result<(E, Log)>> {
+    /// Starts the poller and returns a stream that yields the decoded event and the raw log.
+    ///
+    /// Note that this stream will not return `None` until the provider is dropped.
+    pub fn into_stream(self) -> impl Stream<Item = alloy_sol_types::Result<(E, Log)>> + Unpin {
         self.poller
-            .spawn()
             .into_stream()
             .flat_map(futures_util::stream::iter)
             .map(|log| decode_log(&log).map(|e| (e, log)))
@@ -109,4 +125,60 @@ impl<T: Transport + Clone, E: SolEvent> EventPoller<T, E> {
 
 fn decode_log<E: SolEvent>(log: &Log) -> alloy_sol_types::Result<E> {
     E::decode_raw_log(log.topics.iter().copied(), &log.data, false)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloy_primitives::U256;
+    use alloy_sol_types::sol;
+    use test_utils::{init_tracing, spawn_anvil};
+
+    sol! {
+        // solc v0.8.24; solc a.sol --via-ir --optimize --bin
+        #[sol(rpc, bytecode = "608080604052346100155760be908161001a8239f35b5f80fdfe60808060405260043610156011575f80fd5b5f3560e01c63299d8665146023575f80fd5b346084575f3660031901126084577f4e4cd44610926680098f1b54e2bdd1fb952659144c471173bbb9cf966af3a98860a0826060602a9452600560608201526468656c6c6f60d81b60808201526001602082015263deadbeef6040820152a2005b5f80fdfea2646970667358221220664e55b832143354f058e35b8948668184da14fcc5bf3300afb39dc6188c9add64736f6c63430008180033")]
+        contract MyContract {
+            #[derive(Debug, PartialEq)]
+            event MyEvent(uint64 indexed, string, bool, bytes32);
+
+            function doEmit() external {
+                emit MyEvent(42, "hello", true, bytes32(uint256(0xdeadbeef)));
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn manual_event() {
+        init_tracing();
+        let (provider, _anvil) = spawn_anvil();
+
+        let contract = MyContract::deploy(&provider).await.unwrap();
+        let event: Event<_, _, _, MyContract::MyEvent> = Event::new(&provider, Filter::new());
+        let all = event.query().await.unwrap();
+        assert_eq!(all.len(), 0);
+
+        let poller = event.watch().await.unwrap();
+
+        let _receipt = contract.doEmit().send().await.unwrap().await.unwrap().expect("no receipt");
+        let expected_event = MyContract::MyEvent {
+            _0: 42,
+            _1: "hello".to_string(),
+            _2: true,
+            _3: U256::from(0xdeadbeefu64).into(),
+        };
+
+        let mut stream = poller.into_stream();
+        let (stream_event, stream_log) = stream.next().await.unwrap().unwrap();
+        assert_eq!(stream_event, expected_event);
+        assert_eq!(stream_log.address, *contract.address());
+        assert_eq!(stream_log.block_number, Some(U256::from(2)));
+
+        // This is not going to return `None`
+        // assert!(stream.next().await.is_none());
+
+        let all = event.query().await.unwrap();
+        assert_eq!(all.len(), 1);
+        assert_eq!(all[0].0, expected_event);
+        assert_eq!(all[0].1, stream_log);
+    }
 }
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -49,15 +49,9 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, E: SolEvent> Event<N, 
     /// Watches for events that match the filter.
     ///
     /// Returns a stream of decoded events and raw logs.
+    #[doc(alias = "stream")]
+    #[doc(alias = "stream_with_meta")]
     pub async fn watch(&self) -> TransportResult<EventPoller<T, E>> {
-        let poller = self.provider.watch_logs(&self.filter).await?;
-        Ok(EventPoller::new(poller))
-    }
-
-    /// Watches for events that match the filter.
-    ///
-    /// Returns a stream of decoded events and raw logs.
-    pub async fn subscribe(&self) -> TransportResult<EventPoller<T, E>> {
         let poller = self.provider.watch_logs(&self.filter).await?;
         Ok(EventPoller::new(poller))
     }
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -1,4 +1,4 @@
-use alloy_providers::{new::FilterPoller, Network, Provider};
+use alloy_providers::{new::FilterPollerBuilder, Network, Provider};
 use alloy_rpc_types::{Filter, Log};
 use alloy_sol_types::SolEvent;
 use alloy_transport::{Transport, TransportResult};
@@ -77,7 +77,7 @@ impl<N, T, P: Clone, E> Event<N, T, &P, E> {
 /// Polling configuration is available through the [`poller`](Self::poller) field.
 pub struct EventPoller<T, E> {
     /// The inner poller.
-    pub poller: FilterPoller<T, Log>,
+    pub poller: FilterPollerBuilder<T, Log>,
     _phantom: PhantomData<E>,
 }
 
@@ -93,7 +93,7 @@ impl<T: fmt::Debug, E> fmt::Debug for EventPoller<T, E> {
 impl<T: Transport + Clone, E: SolEvent> EventPoller<T, E> {
     /// Creates a new event poller with the provided filter poller.
     #[allow(clippy::missing_const_for_fn)]
-    pub fn new(poller: FilterPoller<T, Log>) -> Self {
+    pub fn new(poller: FilterPollerBuilder<T, Log>) -> Self {
         Self { poller, _phantom: PhantomData }
     }
 
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -1,0 +1,112 @@
+use alloy_providers::{new::FilterPoller, Network, Provider};
+use alloy_rpc_types::{Filter, Log};
+use alloy_sol_types::SolEvent;
+use alloy_transport::{Transport, TransportResult};
+use futures_util::StreamExt;
+use std::{fmt, marker::PhantomData};
+
+use crate::Error;
+
+/// Helper for managing the event filter before querying or streaming its logs
+#[must_use = "event filters do nothing unless you `query`, `watch`, or `stream` them"]
+pub struct Event<N, T, P, E> {
+    /// The provider to use for querying or streaming logs.
+    pub provider: P,
+    /// The filter to use for querying or streaming logs.
+    pub filter: Filter,
+    _phantom: PhantomData<(T, N, E)>,
+}
+
+impl<N, T, P: fmt::Debug, E> fmt::Debug for Event<N, T, P, E> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Event")
+            .field("provider", &self.provider)
+            .field("filter", &self.filter)
+            .field("event_type", &format_args!("{}", std::any::type_name::<E>()))
+            .finish()
+    }
+}
+
+impl<N: Network, T: Transport + Clone, P: Provider<N, T>, E: SolEvent> Event<N, T, P, E> {
+    /// Creates a new event with the provided provider and filter.
+    #[allow(clippy::missing_const_for_fn)]
+    pub fn new(provider: P, filter: Filter) -> Self {
+        Self { provider, filter, _phantom: PhantomData }
+    }
+
+    /// Queries the blockchain for the selected filter and returns a vector of matching event logs.
+    pub async fn query(&self) -> Result<Vec<(E, Log)>, Error> {
+        let logs = self.query_raw().await?;
+        logs.into_iter().map(|log| Ok((decode_log(&log)?, log))).collect()
+    }
+
+    /// Queries the blockchain for the selected filter and returns a vector of matching event logs,
+    /// without decoding them.
+    pub async fn query_raw(&self) -> TransportResult<Vec<Log>> {
+        self.provider.get_logs(&self.filter).await
+    }
+
+    /// Watches for events that match the filter.
+    ///
+    /// Returns a stream of decoded events and raw logs.
+    pub async fn watch(&self) -> TransportResult<EventPoller<(), E>> {
+        // let poller = self.provider.watch_logs(&self.filter).await?;
+        // Ok(EventPoller::new(poller))
+        todo!()
+    }
+
+    /// Watches for events that match the filter.
+    ///
+    /// Returns a stream of decoded events and raw logs.
+    pub async fn subscribe(&self) -> TransportResult<()> {
+        // let poller = self.provider.watch_logs(&self.filter).await?;
+        // Ok(EventPoller::new(poller))
+        todo!()
+    }
+}
+
+impl<N, T, P: Clone, E> Event<N, T, &P, E> {
+    /// Clones the provider and returns a new event with the cloned provider.
+    pub fn with_cloned_provider(self) -> Event<N, T, P, E> {
+        Event { provider: self.provider.clone(), filter: self.filter, _phantom: PhantomData }
+    }
+}
+
+/// An event poller.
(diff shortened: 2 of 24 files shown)


crates/providers/src/heart.rs
```diff
@@ -16,8 +19,150 @@ use tokio::{
...
+    /// Registers the watching configuration with the provider.
+    ///
+    /// This does not wait for the transaction to be confirmed, but returns a [`PendingTransaction`]
+    /// that can be awaited at a later moment.
+    ///
+    /// See:
+    /// - [`watch`](Self::watch) for watching the transaction without fetching the receipt.
+    /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
+    ///   confirmed.
+    pub async fn register(self) -> TransportResult<PendingTransaction> {
```
DaniPopes, line 118: register -> PendingTransaction
wait -> PendingTransaction.await -> B256
get_receipt -> PendingTransaction.await.await -> Option<Receipt> (turn option into an error?)

Not super sure about names, open to suggestions to the API as well

crates/providers/src/heart.rs
```diff
@@ -16,8 +19,150 @@ use tokio::{
...
+    }
+}
+
+/// A configuration object for watching for transaction confirmation.
+///
+/// This object is not directly usable, but can be used to create a [`PendingTransactionConfig`]
+/// to watch for a transaction.
+#[must_use = "this type does nothing unless you call `with_provider`"]
+#[derive(Debug)]
+pub struct PendingTransactionConfigInner {
```
DaniPopes, line 165: `send_transaction*` APIS currently return the Inner struct because we cannot use `&Self` as the `P: Provider` generic in the return type of a `Provider` method due to weird constraints with either object safety or `auto_impl` (or both).

This means that there's an extra `with_provider` step:
```rust
provider
.send_transaction(tx)
.await?
.with_provider(&provider) // here
.wait() // or one of the other methods in `PendingTransactionConfig`
.await?;
```

crates/providers/src/new.rs
```diff
@@ -30,24 +30,46 @@ pub type WeakProvider<P> = Weak<P>;
 /// A borrowed [`Provider`].
 pub type ProviderRef<'a, P> = &'a P;
 
+/// A task that polls the provider with `eth_getFilterChanges`, returning a list of `R`.
+///
+/// See [`PollerBuilder`] for more details.
+pub type FilterPollerBuilder<T, R> = PollerBuilder<T, (U256,), Vec<R>>;
```
DaniPopes, line 36: Java flashbacks

crates/providers/src/new.rs
```diff
@@ -96,7 +130,87 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
     fn weak_client(&self) -> WeakClient<T>;
 
-    async fn new_pending_transaction(&self, tx_hash: B256) -> TransportResult<PendingTransaction>;
+    /// Watch for the confirmation of a single pending transaction with the given configuration.
+    ///
+    /// Note that this is handled internally rather than calling any specific RPC method.
+    async fn watch_pending_transaction(
+        &self,
+        config: PendingTransactionConfigInner,
+    ) -> TransportResult<PendingTransaction>;
```
DaniPopes, line 139: This is the internal method used by the `PendingTransactionConfig` type which uses the internal heartbeat

crates/providers/src/new.rs
```diff
@@ -96,7 +130,87 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+    /// See also [`watch_logs`](Self::watch_logs) to configure a poller.
+    async fn new_filter(&self, filter: &Filter) -> TransportResult<U256> {
+        self.client().prepare("eth_newFilter", (filter,)).await
+    }
+
+    /// Get a list of values that have been added since the last poll.
+    ///
+    /// The return value depends on what stream `id` corresponds to.
+    /// See [`FilterChanges`] for all possible return values.
+    #[auto_impl(keep_default_for(&, &mut, Rc, Arc, Box))]
```
DaniPopes, line 172: TIL this is necessary for `dyn Provider` to work, because otherwise `auto_impl` puts `U: Provider + Sized` bounds on the generic implementations if `Self: Sized` methods are not marked with `keep_default_for`. I added a test for this in this file

crates/providers/src/new.rs
onbjerg: image

crates/providers/src/heart.rs
onbjerg: what's the issue with auto_impl? otherwise we should prob consider implementing provider for &P manually, this API is not ideal

crates/node-bindings/src/anvil.rs
```diff
@@ -20,14 +20,24 @@ const ANVIL_STARTUP_TIMEOUT_MILLIS: u64 = 10_000;
...
+    child: Child,
     private_keys: Vec<K256SecretKey>,
     addresses: Vec<Address>,
     port: u16,
     chain_id: Option<u64>,
 }
 
 impl AnvilInstance {
+    /// Returns a reference to the child process.
+    pub fn child(&self) -> &Child {
```
onbjerg, line 32: can we spin these changes out into a separate pr?

crates/contract/src/instance.rs
```diff
@@ -97,6 +99,11 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>> ContractInstance<N, T,
         let function = self.interface.get_from_selector(selector)?;
         CallBuilder::new_dyn(&self.provider, function, args)
     }
+
+    /// Returns an [`Event`] builder with the provided filter.
+    pub fn event<E: SolEvent>(&self, filter: Filter) -> Event<N, T, &P, E> {
+        Event::new(&self.provider, filter)
```
onbjerg, line 105: art

onbjerg:
looks ossome

crates/providers/src/new.rs
```diff
@@ -482,34 +600,58 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProviderInner<N, T
...
-    use reqwest::Client;
 
-    struct _ObjectSafe<N: Network>(dyn Provider<N>);
+    extern crate self as alloy_providers;
 
-    fn init_tracing() {
-        let _ = tracing_subscriber::fmt::try_init();
-    }
+    // NOTE: We cannot import the test-utils crate here due to a circular dependency.
+    include!("../../internal-test-utils/src/providers.rs");
```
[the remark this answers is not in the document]
DaniPopes: I think it's because it's a unit test

crates/providers/src/new.rs
```diff
@@ -150,22 +261,26 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
-    /// transaction has been confirmed.
+    /// Broadcasts a transaction to the network.
+    ///
+    /// Returns a type that can be used to configure how and when to await the transaction's
+    /// confirmation.
     async fn send_transaction(
         &self,
         tx: N::TransactionRequest,
-    ) -> TransportResult<PendingTransaction> {
+    ) -> TransportResult<PendingTransactionConfigInner> {
```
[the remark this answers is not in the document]
DaniPopes: needs better naming, the inner struct is held internally, the non-inner one is for better send_transaction API (so you can chain .watch() etc)

crates/providers/src/heart.rs
DaniPopes: Discussion here https://discord.com/channels/900503503162724452/1216893060747886673

crates/node-bindings/src/anvil.rs
DaniPopes: Will do later

crates/providers/src/heart.rs
```diff
@@ -1,23 +1,203 @@
...
+///     .watch()
+///     .await?;
+/// # Ok(())
+/// # }
+/// ```
+#[must_use = "this type does nothing unless you call `register`, `watch` or `get_receipt`"]
 #[derive(Debug)]
-pub struct WatchConfig {
+pub struct PendingTransactionBuilder<N, T, P> {
+    inner: PendingTransactionConfig,
```
gakonst, line 59:
```suggestion
    config: PendingTransactionConfig,
```

crates/providers/src/heart.rs
```diff
@@ -1,23 +1,203 @@
...
+    ///
+    /// This does not wait for the transaction to be confirmed, but returns a [`PendingTransaction`]
+    /// that can be awaited at a later moment.
+    ///
+    /// See:
+    /// - [`watch`](Self::watch) for watching the transaction without fetching the receipt.
+    /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
+    ///   confirmed.
+    #[doc(alias = "build")]
+    pub async fn register(self) -> TransportResult<PendingTransaction> {
```
gakonst, line 153: Feel like this should be the default thing that `send()` returns (which is registered in the heartbeat task)

and `.await`ing it returns the confirmed tx (whether successful or not)

and `.with_receipt().await` returns `(tx, receipt)`

crates/rpc-client/src/client.rs
```diff
@@ -60,21 +76,33 @@ impl<T: Transport> RpcClient<T> {
...
+    ///
+    /// This will create a new client if this instance is not the only reference to the inner
+    /// client.
+    pub fn boxed(self) -> RpcClient<BoxTransport> {
+        let inner = match Arc::try_unwrap(self.0) {
+            Ok(inner) => inner,
+            // TODO: `id` is discarded.
+            Err(inner) => RpcClientInner::new(inner.transport.clone(), inner.is_local),
+        };
+        RpcClient::from_inner(inner.boxed())
```
gakonst, line 107: nit

```suggestion
        match Arc::try_unwrap(self.0) {
            Ok(inner) => RpcClient::from_inner(inner.boxed()),
            // TODO: `id` is discarded.
            Err(inner) => RpcClientInner::new(inner.transport.clone(), inner.is_local),
        }
```

crates/rpc-client/src/client.rs
```diff
@@ -118,15 +146,44 @@ pub struct RpcClientInner<T> {
 
 impl<T> RpcClientInner<T> {
     /// Create a new [`RpcClient`] with the given transport.
+    #[inline]
     pub const fn new(t: T, is_local: bool) -> Self {
         Self { transport: t, is_local, id: AtomicU64::new(0) }
     }
 
+    /// Returns the default poll interval for the client.
+    pub fn default_poll_interval(&self) -> Duration {
```
gakonst: we could also get a blocktime hint here by querying 2 blocks and subtracting their timestamps

crates/rpc-client/src/client.rs
```diff
@@ -175,6 +233,7 @@ impl<T: Transport + Clone> RpcClientInner<T> {
     /// Serialization is done lazily. It will not be performed until the call
     /// is awaited. This means that if a serializer error occurs, it will not
     /// be caught until the call is awaited.
+    #[doc(alias = "request")]
```
gakonst, line 238: ah nice

crates/rpc-client/src/poller.rs
```diff
@@ -15,14 +16,22 @@ use tracing::Instrument;
...
+/// sends the responses to a channel. By default, this is done every 10 seconds, with a channel size
+/// of 16, and no limit on the number of successful polls. This is all configurable.
+///
+/// The builder is consumed using the [`spawn`](Self::spawn) method, which returns a channel to
+/// receive the responses. The task will continue to poll until either the client or the channel is
+/// dropped.
+///
+/// The channel can be converted into a stream using the [`into_stream`](PollChannel::into_stream)
+/// method.
+// TODO: make this be able to be spawned on the current thread instead of forcing a task.
```
gakonst, line 56: why does that matter

crates/providers/src/heart.rs
```diff
@@ -16,8 +19,150 @@ use tokio::{
...
+    /// Sets the timeout.
+    pub fn set_timeout(&mut self, timeout: Option<Duration>) {
+        self.inner.set_timeout(timeout);
+    }
+
+    /// Sets the timeout.
+    pub fn with_timeout(mut self, timeout: Option<Duration>) -> Self {
+        self.set_timeout(timeout);
+        self
+    }
```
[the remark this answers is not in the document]
DaniPopes: Renamed to PendingTransactionBuilder

crates/rpc-client/src/poller.rs
```diff
@@ -156,6 +167,15 @@ where
         fut.instrument(span).spawn_task();
         rx.into()
     }
+
+    /// Starts the poller and returns the stream of responses.
+    ///
+    /// Note that this is currently equivalent to `self.spawn().into_stream()`, but this may change
+    /// in the future.
+    // TODO: can we name this type? This should be a different type from `PollChannel::into_stream`
```
gakonst, line 200: Yeah let's wrap it in a newtype cc @mattsse

crates/providers/src/new.rs
```diff
@@ -96,7 +130,87 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+
+    /// Watch for new logs using the given filter by polling the provider with
+    /// [`eth_getFilterChanges`](Self::get_filter_changes).
+    ///
+    /// Returns a builder that is used to configure the poller. See [`PollerBuilder`] for more
+    /// details.
+    async fn watch_logs(&self, filter: &Filter) -> TransportResult<FilterPollerBuilder<T, Log>> {
+        let id = self.new_filter(filter).await?;
+        Ok(PollerBuilder::new(self.weak_client(), "eth_getFilterChanges", (id,)))
+    }
```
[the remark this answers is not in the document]
DaniPopes: Same name as ethers, and they're mentioned in the respective `get_filter` function.
I've also moved them to the top, and added examples.

crates/rpc-client/src/poller.rs
```diff
@@ -204,7 +224,14 @@ where
     }
 
     /// Convert the poll channel into a stream.
-    pub fn into_stream(self) -> BroadcastStream<Resp> {
+    // TODO: can we name this type?
+    pub fn into_stream(self) -> impl Stream<Item = Resp> + Unpin {
```
gakonst, line 253: yep as above

crates/providers/src/new.rs
```diff
@@ -96,7 +123,148 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+    /// # Examples
+    ///
+    /// Get the next 5 blocks:
+    ///
+    /// ```no_run
+    /// # async fn example<N: alloy_network::Network>(provider: impl alloy_providers::Provider<N>) -> Result<(), Box<dyn std::error::Error>> {
+    /// use futures::StreamExt;
+    ///
+    /// let poller = provider.watch_blocks().await?;
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
```
gakonst, line 149: why is this flat_map needed?

crates/providers/src/new.rs
```diff
@@ -96,7 +123,148 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+    /// # Examples
+    ///
+    /// Get the next 5 pending transactions:
+    ///
+    /// ```no_run
+    /// # async fn example<N: alloy_network::Network>(provider: impl alloy_providers::Provider<N>) -> Result<(), Box<dyn std::error::Error>> {
+    /// use futures::StreamExt;
+    ///
+    /// let poller = provider.watch_pending_transactions().await?;
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
```
gakonst, line 176: wonder if it's worth making the watch_pending_transactions future always do `into_stream`? what else would we use it for

crates/providers/src/new.rs
```diff
@@ -96,7 +123,148 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
+    }
+
+    /// Notify the provider that we are interested in new blocks.
+    ///
+    /// Returns the ID to use with [`eth_getFilterChanges`](Self::get_filter_changes).
+    ///
+    /// See also [`watch_pending_transactions`](Self::watch_pending_transactions) to configure a
+    /// poller.
+    async fn new_pending_transactions_filter(&self) -> TransportResult<U256> {
+        self.client().prepare("eth_newPendingTransactionFilter", ()).await
```
gakonst, line 237: @mattsse now that we have `full: bool` as 2nd argument to this let's add `new_pending_transactions_filter_full`?  or something?

crates/providers/src/new.rs
```diff
@@ -108,13 +276,12 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
         tag: Option<BlockId>,
-    ) -> TransportResult<alloy_primitives::U256> {
+    ) -> TransportResult<U256> {
         self.client().prepare("eth_getTransactionCount", (address, tag.unwrap_or_default())).await
     }
 
     /// Get a block by its number.
-    ///
-    /// TODO: Network associate
+    // TODO: Network associate
```
gakonst, line 284: not a blocker

crates/providers/src/new.rs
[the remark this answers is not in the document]
gakonst: uh OK with this but lets leave a follow on

crates/providers/src/new.rs
```diff
@@ -482,34 +673,58 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProviderInner<N, T
...
+        let boxed = provider.boxed();
+        let num = boxed.get_block_number().await.unwrap();
+        assert_eq!(0, num);
+
+        let boxed_boxdyn = Box::new(boxed) as Box<dyn Provider<Ethereum>>;
+        let num = boxed_boxdyn.get_block_number().await.unwrap();
+        assert_eq!(0, num);
     }
 
     #[tokio::test]
```
gakonst, line 749: we'll want to also test out if we can spawn the tx / events subscription stream

crates/providers/src/new.rs
```diff
@@ -524,9 +739,23 @@ mod tests {
             gas: Some(U256::from(21000)),
             ..Default::default()
         };
+
+        let pending_tx = provider.send_transaction(tx.clone()).await.expect("failed to send tx");
+        let hash1 = *pending_tx.tx_hash();
```
gakonst, line 763: Deref + tx_hash() seems weird, would make this work with just derefing (bc it's convenient) or with `hash()`

gakonst (commented):
gg so far so good

crates/contract/src/event.rs
```diff
@@ -0,0 +1,191 @@
...
+        self.poller
+            .into_stream()
+            .flat_map(futures_util::stream::iter)
+            .map(|log| decode_log(&log).map(|e| (e, log)))
+    }
+}
+
+fn decode_log<E: SolEvent>(log: &Log) -> alloy_sol_types::Result<E> {
+    E::decode_raw_log(log.topics.iter().copied(), &log.data, false)
+}
```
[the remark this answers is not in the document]
DaniPopes: That's `Log`

crates/contract/src/call.rs
```diff
@@ -193,7 +193,8 @@ pub struct CallBuilder<N: Network, T, P, D> {
     request: N::TransactionRequest,
     block: Option<BlockId>,
     state: Option<StateOverride>,
-    provider: P,
+    /// The provider.
+    pub provider: P,
```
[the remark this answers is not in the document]
DaniPopes: Need it exposed for `sol!` to move out of it

crates/contract/src/call.rs
```diff
@@ -393,13 +398,13 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, D: CallDecoder> CallBu
...
+    /// See [`Provider::send_transaction`] for more information.
     pub async fn send(
         &self,
-    ) -> Result<impl IntoFuture<Output = Result<Option<N::ReceiptResponse>>> + '_> {
-        let pending = self.provider.send_transaction(self.request.clone()).await?;
-
-        Ok(pending.and_then(|hash| self.provider.get_transaction_receipt(hash)).map_err(Into::into))
+    ) -> Result<impl Future<Output = Result<Option<N::ReceiptResponse>>> + '_> {
+        let config = self.provider.send_transaction(self.request.clone()).await?;
+        Ok(config.with_provider(&self.provider).get_receipt().map_err(Into::into))
```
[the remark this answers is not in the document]
DaniPopes: maybe `send_raw` does juts `self.provider.send_transaction`, and `send` resolves to contract address?

crates/providers/src/heart.rs
```diff
@@ -1,23 +1,203 @@
...
+/// ```no_run
+/// # async fn example<N: alloy_network::Network>(provider: impl alloy_providers::Provider<N>, tx: N::TransactionRequest) -> Result<(), Box<dyn std::error::Error>> {
+/// // Send a transaction, and configure the pending transaction.
+/// let builder = provider.send_transaction(tx)
+///     .await?
+///     .with_confirmations(2)
+///     .with_timeout(Some(std::time::Duration::from_secs(60)));
+/// # let builder = builder.with_provider(&provider); // TODO
+/// // Register the pending transaction with the provider.
+/// let pending_transaction = builder.register().await?;
```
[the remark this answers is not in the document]
DaniPopes: `watch` does `register().await?.await`, we need to expose both because creating a pending transaction is async too due to use of tokio channels, vs this step not existing in ethers

crates/providers/src/new.rs
DaniPopes: eth_getFilterChanges returns a Vec

crates/rpc-client/src/poller.rs
DaniPopes: this crate depends on tokio only for this spawn+channel, could be an option also to do everything in the current thread

crates/providers/src/new.rs
DaniPopes: it returns a `PollerBuilder` to configure polling interval etc.

crates/providers/src/new.rs
DaniPopes: There is no network block RPC type

crates/providers/src/new.rs
DaniPopes: `impl Deref` is an anti-pattern and implicit copies make me sad, my opinion will not change on this :smiley:

crates/provider/src/new.rs
```diff
@@ -91,13 +121,171 @@ impl<N, T> RootProviderInner<N, T> {
 #[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
 #[auto_impl::auto_impl(&, &mut, Rc, Arc, Box)]
 pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync {
+    /// Returns the root provider.
+    fn root(&self) -> &RootProvider<N, T>;
```
onbjerg, line 125: what's the downside of this? when would you use a custom provider as your root provider?

if we go with this, we should update the `ProviderBuilder` to not accept anything other than `RootProvider` as the final provider (i.e. prob just remove `ProviderBuilder::provider`)

crates/rpc-client/src/client.rs
```diff
@@ -118,15 +148,44 @@ pub struct RpcClientInner<T> {
...
 impl<T> RpcClientInner<T> {
     /// Create a new [`RpcClient`] with the given transport.
+    #[inline]
     pub const fn new(t: T, is_local: bool) -> Self {
         Self { transport: t, is_local, id: AtomicU64::new(0) }
     }
 
+    /// Returns the default poll interval for the client.
+    pub const fn default_poll_interval(&self) -> Duration {
+        if self.is_local {
```
onbjerg, line 158: should make this configurable eventually, follow up material

onbjerg (commented):
also to be clear, this only uses polling and not pubsub?

crates/provider/src/new.rs
```diff
@@ -348,9 +360,9 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
     async fn send_transaction(
         &self,
         tx: N::TransactionRequest,
-    ) -> TransportResult<PendingTransactionConfig> {
+    ) -> TransportResult<PendingTransactionBuilder<'_, N, T>> {
```
mattsse, line 437: can we make pending tx not borrowing?
this makes it impossible to move the pending tx somewhere else

mattsse (changes requested):
I think the pending tx type should take a clone

crates/provider/src/new.rs
mattsse: ah I may have misunderstood this, this is only the first step, this can be converted into an owned by calling register etc..

DaniPopes: yep this is just a local builder

crates/provider/src/new.rs
```diff
@@ -37,18 +43,64 @@ pub struct RootProvider<N, T> {
...
+        self.pubsub_frontend()?.get_subscription(id).await.map(Subscription::from)
+    }
+
+    /// Unsubscribes from the subscription corresponding to the given RPC subscription ID.
+    #[cfg(feature = "pubsub")]
+    pub fn unsubscribe(&self, id: U256) -> TransportResult<()> {
+        self.pubsub_frontend()?.unsubscribe(id)
+    }
+
+    #[cfg(feature = "pubsub")]
```
onbjerg, line 90: lets document this in the crate readme

crates/provider/src/new.rs
```diff
@@ -485,12 +778,20 @@ pub trait RawProvider<N: Network, T: Transport + Clone = BoxTransport>: Provider
     /// Sends a raw JSON-RPC request.
     async fn raw_request<P, R>(&self, method: &'static str, params: P) -> TransportResult<R>
     where
-        P: Serialize + Send + Sync + Clone,
-        R: Serialize + DeserializeOwned + Send + Sync + Unpin + 'static,
-        Self: Sync,
+        P: RpcParam,
```
onbjerg, line 781: right

onbjerg (approved):
generally like this, although we need to figure out a way to make transport layering work again. i think let's get this in, think on it, and if it's not possible, maybe we can move most use cases into provider layers? (retry logic etc)

crates/providers/src/new.rs
gakonst: heh fair, not a strong one on my end, just don't love the `*x.foo()`, but yeah i hear the implicit copy point

crates/provider/src/heart.rs
```diff
@@ -15,9 +17,179 @@ use tokio::{
...
+///
+/// This can also be more concisely written using `watch`:
+/// ```no_run
+/// # async fn example<N: alloy_network::Network>(provider: impl alloy_provider::Provider<N>, tx: N::TransactionRequest) -> Result<(), Box<dyn std::error::Error>> {
+/// let tx_hash = provider.send_transaction(tx)
+///     .await?
+///     .with_confirmations(2)
+///     .with_timeout(Some(std::time::Duration::from_secs(60)))
+///     .watch()
+///     .await?;
```
gakonst, line 49: OK I am fine with this API -- let's open issue for potentially auto-registering if possible, but we can move forward.

crates/provider/src/new.rs
```diff
@@ -37,18 +43,64 @@ pub struct RootProvider<N, T> {
...
+    #[cfg(feature = "pubsub")]
+    fn pubsub_frontend(&self) -> TransportResult<&PubSubFrontend> {
+        let t = self.transport() as &dyn std::any::Any;
+        t.downcast_ref::<PubSubFrontend>()
+            .or_else(|| {
+                t.downcast_ref::<BoxTransport>()
+                    .and_then(|t| t.as_any().downcast_ref::<PubSubFrontend>())
+            })
+            .ok_or_else(TransportErrorKind::pubsub_unavailable)
+    }
```
gakonst, line 99: I am OK with this -- @mattsse @onbjerg any objections?

crates/provider/src/new.rs
```diff
@@ -91,13 +155,223 @@ impl<N, T> RootProviderInner<N, T> {
...
+
+    /// Subscribe to new pending transactions.
+    ///
+    /// Returns a stream of new pending transaction hashes.
+    #[cfg(feature = "pubsub")]
+    async fn subscribe_pending_transactions(&self) -> TransportResult<Subscription<B256>> {
+        self.root().pubsub_frontend()?;
+        let id = self.client().prepare("eth_subscribe", ("newPendingTransactions",)).await?;
+        self.root().get_subscription(id).await
```
gakonst, line 206: nit this makes 2 calls to pubsub_frontend due to the get_subscription(id) call, but should be fine

crates/provider/src/new.rs
```diff
@@ -91,13 +155,223 @@ impl<N, T> RootProviderInner<N, T> {
...
+    ///
+    /// Get the next 5 blocks:
+    ///
+    /// ```no_run
+    /// # async fn example<N: alloy_network::Network>(provider: impl alloy_provider::Provider<N>) -> Result<(), Box<dyn std::error::Error>> {
+    /// use futures::StreamExt;
+    ///
+    /// let poller = provider.watch_blocks().await?;
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
+    /// while let Some(block_hash) = stream.next().await {
```
gakonst, line 257: Can we get rid of the `into_stream().flat_map(futures::stream::iter)`? When do we want to use the stream vs the iter? https://docs.rs/futures/latest/futures/stream/fn.iter.html

gakonst (approved):
Great!!! I am comfortable moving forward with this, let's get all the ethers examples ported now @yash-atreya @zerosnacks and please flag missing pieces to Dani as needed.

crates/provider/src/new.rs
DaniPopes: yes, we don't want to make the request if we know we're not a pubsub

mattsse (approved):
progress 💪

crates/provider/src/new.rs
mattsse: we can probably improve this, but for now this seems fine

crates/providers/src/new.rs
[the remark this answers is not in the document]
DaniPopes: Renamed

crates/providers/src/heart.rs
```diff
@@ -1,23 +1,203 @@
...
+/// ```
+///
+/// This can also be more concisely written using `watch`:
+/// ```no_run
+/// # async fn example<N: alloy_network::Network>(provider: impl alloy_providers::Provider<N>, tx: N::TransactionRequest) -> Result<(), Box<dyn std::error::Error>> {
+/// let tx_hash = provider.send_transaction(tx)
+///     .await?
+///     .with_confirmations(2)
+///     .with_timeout(Some(std::time::Duration::from_secs(60)))
+/// #   .with_provider(&provider) // TODO
```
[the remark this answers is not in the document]
DaniPopes: Removed

crates/providers/src/heart.rs
DaniPopes: `send` now returns `PendingTransactionBuilder`

crates/contract/src/call.rs
```diff
@@ -393,13 +398,13 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, D: CallDecoder> CallBu
     }
```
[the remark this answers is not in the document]
DaniPopes: Done

crates/provider/src/heart.rs
DaniPopes: Not sure what this means

crates/providers/src/new.rs
DaniPopes: https://github.com/alloy-rs/alloy/issues/289

crates/providers/src/new.rs
DaniPopes: Added `full: bool` and `fn watch_pending_transactions_full` for `N::TransactionResponse`

crates/rpc-client/src/client.rs
DaniPopes: https://github.com/alloy-rs/alloy/issues/290

crates/rpc-client/src/client.rs
DaniPopes: Would have to be set from the heartbeat task and the field would have to be a `Arc<RwLock<Duration>>`, which means we can also synchronize all polling intervals globally for that provider? I can see this happening

PR #276, opened 2024-03-12
feat: nonce filling layer

### LINKED ISSUE #275: [Feature] Add nonce filler layer
### Component

providers, pubsub

Add a `NonceLayer` (need a better name) like `SignerLayer` that queries, using the provider, the latest appropriate nonce for a `N::TransactionRequest` and fills it if it is missing.

Some notes:

1. This layer would need to be layered on *before* the `SignerLayer` (i.e. the request goes through `NonceLayer` -> `SignerLayer` -> `RootProvider`)
2. Take note that transactions might be sent concurrently on the same provider, so this needs to be taken into account (see e.g. the atomics usage in ethers: https://docs.rs/ethers-middleware/2.0.14/src/ethers_middleware/nonce_manager.rs.html#10-16)
3. Ideally the layer would work for multiple addresses (i.e. no need to specify an address in `NonceLayer` like we did for `NonceManagerMiddleware` in ethers)

### Additional context

_No response_

onbjerg:
## Motivation

This is a replacement for `NonceManagerMiddleware` in ethers

## Solution

A layer that fills nonces for transaction requests. It does not require you to specify a sender address like in ethers, instead we use a dashmap internally.

I would like some ideas on how to test that this works with sending transaction requests for the same sender from multiple threads at the same time (i.e. a sanity check on the atomics)

Closes #275

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/providers/src/layers/nonce.rs
```diff
@@ -0,0 +1,204 @@
...
+///
+/// [`ProviderBuilder`]: crate::ProviderBuilder
+pub struct ManagedNonceProvider<N, T, P>
+where
+    N: Network,
+    T: Transport + Clone,
+    P: Provider<N, T>,
+{
+    inner: P,
+    nonces: DashMap<Address, AtomicU64>,
```
mattsse, line 74: I believe this raw atomic is redundant because it is not shared

crates/providers/src/layers/nonce.rs
```diff
@@ -0,0 +1,204 @@
...
+    P: Provider<N, T>,
+{
+    async fn get_nonce(&self, from: Address) -> TransportResult<U64> {
+        // initialize the account nonce if it hasn't already been
+        if !self.nonces.contains_key(&from) {
+            let nonce = self.inner.get_transaction_count(from, None).await?;
+
+            // check again in case another thread filled the initial nonce
+            if !self.nonces.contains_key(&from) {
+                self.nonces.insert(from, AtomicU64::new(nonce.to()));
```
mattsse, line 91: there's still a race condition here because contains_key + insert is not atomic.

crates/providers/src/layers/nonce.rs
```diff
@@ -0,0 +1,204 @@
...
+    _phantom: PhantomData<(N, T)>,
+}
+
+impl<N, T, P> ManagedNonceProvider<N, T, P>
+where
+    N: Network,
+    T: Transport + Clone,
+    P: Provider<N, T>,
+{
+    async fn get_nonce(&self, from: Address) -> TransportResult<U64> {
```
mattsse, line 84: this could potentially be problematic, because now once inserted into the map, the nonce is fixed

if you send two tx without a nonce sequentially through this provider, the second one will reuse the previous nonce

onbjerg: that's not the case, as you can see in the test

crates/providers/src/layers/nonce.rs
onbjerg: i guess the best way would be `RwLock<HashMap<Address, AtomicU64>>`, write lock for inits and otherwise fetch adds for updating the nonce, wdyt? this would assume that initialization is infrequent because you send from the same accounts more often

crates/providers/src/layers/nonce.rs
```diff
@@ -0,0 +1,204 @@
...
+        if !self.nonces.contains_key(&from) {
+            let nonce = self.inner.get_transaction_count(from, None).await?;
+
+            // check again in case another thread filled the initial nonce
+            if !self.nonces.contains_key(&from) {
+                self.nonces.insert(from, AtomicU64::new(nonce.to()));
+            }
+        }
+
+        Ok(U64::from(self.nonces.get(&from).unwrap().fetch_add(1, Ordering::SeqCst)))
```
onbjerg, line 95: i should probably also just use `Ordering::Relaxed` here since we do not care about ordering, only atomicity

crates/providers/src/layers/nonce.rs
onbjerg: if I replace this with:

1. `RwLock<HashMap<Address, AtomicU64>>`
2. check if data was inserted between trying to acquire the lock and actually acquiring it, short circuiting if it was
3. initializing the atomic

for initialized values:

1. read lock
2. fetch_add

do you think this would be sound?

DaniPopes: doesn't the entry API work fine here with just dashmap?

crates/providers/src/layers/nonce.rs
DaniPopes: yeah u64 should be just fine

crates/providers/src/layers/nonce.rs
onbjerg: actually, might, since it locks the shard that contains the kv pair

crates/providers/src/layers/nonce.rs
onbjerg: think its only ok if we use the entry api every time which acquires a lock

crates/provider/src/layers/nonce.rs
```diff
@@ -0,0 +1,207 @@
...
+    _phantom: PhantomData<(N, T)>,
+}
+
+impl<N, T, P> ManagedNonceProvider<N, T, P>
+where
+    N: Network,
+    T: Transport + Clone,
+    P: Provider<N, T>,
+{
+    async fn get_nonce(&self, from: Address) -> TransportResult<U64> {
```
mattsse, line 84:
```suggestion
    async fn get_next_nonce(&self, from: Address) -> TransportResult<U64> {
```

crates/provider/src/layers/nonce.rs
```diff
@@ -0,0 +1,207 @@
...
+        let nonce = match self.nonces.entry(from) {
+            Entry::Occupied(entry) => {
+                let next_nonce = entry.get() + 1;
+                entry.replace_entry(next_nonce);
+                next_nonce
+            }
+            Entry::Vacant(entry) => {
+                let nonce: u64 = self.inner.get_transaction_count(from, None).await?.to();
+                entry.insert(nonce);
```
mattsse, line 93: this uses a sync lock across an await point, which can cause issues,

imo a better approach here would be a sync lock over an async lock that guards the nonce

mattsse (changes requested):

crates/provider/src/layers/nonce.rs
```diff
@@ -0,0 +1,211 @@
...
+///     .signer(EthereumSigner::from(signer)) // note the order!
+///     .network::<Ethereum>()
+///     .provider(RootProvider::new(transport));
+///
+/// provider.send_transaction(TransactionRequest::default()).await;
+/// # }
+/// ```
+///
+/// [`SignerLayer`]: crate::layers::SignerLayer
+pub struct ManagedNonceLayer;
```
mattsse, line 41: + debug + clone etc

mattsse (approved):
smol derive nit, lgtm otherwise

zerosnacks:
@onbjerg could it be that the `ManagedNonceLayer` doesn't track contract deployments? I'm getting some strange errors where at once it complains that either the nonce is too high (1) or too low (0) after deploying a contract with the same provider.

For context see: https://github.com/alloy-rs/examples/pull/20/files

onbjerg:
It should provided the function you use to deploy uses `send_transaction`

zerosnacks:
> It should provided the function you use to deploy uses `send_transaction`. Where exactly are you getting the error?

I've opened an issue on Alloy to explore this with a minimal reproduction example: https://github.com/alloy-rs/alloy/issues/406 + https://github.com/zerosnacks/alloy-repro-406

Issue #262, opened 2024-03-11
[Feature] Move `Wallet`/`LocalWallet` etc to its own crate


onbjerg:
### Component

signers

### Describe the feature you would like

`alloy-network` (and by extension `alloy-provider`) cannot be WASM-compatible since https://github.com/alloy-rs/alloy/pull/190 because it depends on `alloy-signers`, which itself is not WASM-compatible because of a dependency on `rand`.

As far as I can tell, this is only for `Wallet`/`LocalWallet` etc., which can live in their own crate.

https://github.com/alloy-rs/alloy/pull/190#discussion_r1516749547

PR #279, opened 2024-03-12
refactor: remove `async_trait` in tx builder

### LINKED ISSUE #268: [Feature] Remove `async_trait` from `TransactionBuilder`
### Component

network, json-rpc

Not needed

See https://github.com/alloy-rs/alloy/pull/190/files#r1517742855

### Additional context

_No response_

onbjerg:
## Motivation

Closes #268

## Solution

We can't use `async fn` because we need the `Send` bound for `NetworkSigner`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/transaction/builder.rs
+++ b/crates/network/src/transaction/builder.rs
@@ -1,7 +1,6 @@
 use super::signer::NetworkSigner;
 use crate::Network;
 use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256, U64};
-use async_trait::async_trait;
 
 /// Error type for transaction builders.
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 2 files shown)


crates/network/src/transaction/builder.rs
```diff
@@ -196,5 +193,8 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
     fn build_unsigned(self) -> BuilderResult<N::UnsignedTx>;
 
     /// Build a signed transaction.
-    async fn build<S: NetworkSigner<N>>(self, signer: &S) -> BuilderResult<N::TxEnvelope>;
+    fn build<S: NetworkSigner<N>>(
+        self,
+        signer: &S,
+    ) -> impl std::future::Future<Output = BuilderResult<N::TxEnvelope>> + Send;
```
mattsse, line 199: does this work for wasm?

mattsse (approved):

crates/network/src/transaction/builder.rs
onbjerg: unsure, but this entire crate does not yet work for wasm because of https://github.com/alloy-rs/alloy/issues/262

onbjerg: ah, I assume it's not, `wasm-bindgen` does not generate `Send` futures. i assume we would need to conditionally compile this, but I'd leave that for the above issue perhaps, wdyt

DaniPopes: related: `impl_future!` macro in `alloy_transport` https://github.com/alloy-rs/alloy/blob/86027c9bb984f3a12a30ffd2a3c5f2f06595f1d6/crates/transport/src/lib.rs#L61

PR #287, opened 2024-03-12
feat: add initial EIP-7547 engine types

Rjected:
## Motivation
Inclusion lists are being implemented by clients for a poc:
https://github.com/ethereum/go-ethereum/commit/3fd819994b654e7dd11000bb8eee7d3f3a3b9ad4

To do this for reth, we'll need to get the proper rpc types, so we can implement the calls in reth.

## Solution

Implemented the types defined in https://github.com/michaelneuder/execution-apis/pull/1

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


prestwich:
procedural question, should in-progress EIPs be homed in `alloy-eips` to minimize intrusion in other parts of the codebase?

Rjected:
> procedural question, should in-progress EIPs be homed in `alloy-eips` to minimize intrusion in other parts of the codebase?

For this one it could be, since it would not cause any circular deps. But if an in progress eip needed a transaction type for example, it would cause problems. IMO despite this one not causing circular issues, it seems best to just have it as a draft PR to the place where it will eventually live (rpc-engine-types).

prestwich:
another option would be to have a specific crate alloy-eipXXXX for these while they're in progress and not accepted

mattsse:
> another option would be to have a specific crate alloy-eipXXXX for these while they're in progress and not accepted

that makes sense!

Rjected:
> > another option would be to have a specific crate alloy-eipXXXX for these while they're in progress and not accepted
> 
> that makes sense!

This is now in the eip7547 crate

Cargo.toml
```diff
@@ -20,6 +20,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 alloy-consensus = { version = "0.1.0", default-features = false, path = "crates/consensus" }
 alloy-contract = { version = "0.1.0", default-features = false, path = "crates/contract" }
 alloy-eips = { version = "0.1.0", default-features = false, path = "crates/eips" }
+alloy-eip7547 = { version = "0.1.0", default-features = false, path = "crates/eip7547" }
```
prestwich, line 23: key here doesn't match package name. different hyphenation. standardize on `eip7547`

Rjected: whoops, just fixed

prestwich (approved):
UTACK. I have not reviewed these types for conformance, but given this is a pre-finalization EIP i don't think that matters much

crates/eip7547/src/summary.rs
```diff
@@ -0,0 +1,186 @@
...
+///
+/// - `address` : `DATA`, 20 Bytes
+/// - `nonce` : `QUANTITY`, 64 Bits
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InclusionListSummaryEntryV1 {
+    /// The address of the inclusion list entry.
+    pub address: Address,
+    /// The nonce of the inclusion list entry.
+    pub nonce: u64,
```
mattsse, line 85: this should be hex serde?

crates/eip7547/src/summary.rs
```diff
@@ -0,0 +1,186 @@
...
+/// - `proposer_index`: `QUANTITY`, 64 Bits
+/// - `parent_hash`: `DATA`, 32 Bytes
+/// - `summary`: `Array of InclusionListSummaryEntryV1`, Array of entries that must be satisfied.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InclusionListSummaryV1 {
+    /// The slot of the inclusion list summary.
+    pub slot: u64,
+    /// The proposer index of the inclusion list summary.
+    pub proposer_index: u64,
```
mattsse, line 111: should be hex?

mattsse (changes requested):
quantity should be hex

https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding

crates/eip7547/src/summary.rs
```diff
@@ -82,7 +82,7 @@ pub struct InclusionListSummaryEntryV1 {
     /// The address of the inclusion list entry.
     pub address: Address,
     /// The nonce of the inclusion list entry.
-    pub nonce: u64,
+    pub nonce: U64,
```
mattsse, line 85: can we use u64 and serde_with alloy_serde::u64_hex for all U64?

mattsse (changes requested):

crates/eip7547/src/summary.rs
Rjected: whoops, yeah

mattsse (approved):

PR #291, opened 2024-03-13
feat(examples): Port `wallets` from `ethers-rs`

zerosnacks:
## Motivation

To add examples from `ethers-rs`: `wallets`

## Solution

Includes setup of the initial example directory structure allowing you to run examples using the syntax: `cargo run --example mnemonic-signer` where `mnemonic-signer` is any example.

Makes two changes in `Alloy`:
- Enables default features for `coins-bip39`, exported from the `signer` crate so it includes wordlists.
- Clarifies comment of `eip712_signing_hash`.

Suggestions for future examples to add:
- `AWS signer`
- `GCP signer`
- `Keystore signer`

Please consider I do not have access to all devices listed making it difficult to guarantee the correct functioning of the examples. All examples are designed to be able to be ran.

It would be great to get feedback on these initial examples.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


examples/wallets/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+alloy-rpc-types = { workspace = true }
+alloy-signer = { workspace = true, features = [
+    "eip712",
+    "mnemonic",
+    "yubihsm",
+] }
+alloy-signer-ledger = { workspace = true }
+alloy-signer-trezor = { workspace = true }
+alloy-sol-types = { workspace = true }
+alloy-transport-http = { workspace = true }
```
DaniPopes, line 29: I think we should be using the `alloy` crates, wdyt @gakonst @mattsse ?

DaniPopes: Also these should be inline syntax `.workspace = true` since we don't enable anything else

onbjerg (approved):

prestwich (changes requested):
i am a strong NACK on examples crates in the workspace right now. These were a huge mess for ethers, and we haven't had a discussion on how to resolve that yet

DaniPopes:
@prestwich I don't see the concern with examples if we don't run them, only compile them. They're just like `no_run` doctests.

prestwich:
> @prestwich I don't see the concern with examples if we don't run them, only compile them. They're just like no_run doctests.
 
When making breaking changes, or updating recommended usage, an in-repo example depending on local packages needs to be updated on every PR, even ones that may never reach GA. Whereas homing examples in another repo and running a chron CI against both published and unpublished versions pushes all maintenance work out of the dev cycle and into the release cycle 

basically these likely belong in an `alloy-examples` and putting them here slows pace of development

prestwich:
Setting up a home for these
https://github.com/alloy-rs/examples

crates/signer/Cargo.toml
```diff
@@ -30,7 +30,7 @@ eth-keystore = { version = "0.5.0", default-features = false, optional = true }
 
 # mnemonic
 coins-bip32 = { version = "0.8.7", default-features = false, optional = true }
-coins-bip39 = { version = "0.8.7", default-features = false, optional = true }
+coins-bip39 = { version = "0.8.7", optional = true }
```
zerosnacks, line 33: Note @zerosnacks 

Make sure to remove this after #309 is merged.

zerosnacks:
Closing in favor of https://github.com/alloy-rs/examples

Issue #263, opened 2024-03-11
[Feature] Default `Network` parameters to `Ethereum` where possible


onbjerg:
### Component

network, json-rpc, providers, pubsub

### Describe the feature you would like

See https://github.com/alloy-rs/alloy/pull/190/files#r1517716953 and https://github.com/alloy-rs/alloy/pull/190/files#r1517746095

prestwich:
I believe this is sufficiently complete

PR #304, opened 2024-03-13
feat: default to Ethereum network in `ProviderBuilder`

DaniPopes:
Ref https://github.com/alloy-rs/alloy/issues/263

### DIFF
--- a/crates/provider/src/builder.rs
+++ b/crates/provider/src/builder.rs
@@ -1,5 +1,5 @@
 use crate::{layers::SignerLayer, Provider, RootProvider};
-use alloy_network::Network;
+use alloy_network::{Ethereum, Network};
 use alloy_rpc_client::RpcClient;
 use alloy_transport::Transport;
 use std::marker::PhantomData;
@@ -70,9 +70,8 @@ where
 ///
 /// [`tower::ServiceBuilder`]: https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html
 #[derive(Debug)]
-pub struct ProviderBuilder<L, N = ()> {
+pub struct ProviderBuilder<L, N = Ethereum> {
     layer: L,
-
     network: PhantomData<N>,
 }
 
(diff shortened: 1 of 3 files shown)


crates/provider/src/builder.rs
```diff
@@ -99,17 +98,16 @@ impl<L, N> ProviderBuilder<L, N> {
     /// [`tower::ServiceBuilder`]. The first layer added will be the first to
     /// see the request.
     ///
-    ///
     /// [`tower::ServiceBuilder::layer`]: https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#method.layer
     /// [`tower::ServiceBuilder`]: https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html
-    pub fn layer<Inner>(self, layer: Inner) -> ProviderBuilder<Stack<Inner, L>> {
+    pub fn layer<Inner>(self, layer: Inner) -> ProviderBuilder<Stack<Inner, L>, N> {
```
DaniPopes, line 103: need to be careful here not to ignore `N`, previously this would reset it back to `()`

crates/provider/src/builder.rs
```diff
@@ -99,17 +98,16 @@ impl<L, N> ProviderBuilder<L, N> {
     /// [`tower::ServiceBuilder`]. The first layer added will be the first to
```
prestwich, line 98: maybe add `pub fn network<Other: Network>(self) -> ProviderBuilder<Stack<Inner, L>, Other>` for lazy users or programmatic switching?

DaniPopes: Already exists https://github.com/alloy-rs/alloy/blob/9a33128261e4c1c48d7e124758758aa218024b20/crates/provider/src/builder.rs#L114-L125

DaniPopes: Just noticed docs need an update. cc @onbjerg

mattsse (approved):

crates/provider/src/builder.rs
prestwich: yes Ethereum by default

PR #309, opened 2024-03-14
Feat: enable default features for `coins_bip39` to export default wordlist

### LINKED ISSUE #308: [Feature] Enable default features for `coins_bip39` to export default wordlist
### Component

signers

Prevents users from having to import `coins_bip39` themselves to use the wordlist, achieving the same user experience as in `ether-rs`.

https://github.com/alloy-rs/alloy/blob/818b3e9692f2a207e456653a67aa8956fe1518a9/crates/signer/src/lib.rs#L37-L38

### Additional context

_No response_

zerosnacks:
## Motivation

Closes #308 

Prevents users from having to import `coins_bip39` themselves to use the wordlist, achieving the same user experience as in `ether-rs`.

https://github.com/alloy-rs/alloy/blob/818b3e9692f2a207e456653a67aa8956fe1518a9/crates/signer/src/lib.rs#L37-L38

## Solution

Enables default features.

This change is already included as part of #291 but if we decide to move the examples folder out this is required as part of it.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/signer/Cargo.toml
```diff
@@ -30,7 +30,7 @@ eth-keystore = { version = "0.5.0", default-features = false, optional = true }
 
 # mnemonic
 coins-bip32 = { version = "0.8.7", default-features = false, optional = true }
-coins-bip39 = { version = "0.8.7", default-features = false, optional = true }
+coins-bip39 = { version = "0.8.7", optional = true }
```
DaniPopes, line 33: This enables all languages: https://docs.rs/crate/coins-bip39/latest/features
Can we just enable the english one?

DaniPopes (changes requested):

crates/signer/Cargo.toml
prestwich: probably want to introduce an `all-bip39-lists` feature that re-exports all wordlists, while keeping english as default

PR #316, opened 2024-03-14
feat: update priority fee estimator

klkvr:
## Motivation

Current estimator version is not flexible and will always return 3 gwei priority fee if base fee is less than 100 gwei which is not correct for L2s (and often pretty high even for Mainnet).

## Solution

New impl was inspired by Metamask estimator and introduces following changes:
1. We no more use `EIP1559_FEE_ESTIMATION_THRESHOLD_MAX_CHANGE` because it could've resulted in overestimated gas prices. Let's consider `rewards` with following contents: `[100, 1, 1, 1, 1]`, in this case we'd filter out all 1's because of spike caused by a single 100 price, and the resulted priority fee estimate will be 100 gwei which is incorrect. Using median should be enough for a fair estimate filtering out potential anomalies.
2. We now simply multiply base fee by 2 to calculate max base fee instead of using `base_fee_surged` oriented on mainnet.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/utils.rs
+++ b/crates/provider/src/utils.rs
@@ -3,7 +3,7 @@
 use alloy_primitives::U256;
 
 /// The number of blocks from the past for which the fee rewards are fetched for fee estimation.
-pub const EIP1559_FEE_ESTIMATION_PAST_BLOCKS: u64 = 5;
+pub const EIP1559_FEE_ESTIMATION_PAST_BLOCKS: u64 = 10;
 /// Multiplier for the current base fee to estimate max base fee for the next block.
 pub const EIP1559_BASE_FEE_MULTIPLIER: f64 = 2.0;
 /// The default percentile of gas premiums that are fetched for fee estimation.
--- a/crates/provider/src/utils.rs
+++ b/crates/provider/src/utils.rs
@@ -20,8 +20,15 @@ fn estimate_priority_fee(rewards: &[Vec<U256>]) -> U256 {
     }
 
     rewards.sort();
+
     // Return the median.
-    *rewards[rewards.len() / 2]
+    let n = rewards.len();
+    
+    if n % 2 == 0 {
+        (*rewards[n / 2 - 1] + *rewards[n / 2]) / U256::from(2)
+    } else {
+        *rewards[(n + 1) / 2]
+    }
 }
 
 /// The default EIP-1559 fee estimator which is based on the work by [MetaMask](https://github.com/MetaMask/core/blob/main/packages/gas-fee-controller/src/fetchGasEstimatesViaEthFeeHistory/calculateGasFeeEstimatesForPriorityLevels.ts#L56)
--- a/crates/provider/src/utils.rs
+++ b/crates/provider/src/utils.rs
@@ -3,14 +3,11 @@
 use alloy_primitives::{I256, U256};
 
 /// The number of blocks from the past for which the fee rewards are fetched for fee estimation.
-pub const EIP1559_FEE_ESTIMATION_PAST_BLOCKS: u64 = 10;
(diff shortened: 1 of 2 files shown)


crates/provider/src/utils.rs
```diff
@@ -23,75 +17,16 @@ fn estimate_priority_fee(rewards: Vec<Vec<U256>>) -> U256 {
...
-    if base_fee_per_gas <= U256::from(40_000_000_000u64) {
-        base_fee_per_gas * U256::from(2)
-    } else if base_fee_per_gas <= U256::from(100_000_000_000u64) {
-        base_fee_per_gas * U256::from(16) / U256::from(10)
-    } else if base_fee_per_gas <= U256::from(200_000_000_000u64) {
-        base_fee_per_gas * U256::from(14) / U256::from(10)
-    } else {
-        base_fee_per_gas * U256::from(12) / U256::from(10)
-    }
+    rewards[rewards.len() / 2]
```
mattsse, line 23: using the median by default is sane

crates/provider/src/utils.rs
```diff
@@ -1,19 +1,13 @@
 //! Provider-related utilities.
 
-use alloy_primitives::{I256, U256};
+use alloy_primitives::U256;
 
 /// The number of blocks from the past for which the fee rewards are fetched for fee estimation.
 pub const EIP1559_FEE_ESTIMATION_PAST_BLOCKS: u64 = 10;
+/// Multiplier for the current base fee to estimate max base fee for the next block.
+pub const EIP1559_BASE_FEE_MULTIPLIER: f64 = 2.0;
```
mattsse, line 8: that should be more reasonable

mattsse (approved):
lgtm

DaniPopes (changes requested):

crates/provider/src/utils.rs
DaniPopes, line 23:
```suggestion
    let n = rewards.len();
    if n % 2 == 0 {
        (rewards[n / 2] + rewards[n / 2 + 1]) / 2
    } else {
        rewards[(n + 1) / 2]
    }
```

DaniPopes (changes requested):

crates/provider/src/utils.rs
klkvr: wouldn't this panic when n == 2?

DaniPopes: sure, just flagging this is not the real median

klkvr: 22e3f0df96b23ac1efde66fbe8ebe61cd50d1bd3

crates/provider/src/utils.rs
```diff
@@ -1,97 +1,40 @@
...
-        potential_max_fee
-    };
-    (max_fee_per_gas, max_priority_fee_per_gas)
+/// The default EIP-1559 fee estimator which is based on the work by [MetaMask](https://github.com/MetaMask/core/blob/main/packages/gas-fee-controller/src/fetchGasEstimatesViaEthFeeHistory/calculateGasFeeEstimatesForPriorityLevels.ts#L56)
+/// (constants for "medium" priority level are used)
+pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>]) -> (U256, U256) {
+    let max_priority_fee_per_gas = estimate_priority_fee(rewards);
+    let potential_max_fee = base_fee_per_gas * U256::from(EIP1559_BASE_FEE_MULTIPLIER);
+    (potential_max_fee, max_priority_fee_per_gas)
 }
```
DaniPopes, line 40: Can we add some sanity tests? I don't see any

DaniPopes (approved):
Needs test otherwise LGTM

PR #319, opened 2024-03-15
feat(json-rpc): use `Cow` instead of `&'static str` for method names

DaniPopes:
This allows dynamic method names in the type representation.


crates/json-rpc/src/request.rs
```diff
@@ -194,8 +195,8 @@ impl SerializedRequest {
     }
 
     /// Returns the request method.
-    pub const fn method(&self) -> &'static str {
-        self.meta.method
+    pub const fn method(&self) -> &Cow<'static, str> {
```
mattsse, line 198: perhaps &str? 
it wouldn't make it const but I think since the caller wants the &str we can just deref here?

mattsse (approved):
imo this makes sense

pending @prestwich

prestwich (approved):

gakonst:
nice

Issue #321, opened 2024-03-15
[Feature] Add `new` method to IpcConnect


yash-atreya:
### Component

transports

### Describe the feature you would like

Related to `alloy-transport-ipc`.

Currently, `IpcConnect` doesn't implement the `new` method. This isn't ideal UX.

PR #322, opened 2024-03-15
add: `new` method to initialize IpcConnect

yash-atreya:
## Motivation

Ref: #321 

## Solution

Impl `new` method for `IpcConnect`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-ipc/src/connect.rs
+++ b/crates/transport-ipc/src/connect.rs
@@ -12,11 +12,11 @@ pub struct IpcConnect<T> {
 impl<T> IpcConnect<T> {
     /// Create a new IPC connection object for any type T that can be converted into
     /// `IpcConnect<T>`.
-    pub fn new(inner: T) -> Self
+    pub const fn new(inner: T) -> Self
     where
-        Self: alloy_pubsub::PubSubConnect + From<T>,
+        Self: alloy_pubsub::PubSubConnect,
     {
-        IpcConnect::from(inner)
+        Self { inner }
     }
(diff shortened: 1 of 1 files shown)


prestwich (changes requested):
there should be some trait bound on `new` to prevent users from creating IpcConnect instances that don't work

prestwich:
```
impl<T> IpcConnect<T> {
    /// Create a new IPC connection object for any type T that can be converted into IpcConnect<T>.
    pub fn new(inner: T) -> Self
    where
        Self: alloy_pubsub::PubSubConnect,
    {
        Self { inner }
    }
}
```

prestwich (approved):

crates/transport-ipc/src/connect.rs
```diff
@@ -10,7 +10,8 @@ pub struct IpcConnect<T> {
 }
 
 impl<T> IpcConnect<T> {
-    /// Create a new IPC connection object for any type T that can be converted into IpcConnect<T>.
+    /// Create a new IPC connection object for any type T that can be converted into
+    /// `IpcConnect<T>`.
     pub fn new(inner: T) -> Self
     where
         Self: alloy_pubsub::PubSubConnect + From<T>,
```
prestwich, line 17: the `From<T>` bound is unnecessary as you can just write out `Self { inner }`

prestwich (approved):

Issue #255, opened 2024-03-08
[Feature] Deduplicate consensus and rpc types AccessList


DaniPopes:
### Component

consensus, eips, genesis, network, json-rpc

### Describe the feature you would like

https://github.com/alloy-rs/alloy/blob/b370cf390e88f893711b1dedd90dd7654846e73a/crates/eips/src/eip2930.rs#L47

https://github.com/alloy-rs/alloy/blob/b370cf390e88f893711b1dedd90dd7654846e73a/crates/rpc-types/src/eth/transaction/access_list.rs#L17

mattsse:
from reth POV I believe there's a similar problem with the Withdrawal type, 

so I think we should find a way to extract the Withdrawal type from rpc and perhaps also move to eip crate or find a better location for these types that have the same representation in rpc and consensus

sambacha:
Checking out the access list generation, it seems there are no checks against preventing that \`tx sender\`, \`tx recipient\`, \`block producer\`, created contracts and precompiles  should \*not\* be included in access lists as they are considered warm. Should I make a separate issue?

sambacha:
Checking out the access list generation, it seems there are no checks against preventing that \`tx sender\`, \`tx recipient\`, \`block producer\`, created contracts and precompiles  should \*not\* be included in access lists as they are considered warm. Should I make a separate issue?

sambacha:
Checking out the access list generation, it seems there are no checks against preventing that \`tx sender\`, \`tx recipient\`, \`block producer\`, created contracts and precompiles  should \*not\* be included in access lists as they are considered warm. Should I make a separate issue?

DaniPopes:
I don't fully understand your comment but that should definitely be in a different issue

mattsse:
from,to are excluded

https://github.com/paradigmxyz/evm-inspectors/blob/d7317c7fbf38d5e5df2393beee0414c987ea440a/src/access_list.rs#L31

mattsse:
AccessList+Withdrawals should be moved to eips crate

JackG-eth:
Happy to pick this up

PR #324, opened 2024-03-15
chore: deduplicate AccessList and Withdrawals types

JackG-eth:
closes https://github.com/alloy-rs/alloy/issues/255

deduplicates AccessList+Withdrawals by moving them into eips crate and reusing them in rpc and consensus


crates/eips/src/withdrawal.rs
mattsse, line 1: let's rename this eip4895.rs and documents, see also eip2930.rs

https://eips.ethereum.org/EIPS/eip-4895

this file also needs changes so serde is optional and we need ssz feature

mattsse (changes requested):
great start,

a few more todos!

crates/eips/Cargo.toml
```diff
@@ -14,9 +14,11 @@ exclude.workspace = true
 [dependencies]
 alloy-primitives = { workspace = true, features = ["rlp"], default-features = false }
 alloy-rlp = { workspace = true, features = ["derive"] }
+alloy-serde.workspace = true
 
 # serde
 serde = { workspace = true, default-features = false, optional = true }
+serde_json.workspace = true
```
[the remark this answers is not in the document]
JackG-eth: moved

crates/eips/src/withdrawal.rs
JackG-eth: Added optionality of serde, is this as some packages need it and others don't pushing more performance?

JackG-eth: Do you mean szz feature for eip2930? as it already exists for withdrawals.

crates/eips/src/eip4895.rs
```diff
@@ -1,27 +1,17 @@
...
-    Default,
-    Hash,
-    RlpEncodable,
-    RlpDecodable,
-    Serialize,
-    Deserialize,
-)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Default, Hash, RlpEncodable, RlpDecodable)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "ssz", derive(ssz_derive::Encode, ssz_derive::Decode))]
```
mattsse, line 15: this feature currently does not exist in this crate, so we need to add this and the required deps, see rpc-types for reference
then we have to enable this in rpc-types if the ssz feature is enabled there

mattsse (changes requested):

mattsse (approved):

mattsse:
fyi @klkvr with this AccessList types are deduplicated

onbjerg:
can you also remove https://github.com/alloy-rs/alloy/blob/e8c622c713938cd2ede29d6b38d2aaa009dd3ad5/crates/network/src/ethereum/builder.rs#L229 which was part of the motivation for this PR?

onbjerg (changes requested):
(see above)

mattsse:
done

onbjerg (approved):
lgtm

Issue #323, opened 2024-03-15
[Feature] Impl `GasEstimatorLayer`


yash-atreya:
### Component

provider, pubsub

### Describe the feature you would like

Just like how `ManagedNonceLayer` automatically manages the `nonce` field in `TransactionRequest` if unset by the user. `GasEstimatorLayer` would populate the `gas_price`, `gas`, `max_fee_per_gas` and `max_priority_fee_per_gas`.

Using this layer would remove the need for users to set these critical fields themselves.

PR #326, opened 2024-03-15
feat(layers): GasEstimationLayer

yash-atreya:
## Motivation

Ref: #323 

## Solution

Implement a `layer` that populates the `gas_price`, `gas`, `max_fee_per_gas` and `max_priority_fee_per_gas`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -1,5 +1,5 @@
 use crate::{PendingTransactionBuilder, Provider, ProviderLayer, RootProvider};
-use alloy_network::Network;
+use alloy_network::{Network, TransactionBuilder};
 use alloy_primitives::U256;
 use alloy_transport::{Transport, TransportResult};
 use async_trait::async_trait;
@@ -17,11 +17,12 @@ where
     T: Transport + Clone,
 {
     type Provider = GasEstimatorProvider<N, T, P>;
-    fn layer(&self, _inner: P) -> Self::Provider {
-        todo!()
+    fn layer(&self, inner: P) -> Self::Provider {
+        GasEstimatorProvider { inner, _phantom: PhantomData }
     }
 }
 
+/// A provider that provides gas estimation for transactions.
 #[derive(Debug, Clone)]
 pub struct GasEstimatorProvider<N, T, P>
 where
@@ -40,18 +41,13 @@ where
     P: Provider<N, T>,
 {
     async fn get_gas_price(&self) -> TransportResult<U256> {
-        todo!()
+        let gas_price = self.inner.get_gas_price().await?;
+        Ok(gas_price)
     }
 
-    async fn get_gas_estimate(&self, _tx: N::TransactionRequest) -> TransportResult<U256> {
-        todo!()
-    }
-
-    async fn get_1559_estimate_fees(
-        &self,
-        _tx: N::TransactionRequest,
-    ) -> TransportResult<(U256, U256)> {
-        todo!()
+    async fn get_gas_estimate(&self, tx: &N::TransactionRequest) -> TransportResult<U256> {
+        let gas_estimate = self.inner.estimate_gas(tx, None).await?;
+        Ok(gas_estimate)
     }
 }
 
@@ -71,7 +67,60 @@ where
         &self,
         mut tx: N::TransactionRequest,
     ) -> TransportResult<PendingTransactionBuilder<'_, N, T>> {
-        // TODO: Get estimates and set the gas for the tx.
+        if tx.gas_price().is_none() {
+            let gas_price = self.get_gas_price().await?;
+            tx.set_gas_price(gas_price);
+        }
+        if tx.gas_limit().is_none() {
+            let gas_estimate = self.get_gas_estimate(&tx).await?;
+            tx.set_gas_limit(gas_estimate);
+        }
+
         self.inner.send_transaction(tx).await
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{layers::ManagedNonceLayer, ProviderBuilder};
+    use alloy_network::EthereumSigner;
+    use alloy_node_bindings::Anvil;
+    use alloy_primitives::{address, U128, U256};
+    use alloy_rpc_client::RpcClient;
+    use alloy_rpc_types::TransactionRequest;
+    use alloy_transport_http::Http;
+    use reqwest::Client;
+
+    #[tokio::test]
+    async fn no_gas_price_or_limit() {
+        let anvil = Anvil::new().spawn();
+        let url = anvil.endpoint().parse().unwrap();
+        let http = Http::<Client>::new(url);
+
+        let wallet = alloy_signer_wallet::Wallet::from(anvil.keys()[0].clone());
+
+        let provider = ProviderBuilder::new()
+            .layer(ManagedNonceLayer)
+            .layer(GasEstimatorLayer)
+            .signer(EthereumSigner::from(wallet))
+            .provider(RootProvider::new(RpcClient::new(http, true)));
+
+        let tx = TransactionRequest {
+            from: Some(anvil.addresses()[0]),
+            value: Some(U256::from(100)),
+            to: address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into(),
+            ..Default::default()
+        };
+
+        let tx = provider.send_transaction(tx).await.unwrap();
+
+        let tx = tx.get_receipt().await.unwrap();
+
+        let set_gas_price = U128::from(0x77359400);
+        let set_gas_limit = U256::from(0x5208);
+
+        assert_eq!(tx.effective_gas_price, set_gas_price);
+        assert_eq!(tx.gas_used, Some(set_gas_limit));
     }
 }
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -49,6 +49,13 @@ where
         let gas_estimate = self.inner.estimate_gas(tx, None).await?;
         Ok(gas_estimate)
     }
+
+    async fn get_eip1159_fees_estimate(&self) -> TransportResult<(U256, U256)> {
+        let (max_fee_per_gas, max_priority_fee_per_gas) =
+            self.inner.estimate_eip1559_fees(None).await?;
+
+        Ok((max_fee_per_gas, max_priority_fee_per_gas))
+    }
 }
 
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
@@ -76,6 +83,13 @@ where
             tx.set_gas_limit(gas_estimate);
         }
 
+        if tx.max_fee_per_gas().is_none() && tx.max_priority_fee_per_gas().is_none() {
+            let (max_fee_per_gas, max_priority_fee_per_gas) =
+                self.get_eip1159_fees_estimate().await?;
+
+            tx.set_max_fee_per_gas(max_fee_per_gas);
+            tx.set_max_priority_fee_per_gas(max_priority_fee_per_gas);
+        }
         self.inner.send_transaction(tx).await
     }
 }
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -135,7 +135,7 @@ where
         let gas_price = self.get_gas_price();
 
         if tx.gas_limit().is_none() {
-            let gas_estimate = self.get_gas_estimate(&tx);
+            let gas_estimate = self.get_gas_estimate(tx);
 
             match futures::join!(gas_price, gas_estimate) {
                 (Ok(gas_price), Ok(gas_estimate)) => {
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -1,0 +1,77 @@
+use crate::{PendingTransactionBuilder, Provider, ProviderLayer, RootProvider};
+use alloy_network::Network;
+use alloy_primitives::U256;
+use alloy_transport::{Transport, TransportResult};
+use async_trait::async_trait;
+use std::marker::PhantomData;
+
+/// A layer that provides gas estimation for transactions.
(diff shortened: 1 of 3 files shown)


crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,126 @@
...
+            to: address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into(),
+            ..Default::default()
+        };
+
+        let tx = provider.send_transaction(tx).await.unwrap();
+
+        let tx = tx.get_receipt().await.unwrap();
+
+        let set_gas_price = U128::from(0x77359400);
+        let set_gas_limit = U256::from(0x5208);
```
gakonst, line 121: We should make the tx type have the `u128` and `u64` types where possible i think instead of `U128`?

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,126 @@
...
+        mut tx: N::TransactionRequest,
+    ) -> TransportResult<PendingTransactionBuilder<'_, N, T>> {
+        if tx.gas_price().is_none() {
+            let gas_price = self.get_gas_price().await?;
+            tx.set_gas_price(gas_price);
+        }
+        if tx.gas_limit().is_none() {
+            let gas_estimate = self.get_gas_estimate(&tx).await?;
+            tx.set_gas_limit(gas_estimate);
+        }
```
gakonst, line 84: this should be done with `match` and `join`'ing the 2 futures if both are present, to make it faster.

i also think we should do this for `nonce`, and perhaps this should be a general layer that fills both all fields

gakonst (changes requested):

crates/provider/src/layers/gas.rs
yash-atreya: Could be done but I think that should be a separate PR. Bunch of places where we're using U128 and U64.

crates/provider/src/layers/gas.rs
yash-atreya: > i also think we should do this for `nonce`, and perhaps this should be a general layer that fills both all fields
So make generalized tx builder layer and do nonce management and gas estimation in one?

gakonst: Yes I think so cc @prestwich @DaniPopes @mattsse for their view

crates/provider/src/layers/gas.rs
gakonst: yeah separate OK

crates/provider/src/layers/gas.rs
yash-atreya: Using `join!` now.

yash-atreya:
Decided to make this a more generic tx builder layer. Ref: https://github.com/alloy-rs/alloy/pull/326#discussion_r1526791175

Renaming it to `FillTxLayer`.
This layer will manage nonces and estimates the gas.

yash-atreya:
> Decided to make this a more generic tx builder layer. Ref: [#326 (comment)](https://github.com/alloy-rs/alloy/pull/326#discussion_r1526791175)
> 
> Renaming it to `FillTxLayer`. This layer will manage nonces and estimates the gas.

Undo above.

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,207 @@
...
+    async fn send_transaction(
+        &self,
+        mut tx: N::TransactionRequest,
+    ) -> TransportResult<PendingTransactionBuilder<'_, N, T>> {
+        let (gas_price, gas_limit, eip1559_fees) = match (
+            tx.gas_price().is_none(),
+            tx.gas_limit().is_none(),
+            (tx.max_fee_per_gas().is_none() && tx.max_priority_fee_per_gas().is_none()),
+        ) {
+            (true, true, true) => {
```
onbjerg, line 82: pls add comments to make it clearer what tx types we are filling

onbjerg (changes requested):

crates/provider/src/layers/gas.rs
yash-atreya: @onbjerg by tx types do you mean `EIP1559`, or `legacy` or do you mean specifying the fields being populated?

prestwich: for this, we can assume that there are 2 intended ways:
- 1559 style (assumed default)
- legacy style (IF and ONLY IF the user has already set the `gas_price` key).

the builder should have a type-level guarantee that `gas_price` and the 1559 gas fields are mutually exclusive, i.e. `(true,true,true)` should never be permitted at all.

as a result we should be able to simplify the 8 match arms to an `if let Some(price) = tx.gas_price` or similar

prestwich: in the future, we might want to allow a network-associated `GasStrategy` to handle this 🤔

onbjerg: I don't quite understand this part though, if this is supposed to fill `gas_price` or `max_fee_per_gas`/`max_prio_fee_per_gas`, would they not always be unset? i.e. it would never be a legacy tx unless you set a gas price, in which case, we probably shouldn't override it?

onbjerg: > @onbjerg by tx types do you mean EIP1559, or legacy or do you mean specifying the fields being populated?

i mean, please add comments on the arms to explain what we are matching for, it is quite hard to parse what `(true, true, false)` etc means

yash-atreya: > I don't quite understand this part though, if this is supposed to fill `gas_price` or `max_fee_per_gas`/`max_prio_fee_per_gas`, would they not always be unset? i.e. it would never be a legacy tx unless you set the `gas_price` field, in which case, we probably shouldn't override it?

Agree with @onbjerg's, point here. 

How about?
```rust
if tx.gas_price().is_none() {
   // Assume its a EIP1559 tx
   // Populate the following fields: gas_limit, max_fee_per_gas and max_priority_fee_per_gas 
} else {
   // Assume its a legacy tx
   // Populate only the gas_limit field if unset.
 }
```

prestwich: > > I don't quite understand this part though, if this is supposed to fill `gas_price` or `max_fee_per_gas`/`max_prio_fee_per_gas`, would they not always be unset? i.e. it would never be a legacy tx unless you set the `gas_price` field, in which case, we probably shouldn't override it?
> 
> Agree with @onbjerg's, point here.
> 
> How about?
> 
> ```rust
> if tx.gas_price().is_none() {
>    // Assume its a EIP1559 tx
>    // Populate the following fields: gas_limit, max_fee_per_gas and max_priority_fee_per_gas 
> } else {
>    // Assume its a legacy tx
>    // Populate only the gas_limit field if unset.
>  }
> ```

Yes, this is the correct strategy. Recommend decomposing into `fill_legacy` and `fill_1559` functions

yash-atreya: Pushed changes. 

In this update, if the `gas_price` is not set (i.e assumed to be 1559 tx) but the network doesn't support eip1559 txs then it falls back to being process as a legacy tx.

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,367 @@
...
+    ) -> Result<N::TransactionRequest, TransportError> {
+        // Check if the user has set the gas_limit, max_fee_per_gas and max_priority_fee_per_gas
+        // fields.
+        let gas_limit_unset = tx.gas_limit().is_none();
+        // eip1559_fees_unset is true if any one of the two max_fee_per_gas or
+        // max_priority_fee_per_gas is unset.
+        let eip1559_fees_unset =
+            tx.max_fee_per_gas().is_none() || tx.max_priority_fee_per_gas().is_none();
+
+        match (gas_limit_unset, eip1559_fees_unset) {
```
prestwich, line 108: remove the combinatoric match please

follow the pattern in `populate_gas` and `populate_gas_eip1559` in `provider.rs`

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,367 @@
...
+                Ok(tx)
+            }
+        }
+    }
+
+    /// Populates the gas_price and only populates the gas_limit field if unset.
+    /// This method always assumes that the gas_price is unset.
+    async fn handle_legacy_tx(
+        &self,
+        mut tx: N::TransactionRequest,
```
prestwich, line 199: `&mut`

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,367 @@
...
+    async fn get_eip1559_fees_estimate(&self) -> TransportResult<(U256, U256)> {
+        self.inner.estimate_eip1559_fees(None).await
+    }
+
+    /// Populates the gas_limit, max_fee_per_gas and max_priority_fee_per_gas fields if unset.
+    /// Requires the chain_id to be set in the transaction request to be processed as a EIP-1559 tx.
+    /// If the network does not support EIP-1559, it will process it as a legacy tx.
+    async fn handle_eip1559_tx(
+        &self,
+        mut tx: N::TransactionRequest,
```
prestwich, line 98: &mut

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,367 @@
...
+    async fn get_eip1559_fees_estimate(&self) -> TransportResult<(U256, U256)> {
+        self.inner.estimate_eip1559_fees(None).await
+    }
+
+    /// Populates the gas_limit, max_fee_per_gas and max_priority_fee_per_gas fields if unset.
+    /// Requires the chain_id to be set in the transaction request to be processed as a EIP-1559 tx.
+    /// If the network does not support EIP-1559, it will process it as a legacy tx.
+    async fn handle_eip1559_tx<'a>(
+        &'a self,
+        tx: &'a mut N::TransactionRequest,
```
prestwich, line 98: these must have different lifetimes

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,367 @@
...
+                Ok(tx)
+            }
+        }
+    }
+
+    /// Populates the gas_price and only populates the gas_limit field if unset.
+    /// This method always assumes that the gas_price is unset.
+    async fn handle_legacy_tx<'a>(
+        &'a self,
+        tx: &'a mut N::TransactionRequest,
```
prestwich, line 199: different lifetimes

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    /// Populates the gas_price and only populates the gas_limit field if unset.
+    /// This method always assumes that the gas_price is unset.
+    async fn handle_legacy_tx<'a, 'b>(
+        &'a self,
+        tx: &'b mut N::TransactionRequest,
+    ) -> Result<&'b mut N::TransactionRequest, TransportError> {
+        let gas_price = self.get_gas_price();
```
prestwich, line 135: here's a nice little pattern for future usage:

```
        let gas_limit_fut = if let Some(gas_limit) = tx.gas_limit() {
            futures::future::Either::Left(async move { gas_limit })
        } else {
            futures::future::Either::Right(async { self.estimate_gas(&*tx, None).await })
        };
```

prestwich: with FuturesExt trait

```
let gas_limit_fut = tx.gas_limit() {
            async move { gas_limit }.left_future()
        } else {
            async { self.estimate_gas(&*tx, None).await }.right_future()
        };
```

prestwich: that way you can run the `join!` without all the clunky if block

yash-atreya: This was a really cool thing I learned! Thanks.

crates/provider/src/layers/gas.rs
```diff
@@ -144,30 +144,20 @@ where
...
-            match gas_price {
-                Ok(gas_price) => {
-                    tx.set_gas_price(gas_price);
-                    Ok(tx)
-                }
-                Err(err) => Err(err),
+            async { self.get_gas_estimate(tx).await }.right_future()
+        };
+
+        match futures::try_join!(gas_price_fut, gas_limit_fut) {
```
prestwich, line 154: this `match` block is equivalent to a `.map` right?

yash-atreya: Oh yes!

crates/provider/src/layers/gas.rs
```diff
@@ -144,30 +144,20 @@ where
         &'a self,
         tx: &'b mut N::TransactionRequest,
     ) -> Result<&'b mut N::TransactionRequest, TransportError> {
```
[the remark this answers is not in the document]
yash-atreya: The arg is `&mut`, and returning a `N::TransactionRequest` would require cloning.

prestwich (approved):

crates/provider/src/layers/gas.rs
```diff
@@ -0,0 +1,296 @@
...
+        match futures::try_join!(gas_estimate_fut, eip1559_fees_fut) {
+            Ok((gas_limit, eip1559_fees)) => {
+                tx.set_gas_limit(gas_limit);
+                tx.set_max_fee_per_gas(eip1559_fees.max_fee_per_gas);
+                tx.set_max_priority_fee_per_gas(eip1559_fees.max_priority_fee_per_gas);
+                Ok(())
+            }
+            Err(err) => {
+                if err.is_transport_error()
+                    && err.to_string() == *"EIP-1559 not activated".to_string()
```
gakonst, line 127: I'd love this to a helper method on the error re 1559 but we can do as followup

gakonst (approved):

crates/provider/src/layers/gas.rs
prestwich: actually this is a bit of a hole in the API. 1559 not activated should not be a transport error. "endpoint doesn't exist" falls in a hole between transport error and rpc error 🤔

PR #330, opened 2024-03-16
Fix subscribe blocks

0xc0ffeebabeeth:
## Motivation
Some RPC Providers (Alchemy and Reth were tested) don't have the "uncles" field in their newHeads subscription, this breaks the subscribe_blocks() stream.

## Solution

Providing the serde(default) flag on the uncles field allows graceful handling of both existing and missing uncle fields in the subscription response.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -233,7 +233,7 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
     /// # }
     /// ```
     #[cfg(feature = "pubsub")]
-    async fn subscribe_blocks(&self) -> TransportResult<Subscription<Block>> {
+    async fn subscribe_blocks(&self) -> TransportResult<Subscription<N::HeaderResponse>> {
         self.root().pubsub_frontend()?;
         let id = self.client().request("eth_subscribe", ("newHeads",)).await?;
         self.root().get_subscription(id).await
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):
Thanks! Ref https://t.me/ethers_rs/36155
Pending @mattsse or @prestwich, should this also be `skip_serializing_if`?

prestwich:
> Thanks! Ref https://t.me/ethers_rs/36155 Pending @mattsse or @prestwich, should this also be `skip_serializing_if`?

Given that we are post-merge, does it now ALWAYS omit that field in the `newHeads` objects? The field is present in `getBlockByNumber` even for recent blocks, so the `newHeads` notification object is NOT a valid block? cc @mattsse what's reth do?

prestwich:
my initial take is that we should NOT do `skip_serializing_if`, and just leave it as `default`

DaniPopes:
Actually should this be `Header` instead of block?

@prestwich https://github.com/paradigmxyz/reth/blob/4e49aa9df66dfc19907c7e788f1817e83f2f61c9/crates/rpc/rpc/src/eth/pubsub.rs#L292

prestwich:
> Actually should this be Header instead of block?

this is quite likely

0xc0ffeebabeeth:
The newHeads object does seem to contain withdrawals though, which aren't included in the Header object

prestwich:
so its a separate object? 😮‍💨

DaniPopes:
most consistent eth rpc api

DaniPopes (approved):

prestwich (approved):

PR #331, opened 2024-03-17
fix: force clippy to stable

prestwich:
# Motivation

Prevent future random CI breakage

# Solution

Set clippy to stable in CI

there may be a better way to achieve this, but this way is straightforward

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
lgtm since we decided to do it in the other repo

mattsse (approved):

PR #337, opened 2024-03-18
fix: make `TransactionReceipt::transaction_hash` field mandatory

klkvr:
## Motivation

Option initially came from Reth, but it should be mandatory

Ref https://github.com/ethereum/go-ethereum/blob/ba2dd9385c2a51134e520083dc732787a813b107/core/types/receipt.go#L62-L62

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/signer.rs
+++ b/crates/provider/src/layers/signer.rs
@@ -144,7 +144,7 @@ mod tests {
 
         let receipt =
             provider.get_transaction_receipt(local_hash2).await.unwrap().expect("no receipt");
-        let receipt_hash = receipt.transaction_hash.expect("no receipt hash");
+        let receipt_hash = receipt.transaction_hash;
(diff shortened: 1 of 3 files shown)


prestwich (commented):
before we make this change, there were some concessions made to the goofy optimism deposit transaction format. are we absolutely certain that this is not one of those concessions?

klkvr:
per https://specs.optimism.io/protocol/deposits.html#deposit-receipt:
>Transaction receipts use standard typing as per [EIP-2718](https://eips.ethereum.org/EIPS/eip-2718). The Deposit transaction receipt type is equal to a regular receipt, but extended with an optional depositNonce field.

DaniPopes (approved):

mattsse (approved):
yep, this should not be optional.

a tx must have a hash

PR #338, opened 2024-03-18
feat: `impl From<Transaction> for TransactionRequest` + small type updates

klkvr:
## Motivation

This was in ethers and is being used in foundry: https://github.com/foundry-rs/foundry/blob/master/crates/common/src/transactions.rs#L43

`Transaction` does not include sidecar so this conversion is not lossless

I've also updated several field types on `Transaction` so they match `TransactionRequest` types. 

- Gas price related fields were updated to U256
- `Transaction::chain_id` changed to u64 to match `TransactionRequest`
- `Transaction::access_list` changed to use `AccessList` struct
- `Transaction::transaction_type` changed to `U8` (Ethereum spec defines it as a number <=127, Optimism deposit transaction has type 126).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/request.rs
+++ b/crates/rpc-types/src/eth/transaction/request.rs
@@ -1,9 +1,9 @@
 //! Alloy basic Transaction Request type.
 
-use alloy_primitives::{Address, Bytes, ChainId, B256, U256, U8};
 use crate::{
     eth::transaction::AccessList, other::OtherFields, BlobTransactionSidecar, Transaction,
 };
+use alloy_primitives::{Address, Bytes, ChainId, B256, U256, U8};
 use serde::{Deserialize, Serialize};
 use std::hash::Hash;
 


mattsse (approved):
lgtm,

pending @onbjerg

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -237,6 +239,28 @@ impl From<Option<Bytes>> for TransactionInput {
     }
 }
 
+impl From<Transaction> for TransactionRequest {
+    fn from(tx: Transaction) -> TransactionRequest {
```
mattsse, line 243: could we add a function for this on Transaction as well?
this way we can properly document the limitations?

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -43,18 +43,18 @@ pub struct Transaction {
...
-    pub max_fee_per_gas: Option<U128>,
+    pub max_fee_per_gas: Option<U256>,
     /// The miner's tip.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_priority_fee_per_gas: Option<U128>,
+    pub max_priority_fee_per_gas: Option<U256>,
     /// Configured max fee per blob gas for eip-4844 transactions
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub max_fee_per_blob_gas: Option<U128>,
+    pub max_fee_per_blob_gas: Option<U256>,
```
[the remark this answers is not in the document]
klkvr: I haven't found this data in spec for gasPrice, maxFeePerGas, etc

but 4844 spec says that maxFeePerBlobGas is uint256: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md#blob-transaction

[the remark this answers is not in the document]
mattsse: yeah it's not really consistent, so we should stick to the spec and use u256

crates/rpc-types/src/eth/transaction/request.rs
klkvr: I think we can, but this would have to be `TryFrom` as `TransactionRequest` mostly has `Option`s

klkvr: Or you meant not the reverse conversion function but something like `Transaction::to_request`?

mattsse: yeah, which we then can reuse as the From impl, 
the benefit of this is that there's now a dedicated function we can document

mattsse (approved):
lgtm

pending @prestwich @onbjerg

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -86,6 +87,33 @@ pub struct Transaction {
     pub other: OtherFields,
 }
 
+impl Transaction {
+    /// Converts [Transaction] into [TransactionRequest].
+    ///
+    /// During this conversion data for [TransactionRequest::sidecar] is not populated as it is not
+    /// part of [Transaction].
+    pub fn into_request(self) -> TransactionRequest {
```
onbjerg, line 95: any reason we have this instead of just the auto-impl'd `Into` we get from `From`?

klkvr: We wanted to have it documented on `Transaction`  

Ref https://github.com/alloy-rs/alloy/pull/338#discussion_r1529052619

crates/contract/src/event.rs
```diff
@@ -65,6 +65,15 @@ impl<N: Network, T: Transport + Clone, P: Provider<N, T>, E: SolEvent> Event<N,
         let poller = self.provider.watch_logs(&self.filter).await?;
         Ok(EventPoller::new(poller))
     }
+
+    /// Subscribes to the stream of events that match the filter.
+    ///
+    /// Returns a stream of decoded events and raw logs.
+    #[cfg(feature = "pubsub")]
```
mattsse, line 72: looks like something went wrong while rebasing this branch?

mattsse (approved):

onbjerg (approved):

PR #339, opened 2024-03-19
feat: `Provider::subscribe_logs`

klkvr:
## Motivation

Adds `Provider::subscribe_logs` and `Event::subscribe` methods

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (changes requested):
lgtm, I would want `Event::subscribe` too

onbjerg (approved):
lgtm

mattsse (approved):
lgtm

pending @DaniPopes

prestwich (commented):
@DaniPopes can we somehow unify the usage of `Subscription<T>` and `PollChannel<T>` to have a single front-end that works for both polling and pubsub?

crates/contract/src/event.rs
```diff
@@ -154,7 +220,7 @@ mod tests {
     #[tokio::test]
     async fn event_filters() {
         init_tracing();
-        let (provider, _anvil) = spawn_anvil();
+        let (provider, anvil) = spawn_anvil();
```
DaniPopes, line 223: Keep underscore for `not(feature = "pubsub")`

```suggestion
        let (provider, _anvil) = spawn_anvil();
```

DaniPopes (approved):

DaniPopes:
@prestwich I think so, but I feel that would complicate things quite a bit for not much benefit?

prestwich:
it means that users would follow the same API regardless of provider and get automatically upgraded to pubsub subscriptions when they change endpoints (unless they specifically want to downgrade to polling for some reason)

onbjerg:
i agree w prestwich here, this would most likely also give us transport stacking back?

crates/contract/src/event.rs
```diff
@@ -131,10 +140,67 @@ fn decode_log<E: SolEvent>(log: &Log) -> alloy_sol_types::Result<E> {
...
+                .field("event_type", &format_args!("{}", std::any::type_name::<E>()))
+                .finish()
+        }
+    }
+
+    impl<E: SolEvent> EventSubscription<E> {
+        /// Creates a new event subscription with the provided subscription.
+        #[allow(clippy::missing_const_for_fn)]
+        #[inline]
+        pub fn new(sub: Subscription<Log>) -> Self {
```
prestwich, line 186: `From` impl preferred here

crates/contract/src/event.rs
```diff
@@ -131,10 +140,67 @@ fn decode_log<E: SolEvent>(log: &Log) -> alloy_sol_types::Result<E> {
...
+            Self { sub, _phantom: PhantomData }
+        }
+
+        /// Converts the subscription into a stream.
```
prestwich, line 190: This results in an unnameable type. Is there any situation in which a user would want to name the resulting stream?

prestwich (changes requested):

prestwich:
Regardless, that's not in scope for this PR. Opened #346 to capture.

crates/contract/src/event.rs
```diff
@@ -131,10 +140,67 @@ fn decode_log<E: SolEvent>(log: &Log) -> alloy_sol_types::Result<E> {
...
+    /// An event subscription.
+    ///
+    /// Underlying subscription is available through the [`sub`](Self::sub) field.
+    pub struct EventSubscription<E> {
+        /// The inner poller.
+        pub sub: Subscription<Log>,
+        _phantom: PhantomData<E>,
+    }
+
+    impl<E> std::ops::Deref for EventSubscription<E> {
```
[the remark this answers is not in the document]
klkvr: I think we can, @DaniPopes wdyt?

crates/contract/src/event.rs
klkvr: added for both `EventPoller` and `EventSubscription` in d37f0503918e7a7d9e953e664c3b4293f25a1bad

crates/contract/src/event.rs
mattsse: I guess we want a similar thing as `SubscriptionStream` here with the type we decode on poll_next?

prestwich: ideally we have these all work the same way. can convert this to issue?

crates/contract/Cargo.toml
```diff
@@ -26,10 +26,15 @@ futures-util.workspace = true
...
-alloy-rpc-client.workspace = true
+alloy-rpc-client = { workspace = true, features = ["pubsub", "ws"] }
 alloy-transport-http.workspace = true
 alloy-node-bindings.workspace = true
 test-utils.workspace = true
 reqwest.workspace = true
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
+[features]
+pubsub = ["alloy-provider/pubsub", "dep:alloy-pubsub"]
```
prestwich, line 40: nit: should have a newline

crates/contract/src/event.rs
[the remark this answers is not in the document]
DaniPopes: no real reason other than allowing `x.set_thing` rather than `x.inner.set_thing`

prestwich:
re-running CI before merge

PR #341, opened 2024-03-19
Fix: change nonce from `U64` to `u64` 

### LINKED ISSUE #335: [Suggestion] Change `nonce` field to be `u64` compatible
### Component

provider, pubsub

Instead of `.nonce(U64::from(1))` we would want to be able to do just `.nonce(1)`.

See: https://github.com/alloy-rs/examples/pull/2#discussion_r1526812484
See: https://github.com/alloy-rs/examples/pull/2#discussion_r1526813443
See: https://github.com/alloy-rs/examples/pull/8#discussion_r1526788047

There are likely also other places where similar user facing values are desired to be native types: 

- `chain_id`
- `transaction_type`
- etc..

See: https://github.com/alloy-rs/examples/pull/2#discussion_r1526813443

### Additional context

Once implemented make sure to fix in examples: https://github.com/alloy-rs/examples/pull/2

There is some complexity around this, FWD from @prestwich in https://github.com/alloy-rs/examples/pull/8#discussion_r1528910529:

>rpc types must store nonce as U64 (or use a serde(with = ...)) to be on-spec so this probably propagates outward from there 😮‍💨 we should settle on a specific type for these and audit codeplace for compliance. prefer u64 in all non-rpc structs

zerosnacks:
## Motivation

Closes #335 

## Solution

Uses Serde's `with =` attribute and the existing `u64_hex` / `u64_hex_opt` to make sure RPC structs stay compliant.

~~Opening this PR as a draft first to get initial feedback.~~ ~~If implementation is correct I can add the other fields (`chain_id`, `transaction_type`, u128 native fields?)~~ (largely implemented in #338)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -20,7 +20,7 @@
 use alloy_primitives::{Address, Bytes, B256, U256};
 use alloy_serde::{
     json_u256::{deserialize_json_ttd_opt, deserialize_json_u256},
-    num::{u64_hex_or_decimal, u64_hex_or_decimal_opt},
+    num::{u64_hex, u64_hex_opt, u64_hex_or_decimal, u64_hex_or_decimal_opt},
     storage::deserialize_storage_map,
 };
 use serde::{Deserialize, Serialize};
@@ -34,7 +34,7 @@ pub struct Genesis {
     #[serde(default)]
     pub config: ChainConfig,
     /// The genesis header nonce.
-    #[serde(with = "u64_hex_or_decimal")]
+    #[serde(with = "u64_hex")]
     pub nonce: u64,
     /// The genesis header timestamp.
     #[serde(with = "u64_hex_or_decimal")]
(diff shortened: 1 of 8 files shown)


crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -28,7 +29,8 @@ pub struct Transaction {
     /// Hash
     pub hash: B256,
     /// Nonce
-    pub nonce: U64,
+    #[serde(with = "u64_hex_or_decimal")]
```
mattsse, line 32: this should just be u64 hex, not decimal

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -33,7 +34,8 @@ pub struct TransactionRequest {
     #[serde(default, flatten)]
     pub input: TransactionInput,
     /// The nonce of the transaction.
-    pub nonce: Option<U64>,
+    #[serde(with = "u64_hex_or_decimal_opt")]
```
mattsse, line 37:
```suggestion
    #[serde(with = "alloy_serde::u64_hex_opt")]
```

mattsse (approved):
lgtm

crates/genesis/src/lib.rs
```diff
@@ -34,7 +34,7 @@ pub struct Genesis {
     #[serde(default)]
     pub config: ChainConfig,
     /// The genesis header nonce.
-    #[serde(with = "u64_hex_or_decimal")]
+    #[serde(with = "u64_hex")]
```
mattsse, line 37: these we need to keep, because the genesis format is a bit cursed and both variants are possible

mattsse (changes requested):

crates/genesis/src/lib.rs
zerosnacks: reverted the changes

mattsse (approved):

Issue #318, opened 2024-03-14
[Feature] Provider::connect API


yash-atreya:
### Component

provider, pubsub

### Motivation

Currently the DX of instantiating `Provider`'s is cumbersome. 

In order to initialize a simple `HttpProvider` a user needs to do the following: 
`RpcClient` which needs a `Transport` type (http / ws)  which needs a reqwest client. 

We also need to specify the `Network` as a generic.

```rust
fn init() -> (HttpProvider<Ethereum>, AnvilInstance) {
    let anvil = Anvil::new().fork("https://eth.merkle.io").spawn();
    let url = "https://eth.merkle.io".parse().unwrap();
    let http = Http::<Client>::new(url);
    (HttpProvider::new(RpcClient::new(http, true)), anvil)
}
```

The above applies to instantiating WS and IPC providers as well.

### Solution

Implement a simpler API such that a user can just pass a RPC_URL (http/ws) or IPC_PATH and a provider is generated accordingly. 

Ideal interface to achieve this UX is implementing a `connect` method in `Provider` that accepts a `String` or `Url` type and returns a `Provider`.

```rust
let http_provider = Provider::connect("http://eth.merkle.io")
let ws_provider = Provider::connect("was://eth.ws-provider.io")
let ipc_provider = Provider::connect("/tmp/reth.ipc")
```

prestwich:
```
let http_provider = Provider::connect("http://eth.merkle.io")
let ws_provider = Provider::connect("was://eth.ws-provider.io")
let ipc_provider = Provider::connect("/tmp/reth.ipc")
```

this is not an achievable API unless the type of all 3 contains `BoxTransport` instead of the actual transport types.

DaniPopes:
@prestwich `ProviderBuilder::connect -> Provider<BoxedTransport>`?

prestwich:
not a good api, because that would lead to users using the connect function, and then accidentally blowing up their alchemy bill via polling instead of subscribing over ws. We need to force the user to know their transport type, unless they're specifically intending to box it.

yash-atreya:
> not a good api, because that would lead to users using the connect function, and then accidentally blowing up their alchemy bill via polling instead of subscribing over ws. We need to force the user to know their transport type, unless they're specifically intending to box it.

To that point, we could have `Provider::connect_http(&str) -> RootProvider<BoxedTransport>` and it would still be better than what it is right now and also force the users to know what they are using.

[...]

PR #342, opened 2024-03-19
feat(providers): connect_boxed api

yash-atreya:
## Motivation

Ref: #318 

## Solution

Implement a `connect_boxed(&str) -> RootProvider<N, BoxTransport>` method for `RootProvider<N, T>`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/provider/src/utils.rs
```diff
@@ -39,6 +39,17 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
     (potential_max_fee, max_priority_fee_per_gas)
 }
 
+/// Identifies the intended transport type from the given string.
+pub fn parse_str_to_tranport_type(s: &str) -> &str {
```
prestwich, line 43: this should return an error instead of invalid output

crates/provider/src/provider.rs
```diff
@@ -69,6 +71,41 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+    /// Creates a new root provider with a boxed transport from a string.
+    /// BoxTransport is used to allow for different transport types.
+    pub async fn connect_boxed(conn_str: &str) -> TransportResult<RootProvider<N, BoxTransport>> {
+        RootProvider::<N, _>::connect(conn_str).await
+    }
+}
+
+impl<N: Network> RootProvider<N, BoxTransport> {
+    /// Creates a new root provider with a boxed transport from a string.
+    pub async fn connect(conn_str: &str) -> TransportResult<RootProvider<N, BoxTransport>> {
```
prestwich, line 84: `connect_boxed`

crates/provider/src/provider.rs
```diff
@@ -69,6 +71,41 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+        match utils::parse_str_to_tranport_type(conn_str) {
+            "http" => {
+                let provider =
+                    RootProvider::<N, _>::new_http(reqwest::Url::parse(conn_str).unwrap());
```
prestwich, line 88: can panic on malformatted input. should not be able to panic

prestwich (changes requested):

Cargo.toml
```diff
@@ -25,8 +25,10 @@ alloy-json-rpc = { version = "0.1.0", default-features = false, path = "crates/j
 alloy-network = { version = "0.1.0", default-features = false, path = "crates/network" }
 alloy-node-bindings = { version = "0.1.0", default-features = false, path = "crates/node-bindings" }
 alloy-provider = { version = "0.1.0", default-features = false, path = "crates/provider" }
-alloy-pubsub = { version = "0.1.0", default-features = false, path = "crates/pubsub" }
-alloy-rpc-client = { version = "0.1.0", default-features = false, path = "crates/rpc-client" }
+alloy-pubsub = { version = "0.1.0", path = "crates/pubsub" }
+alloy-rpc-client = { version = "0.1.0", features = [
```
prestwich, line 29: is this necessary for all usages of provider?

crates/provider/src/provider.rs
```diff
@@ -69,6 +71,41 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+                let ws_client = RpcClient::connect_pubsub(ws).await?;
+
+                Ok(RootProvider::<N, _>::new(ws_client).boxed())
+            }
+            "ipc" => {
+                let ipc = IpcConnect::new(conn_str.to_string());
+
+                let ipc_client = RpcClient::connect_pubsub(ipc).await?;
+
+                Ok(RootProvider::<N, _>::new(ipc_client).boxed())
```
prestwich, line 104: move instantiation of the provider out of the match block

```
let boxed_transport = match { ... };

RootProvider::new(boxed_transport)
```

crates/provider/src/utils.rs
```diff
@@ -39,6 +39,17 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
...
+    if s.starts_with("http://") || s.starts_with("https://") {
+        "http"
+    } else if s.starts_with("ws://") || s.starts_with("wss://") {
+        "ws"
+    } else {
+        "ipc"
```
mattsse, line 49: this should check if the file exists

crates/provider/src/utils.rs
```diff
@@ -39,6 +39,17 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
     (potential_max_fee, max_priority_fee_per_gas)
 }
 
+/// Identifies the intended transport type from the given string.
+pub fn parse_str_to_tranport_type(s: &str) -> &str {
+    if s.starts_with("http://") || s.starts_with("https://") {
+        "http"
```
mattsse, line 45: I think we also want to check for things like "localhost:port"
or if the input is a `SocketAddress`, then prefix with http,

so the Http variant should contain a String

mattsse (changes requested):

crates/provider/src/utils.rs
prestwich: also need to extract any auth info for WsConnect

some useful methods:
https://docs.rs/url/latest/url/struct.Url.html#method.has_authority
https://docs.rs/url/latest/url/struct.Url.html#method.origin
https://docs.rs/url/latest/url/struct.Url.html#method.scheme

prestwich: do we have `parse` implementations on the relevant `Connect` structs?

crates/provider/src/utils.rs
```diff
@@ -48,6 +53,63 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
...
+            Ok(BuiltInTransportType::Http(s.to_string()))
+        } else if s.starts_with("ws://") || s.starts_with("wss://") {
+            Ok(BuiltInTransportType::Ws)
+        } else {
+            Ok(BuiltInTransportType::Ipc)
+        }
+    }
+}
+
+impl Clone for BuiltInTransportType {
```
gakonst, line 103: Derive?

crates/rpc-client/tests/it/ws.rs
```diff
@@ -1,7 +1,7 @@
 use alloy_node_bindings::Anvil;
 use alloy_primitives::U64;
 use alloy_rpc_client::{ClientBuilder, RpcCall};
-use alloy_transport_ws::WsConnect;
+use alloy_transport::WsConnect;
```
gakonst, line 4: Failing in CI

Cargo.toml
prestwich: don't enable features in root cargo.toml

crates/provider/src/utils.rs
```diff
@@ -48,6 +53,63 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
...
+    Ws,
+    /// IPC transport.
+    Ipc,
+}
+
+impl FromStr for BuiltInTransportType {
+    type Err = RpcError<TransportErrorKind>;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.starts_with("http://") || s.starts_with("https://") {
```
prestwich, line 93: shouldn't this impl use `parse_str_to_tranport_type` ?

crates/provider/src/utils.rs
```diff
@@ -48,6 +53,63 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
...
+    // Return the transport type or error
+    BuiltInTransportType::from_str(&s)
+        .map_err(|_| TransportErrorKind::custom_str("Invalid transport type"))
+}
+
+/// The built-in transport types.
+#[derive(Debug, PartialEq, Eq)]
+pub enum BuiltInTransportType {
+    /// HTTP transport.
+    Http(String),
```
prestwich, line 82: ideally, each variant should hold the authenticated connection information. We may also want to implement a `fn connect(self) -> BoxTransport` for this. The ideal experience here would be

```
let transport: BoxTransport = my_str.parse::<BuiltinTransportDetails>()?.connect();

File: crates/provider/src/utils.rs [hunk 12]
```diff
@@ -91,4 +153,56 @@ mod tests {
...
+        assert_eq!(
+            super::parse_str_to_tranport_type("http://127.0.0.1:8545").unwrap(),
+            BuiltInTransportType::Http("http://127.0.0.1:8545".to_string())
+        );
+        assert_eq!(
+            super::parse_str_to_tranport_type("ws://127.0.0.1:8545").unwrap(),
+            BuiltInTransportType::Ws
+        );
+        assert_eq!(
+            super::parse_str_to_tranport_type("http://localhost").unwrap(),
```
prestwich, line 192: add test with HTTP basic auth `user:password` pairs

crates/provider/src/provider.rs
```diff
@@ -69,6 +74,42 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
     pub fn new(client: RpcClient<T>) -> Self {
         Self { inner: Arc::new(RootProviderInner::new(client)) }
     }
+
+    /// Creates a new root provider with a boxed transport from a string.
+    pub async fn connect_boxed(conn_str: &str) -> TransportResult<RootProvider<N, BoxTransport>> {
+        let boxed_client = match BuiltInTransportType::from_str(conn_str) {
+            Ok(BuiltInTransportType::Http(conn_str)) => {
+                RpcClient::new_http(reqwest::Url::parse(&conn_str).unwrap()).boxed()
```
mattsse, line 82: no unwraps please

crates/provider/src/provider.rs
```diff
@@ -69,6 +74,42 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+            }
+            Ok(BuiltInTransportType::Ws(conn_str)) => {
+                // Extract auth info if any
+                let url = reqwest::Url::parse(&conn_str).unwrap();
```
mattsse, line 86: same

crates/provider/src/provider.rs
```diff
@@ -69,6 +74,42 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+                    Some(Authorization::basic(username, pass))
+                } else {
+                    None
+                };
+
+                let ws = WsConnect::with_auth(conn_str, auth);
+
+                let ws_client = RpcClient::connect_pubsub(ws).await?;
+
+                ws_client.boxed()
```
mattsse, line 99: this should be extracted to new function

crates/provider/src/provider.rs
```diff
@@ -69,6 +74,42 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
...
+            }
+            Ok(BuiltInTransportType::Ipc(conn_str)) => {
+                let ipc = IpcConnect::new(conn_str);
+
+                let ipc_client = RpcClient::connect_pubsub(ipc).await?;
+
+                ipc_client.boxed()
```
mattsse, line 106: same

crates/provider/src/utils.rs
```diff
@@ -48,6 +53,51 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
...
 }
 
+/// Identifies the intended transport type from the given string.
+pub fn parse_str_to_tranport_type(s: &str) -> Result<String, RpcError<TransportErrorKind>> {
+    // Parse if string is a socket address e.g. "127.0.0.1:8545"
+    let s = s.parse::<SocketAddr>().ok().map_or_else(|| s.to_string(), |s| format!("http://{}", s));
+
+    // Check if s is a path and it exists
+    let path = Path::new(&s);
+    if path.exists() && path.is_file() {
```
mattsse, line 63:
```suggestion
    if path.is_file() {
```

crates/provider/src/utils.rs
```diff
@@ -48,6 +53,51 @@ pub fn eip1559_default_estimator(base_fee_per_gas: U256, rewards: &[Vec<U256>])
     (potential_max_fee, max_priority_fee_per_gas)
 }
 
+/// Identifies the intended transport type from the given string.
+pub fn parse_str_to_tranport_type(s: &str) -> Result<String, RpcError<TransportErrorKind>> {
+    // Parse if string is a socket address e.g. "127.0.0.1:8545"
+    let s = s.parse::<SocketAddr>().ok().map_or_else(|| s.to_string(), |s| format!("http://{}", s));
+
```
mattsse, line 60: this can return early

mattsse (changes requested):

yash-atreya:
Need some help with fixing ci. Not sure why it's failing. 
`WsConnect` is to be imported from `alloy_transport_ws`.
That's how it is on `main` https://github.com/alloy-rs/alloy/blob/3108462e177741bbf2f72548fe1f84f98686b838/crates/rpc-client/tests/it/ws.rs#L4

crates/provider/Cargo.toml
```diff
@@ -14,12 +14,13 @@ exclude.workspace = true
 [dependencies]
 alloy-json-rpc.workspace = true
 alloy-network.workspace = true
-alloy-rpc-client = { workspace = true, features = ["reqwest"] }
+alloy-rpc-client = { workspace = true, features = ["reqwest", "pubsub"] }
```
prestwich, line 17: > Need some help with fixing ci. Not sure why it's failing. `WsConnect` is to be imported from `alloy_transport_ws`. That's how it is on `main`
> 
> https://github.com/alloy-rs/alloy/blob/3108462e177741bbf2f72548fe1f84f98686b838/crates/rpc-client/tests/it/ws.rs#L4

`dep:alloy-transport-ws` is under the `ws` feature, but the test is `cfg(feature = "pubsub")` so when the pubsub feature is active, but the `ws` feature is not, there's an error. This is triggered by this feature line here, I believe

prestwich: basically, this feature activation causes the cargo resolver to activate ONLY `pubsub` on alloy-rpc-client, and not `ws`, which causes the alloy-transport-ws to be missing when the test runs. This feature should not be activated here, it should instead be activated under the `pubsub` feature n this crate

we should also change the cfg in `rpc-client/tests/it/main.rs`

prestwich: summary:
- remove this feature activation
- update cfgs in `rpc-client/tests/it/main.rs`

prestwich: cc @mattsse @DaniPopes should `ws` and `ipc` be default features for the `provider` crate?

yash-atreya: > summary:
> 
> * remove this `pubsub` feature activation
> * update cfgs in `rpc-client/tests/it/main.rs`

We need the `pubsub` feature enabled on alloy-rpc-client as it is required for `connect_pubsub`

prestwich: that implies that `connect_pubsub` should be cfg gated behind the alloy-provider `pubsub` feature

prestwich: https://github.com/alloy-rs/alloy/pull/342/files/072ff5ea130fe966660af428ff380a9febaac698#diff-2ba22fae76dc97fc89caff6fcd110d232b06d71b6d3c94717cdf80aa178cbe02R49

prestwich: and the WS/IPC components also need cfg gating

prestwich (changes requested):
1 more feature: 

The `connect_boxed` shortcut should also exist in the  `ProviderBuilder` as `fn connect_boxed(self, s: &str) -> Result<L::Provider>`

crates/provider/Cargo.toml
yash-atreya: > that implies that `connect_pubsub` should be cfg gated behind the alloy-provider `pubsub` feature

If I'm not mistaken, this means disabling the `connect_boxed` method as it uses `connect_pubsub` for ws and ipc. 

Not seeing a way to gate connect_pubsub using the pubsub feature without sacrificing `connect_boxed`.

prestwich: it means changing the behavior to produce ONLY enabled trandport types. eg. only http when pubsub is disabled

crates/provider/src/provider.rs
```diff
@@ -69,6 +72,34 @@ impl<N: Network, T: Transport> RootProvider<N, T> {
     pub fn new(client: RpcClient<T>) -> Self {
         Self { inner: Arc::new(RootProviderInner::new(client)) }
     }
+
+    /// Creates a new root provider with a boxed transport from a string.
+    pub async fn connect_boxed(conn_str: &str) -> TransportResult<RootProvider<N, BoxTransport>> {
+        let boxed_client = match BuiltInTransportType::from_str(conn_str) {
```
prestwich, line 78: this match could have a `?`

crates/provider/src/utils.rs
```diff
@@ -52,6 +57,62 @@ pub fn eip1559_default_estimator(
...
+    if path.is_file() {
+        // IPC if it exists
+        return Ok(s.to_string());
+    }
+
+    // Parse the URL or return an error
+    Url::parse(s).map_err(TransportErrorKind::custom)?;
+
+    Ok(s.to_string())
+}
```
mattsse, line 87: what does this do?

this function seems redundant, should be integrated in fromstr

crates/provider/src/utils.rs
```diff
@@ -52,6 +57,62 @@ pub fn eip1559_default_estimator(
...
+
+/// The built-in transport types.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BuiltInTransportType {
+    /// HTTP transport.
+    Http(String),
```
mattsse, line 93: this should be URL

crates/provider/src/utils.rs
```diff
@@ -98,4 +159,70 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_parsing_urls() {
+        assert_eq!(
+            BuiltInTransportType::from_str("http://localhost:8545").unwrap(),
```
mattsse, line 166: needs test case for "localhost:8545"

crates/provider/src/utils.rs
```diff
@@ -52,6 +57,62 @@ pub fn eip1559_default_estimator(
     Eip1559Estimation { max_fee_per_gas: potential_max_fee, max_priority_fee_per_gas }
 }
 
+/// Extracts the authorization information from the given URL.
+pub fn extract_auth_info(url: Url) -> Option<Authorization> {
```
mattsse, line 61:
```suggestion
pub fn extract_auth_info(url: &Url) -> Option<Authorization> {
```

mattsse (changes requested):

prestwich:
did a significant cleanup here:
- `BuiltInConnectionString` -> alloy-rpc-client
- improve parsing
- `impl BoxTransportConnect for BuiltinConnectionString`
- passthrough features appropriately
- `ProviderBuilder::on_builtin` to attempt to connect to a builting transport type via the connection string

crates/rpc-client/src/builtin.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Tries to parse the given string as an IPC path, returning an error if
+    /// the path does not exist.
+    #[cfg(feature = "ipc")]
+    pub fn try_as_ipc(s: &str) -> Result<Self, TransportError> {
+        // Check if s is a path and it exists
+        let path = std::path::Path::new(&s);
+
+        path.is_file()
+            .then_some(Self::Ipc(s.to_string()))
+            .ok_or_else(|| TransportErrorKind::custom_str("Invalid IPC path. File does not exist."))
```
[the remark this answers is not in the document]
prestwich: done 🖖

crates/alloy/Cargo.toml
```diff
@@ -21,11 +21,11 @@ alloy-genesis = { workspace = true, default-features = false, optional = true }
 alloy-network = { workspace = true, default-features = false, optional = true }
 alloy-node-bindings = { workspace = true, default-features = false, optional = true }
 alloy-provider = { workspace = true, default-features = false, optional = true }
-alloy-pubsub = { workspace = true, default-features = false, optional = true }
+alloy-pubsub = { workspace = true, optional = true }
 
 # rpc
 alloy-json-rpc = { workspace = true, default-features = false, optional = true }
-alloy-rpc-client = { workspace = true, default-features = false, optional = true }
+alloy-rpc-client = { workspace = true, optional = true }
```
prestwich, line 28: this two lines now cause warnings as the root cargo toml doesn't set default-features

prestwich (approved):

prestwich (approved):

PR #343, opened 2024-03-19
feat: `impl TryFrom<Transaction> for TxEnvelope`

klkvr:
## Motivation

1. Adds conversion from `alloy_rpc_types::Signature` to `alloy_primitives::Signature`
2. Adds conversion from `alloy_rpc_types::AccessList` to `alloy_eips::AccessList`
3. Adds conversion from `alloy_rpc_types::Transaction` to `TxEnvelope`

Blocked by #338

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -15,8 +15,7 @@ exclude.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-eips.workspace = true
-alloy-rpc-types.workspace = true
-thiserror.workspace = true
+thiserror = { workspace = true, optional = true }
 c-kzg = { version = "1.0", features = ["serde"], optional = true }
 sha2 = { version = "0.10", optional = true }
 
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -15,6 +15,7 @@ exclude.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-eips.workspace = true
+alloy-rpc-types.workspace = true
 thiserror = { workspace = true, optional = true }
 c-kzg = { version = "1.0", features = ["serde"], optional = true }
 sha2 = { version = "0.10", optional = true }
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -16,7 +16,7 @@ alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-eips.workspace = true
 alloy-rpc-types.workspace = true
-thiserror = { workspace = true, optional = true }
+thiserror.workspace = true
 c-kzg = { version = "1.0", features = ["serde"], optional = true }
 sha2 = { version = "0.10", optional = true }
 
@@ -32,5 +32,5 @@ tokio = { workspace = true, features = ["macros"] }
 
 [features]
 k256 = ["alloy-primitives/k256"]
-kzg = ["dep:c-kzg", "dep:sha2", "dep:thiserror"]
+kzg = ["dep:c-kzg", "dep:sha2"]
 arbitrary = ["dep:arbitrary", "alloy-eips/arbitrary"]
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -31,5 +31,5 @@ tokio = { workspace = true, features = ["macros"] }
 
 [features]
 k256 = ["alloy-primitives/k256"]
-kzg = ["dep:c-kzg", "dep:sha2"]
+kzg = ["dep:c-kzg", "dep:sha2", "dep:thiserror"]
 arbitrary = ["dep:arbitrary", "alloy-eips/arbitrary"]
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -260,8 +260,8 @@ impl TryFrom<Transaction> for TxEnvelope {
         match tx.transaction_type.map(|t| t.to()) {
             None | Some(0) => {
                 let tx = TxLegacy {
-                    chain_id: tx.chain_id,
-                    nonce: tx.nonce.to(),
+                    chain_id: tx.chain_id.map(|c| c.to()),
+                    nonce: tx.nonce,
                     gas_price: tx.gas_price.ok_or(Eip2718Error::MissingGasPrice)?.to(),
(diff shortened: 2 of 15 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -245,6 +247,92 @@ impl Encodable2718 for TxEnvelope {
...
+    type Error = Eip2718Error;
+
+    fn try_from(tx: Transaction) -> Result<Self, Self::Error> {
+        let signature = tx
+            .signature
+            .ok_or(Eip2718Error::MissingSignature)?
+            .try_into()
+            .map_err(Eip2718Error::SignatureError)?;
+
+        match tx.transaction_type.map(|t| t.to()) {
```
prestwich, line 260: each match arm should delegate to another TryFrom impl

prestwich: should not match on integers, instead de-duplicate the `TransactionType` enums from consensus and rpc-types

crates/eips/src/eip2718.rs
```diff
@@ -18,19 +18,51 @@ const TX_TYPE_BYTE_MAX: u8 = 0x7f;
...
+    /// Missing `maxFeePerGas` field for EIP-1559 transaction.
+    MissingMaxFeePerGas,
+    /// Missing `maxPriorityFeePerGas` field for EIP-1559 transaction.
+    MissingMaxPriorityFeePerGas,
+    /// Missing `maxFeePerBlobGas` field for EIP-4844 transaction.
+    MissingMaxFeePerBlobGas,
+    /// Missing `chainId` field for EIP-155 transaction.
+    MissingChainId,
+    /// Signature error from [`alloy_primitives`].
+    SignatureError(SignatureError),
```
prestwich, line 42: none of these are 2718 errors. they should not be in this enum

klkvr: What'd be the best place for them?

crates/consensus/src/transaction/envelope.rs
klkvr: tx.transaction_type is `Option<U64>`:
https://github.com/alloy-rs/alloy/blob/218ece4ca60698c04f6fe29eda629add642aeebd/crates/rpc-types/src/eth/transaction/mod.rs#L76-L81

would it make sense to add `impl TryFrom<Option<U64>> for TxType`?

prestwich: missing type always implies 0, right? so you can use `unwrap_or_default()` on the option before converting.

klkvr: yep, but 0 seems to be treated as invalid type for this conversion

https://github.com/alloy-rs/alloy/blob/218ece4ca60698c04f6fe29eda629add642aeebd/crates/consensus/src/transaction/envelope.rs#L40-L51

prestwich: cc @Rjected some tools use `type: 0` to reflect legacy txns. shouldn't we support this?

crates/consensus/Cargo.toml
```diff
@@ -15,7 +15,8 @@ exclude.workspace = true
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-eips.workspace = true
-thiserror = { workspace = true, optional = true }
+alloy-rpc-types.workspace = true
```
prestwich, line 18: this dependency should not exist. RPC types should rely on consensus, not vice versa

crates/consensus/src/transaction/envelope.rs
Rjected: > cc @Rjected some tools use `type: 0` to reflect legacy txns. shouldn't we support this?

yes, just need to make sure that we still disallow decode attempts for "type 0" 2718 txs. But this can be done in decoding and not here, so try_from can support type 0

prestwich: dope. so this `try_from` can be updated because there is a corresponding check in the `impl Decodable2718 for TxEnvelope` match arm for the legacy type 👍

prestwich:
this could be done by embedding a `TxEnvelope` in the `Transaction`

klkvr:
wouldn't such approach cause issues with deposit transaction? we have optional signature for this case I believe

prestwich:
Optimism is on the hook for making their own rpc `Transaction` struct, which could have different semantics

mattsse:
> the other way to approach this would be to enforce completeness requirements during serde deser.

since all fields are public, I don't think we can guarantee this easily,
I think we want the validation checks when we convert from rpc -> consensus

prestwich:
> it's also possible that impls are faulty/incomplete for this endpoint, like missing field, I've seen this on a few newer networks, but this should not affect the rpc response imo and only be a problem when converting to consensus tx

we already have validity and correctness checks for the RPC response member data types, e.g. we reject an object in the `gas_limit` field. So why not check for type-level correctness? Is it just because we want to support faulty and/or off-spec nodes? That would imply we should significantly relax RPC deserialization overall, and seems like a non-goal anyway

prestwich:
well wed just need a _:() to prevent incorrect instantiation?

mattsse:
this only solves instantiation, but not mut access, for example:

```rust
let mut tx = p.tx_byhash(<>);
tx.max_fee_per_gas.take();
```

we could solve this by changing the rpc::Tx to an enum instead, but I'd really prefer a simple object here and keep conversion fallible

prestwich:
ya, id like to revisit this in the future tho

crates/rpc-types/src/eth/transaction/access_list.rs
```diff
@@ -48,6 +48,13 @@ impl AccessList {
     }
 }
 
+impl From<AccessList> for alloy_eips::eip2930::AccessList {
+    fn from(value: AccessList) -> Self {
+        // SAFETY: Same repr and size
```
prestwich, line 53: can we make a test that tests this safety assumption?

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -115,6 +120,123 @@ impl Transaction {
     }
 }
 
+impl TryFrom<Transaction> for Signed<TxLegacy> {
```
prestwich, line 123: much improved 👌

crates/rpc-types/src/eth/transaction/access_list.rs
mattsse: this should no longer be necessary after #324

can take over #324 and push that through

prestwich (approved):

PR #347, opened 2024-03-20
chore: check for reqwest http success

mattsse:
same check as hyper http


crates/transport-http/src/reqwest.rs
```diff
@@ -19,8 +19,17 @@ impl Http<reqwest::Client> {
                 .and_then(Response::error_for_status)
```
prestwich, line 19: is this not handled here?

mattsse:
my bad, overlooked this

PR #355, opened 2024-03-21
fix: mandatory `to` on `TxEip4844`

klkvr:
## Motivation

Per https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md#blob-transaction:
> The field `to` deviates slightly from the semantics with the exception that it MUST NOT be `nil` and therefore must always represent a 20-byte address. This means that blob transactions cannot have the form of a create transaction.

## Solution

Change `TxKind` to `Address`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip4844.rs
+++ b/crates/consensus/src/transaction/eip4844.rs
@@ -1041,7 +1041,7 @@ mod tests {
     use crate::{SignableTransaction, TxEnvelope};
     #[cfg(not(feature = "kzg"))]
     use alloy_eips::eip4844::{Blob, Bytes48};
-    use alloy_primitives::{Signature, TxKind, U256};
+    use alloy_primitives::{Signature, U256};
     use alloy_rlp::{Decodable, Encodable};
     #[cfg(feature = "kzg")]
     use c_kzg::{Blob, Bytes48};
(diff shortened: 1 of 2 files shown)


mattsse (approved):
oh nice find we also have this wrong in reth

ref https://github.com/ethereum/go-ethereum/blob/ba2dd9385c2a51134e520083dc732787a813b107/core/types/tx_blob.go#L38-L38

compared to 

https://github.com/ethereum/go-ethereum/blob/ba2dd9385c2a51134e520083dc732787a813b107/core/types/tx_dynamic_fee.go#L34-L34

prestwich (commented):
[ref](https://eips.ethereum.org/EIPS/eip-4844#blob-transaction)

Do we need to address this in any tx builder as well? I.e. do we have any tx builder where we can make a build error when `to` is unset and `blob` is set?

prestwich:
good catch :)

mattsse:
> [ref](https://eips.ethereum.org/EIPS/eip-4844#blob-transaction)
> 
> Do we need to address this in any tx builder as well? I.e. do we have any tx builder where we can make a build error when `to` is unset and `blob` is set?

this is covered in the builder step:

https://github.com/alloy-rs/alloy/pull/355/files#diff-b7dd7b7cffb00784af1e996954e9e5b3997d1edee8ae3ee1b289468f1ddd2336R214

prestwich:
oh dope I missed that 👍

PR #359, opened 2024-03-21
feat(node-bindings): add methods for returning instance urls

yash-atreya:
## Motivation

The existing methods such as `endpoint()` and `ws_endpoint()` would return the URLs as a `String`. One would have to parse the String as a URL and then pass it to the provider. 

## Solution

Adds two methods `endpoint_url()` and `ws_endpoint_url()`  which return the `HTTP` and `WS` urls respectively for both `anvil` and `geth` instances.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/geth.rs
+++ b/crates/node-bindings/src/geth.rs
@@ -691,7 +691,6 @@ mod tests {
     fn run_with_tempdir(f: impl Fn(&Path)) {
         let temp_dir = tempfile::tempdir().unwrap();
         let temp_dir_path = temp_dir.path();
-        println!("temp dir: {:?}", temp_dir_path);
         f(temp_dir_path);
         #[cfg(not(windows))]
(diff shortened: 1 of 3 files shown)


yash-atreya:
cc @mattsse

crates/node-bindings/src/geth.rs
```diff
@@ -680,6 +691,7 @@ mod tests {
     fn run_with_tempdir(f: impl Fn(&Path)) {
         let temp_dir = tempfile::tempdir().unwrap();
         let temp_dir_path = temp_dir.path();
+        println!("temp dir: {:?}", temp_dir_path);
```
mattsse, line 694:
```suggestion
```

mattsse (changes requested):
smol nit

PR #363, opened 2024-03-22
fix: remove app-layer usage of transport error

prestwich:
## Motivation

TransportError should capture `Transport`-layer errors, not application or RPC semantics errors. RPC semantics errors should be captured by `RpcError` (note that `TransportErrorKind` instantiates an `RpcError<TransportError>`, not a `TransportError`). Application errors should be captured by either RpcError (e.g. when the application expects a non-null response) or by an ErrResp

## Solution

- Remove usages of `TransportErrorKing::custom_str`
- Remove `MissingReceipt` variant from TransportError
- Add `NullResp` variant to `RpcError`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -1,6 +1,7 @@
 use crate::{
     utils::Eip1559Estimation, PendingTransactionBuilder, Provider, ProviderLayer, RootProvider,
 };
+use alloy_json_rpc::RpcError;
 use alloy_network::{Network, TransactionBuilder};
 use alloy_primitives::U256;
 use alloy_transport::{Transport, TransportError, TransportResult};
@@ -122,16 +123,8 @@ where
                 tx.set_max_priority_fee_per_gas(eip1559_fees.max_priority_fee_per_gas);
                 Ok(())
             }
-            Err(err) => {
-                if err.is_transport_error()
-                    && err.to_string() == *"EIP-1559 not activated".to_string()
-                {
-                    // If EIP-1559 is not activated, it will process as a legacy tx.
-                    self.handle_legacy_tx(tx).await
-                } else {
-                    Err(err)
-                }
-            }
+            Err(RpcError::NullResp) => self.handle_legacy_tx(tx).await,
+            Err(e) => Err(e),
         }
     }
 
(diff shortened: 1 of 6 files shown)


crates/provider/src/provider.rs
```diff
@@ -762,41 +762,25 @@ pub trait Provider<N: Network, T: Transport + Clone = BoxTransport>: Send + Sync
...
-        };
-
-        Ok(Eip1559Estimation { max_fee_per_gas, max_priority_fee_per_gas })
+        )?;
+
+        let base_fee_per_gas = bf
+            .ok_or(RpcError::NullResp)?
+            .header
+            .base_fee_per_gas
+            .ok_or(RpcError::UnsupportedFeature("eip1559"))?;
```
mattsse, line 778: I think this makes sense

crates/provider/src/layers/gas.rs
```diff
@@ -122,16 +123,8 @@ where
...
-                if err.is_transport_error()
-                    && err.to_string() == *"EIP-1559 not activated".to_string()
-                {
-                    // If EIP-1559 is not activated, it will process as a legacy tx.
-                    self.handle_legacy_tx(tx).await
-                } else {
-                    Err(err)
-                }
-            }
+            Err(RpcError::UnsupportedFeature("eip1559")) => self.handle_legacy_tx(tx).await,
```
mattsse, line 126: I think this could also fail with a
-32601, message: Method not found

rpc error for the fee_history call?

some providers also return -32600, message: Unsupported method:

so perhaps we should be less restrictive here and also check the rpc error response?

mattsse: I think that makes sense, so the unsupported feature check is okay because this looks at the base fee value of the header

mattsse (approved):

crates/provider/src/layers/gas.rs
prestwich: we need to write down somewhere that 1559 is somewhat privileged by the library in ways that other tx types and HF features are not. I.e. we generally assume that 1559 is active and proactively attempt to use it in gas estimation even tho we acknowledge that there may be chains that don't support it

PR #370, opened 2024-03-22
fix: correctly process chainId field

klkvr:
## Motivation

WIthout `serde(default)` legacy transactions without chainId field can't be deserialized right now

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/mod.rs
+++ b/crates/rpc-types/src/eth/transaction/mod.rs
@@ -71,6 +71,7 @@ pub struct Transaction {
     /// The chain id of the transaction, if any.
     #[serde(default)]
     #[serde(with = "alloy_serde::u64_hex_opt")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub chain_id: Option<u64>,
     /// Contains the blob hashes for eip-4844 transactions.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -315,4 +317,24 @@ mod tests {
...
+            from: Address::with_last_byte(6),
+            value: U256::from(8),
+            gas: U256::from(10),
+            input: Bytes::from(vec![11, 12, 13]),
+            ..Default::default()
+        };
+        let serialized = serde_json::to_string(&transaction).unwrap();
+        assert_eq!(
+            serialized,
+            r#"{"hash":"0x0000000000000000000000000000000000000000000000000000000000000001","nonce":"0x2","blockHash":null,"blockNumber":null,"transactionIndex":null,"from":"0x0000000000000000000000000000000000000006","to":null,"value":"0x8","gas":"0xa","input":"0x0b0c0d"}"#
```
onbjerg, line 335: where is this fixture from

klkvr: that's just a fixture from the test above with all options set to None, the idea was to test that we can handle any of the optional fields missing

I can change it to a valid legacy tx payload so we avoid unexpected test failures when modifying `Transaction` in the future

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -69,7 +69,9 @@ pub struct Transaction {
     #[serde(flatten, skip_serializing_if = "Option::is_none")]
     pub signature: Option<Signature>,
     /// The chain id of the transaction, if any.
+    #[serde(default)]
     #[serde(with = "alloy_serde::u64_hex_opt")]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
mattsse, line 74: can we collapse them into a single serde attribute?
the skip is also fine because this field should be ignored

mattsse (approved):
pedantic style nit, otherwise lgtm

Issue #349, opened 2024-03-20
[Feature] Move Mainnet KZG group and `Lazy<KzgSettings>`


prestwich:
### Component

consensus, eips, genesis

### Describe the feature you would like

4844 constants in alloy-eips needs the following
- KZG constants and files from [revm-primitives](https://github.com/bluealloy/revm/tree/main/crates/primitives/src/kzg)
- `Lazy<Arc<KzgSettings>>` and other loading methods from [reth](https://github.com/paradigmxyz/reth/blob/c868c52063a7f507f2a455a752e159c9b5f08402/crates/primitives/src/constants/eip4844.rs#L8)

### Additional context

Should be feature-gated behind `c_kzg`

developeruche:
```rust 
#[cfg(feature = "kzg")]
use c_kzg::{Blob, Bytes48, KzgCommitment, KzgProof, KzgSettings};
```

this is already behind the`kzg` features flag, do you intend the feature should be changed to `c_kzg`?

prestwich:
no, use whatever existing feature name 🤝

developeruche:
I don't think, 

1. [g1_points](https://github.com/bluealloy/revm/blob/main/crates/primitives/src/kzg/g1_points.bin)
2. [g2_points](https://github.com/bluealloy/revm/blob/main/crates/primitives/src/kzg/g2_points.bin)
3. [trusted_setup](https://github.com/bluealloy/revm/blob/main/crates/primitives/src/kzg/trusted_setup.txt)

would be necessary in the `alloy-eips` module, what do you think?

Secondly, I am looking to add;

1. Re-export `EnvKzgSettings` from the `revm-primitive` env_settings.rs
2. Re-export all public const from `revm-primitive` trusted_setup_points.rs


what are your thoughts?

developeruche:
@prestwich

mattsse:
all of the utils and constants defined on revm should be moved here, then once we have a release we can replace them in revm with our crate

developeruche:
Alright... that makes sense

PR #375, opened 2024-03-22
feat: re-export EnvKzgSettings

prestwich:
## Motivation

Quality of life improvement post #349 

## Solution

- re-export `EnvKzgSettings` from alloy-consensus to allow easy use with `SidecarBuilder`
- refactor `SidecarBuilder::build` to use default settings
- add `SidecarBuilder::build_with_settings`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip4844/builder.rs
+++ b/crates/consensus/src/transaction/eip4844/builder.rs
@@ -1,7 +1,7 @@
 #[cfg(not(feature = "kzg"))]
 use alloy_eips::eip4844::Blob;
 #[cfg(feature = "kzg")]
-use c_kzg::{Blob, KzgCommitment, KzgProof, KzgSettings};
+use c_kzg::{Blob, KzgCommitment, KzgProof};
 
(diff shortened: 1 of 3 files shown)


crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -339,12 +339,13 @@ impl<T: SidecarCoder> SidecarBuilder<T> {
...
     }
 
+    /// Build the sidecar from the data with the provided settings.
     #[cfg(feature = "kzg")]
-    /// Build the sidecar from the data.
-    pub fn build(
+    pub fn build_with_settings(
         self,
-        settings: &KzgSettings,
+        settings: &crate::EnvKzgSettings,
```
DaniPopes, line 346: What if we kept `KzgSettings` here and called `.get()` in `build`? This way we don't force our own Env type on this function

prestwich: sure makes sense

DaniPopes (approved):

PR #379, opened 2024-03-23
feat: make HTTP provider optional

### LINKED ISSUE #199: [Feature] Make HTTP provider optional, on by default
### Component

providers, pubsub

Currently `alloy-providers` depends on `alloy-transport-http`, only for some trait impls and type aliases. This should be made optional since it is generic over `T: Transport`.

### Additional context

_No response_

leruaa:
Closes #199

### DIFF
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -55,6 +55,6 @@ reqwest = [
     "dep:alloy-transport-http",
     "alloy-rpc-client/reqwest",
 ]
-hyper = ["dep:alloy-transport-http", "dep:url", "alloy-rpc-client/hyper"]
+hyper = ["dep:alloy-transport-http", "dep:url", "alloy-rpc-client/hyper",]
 ws = ["pubsub", "alloy-rpc-client/ws", "alloy-transport-ws"]
 ipc = ["pubsub", "alloy-rpc-client/ipc", "alloy-transport-ipc"]
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -20,6 +20,7 @@ alloy-rpc-types.workspace = true
 alloy-transport-http = { workspace = true, optional = true }
 alloy-transport-ipc = { workspace = true, optional = true }
 alloy-transport-ws = { workspace = true, optional = true }
+alloy-pubsub = { workspace = true, optional = true }
 alloy-transport.workspace = true
 alloy-primitives.workspace = true
 
@@ -33,8 +34,7 @@ reqwest = { workspace = true, optional = true }
 serde_json.workspace = true
 tokio = { workspace = true, features = ["sync", "macros"] }
 tracing.workspace = true
-
-alloy-pubsub = { workspace = true, optional = true }
+url = { workspace = true, optional = true }
 
 [dev-dependencies]
 alloy-consensus.workspace = true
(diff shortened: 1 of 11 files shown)


crates/provider/src/lib.rs
```diff
@@ -15,11 +15,19 @@
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
-use alloy_transport_http::Http;
-use reqwest::Client as ReqwestClient;
+#[cfg(feature = "http")]
+mod http {
+    use alloy_transport_http::Http;
+    use reqwest::Client as ReqwestClient;
```
mattsse, line 21: this alias is kinda pointless, let's just use reqwest::clinet in the pub type HttpProvider

mattsse (changes requested):
pedantic nit

otherwise sounds reasonable to me

mattsse (approved):
lgtm, pending @DaniPopes

onbjerg (approved):
lgtm too

crates/provider/Cargo.toml
```diff
@@ -45,6 +45,13 @@ tokio = { workspace = true, features = ["macros"] }
 tracing-subscriber = { workspace = true, features = ["fmt"] }
 
 [features]
+default = ["http"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
+http = [
```
prestwich, line 50: how does this interact with hyper?

prestwich:
@leruaa this has conflicts with #318 which was just merged. I can clean them up if you don't feel like you understand the code. let me know

leruaa:
I'll give a try

crates/provider/Cargo.toml
leruaa: Interesting question! Right now it doesn't iteract with hyper at all... But I noticed features "[reqwest](https://github.com/alloy-rs/alloy/blob/7e39c85f9f51e6449a8b661f54df0ac213f18639/crates/provider/src/builder.rs#L228-L229)" and "[hyper](https://github.com/alloy-rs/alloy/blob/7e39c85f9f51e6449a8b661f54df0ac213f18639/crates/provider/src/builder.rs#L238-L239)" has been added in #342 but they aren't declared in the Cargo.toml.

Should I add these 2 new features that both enable "http" ?

leruaa:
@prestwich I think the merge went fine

crates/provider/Cargo.toml
```diff
@@ -14,10 +14,10 @@ exclude.workspace = true
 [dependencies]
 alloy-json-rpc.workspace = true
 alloy-network.workspace = true
-alloy-rpc-client = { workspace = true, features = ["reqwest"] }
+alloy-rpc-client = { workspace = true }
```
prestwich, line 17: nit: change to `.workspace = true`

crates/provider/Cargo.toml
prestwich: Yes, that would be my preferred approach 👍

prestwich (changes requested):

crates/provider/src/builder.rs
```diff
@@ -225,23 +226,25 @@ impl<L, N> ProviderBuilder<L, N> {
         Ok(self.on_client(client))
     }
 
+    /// Build this provider with an Reqwest HTTP transport.
     #[cfg(feature = "reqwest")]
     pub fn on_reqwest_http(self, url: url::Url) -> Result<L::Provider, TransportError>
     where
         L: ProviderLayer<RootProvider<N, BoxTransport>, N, BoxTransport>,
```
prestwich, line 233: this is probably on me, but we should not box here, given the user has explicitly specified the transport type to be used. replace these `BoxTransport` usages with naming the Http type

crates/provider/src/builder.rs
```diff
@@ -225,23 +226,25 @@ impl<L, N> ProviderBuilder<L, N> {
...
-        let client = ClientBuilder::default().reqwest_http(url);
+        let client = ClientBuilder::default().reqwest_http(url).boxed();
         Ok(self.on_client(client))
     }
 
+    /// Build this provider with an Hyper HTTP transport.
     #[cfg(feature = "hyper")]
     pub fn on_hyper_http(self, url: url::Url) -> Result<L::Provider, TransportError>
     where
         L: ProviderLayer<RootProvider<N, BoxTransport>, N, BoxTransport>,
```
prestwich, line 244: here too

prestwich: this may require reproducing generics from down the stack I think. if it gets ugly or annoying go ahead and skip this

prestwich:
actually I'm gonna go ahead and rebase and do this :)

prestwich (approved):

PR #381, opened 2024-03-23
fix: correctly treat `confirmation` for `watch_pending_transaction`

klkvr:
## Motivation

Current logic was not treating `confirmations` in the expected way (tx being included = 1 confirmation)
Also `split_off` splits off keys including the one at which split is being performed, thus we actually notified `TxWatcher` about txses one block later

Two open questions:
1. Currently if tx is sent, included into block and one more block is mined (basically tx included into block X and latest block is X+1), then call to `watch_pending_transaction` will indefinitely wait for a block with targeted tx. Is this expected? Feels like a footgun, especially considering that we don't have a default timeout
2. It feels like `watch_pending_transaction` users specifying `confirmations > 1` would also be interested in ensuring that no reorg has happened and that tx is actually on-chain, should we add a parameter for that?


prestwich:
please put open questions in issues 👍

mattsse (approved):
I believe this is correct

PR #383, opened 2024-03-24
feat: add `AnyNetwork`

### LINKED ISSUE #81: [Feature] `AnyNetwork`
### Component

providers, rpc-types

implement the network trait from #80 for an `AnyNetwork` which contains the standard ethereum JSON-RPC types + an `extra_fields: HashMap<String, Value>`. If necessary, differentiate `AnyTransaction` from the transaction types described in #73 

### Additional context

_No response_

leruaa:
Closes #81


crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -81,6 +81,12 @@ impl Hash for TransactionRequest {
     }
 }
 
+impl Default for Extra<TransactionRequest> {
```
onbjerg, line 84: impl Default for any Extra<T> where T: Default instead imo

crates/rpc-types/src/extra.rs
```diff
@@ -0,0 +1,44 @@
...
+    ops::{Deref, DerefMut},
+};
+
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Wrapper allowing to catch all fields missing on the inner struct while
+/// deserialize.
+pub struct Extra<T> {
```
onbjerg, line 12: i think we already have this struct: https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types/src/eth/other.rs

just rename the one in that mod instead

crates/network/src/any/mod.rs
```diff
@@ -0,0 +1,35 @@
+use alloy_rpc_types::{Extra, Header, Transaction, TransactionReceipt, TransactionRequest};
+
+use crate::{Network, ReceiptResponse};
+
+mod builder;
+
+/// Types for a catch-all network.
```
onbjerg, line 7: please expand the docs a bit on when you would use this (and also discourage its usage), and what it does

you would only use it if you do not know the network type at compile time

onbjerg (changes requested):

onbjerg (changes requested):
please also remove the catch all fields currently present on some of the rpc-types, e.g. 

https://github.com/alloy-rs/alloy/blob/28cab90c70ef2df16571e1ec131aebbba7bc4412/crates/rpc-types/src/eth/block.rs#L962-L963

https://github.com/alloy-rs/alloy/blob/28cab90c70ef2df16571e1ec131aebbba7bc4412/crates/rpc-types/src/eth/transaction/request.rs#L55-L57

https://github.com/alloy-rs/alloy/blob/28cab90c70ef2df16571e1ec131aebbba7bc4412/crates/rpc-types/src/eth/transaction/receipt.rs#L56-L58

crates/rpc-types/src/extra.rs
leruaa: Not sure if I understood your comment correctly. Ended up moving `Extra<T>` to [rpc-types/src/eth/other.rs](https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types/src/eth/other.rs), renaming to `WithOtherFields<T>` with a `pub other: OtherFields` field.

leruaa:
Not sure if this should be removed:

https://github.com/alloy-rs/alloy/blob/28cab90c70ef2df16571e1ec131aebbba7bc4412/crates/rpc-types/src/eth/block.rs#L962-L963

as `Block` is not part of `Network` types

onbjerg:
@leruaa Makes sense, pls flag in that issue that we need to make the proper adjustments  it for `AnyNetwork`

crates/rpc-types/src/eth/other.rs
```diff
@@ -137,3 +137,46 @@ impl<'a> IntoIterator for &'a OtherFields {
         self.as_ref().iter()
     }
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// Wrapper allowing to catch all fields missing on the inner struct while
+/// deserialize.
```
onbjerg, line 143: nit: reorder these (docs on top)

onbjerg (approved):
lgtm

crates/network/src/any/builder.rs
```diff
@@ -0,0 +1,118 @@
+use std::ops::{Deref, DerefMut};
+
+use alloy_consensus::BlobTransactionSidecar;
+use alloy_primitives::U256;
+use alloy_rpc_types::{TransactionRequest, WithOtherFields};
+
+use crate::{ethereum::build_unsigned, BuilderResult, Network, TransactionBuilder};
+
+use super::AnyNetwork;
```
DaniPopes, line 9: please squash imports together

crates/network/src/any/mod.rs
```diff
@@ -0,0 +1,41 @@
+use alloy_rpc_types::{
+    Header, Transaction, TransactionReceipt, TransactionRequest, WithOtherFields,
+};
+
```
DaniPopes, line 4:
```suggestion
```

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -255,6 +246,8 @@ pub struct TransactionInputError;
 
 #[cfg(test)]
 mod tests {
+    use crate::WithOtherFields;
+
```
DaniPopes, line 250:
```suggestion
```

crates/rpc-types/src/with_other.rs
```diff
@@ -0,0 +1,48 @@
+use std::ops::{Deref, DerefMut};
+
+use serde::{Deserialize, Serialize};
+
```
DaniPopes, line 4: same

DaniPopes (approved):

crates/rpc-types/src/with_other.rs
```diff
@@ -0,0 +1,46 @@
+use crate::other::OtherFields;
+use serde::{Deserialize, Serialize};
+use std::ops::{Deref, DerefMut};
+
+/// Wrapper allowing to catch all fields missing on the inner struct while
+/// deserialize.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct WithOtherFields<T> {
+    #[serde(flatten)]
+    inner: T,
```
mattsse, line 10: This needs to pub or we need another away for deconstructing all fields, currently there's no easy way to do this.

so either pub or like a split() -> (T, Other) function

mattsse (changes requested):

crates/network/src/ethereum/builder.rs
```diff
@@ -146,6 +120,37 @@ impl TransactionBuilder<Ethereum> for alloy_rpc_types::TransactionRequest {
...
+    match (
+        request.gas_price.as_ref(),
+        request.max_fee_per_gas.as_ref(),
+        request.access_list.as_ref(),
+        request.max_fee_per_blob_gas.as_ref(),
+        request.blob_versioned_hashes.as_ref(),
+        request.sidecar.as_ref(),
+    ) {
+        // Legacy transaction
+        (Some(_), None, None, None, None, None) => build_legacy(request).map(Into::into),
```
prestwich, line 138: while we're touching this code, we should decompose this into a set of functions `fn will_build_legacy() -> bool` that capture our criteria for building each style of transaction. then this becomes a fallthrough if

I also think our criteria here are not most-friendly to users.

- if a sidecar is set, we should ALWAYS build 4844, regardless of if legacy gas is indicated
- if `max_fee_per_blob_gas` is set, but no sidecar/blob is set, that should be a build error
- 1559 should be our default tx type in the fallthrough case. legacy should be chosen only if gas_price is set.


function should look something as follows:
```
if self.will_build_4844() { self.build_4844() } 
else if self.will_build_2930() { self.build_2930() }
else if self.will_build_legacy() { self.build_legacy() }
else { self.build_1559() } 
```

prestwich: opened as #423

mattsse: is this PR blocked by #423?

prestwich: no, should be separate work

leruaa: Is something missing for this PR to be merged?

prestwich (approved):
Approving, with a followup issue to add tests with various networks' receipt types

leruaa:
The PR shows a requested change from @mattsse, but I marked all conversations as resolved. Is this a bug?

prestwich:
no change request live until the reviewer dismisses it themselves. cc @mattsse Im cool with merge when you are

mattsse (approved):

PR #386, opened 2024-03-24
chore: clean up kzg and features

DaniPopes:
The current implementation uses an Arc so it can never be no-std, and the `MAINNET` static is unnecessary as `EnvKzgSettings::Default.get` should be used instead.

cc @developeruche we should switch out `MAINNET` static usage in Reth to the enum

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,6 +57,7 @@ ethereum_ssz_derive = "0.5"
 ethereum_ssz = "0.5"
 
 # crypto
+c-kzg = { version = "1.0", default-features = false }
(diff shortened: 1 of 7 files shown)


prestwich (approved):

mattsse (approved):

mattsse (approved):
features 👍

developeruche:
> The current implementation uses an Arc so it can never be no-std, and the `MAINNET` static is unnecessary as `EnvKzgSettings::Default.get` should be used instead.
> 
> cc @developeruche we should switch out `MAINNET` static usage in Reth to the enum

Alright... would get on with this

PR #388, opened 2024-03-24
fix: small fixes for `Transaction`

klkvr:
## Motivation

1. Make `Transaction::blob_versioned_hashes` an `Option` so that we can determine if it was not present or just empty. This field has similar semantics to `access_list`
2. EIP-1559 txs include deprecated `gasPrice` field: 
    image
    We have to account for it when converting to `TransactionRequest` to not end up with invalid request having both `gasPrice` and `maxFeePerGas`.


prestwich:
please link source of screenshot

klkvr:
@prestwich https://ethereum.github.io/execution-apis/api-documentation/ eth_getTransactionByHash section
https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L124

mattsse (approved):
lgtm,

this is better, this should indeed be skipped if None and using an option makes it possible to distinguish between blob vs non blob

onbjerg (approved):
cursed spec

Issue #394, opened 2024-03-26
[Feature] `Rpc` type cast to `Consensus` type


rkdud007:
### Component

rpc-type

### Describe the feature you would like

The type return from RPC and the type return from consensus type are different. And as the rlp encode/decode is bound with consensus type, even though fetch data in rpc type, faced lots of cases that required type cast from rpc -> consensus type. 

### Ideal solution

For example in transaction, would like to have : 
```rust
impl TryFrom<alloy_rpc_types::Transaction> for alloy_consensus::TxEnvelope { 
...
}
```

so that can use like this
```rust
let tx_for_consensus: TxEnvelope = tx_from_rpc.try_into();
```

rkdud007:
sorry just noticed it was added in late commit #343

PR #395, opened 2024-03-26
feat: `impl TryFrom<TransactionReceipt> for ReceiptEnvelope`

rkdud007:
## Motivation

- add conversion from `alloy_rpc_types::ReceiptEnvelope` to `alloy_consensus::ReceiptEnvelope`

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/receipt.rs
+++ b/crates/rpc-types/src/eth/transaction/receipt.rs
@@ -1,7 +1,6 @@
 use crate::{other::OtherFields, ConversionError, Log};
 use alloy_consensus::{Receipt, ReceiptEnvelope, ReceiptWithBloom, TxType};
 use alloy_primitives::{Address, Bloom, B256, U128, U256, U64, U8};
-use itertools::Itertools;
 use serde::{Deserialize, Serialize};
 
 /// Transaction receipt
@@ -81,13 +80,31 @@ impl TransactionReceipt {
         }
     }
 
-    /// Returns an iterator over the logs for prmitives type conversion.
-    fn logs_iter(
-        &self,
-    ) -> impl Iterator<Item = alloy_primitives::Log<alloy_primitives::LogData>> + '_ {
-        self.logs.iter().map(|log| {
-            alloy_primitives::Log::new_unchecked(log.address, log.topics.clone(), log.data.clone())
-        })
+    /// Consumes the type and returns a vector of logs.
+    fn into_logs(self) -> Vec<alloy_primitives::Log<alloy_primitives::LogData>> {
+        self.logs
+            .into_iter()
+            .map(|log| {
+                alloy_primitives::Log::new_unchecked(
+                    log.address,
+                    log.topics.clone(),
+                    log.data.clone(),
+                )
+            })
+            .collect()
+    }
+
+    /// Consumes the type and converts the transaction receipt into a receipt with bloom.
+    fn into_receipt_with_bloom(self) -> ReceiptWithBloom {
+        let logs_bloom = self.logs_bloom;
+        ReceiptWithBloom::new(
+            Receipt {
+                success: self.success(),
+                cumulative_gas_used: self.cumulative_gas_used.to::<u64>(),
+                logs: self.into_logs(),
+            },
+            logs_bloom,
+        )
     }
 }
 
@@ -95,15 +112,7 @@ impl TryFrom<TransactionReceipt> for ReceiptWithBloom {
     type Error = ConversionError;
 
     fn try_from(tx_receipt: TransactionReceipt) -> Result<Self, Self::Error> {
-        let receipt_with_bloom = ReceiptWithBloom {
-            receipt: Receipt {
-                success: tx_receipt.success(),
-                cumulative_gas_used: tx_receipt.cumulative_gas_used.to::<u64>(),
-                logs: tx_receipt.logs_iter().collect_vec(),
-            },
-            bloom: tx_receipt.logs_bloom,
-        };
-        Ok(receipt_with_bloom)
+        Ok(tx_receipt.into_receipt_with_bloom())
     }
 }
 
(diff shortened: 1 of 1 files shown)


rkdud007:
related to #394 , closed it thought there every conversion had implemented, but realize only for tx

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -69,4 +70,90 @@ impl TransactionReceipt {
         }
         Some(self.from.create(nonce))
     }
+
+    /// Returns `true` if the transaction was successful.
+    /// A transaction is considered successful if the status code is `1`.
+    fn success(&self) -> bool {
+        match &self.status_code {
```
mattsse, line 77: I'm okay with changing the `status_code` from U64 to u64 and use a serde_with hex_opt for this

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -69,4 +70,90 @@ impl TransactionReceipt {
...
+        }
+    }
+
+    /// Returns the logs emitted by the transaction.
+    /// Converts the logs from the RPC type to the internal type.
+    fn logs(&self) -> Vec<PrimitivesLog<LogData>> {
+        let mut logs = Vec::new();
+        for log in &self.logs {
+            let rpc_log: Log = log.clone();
+            let log_data = LogData::try_from(rpc_log).unwrap_or_default();
```
mattsse, line 89: I'm okay with using new_unchecked checked here, because we assume the topics in a receipt are correct

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -69,4 +70,90 @@ impl TransactionReceipt {
...
+    fn success(&self) -> bool {
+        match &self.status_code {
+            Some(status) => status == &U64::from(1),
+            None => false,
+        }
+    }
+
+    /// Returns the logs emitted by the transaction.
+    /// Converts the logs from the RPC type to the internal type.
+    fn logs(&self) -> Vec<PrimitivesLog<LogData>> {
```
mattsse, line 85: this can be converted into an iter `logs_iter` that this function then collects

mattsse (changes requested):
supportive,

some nits

crates/rpc-types/src/eth/transaction/receipt.rs
rkdud007: fixed https://github.com/alloy-rs/alloy/pull/395/commits/720162dd3b00b5a8c55dcc74573b8f7a2d59649e

crates/rpc-types/src/eth/transaction/receipt.rs
rkdud007: fixed https://github.com/alloy-rs/alloy/pull/395/commits/720162dd3b00b5a8c55dcc74573b8f7a2d59649e

crates/rpc-types/src/eth/transaction/receipt.rs
rkdud007: i don't get what does it mean by "use a serde_with hex_opt" in this case. Could you elaborate more?

rkdud007: you mean `#[serde(with = "hex")]`?

mattsse: yeah, but we can do this separately

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -75,23 +76,18 @@ impl TransactionReceipt {
...
-            let log_data = LogData::try_from(rpc_log).unwrap_or_default();
-            let result = PrimitivesLog { address: log.address, data: log_data };
-            logs.push(result);
-        }
-
-        logs
+    /// Returns an iterator over the logs for prmitives type conversion.
+    fn logs_iter(
+        &self,
+    ) -> impl Iterator<Item = alloy_primitives::Log<alloy_primitives::LogData>> + '_ {
```
mattsse, line 87: we also want into_logs, so we can consume the type in TryFrom

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -69,4 +71,85 @@ impl TransactionReceipt {
...
+                cumulative_gas_used: tx_receipt.cumulative_gas_used.to::<u64>(),
+                logs: tx_receipt.logs_iter().collect_vec(),
+            },
+            bloom: tx_receipt.logs_bloom,
+        };
+        Ok(receipt_with_bloom)
+    }
+}
+
+impl TryFrom<TransactionReceipt> for ReceiptEnvelope {
```
mattsse, line 110: this can also be implemented for &TransactionReceipt, and I'd like to have them as functions of the `TransactionReceipt` typ

```
fn into_(self
fn as_(&self
```

mattsse (changes requested):

crates/rpc-types/src/eth/transaction/receipt.rs
prestwich: given that status is always 0 or 1, I would also be comfortable just deserializing it to a bool. "real" future compatibility would be a non-exhaustive enum imho

rkdud007: checked #396 , will leave this nit on that PR

prestwich: ya sorry for stepping on your PR's toes a bit here. I feel like the infallible conversion will serve your needs better? 😅

rkdud007: no worries, i like embed approach better. 
one q is I had checked `impl AsRef<ReceiptEnvelope> for TransactionReceipt` meet the goal what this PR had supposed to, should I have to close the PR?

prestwich: we will talk internally, and then close one of them. If you like the embed approach better, we'll likely go in that direction :)

thank you for making the issue and PR regardless!

prestwich:
closing in favor of #396

PR #396, opened 2024-03-26
feature: embed primitives Log in rpc Log and consensus Receipt in rpc Receipt

### LINKED ISSUE #399: [Feature] Embed consensus types in rpc types wherever possible
the way to enforce it as a bound on network (if we even want to do that) would be to have `type TransactionResponse: RpcObject + AsRef<Self::TxEnvelope>`, basically. That would ensure that any response object contained an envelope

I think that the longer we go **without** embedding consensus objects in the RPC types, the worse of a time we'll have making the change

My primary target list is 
- Log
- TxEnvelope (deferred for now)
- ReceiptEnvelope

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/361#discussion_r1538256567_

prestwich:
Supersedes #395, closes #399

## Motivation

Prevent accidental divergence of consensus and rpc types by embedding the consensus type in the RPC type

## Solution

- [x] Embed `alloy_primitives::Log` into `alloy_rpc_types::Log`
- [x] Embed `alloy_consensus::ReceiptEnvelope` into `alloy_rpc_types::Receipt` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -5,13 +5,15 @@ use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable};
 /// Receipt containing result of transaction execution.
 #[derive(Clone, Debug, PartialEq, Eq, Default)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct Receipt {
     /// If transaction is executed successfully.
     ///
     /// This is the `statusCode`
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity_bool"))]
-    pub success: bool,
+    pub status: bool,
     /// Gas used
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_hex_or_decimal"))]
     pub cumulative_gas_used: u64,
     /// Log send from contracts.
     pub logs: Vec<Log>,
@@ -33,7 +35,7 @@ impl Receipt {
 
 impl TxReceipt for Receipt {
     fn success(&self) -> bool {
-        self.success
+        self.status
     }
 
     fn bloom(&self) -> Bloom {
@@ -56,24 +58,27 @@ impl TxReceipt for Receipt {
 ///
 /// [`Sealed`]: crate::sealed::Sealed
 #[derive(Clone, Debug, PartialEq, Eq, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct ReceiptWithBloom {
+    #[cfg_attr(feature = "serde", serde(flatten))]
     /// The receipt.
     pub receipt: Receipt,
     /// The bloom filter.
-    pub bloom: Bloom,
+    pub logs_bloom: Bloom,
 }
 
 impl TxReceipt for ReceiptWithBloom {
     fn success(&self) -> bool {
-        self.receipt.success
+        self.receipt.status
     }
 
     fn bloom(&self) -> Bloom {
-        self.bloom
+        self.logs_bloom
     }
 
     fn bloom_cheap(&self) -> Option<Bloom> {
-        Some(self.bloom)
+        Some(self.logs_bloom)
     }
 
     fn cumulative_gas_used(&self) -> u64 {
@@ -88,14 +93,14 @@ impl TxReceipt for ReceiptWithBloom {
 impl From<Receipt> for ReceiptWithBloom {
     fn from(receipt: Receipt) -> Self {
         let bloom = receipt.bloom_slow();
-        ReceiptWithBloom { receipt, bloom }
+        ReceiptWithBloom { receipt, logs_bloom: bloom }
     }
 }
 
 impl ReceiptWithBloom {
     /// Create new [ReceiptWithBloom]
     pub const fn new(receipt: Receipt, bloom: Bloom) -> Self {
-        Self { receipt, bloom }
+        Self { receipt, logs_bloom: bloom }
     }
 
     /// Consume the structure, returning only the receipt
@@ -107,13 +112,13 @@ impl ReceiptWithBloom {
     /// Consume the structure, returning the receipt and the bloom filter
     #[allow(clippy::missing_const_for_fn)] // false positive
     pub fn into_components(self) -> (Receipt, Bloom) {
-        (self.receipt, self.bloom)
+        (self.receipt, self.logs_bloom)
     }
 
     fn payload_len(&self) -> usize {
-        self.receipt.success.length()
+        self.receipt.status.length()
             + self.receipt.cumulative_gas_used.length()
-            + self.bloom.length()
+            + self.logs_bloom.length()
             + self.receipt.logs.length()
     }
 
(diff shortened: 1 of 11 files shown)


crates/rpc-types/src/eth/log.rs
```diff
@@ -26,33 +24,67 @@ pub struct Log {
...
 }
 
-impl TryFrom<Log> for alloy_primitives::LogData {
-    type Error = LogError;
+impl<T> Log<T>
+where
+    for<'a> &'a T: Into<alloy_primitives::LogData>,
+{
+    /// Reserialize the data.
+    pub fn reserialize(&self) -> Log<alloy_primitives::LogData> {
```
mattsse, line 32: I see, this can be used to convert anything other than LogData in the default `Log`

could add this as context to "deserialize the data into the default Log type"

if we have &self, we can also add an `Into<Log>` either by impl Into trait (perhaps not possible) or `into_reserialize`

crates/rpc-types/src/eth/log.rs
```diff
@@ -1,16 +1,14 @@
...
-    /// Address
-    pub address: Address,
-    /// All topics of the log
-    pub topics: Vec<B256>,
-    /// Additional data fields of the log
-    pub data: Bytes,
+pub struct Log<T = alloy_primitives::LogData> {
+    #[serde(flatten)]
+    /// Consensus log object
+    pub inner: alloy_primitives::Log<T>,
```
mattsse, line 10: do we want getters for address, topics and data in `impl Log`?

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -1,63 +1,65 @@
...
 pub struct TransactionReceipt {
+    /// The receipt envelope, which contains the consensus receipt data..
+    #[serde(flatten)]
+    pub inner: ReceiptEnvelope,
+
     /// Transaction Hash.
     pub transaction_hash: B256,
     /// Index within the block.
-    pub transaction_index: U64,
+    #[serde(with = "alloy_serde::u64_hex_or_decimal")]
```
mattsse, line 17: should enforce QUANTITY on rpc
```suggestion
    #[serde(with = "alloy_serde::u64_hex")]
```

crates/consensus/src/receipt/receipts.rs
```diff
@@ -4,12 +4,16 @@ use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable};
...
 pub struct Receipt {
     /// If transaction is executed successfully.
     ///
     /// This is the `statusCode`
-    pub success: bool,
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity_bool"))]
+    pub status: bool,
     /// Gas used
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_hex_or_decimal"))]
     pub cumulative_gas_used: u64,
```
mattsse, line 17:
```suggestion
    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_hex"))]
    pub cumulative_gas_used: u64,
```

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -1,63 +1,65 @@
...
+    pub block_number: Option<u64>,
     /// Gas used by this transaction alone.
-    pub gas_used: Option<U256>,
+    #[serde(with = "alloy_serde::u64_hex_opt")]
+    pub gas_used: Option<u64>,
     /// The price paid post-execution by the transaction (i.e. base fee + priority fee). Both
     /// fields in 1559-style transactions are maximums (max fee + max priority fee), the amount
     /// that's actually paid by users can only be determined post-execution
-    pub effective_gas_price: U128,
+    #[serde(with = "alloy_serde::u64_hex_or_decimal")]
```
mattsse, line 31:
```suggestion
    #[serde(with = "alloy_serde::u64_hex")]
```

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -1,63 +1,65 @@
...
     #[serde(skip_serializing_if = "Option::is_none", rename = "root")]
     pub state_root: Option<B256>,
-    /// Status: either 1 (success) or 0 (failure). Only present after activation of EIP-658
-    #[serde(skip_serializing_if = "Option::is_none", rename = "status")]
-    pub status_code: Option<U64>,
-    /// EIP-2718 Transaction type.
-    ///
-    /// For legacy transactions this returns `0`. For EIP-2718 transactions this returns the type.
-    #[serde(rename = "type")]
-    pub transaction_type: U8,
```
mattsse, line 55: is this covered by `ReceiptEnvelope`?

mattsse (changes requested):
could we add some sanity serde checks to ensure this works for eth_getTransactionReceipt

random tx `cast rpc eth_getTransactionReceipt "0x21f6554c28453a01e7276c1db2fc1695bb512b170818bfa98fa8136433100616"`

```
{"transactionHash":"0x21f6554c28453a01e7276c1db2fc1695bb512b170818bfa98fa8136433100616","blockHash":"0x4acbdefb861ef4adedb135ca52865f6743451bfbfa35db78076f881a40401a5e","blockNumber":"0x129f4b9","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000200000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000400000800000000000000000000000000000000004000000000000000000800000000100000020000000000000000000080000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000010000000000000000000000000000","gasUsed":"0xbde1","contractAddress":null,"cumulativeGasUsed":"0xa42aec","transactionIndex":"0x7f","from":"0x9a53bfba35269414f3b2d20b52ca01b15932c7b2","to":"0xdac17f958d2ee523a2206206994597c13d831ec7","type":"0x2","effectiveGasPrice":"0xfb0f6e8c9","logs":[{"blockHash":"0x4acbdefb861ef4adedb135ca52865f6743451bfbfa35db78076f881a40401a5e","address":"0xdac17f958d2ee523a2206206994597c13d831ec7","logIndex":"0x118","data":"0x00000000000000000000000000000000000000000052b7d2dcc80cd2e4000000","removed":false,"topics":["0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925","0x0000000000000000000000009a53bfba35269414f3b2d20b52ca01b15932c7b2","0x00000000000000000000000039e5dbb9d2fead31234d7c647d6ce77d85826f76"],"blockNumber":"0x129f4b9","transactionIndex":"0x7f","transactionHash":"0x21f6554c28453a01e7276c1db2fc1695bb512b170818bfa98fa8136433100616"}],"status":"0x1"}
```

crates/rpc-types/src/eth/log.rs
prestwich: will add some for convenience

crates/rpc-types/src/eth/transaction/receipt.rs
prestwich: yes, it is implicit in the variant. Can add a getter for it if too

prestwich:
converting to draft as I need to resolve a type structure issue that results in the deserialized logs not including the RPC log metadata. need to put a generic in `ReceiptEnvelope` and push it down to `ReceiptWithBloom` and `Receipt`

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -70,3 +87,46 @@ impl TransactionReceipt {
...
+
+        let receipt: TransactionReceipt = serde_json::from_str(json_str).unwrap();
```
mattsse, line 103: can we add a check for:

`assert_eq(json::to_value(receipt, json::from_str(json_str))` to ensure we're compatible in both directions

mattsse (approved):
this should make things a bit easier,

one thing to note is that this changes the order of the json keys.
there's no spec for this and it doesn't matter really.

I'd like one more check to ensure we're also serializing properly

crates/rpc-types/src/eth/transaction/receipt.rs
prestwich: done 🖖

prestwich:
rebasing now and adding one last generic

PR #401, opened 2024-03-26
chore: dedupe blob in consensus and rpc

### LINKED ISSUE #400: Dedupe blob transaction sidecar
Can we proactively dedupe this as part of this PR?

current code in RPC types is as follows
```
/// This represents a set of blobs, and its corresponding commitments and proofs.
#[derive(Clone, Debug, PartialEq, Eq, Hash, Default, Serialize, Deserialize)]
#[repr(C)]
pub struct BlobTransactionSidecar {
    /// The blob data.
    pub blobs: Vec<Blob>,
    /// The blob commitments.
    pub commitments: Vec<Bytes48>,
    /// The blob proofs.
    pub proofs: Vec<Bytes48>,
}
```

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/361#discussion_r1540035588_

prestwich:
## Motivation

closes #400 

## Solution

Make an SSZ helper struct that can be transmuted from BlobsBundleV1

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


mattsse (approved):
cleaner

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -5,25 +5,29 @@ use alloy_consensus::{
...
+pub use optimism::OptimismTransactionReceiptFields;
+
 mod receipt;
+pub use receipt::TransactionReceipt;
+
 pub mod request;
+pub use request::{TransactionInput, TransactionRequest};
+
 mod signature;
+pub use signature::{Parity, Signature};
```
DaniPopes, line 30: :sunglasses:

DaniPopes (approved):

PR #405, opened 2024-03-27
feat: implement `admin_trait` 

AbnerZheng:
## Motivation
This Pull Request introduces AdminApi as an extension trait for the Provider, thereby enabling support for geth admin RPCs.

The NodeInfo type is taken from the `ethers-providers` package, with minor modifications to ensure compatibility:
   1. use `B256` instead of `H256`
   2. use `deserialize_json_u256` instead of `deserialize_stringified_numeric` to deserialize U256
   3. derive `copy` for some types because of `missing_copy_implementations`

If it is in the right direction, I would add the test for `add_peers`. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/lib.rs
+++ b/crates/provider/src/lib.rs
@@ -38,6 +38,7 @@ mod provider;
 pub use provider::{FilterPollerBuilder, Provider, RootProvider};
 
 pub mod utils;
+mod admin;
 
 #[doc(no_inline)]
 pub use alloy_network::{self as network, Network};
--- a/crates/provider/src/lib.rs
+++ b/crates/provider/src/lib.rs
@@ -38,7 +38,7 @@ mod provider;
 pub use provider::{FilterPollerBuilder, Provider, RootProvider};
 
 pub mod utils;
-mod admin;
+pub mod admin;
 
 #[doc(no_inline)]
 pub use alloy_network::{self as network, Network};
--- a/crates/provider/src/lib.rs
+++ b/crates/provider/src/lib.rs
@@ -37,8 +37,8 @@ pub use heart::{PendingTransaction, PendingTransactionBuilder, PendingTransactio
 mod provider;
 pub use provider::{FilterPollerBuilder, Provider, RootProvider};
 
-pub mod utils;
 pub mod admin;
+pub mod utils;
 
 #[doc(no_inline)]
 pub use alloy_network::{self as network, Network};
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -995,6 +995,26 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
     }
 }
 
+/// Admin namespace rpc interface that gives access to several non-standard RPC methods.
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[allow(unused, unreachable_pub)]
+pub trait AdminApi<N: Network, T: Transport + Clone = BoxTransport>: Provider<N, T> {
+    /// Adds the given node record to the peerset.
+    async fn add_peer(&self, record: String) -> TransportResult<bool> {
+        self.client().request("admin_addPeer", (record,)).await
+    }
+
+    /// Returns the ENR of the node.
+    async fn node_info(&self) -> TransportResult<String> {
+        self.client().request("admin_nodeInfo", ()).await
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl<N: Network, T: Transport + Clone> AdminApi<N, T> for RootProvider<N, T> {}
+
 #[cfg(test)]
 #[allow(clippy::missing_const_for_fn)]
 mod tests {
@@ -1419,5 +1439,15 @@ mod tests {
                 );
             }
         }
+    }
+
+    #[tokio::test]
+    async fn adds_peer() {
+        use super::AdminApi;
+        init_tracing();
+        let (provider, _anvil) = spawn_anvil();
+
+        let added = provider.add_peer("enr:-".to_string()).await.unwrap();
+        println!("{added:?}");
     }
 }
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -998,7 +998,7 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
 
 /// Admin namespace rpc interface that gives access to several non-standard RPC methods.
 #[allow(unused, unreachable_pub)]
-pub trait AdminApi {
+pub trait AdminApi<N, T> {
     /// Requests adding the given peer, returning a boolean representing
     /// whether or not the peer was accepted for tracking.
     fn add_peer(&self, record: &str) -> impl_future!(<Output = TransportResult<bool>>);
@@ -1024,29 +1024,35 @@ pub trait AdminApi {
     fn node_info(&self) -> impl_future!(<Output= TransportResult<NodeInfo>>);
 }
 
-impl<N: Network, T: Transport + Clone> AdminApi for RootProvider<N, T> {
+impl<N, T, P> AdminApi<N, T> for P
+where
+    N: Network,
+    T: Transport + Clone,
+    P: Provider<N, T>,
+{
     async fn add_peer(&self, record: &str) -> TransportResult<bool> {
-        self.inner.client_ref().request("admin_addPeer", (record,)).await
+        // self.client().
+        self.client().request("admin_addPeer", (record,)).await
     }
 
     async fn add_trusted_peer(&self, record: &str) -> TransportResult<bool> {
-        self.inner.client_ref().request("admin_addTrustedPeer", (record,)).await
+        self.client().request("admin_addTrustedPeer", (record,)).await
     }
 
     async fn remove_peer(&self, record: &str) -> TransportResult<bool> {
-        self.inner.client_ref().request("admin_removePeer", (record,)).await
+        self.client().request("admin_removePeer", (record,)).await
     }
 
     async fn remove_trusted_peer(&self, record: &str) -> TransportResult<bool> {
-        self.inner.client_ref().request("admin_removeTrustedPeer", (record,)).await
+        self.client().request("admin_removeTrustedPeer", (record,)).await
     }
 
     async fn peers(&self) -> TransportResult<Vec<PeerInfo>> {
-        self.inner.client_ref().request("admin_peers", ()).await
+        self.client().request("admin_peers", ()).await
     }
 
     async fn node_info(&self) -> TransportResult<NodeInfo> {
-        self.inner.client_ref().request("admin_nodeInfo", ()).await
+        self.client().request("admin_nodeInfo", ()).await
     }
 }
 
@@ -1489,7 +1495,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn peers() {
+    async fn admin_peers() {
         use super::AdminApi;
         init_tracing();
         let temp_dir = tempfile::TempDir::with_prefix("reth-test-1").unwrap();
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -37,6 +37,7 @@ use std::{
 
 #[cfg(feature = "pubsub")]
 use alloy_pubsub::{PubSubFrontend, Subscription};
+use alloy_rpc_types::admin::NodeInfo;
 
 /// A task that polls the provider with `eth_getFilterChanges`, returning a list of `R`.
 ///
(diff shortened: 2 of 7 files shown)


crates/provider/src/provider.rs
```diff
@@ -995,10 +996,35 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
     }
 }
 
+/// Admin namespace rpc interface that gives access to several non-standard RPC methods.
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[allow(unused, unreachable_pub)]
+pub trait AdminApi {
```
onbjerg, line 1003: how well supported is this namespace? if all clients support it, it should just be a part of the provider trait itself

mattsse: this is only ever enabled locally because privileged access.
An extension trait makes sense here imo

crates/rpc-types/src/eth/admin.rs
```diff
@@ -0,0 +1,410 @@
...
+use alloy_primitives::{B256, U256};
+use alloy_serde::json_u256::deserialize_json_u256;
+use enr::{k256::ecdsa::SigningKey, Enr};
+use serde::{Deserialize, Serialize};
+use std::net::{IpAddr, SocketAddr};
+
+/// This includes general information about a running node, spanning networking and protocol
+/// details.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct NodeInfo {
```
mattsse, line 12: this can be extracted from reth,

including the NodeRecord type

https://github.com/paradigmxyz/reth/blob/4d798c7b655255b84cbbcd3af2e8827de81e105e/crates/rpc/rpc-types/src/admin.rs#L16-L16

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,35 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
...
+    /// Adds the given node record to the peerset.
+    async fn add_peer(&self, record: String) -> TransportResult<bool>;
```
mattsse, line 1005: we have a few more admin endpoints

https://github.com/paradigmxyz/reth/blob/4d798c7b655255b84cbbcd3af2e8827de81e105e/crates/rpc/rpc-api/src/admin.rs#L8-L8

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,35 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
     }
 }
 
+/// Admin namespace rpc interface that gives access to several non-standard RPC methods.
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
```
mattsse, line 1001: async_trait is no longer required, but we need to do use this macro https://github.com/alloy-rs/alloy/blob/f7333c4344fdf6a16f620bc4392ba81b64684017/crates/pubsub/src/connect.rs#L18-L18

mattsse (changes requested):

crates/rpc-types/src/eth/admin.rs
AbnerZheng: `NodeRecord` includes a lot of logic to parse `enode`, which I remember can't handle edge case like missing ports.

crates/provider/src/provider.rs
AbnerZheng: Yeah, I can implement them later.

crates/provider/src/provider.rs
AbnerZheng: Done. Should I remove the usage of `async_trait` from `provider` and `rootProvider` too?

mattsse: not yet

crates/rpc-types/src/eth/admin.rs
```diff
@@ -0,0 +1,410 @@
...
+    pub id: B256,
+
+    /// The node's user agent, containing a client name, version, OS, and other metadata.
+    pub name: String,
+
+    /// The enode URL of the connected node.
+    pub enode: String,
+
+    /// The [ENR](https://eips.ethereum.org/EIPS/eip-778) of the running client.
+    pub enr: Enr<SigningKey>,
```
mattsse, line 23: I think, if we replace this with String, we can get rid of the enr dep entirely

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,66 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
...
+    /// whether or not the peer was accepted for tracking.
+    fn add_peer(&self, record: String) -> impl_future!(<Output = TransportResult<bool>>);
+
+    /// Requests adding the given peer as a trusted peer, which the node will
+    /// always connect to even when its peer slots are full.
+    fn add_trusted_peer(&self, record: String) -> impl_future!(<Output = TransportResult<bool>>);
+
+    /// Requests to remove the given peer, returning true if the enode was successfully parsed and
+    /// the peer was removed.
+    fn remove_peer(&self, record: String) -> impl_future!(<Output = TransportResult<bool>>);
```
mattsse, line 1012: replace string with &str?

mattsse (changes requested):
imo we should leave enr parsing up to the user and just use String for all those types

crates/provider/src/provider.rs
AbnerZheng: Done.

crates/rpc-types/src/eth/admin.rs
AbnerZheng: Done.

crates/provider/src/provider.rs
AbnerZheng: I tried to implement `subscribe_peer_events` as below:
```rust    
#[cfg(feature = "pubsub")]
  async fn subscribe_peer_events(&self) -> TransportResult<Subscription<String>> {
      self.root().pubsub_frontend()?;
      let id = self.client().request("admin_peerEvents", ()).await?;
      self.root().get_subscription(id).await
  }
```
and try to subscribe the peer events in tests by:
```rust
let ws = alloy_rpc_client::WsConnect::new(geth2.ws_endpoint());
let client = RpcClient::connect_pubsub(ws).await.unwrap();
let provider2 = RootProvider::<Ethereum, _>::new(client);
let subscription = provider2.subscribe_peer_events().await.unwrap().into_stream();
```
And run the test with `--all-features` but failed with an error message,
```sh
called `Result::unwrap()` on an `Err` value: ErrorResp(ErrorPayload { code: -32601, message: "the method admin_peerEvents does not exist/is not available", data: None })
```
Don't know how to enable it and I find the methods is not list in https://github.com/ethereum/go-ethereum/blob/f1801a9feda8f81532c92077d2c9a8b785fd699b/internal/web3ext/web3ext.go#L183

Maybe leave this method unimplemented now?

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,65 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
     }
 }
 
+/// Admin namespace rpc interface that gives access to several non-standard RPC methods.
+#[allow(unused, unreachable_pub)]
+pub trait AdminApi {
```
DaniPopes, line 1001: Please use `async_trait` to make it object-safe, see above.

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,65 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
...
+    fn peers(&self) -> impl_future!(<Output = TransportResult<Vec<PeerInfo>>>);
+
+    /// Returns general information about the node as well as information about the running p2p
+    /// protocols (e.g. `eth`, `snap`).
+    fn node_info(&self) -> impl_future!(<Output= TransportResult<NodeInfo>>);
+}
+
+impl<N: Network, T: Transport + Clone> AdminApi for RootProvider<N, T> {
+    async fn add_peer(&self, record: &str) -> TransportResult<bool> {
+        self.inner.client_ref().request("admin_addPeer", (record,)).await
```
DaniPopes, line 1029: `self.client()`

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,65 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
...
+
+    /// Returns the list of peers currently connected to the node.
+    fn peers(&self) -> impl_future!(<Output = TransportResult<Vec<PeerInfo>>>);
+
+    /// Returns general information about the node as well as information about the running p2p
+    /// protocols (e.g. `eth`, `snap`).
+    fn node_info(&self) -> impl_future!(<Output= TransportResult<NodeInfo>>);
+}
+
+impl<N: Network, T: Transport + Clone> AdminApi for RootProvider<N, T> {
```
DaniPopes, line 1027: implement for `P: Provider`

AbnerZheng: Done.

crates/provider/src/provider.rs
AbnerZheng: Done.

crates/provider/src/provider.rs
AbnerZheng: I used `impl_future!` for every methods, shouldn't it be enough?

mattsse: hmm, that's a valid point although since we can box the provider and impl the trait for any provider this trait doesn't really need to be object safe, but fine with using async_trait here as well.

AbnerZheng: Done.

crates/provider/src/provider.rs
```diff
@@ -995,10 +996,75 @@ impl<N: Network, T: Transport + Clone> Provider<N, T> for RootProvider<N, T> {
     }
 }
 
+/// Admin namespace rpc interface that gives access to several non-standard RPC methods.
+#[allow(unused, unreachable_pub)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+pub trait AdminApi<N, T>: Send + Sync {
```
mattsse, line 1003: can we extract this to admin.rs?

crates/provider/src/provider.rs
```diff
@@ -1420,4 +1486,38 @@ mod tests {
             }
         }
     }
+
+    #[tokio::test]
+    async fn node_info() {
+        use super::AdminApi;
```
mattsse, line 1492: should also be moved to admin.rs

crates/rpc-types/src/eth/admin.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    /// The node's user agent, containing a client name, version, OS, and other metadata.
+    pub name: String,
+
+    /// The enode URL of the connected node.
+    pub enode: String,
+
+    /// The [ENR](https://eips.ethereum.org/EIPS/eip-778) of the running client.
+    pub enr: String,
+
```
mattsse, line 23: no empty lines

crates/rpc-types/src/eth/admin.rs
```diff
@@ -0,0 +1,409 @@
...
+use alloy_serde::json_u256::deserialize_json_u256;
+use serde::{Deserialize, Serialize};
+use std::net::{IpAddr, SocketAddr};
+
+/// This includes general information about a running node, spanning networking and protocol
+/// details.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct NodeInfo {
+    /// The node's private key.
+    pub id: B256,
```
mattsse, line 13: this is the wrong type, should be B512

crates/rpc-types/src/eth/admin.rs
```diff
@@ -0,0 +1,409 @@
...
+use alloy_genesis::ChainConfig;
+use alloy_primitives::{B256, U256};
+use alloy_serde::json_u256::deserialize_json_u256;
+use serde::{Deserialize, Serialize};
+use std::net::{IpAddr, SocketAddr};
+
+/// This includes general information about a running node, spanning networking and protocol
+/// details.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct NodeInfo {
```
mattsse, line 11: can we double check the field order of this type?

https://github.com/ethereum/go-ethereum/blob/767b00b0b514771a663f3362dd0310fc28d40c25/p2p/server.go#L1072-L1085

AbnerZheng: Just checked, the order here is correct, should I fire a PR for reth?

https://github.com/paradigmxyz/reth/blob/0a803c49a9fac1bd44b94a1883cf352ced6816ee/crates/rpc/rpc-types/src/admin.rs#L16-L21

crates/rpc-types/src/eth/admin.rs
AbnerZheng: Yes, you are right. I checked both reth and web3.py are using B512. 

https://web3py.readthedocs.io/en/stable/web3.geth.html#web3.geth.admin.node_info

Ethers-rs is using H256 here, so I assume it should be updated?
https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/ext/admin.rs#L14

crates/rpc-types/src/eth/admin.rs
AbnerZheng: Done.

crates/provider/src/provider.rs
AbnerZheng: Done.

crates/provider/src/provider.rs
AbnerZheng: Done.

crates/rpc-types/src/eth/admin.rs
mattsse: not necessary, we'll replace reth's types with this

crates/rpc-types/src/eth/admin.rs
mattsse: I see, for the pr we need to figure out what this is set to

crates/provider/src/lib.rs
```diff
@@ -38,6 +38,7 @@ mod provider;
 pub use provider::{FilterPollerBuilder, Provider, RootProvider};
 
 pub mod utils;
+mod admin;
```
mattsse, line 41:
```suggestion
pub mod admin;
```

mattsse (approved):
lgtm

PR #414, opened 2024-03-27
perf: remove getBlock request in feeHistory

mattsse:
the basefee of the "Latest" block is the second last entry in the vec, if the feeHistory response is correct then we don't need to fetch the block.

this keeps the block request as fallback but perhaps we should just error here?

see MM for ref:

https://github.com/MetaMask/core/blob/main/packages/gas-fee-controller/src/fetchGasEstimatesViaEthFeeHistory/calculateGasFeeEstimatesForPriorityLevels.ts#L62

### DIFF
--- a/crates/rpc-types/src/eth/fee.rs
+++ b/crates/rpc-types/src/eth/fee.rs
@@ -64,6 +64,13 @@ pub struct FeeHistory {
 }
 
 impl FeeHistory {
+    /// Returns the base fee of the requested block in the `eth_feeHistory` request.
+    pub fn latest_block_base_fee(&self) -> Option<U256> {
+        // the base fee of what is considered the "latest" block is the second last element in the
+        // list
+        self.base_fee_per_gas.iter().rev().skip(1).next().copied()
+    }
+
     /// Returns the base fee of the next block.
     pub fn next_block_base_fee(&self) -> Option<U256> {
         self.base_fee_per_gas.last().copied()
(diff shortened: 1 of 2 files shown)


mattsse:
the failing test looks more like an anvil issue

```
 FeeHistory {
    base_fee_per_gas: [
        0x0_U256,
        0x0_U256,
    ],
    gas_used_ratio: [
        0.0,
    ],
```

mattsse:
Update:

basefee entries are 0 pre EIP-1559 for eth_feeHistory

https://github.com/ethereum/go-ethereum/blob/767b00b0b514771a663f3362dd0310fc28d40c25/eth/gasprice/feehistory.go#L83-L83

so this checks for is_zero to fallback to getBlock to confirm, although we could just return an error here because 0 is unreachable anyway

crates/rpc-types/src/eth/fee.rs
```diff
@@ -64,6 +64,13 @@ pub struct FeeHistory {
 }
 
 impl FeeHistory {
+    /// Returns the base fee of the requested block in the `eth_feeHistory` request.
```
onbjerg, line 67:
```suggestion
    /// Returns the base fee of the latest block in the `eth_feeHistory` request.
```

crates/rpc-types/src/eth/fee.rs
```diff
@@ -81,4 +88,18 @@ impl FeeHistory {
             })
             .copied()
     }
+
+    /// Returns the blob fee of the requested block in the `eth_feeHistory` request.
```
onbjerg, line 92:
```suggestion
    /// Returns the blob fee of the latest block in the `eth_feeHistory` request.
```

onbjerg (approved):

PR #418, opened 2024-03-28
Move Otterscan types to alloy

nkysg:
## Motivation
Move Otterscan types to alloy
https://github.com/paradigmxyz/reth/issues/7372


mattsse (approved):
ty

Issue #416, opened 2024-03-28
[Bug] Fix unknown/unused `cfg`s


DaniPopes:
### Component

consensus, eips, genesis

### Operating System

None

### Describe the bug

`cargo check -Zcheck-cfg` reports a lot of unexpected cfg values. Consider also adding this to CI.

DaniPopes:
cc @yash-atreya

yash-atreya:
@developeruche 
> Could you help me with a reference to creating an RpcClient using the Http as the transport layer

If you've enabled the `reqwest` feature on `alloy-rpc-client` then you can do the following to instantiate a `RpcClient<Http<Client>>`:

```rust
let client = Rpc<Http<Client>>::new_http(URL);
```

Also can you please share which version of alloy you're using i.e the commit hash you're pinned to?

developeruche:
> > hey @DaniPopes, on enabling the "reqwest" feature
> > ```rust
> >  #[cfg(feature = "reqwest")]
> >     pub fn reqwest_http(self, url: url::Url) -> TransportResult<RpcClient<L::Service>>
> >     where
> >         L: Layer<alloy_transport_http::Http<reqwest::Client>>,
> >         L::Service: Transport,
> >     {
> >         let transport = alloy_transport_http::Http::new(url);
> >         let is_local = transport.guess_local();
> > 
> >         Ok(self.transport(transport, is_local))
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am get this error;
> > crates/provider/src/builder.rs:236:27 | 236 | Ok(self.on_client(client)) | --------- ^^^^^^ expected `RpcClient<BoxTransport>`, found `RpcClient<Http<Client>>` | | | arguments to this method are incorrect | = note: expected struct `RpcClient<BoxTransport>` found struct `RpcClient<alloy_transport_http::Http<Client>>`
> > Could you help me with a reference to creating an `RpcClient` using the Http as the transport layer
> > cc: @DaniPopes
> 
> @developeruche Implemented a fix here: #419

Awesome 🙌

PR #419, opened 2024-03-28
Adds `check -Zcheck-cfg ` job

yash-atreya:
## Motivation

Ref: #416 

And fixes: https://github.com/alloy-rs/alloy/issues/416#issuecomment-2025723286

## Solution

Adds `reqwest` and `hyper` features in `alloy-provider`.
@prestwich  Unsure why but adding these features forced the `ProviderBuilder::on_client(&self, client: RpcClient<T>)` to only accept `RpcClient<BoxTransport>`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


yash-atreya:
Also, I think we should make the `reqwest` feature default in `alloy-provider` just like its in `alloy-rpc-client`

crates/provider/src/builder.rs
```diff
@@ -225,23 +225,25 @@ impl<L, N> ProviderBuilder<L, N> {
...
 
+    /// Build this provider with hyper HTTP transport.
     #[cfg(feature = "hyper")]
     pub fn on_hyper_http(self, url: url::Url) -> Result<L::Provider, TransportError>
     where
         L: ProviderLayer<RootProvider<N, BoxTransport>, N, BoxTransport>,
         N: Network,
     {
-        let client = ClientBuilder::default().hyper_http(url);
+        let client = ClientBuilder::default().hyper_http(url).boxed();
```
DaniPopes, line 246: I don't think this should be boxed

zerosnacks:
> Also, I think we should make the `reqwest` feature default in `alloy-provider` just like its in `alloy-rpc-client`

Added here: https://github.com/alloy-rs/alloy/pull/421

crates/provider/src/builder.rs
yash-atreya: >@prestwich Unsure why but adding these features forced the ProviderBuilder::on_client(&self, client: RpcClient<T>) to only accept RpcClient<BoxTransport>

Yes it shouldn't.

prestwich: oh I just saw this in #379 and was confused about it 🤔

prestwich: i think what happened is that the blocks were previously entirely non-functional, and therefore were written incorrectly and we didn't notice (because they were non-functional and never checked by the compiler). Now that the cfg blocks are effective, a compiler warning appears. I'll fix in #379 in a second

prestwich:
@yash-atreya would you rebase now that #379 is in?

yash-atreya:
> @yash-atreya would you rebase now that #379 is in?

Yeah I'll take care of it

prestwich (approved):

PR #428, opened 2024-03-30
feat: add network-parameterized block responses

### LINKED ISSUE #267: [Feature] Add network-parameterized block responses
### Component

network, json-rpc, providers, pubsub

#190 replaced the provider with a network-parameterized provider, but block responses are still just Ethereum headers and Ethereum transactions. We need a `Block` type generic for networks.

We have one in `alloy-network` but we need to check that it is correct and use it in `alloy-providers`:

https://github.com/alloy-rs/alloy/blob/63af0c4dd43a94126348afaf94b744b98608c293/crates/network/src/lib.rs#L36-L51

### Additional context

_No response_

leruaa:
Closes #267


crates/network/src/lib.rs
```diff
@@ -33,24 +34,22 @@ pub use any::AnyNetwork;
...
-/// A block response
-#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
-pub struct BlockResponse<N: Network> {
-    #[serde(flatten)]
-    header: N::HeaderResponse,
-    transactions: TransactionList<N::TransactionResponse>,
+/// A header response.
+pub trait HeaderResponse {
+    /// Base fee per unit of gas (if past London)
+    fn base_fee_per_gas(&self) -> &Option<U256>;
```
mattsse, line 48:
```suggestion
    fn base_fee_per_gas(&self) -> Option<U256>;
```

crates/network/src/lib.rs
```diff
@@ -33,24 +34,22 @@ pub use any::AnyNetwork;
...
+    /// Returns the blob fee for the next block according to the EIP-4844 spec.
+    ///
+    /// Returns `None` if `excess_blob_gas` is None.
+    fn next_block_blob_fee(&self) -> Option<u128>;
```
mattsse, line 52: type mixing for fees is not ideal, perhaps we should just use U256 everywhere for fees  @prestwich

mattsse, line 52: for header we need a few more things, but all of these should be moved to a header trait I believe

crates/network/src/lib.rs
```diff
@@ -60,7 +59,7 @@ pub struct BlockResponse<N: Network> {
 /// [`TxReceipt`]: alloy_consensus::TxReceipt
 pub trait ReceiptResponse {
     /// Address of the created contract, or `None` if the transaction was not a deployment.
-    fn contract_address(&self) -> Option<Address>;
+    fn contract_address(&self) -> &Option<Address>;
```
mattsse, line 62:
```suggestion
    fn contract_address(&self) -> Option<Address>;
```

crates/rpc-types/src/transaction_list.rs
```diff
@@ -0,0 +1,40 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+
+/// A list of transactions, either hydrated or hashes.
```
mattsse, line 4:
```suggestion
/// A list of transactions, either full, hashes or uncle for uncle blocks pre-merge.
```

crates/network/src/lib.rs
```diff
@@ -33,24 +34,22 @@ pub use any::AnyNetwork;
...
-#[serde(untagged)]
-pub enum TransactionList<T> {
-    /// Hashes only.
-    Hashes(Vec<B256>),
-    /// Hydrated tx objects.
-    Hydrated(Vec<T>),
-    /// Special case for uncle response
-    Uncled,
+/// A block response.
+pub trait BlockResponse<N: Network> {
```
mattsse, line 38: here we can also integrate a few more helper functions that currently exist on block, for iterating over txs for example.

mattsse (changes requested):
a few suggestions,

it looks like we also need traits for header+block and enshrine Self::Header, Self::Block in the rpc response objects @prestwich ?

crates/rpc-types/src/transaction_list.rs
```diff
@@ -0,0 +1,40 @@
+use alloy_primitives::B256;
+use serde::{Deserialize, Serialize};
+
+/// A list of transactions, either full, hashes or uncle for uncle blocks pre-merge.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum TransactionList<T> {
```
prestwich, line 7: We can set a default type here

crates/rpc-types/src/eth/block.rs
```diff
@@ -144,20 +138,7 @@ impl Header {
...
-    /// Only hashes
-    Hashes(Vec<B256>),
-    /// Full transactions
-    Full(Vec<Transaction>),
-    /// Special case for uncle response.
-    Uncle,
-}
-
-impl BlockTransactions {
+impl TransactionList<Transaction> {
```
prestwich, line 141: could some of this be `impl<T: Transaction> TransactionList<T>` ?

prestwich (changes requested):

crates/rpc-types/src/eth/block.rs
prestwich: for some `Transaction` trait 🤔

prestwich:
looking pretty good. trying to figure out what's both maximally simple and maximally reusable for other network implementers

crates/network/src/lib.rs
prestwich: we need to action #433

leruaa:
@prestwich I'm starting to get confused about what is left to be done in this PR, and if I understood the comments above correctly. In particular:

- Was I correct to rename `BlockResponse` and `HeaderResponse` traits to just `Block` and `Header`?
- I noticed there is a associated `Header` type on `Network` but it's used nowhere. Where is it supposed to be used?
- Should I add associated  `Block` and `Traansaction` types on `Network`?

Thanks in advance for your time

PR #431, opened 2024-03-30
feat: refactor request builder workflow

### LINKED ISSUE #423: Improve tx request builder
while we're touching this code, we should decompose this into a set of functions `fn will_build_legacy() -> bool` that capture our criteria for building each style of transaction. then this becomes a fallthrough if

I also think our criteria here are not most-friendly to users.

- if a sidecar is set, we should ALWAYS build 4844, regardless of if legacy gas is indicated
- if `max_fee_per_blob_gas` is set, but no sidecar/blob is set, that should be a build error
- if `access_list` is set, we should ALWAYS build 2930, regardless of if legacy gas is indicated
- if `sidecar` and `access_list` are both set throw a build error
- 1559 should be our default tx type in the fallthrough case. legacy should be chosen only if gas_price is set.

function should look something as follows:
```
if self.will_build_4844() { self.build_4844() } 
else if self.will_build_2930() { self.build_2930() }
else if self.will_build_legacy() { self.build_legacy() }
else { self.build_1559() } 
```

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/383#discussion_r1544689297_

leruaa:
Closes #423

### DIFF
--- a/crates/network/src/ethereum/builder.rs
+++ b/crates/network/src/ethereum/builder.rs
@@ -1,6 +1,5 @@
 use crate::{
-    BuilderResult, Ethereum, Network, NetworkSigner, TransactionBuilder, TransactionBuilderError,
-    Unbuilt,
+    BuildResult, Ethereum, Network, NetworkSigner, TransactionBuilder, TransactionBuilderError,
 };
 use alloy_consensus::{BlobTransactionSidecar, TxType, TypedTransaction};
 use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
@@ -151,7 +150,7 @@ impl TransactionBuilder<Ethereum> for TransactionRequest {
         self.populate_blob_hashes();
     }
 
-    fn build_unsigned(self) -> Result<TypedTransaction, Unbuilt<Self>> {
+    fn build_unsigned(self) -> BuildResult<TypedTransaction, Ethereum> {
         if let Err((tx_type, missing)) = self.missing_keys() {
             return Err((
                 self,
@@ -164,7 +163,7 @@ impl TransactionBuilder<Ethereum> for TransactionRequest {
     async fn build<S: NetworkSigner<Ethereum>>(
         self,
         signer: &S,
-    ) -> BuilderResult<<Ethereum as Network>::TxEnvelope> {
+    ) -> Result<<Ethereum as Network>::TxEnvelope, TransactionBuilderError<Ethereum>> {
         Ok(signer.sign_request(self).await?)
     }
 }
--- a/crates/network/src/ethereum/builder.rs
+++ b/crates/network/src/ethereum/builder.rs
@@ -114,6 +114,15 @@ impl TransactionBuilder<Ethereum> for TransactionRequest {
         self.sidecar = Some(sidecar);
     }
 
+    fn complete_type(&self, ty: TxType) -> Result<(), Vec<&'static str>> {
+        match ty {
+            TxType::Legacy => self.complete_legacy(),
+            TxType::Eip2930 => self.complete_2930(),
+            TxType::Eip1559 => self.complete_1559(),
+            TxType::Eip4844 => self.complete_4844(),
+        }
+    }
+
     fn can_submit(&self) -> bool {
         // value and data may be None. If they are, they will be set to default.
         // gas fields and nonce may be None, if they are, they will be populated
--- a/crates/network/src/ethereum/builder.rs
+++ b/crates/network/src/ethereum/builder.rs
@@ -141,6 +141,22 @@ impl TransactionBuilder<Ethereum> for TransactionRequest {
         common && (legacy || eip2930 || eip1559 || eip4844)
     }
 
+    fn output_tx_type(&self) -> TxType {
+        todo!()
+    }
+
+    fn output_tx_type_checked(&self) -> BuilderResult<TxType> {
+        if will_build_4844(self)? {
+            Ok(TxType::Eip4844)
+        } else if will_build_2930(self)? {
+            Ok(TxType::Eip2930)
+        } else if will_build_legacy(self)? {
+            Ok(TxType::Legacy)
+        } else {
+            Ok(TxType::Eip1559)
+        }
+    }
+
     fn build_unsigned(self) -> BuilderResult<TypedTransaction> {
         build_unsigned::<Ethereum>(self)
     }
@@ -159,14 +175,11 @@ where
     N: Network,
     N::UnsignedTx: From<TxLegacy> + From<TxEip1559> + From<TxEip2930> + From<TxEip4844Variant>,
 {
-    if will_build_4844(&request)? {
-        build_4844(request).map(TxEip4844Variant::from).map(Into::into)
-    } else if will_build_2930(&request)? {
-        build_2930(request).map(Into::into)
-    } else if will_build_legacy(&request)? {
-        build_legacy(request).map(Into::into)
-    } else {
-        build_1559(request).map(Into::into)
+    match request.output_tx_type_checked()? {
+        TxType::Legacy => build_legacy(request).map(Into::into),
+        TxType::Eip2930 => build_2930(request).map(Into::into),
+        TxType::Eip1559 => build_1559(request).map(Into::into),
+        TxType::Eip4844 => build_4844(request).map(TxEip4844Variant::from).map(Into::into),
     }
 }
 
--- a/crates/network/src/ethereum/builder.rs
+++ b/crates/network/src/ethereum/builder.rs
@@ -199,78 +199,77 @@ fn get_invalid_1559_fields(request: &TransactionRequest) -> Vec<InvalidTransacti
 }
 
 fn will_build_legacy(request: &TransactionRequest) -> Result<bool, TransactionBuilderError> {
-    if request.gas_price.is_some() {
-        let errors = get_invalid_common_fields(request);
+    if request.gas_price.is_none() {
+        return Ok(false);
+    }
 
-        if errors.is_empty() {
-            Ok(true)
-        } else {
-            Err(TransactionBuilderError::InvalidTransactionRequest(TxType::Legacy, errors))
-        }
+    let errors = get_invalid_common_fields(request);
+
+    if errors.is_empty() {
+        Ok(true)
     } else {
-        Ok(false)
+        Err(TransactionBuilderError::InvalidTransactionRequest(TxType::Legacy, errors))
     }
 }
 
 fn will_build_2930(request: &TransactionRequest) -> Result<bool, TransactionBuilderError> {
-    if request.access_list.is_some() {
-        let mut errors = get_invalid_common_fields(request);
+    if request.access_list.is_none() {
+        return Ok(false);
+    }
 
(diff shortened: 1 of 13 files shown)


crates/network/src/ethereum/builder.rs
```diff
@@ -127,30 +127,34 @@ where
...
-        // If only accesslist is set, and there are no EIP-1559 fees
-        (_, None, Some(_), None, None, None) => build_2930(request).map(Into::into),
-        // EIP-1559
-        // If EIP-4844 fields are missing
-        (None, _, _, None, None, None) => build_1559(request).map(Into::into),
-        // EIP-4844
-        // All blob fields required
-        (None, _, _, Some(_), Some(_), Some(_)) => {
-            build_4844(request).map(TxEip4844Variant::from).map(Into::into)
+        (_, Some(_), None, _) => Err(TransactionBuilderError::MissingKey("blob_versioned_hashes")),
```
prestwich, line 135: can we turn these into functions as well? even with 2 things gone, this match block is really bad code smell

crates/network/src/ethereum/builder.rs
```diff
@@ -194,6 +198,10 @@ fn build_1559(request: TransactionRequest) -> Result<TxEip1559, TransactionBuild
     })
 }
 
+const fn will_build_2930(request: &TransactionRequest) -> bool {
```
prestwich, line 201: feature req:

`fn can_build_*() -> Result<(), TransactionBuilderError>`

Should return a list of missing keys required to build that transaction type

crates/network/src/ethereum/builder.rs
```diff
@@ -214,6 +222,10 @@ fn build_2930(request: TransactionRequest) -> Result<TxEip2930, TransactionBuild
     })
 }
 
+const fn will_build_4844(request: &TransactionRequest) -> bool {
+    request.sidecar.is_some()
+}
+
 /// Build an EIP-4844 transaction.
 fn build_4844(request: TransactionRequest) -> Result<TxEip4844, TransactionBuilderError> {
```
prestwich, line 230: related to other comment:
I don't like that these functions will return only 1 missing key per invocation. makes it annoying to debug fill strategies, etc. It'd be nice if we could use `can_build_4844` to find all missing keys in one invocation

prestwich (changes requested):
good direction 🦺

crates/network/src/ethereum/builder.rs
leruaa: If I understand correctly, `TransactionBuilderError::MissingKey(&'static str)` should be changed to   `TransactionBuilderError::MissingKeys(Vec<&'static str>)`, right ?

prestwich: yes, that is how i would do it

crates/network/src/ethereum/builder.rs
```diff
@@ -126,48 +127,128 @@ where
...
+
+    if request.max_fee_per_gas.is_none() {
+        errors.push(InvalidTransactionRequestError::MissingKey("max_fee_per_gas"));
+    }
+
+    errors
+}
+
+fn will_build_legacy(request: &TransactionRequest) -> Result<bool, TransactionBuilderError> {
+    if request.gas_price.is_some() {
```
prestwich, line 170: would prefer inverting and early-return these `if`s to avoid nesting

```
if request.gas_price.is_none { 
    return Ok(false0
}

let errors = get_invalid_common_fields(request);
if errors.is_empty() {
    ....
} else {
    ....
}

```

crates/network/src/ethereum/builder.rs
```diff
@@ -126,48 +127,128 @@ where
...
+    }
+
+    if request.gas.is_none() {
+        errors.push(InvalidTransactionRequestError::MissingKey("gas_limit"));
+    }
+
+    InvalidTransactionRequestErrors(errors)
+}
+
+fn get_invalid_1559_fields(request: &TransactionRequest) -> Vec<InvalidTransactionRequestError> {
```
prestwich, line 155: these could be validity-checking methods on `TransactionRequest`

crates/network/src/transaction/builder.rs
```diff
@@ -1,15 +1,20 @@
...
 use futures_utils_wasm::impl_future;
 
 /// Error type for transaction builders.
 #[derive(Debug, thiserror::Error)]
 pub enum TransactionBuilderError {
-    /// A required key is missing.
-    #[error("A required key is missing: {0}")]
-    MissingKey(&'static str),
+    /// Invalid transaction request
+    #[error("{0:?} transaction can't be built: {1}")]
```
prestwich, line 16: Go ahead and add a `impl Display for TxType` :)

crates/network/src/transaction/builder.rs
```diff
@@ -38,6 +39,44 @@ impl TransactionBuilderError {
     }
 }
 
+/// Wrapper for [`InvalidTransactionRequestError`]s.
+#[derive(Debug)]
+pub struct InvalidTransactionRequestErrors(pub Vec<InvalidTransactionRequestError>);
```
prestwich, line 44: it seems like this newtype is not adding any functionality, or preventing any common mistakes. It can be omitted, with the buildererror variant containing an unwrapped Vec

crates/network/src/ethereum/builder.rs
```diff
@@ -126,48 +127,128 @@ where
...
+            Ok(true)
+        } else {
+            Err(TransactionBuilderError::InvalidTransactionRequest(TxType::Legacy, errors))
+        }
+    } else {
+        Ok(false)
+    }
+}
+
+fn will_build_2930(request: &TransactionRequest) -> Result<bool, TransactionBuilderError> {
```
prestwich, line 183: these can be `pub`

prestwich (changes requested):
feature req:

```
/// Returns the transaction type that this builder will attempt to build. This does not imply 
/// that the builder is ready to build.
pub fn output_tx_type(&self) -> TxType

/// Returns the transaction type that this builder will build. `Err` if the builder
/// is not ready to build, containing the attempt type and missing/conflicted information
pub fn output_tx_type_checked(&self) -> Result<TxType, Vec<TransactionRequestError>>;
```

crates/network/src/transaction/builder.rs
leruaa: It's for the `Display` impl

prestwich: let's manually impl that display?

crates/network/src/ethereum/builder.rs
Evalir: Let's also add some docs

prestwich:
the more i look at this the more complexity I see. e.g. a `TransactionRequest` may be valid for submission via `send_transaction` but not sufficiently filled to allow for signing via `build`

I may do some work on this 🤔

leruaa:
@prestwich Did you have time to have a look at this? Or maybe you can give pointers on the remaining work needed?

prestwich:
i'm still traveling for the next few days, so I don't have time to properly review :( I will be working on this and may take it over somewhat. sorry for taking so long and then stepping on your toes

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -166,10 +148,327 @@ impl TransactionRequest {
...
+    pub fn preferred_type(&self) -> TxType {
+        if self.sidecar.is_some() || self.max_fee_per_blob_gas.is_some() {
+            TxType::Eip4844
+        } else if self.access_list.is_some() {
+            TxType::Eip2930
+        } else if self.gas_price.is_some() {
+            TxType::Legacy
+        } else {
+            TxType::Eip1559
+        }
```
mattsse, line 354: should 2930 only be if gas price is also set?

otherwise all txs with 1559 fee fields and an accesslist will be 2930

crates/network/src/transaction/builder.rs
```diff
@@ -227,8 +230,24 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
     /// a valid transaction.
     fn can_build(&self) -> bool;
 
+    /// Returns the transaction type that this builder will attempt to build.
+    /// This does not imply that the builder is ready to build.
+    fn output_tx_type(&self) -> TxType;
```
mattsse, line 235: hmm, can we use the txtype in the trait here?

because txtypes are network specific

could operate on u8 or make txtype part of the network trait

mattsse (changes requested):
this flow makes sense, but I worry the preferred type detection has unexpected results

how to enforce `build` will actually result in specific type?

I think form a builder pov, I'd also like to have something like
```rust
 builder().
   ....
   .ensure_type(ty: impl AsRef<u8>)?
   .build_preferred(, ty: impl AsRef<u8>)
```

crates/network/src/transaction/builder.rs
prestwich: good catch, we cannot. 

> could operate on u8 or make txtype part of the network trait

I am partial to the idea of using `TxType` on the network trait, but at that point, i would also want to incorporate into `En/Decodable2718`, and make something like `trait En/Decodable2718<T: Into<u8> + TryFrom<u8, Error = _>>` or similar so that we can enforce that the 2718 implementation uses the network `TxType` via a `type TxEnvelope: Encodable2718<Self::TxType>` bound 

let me think about it a bit. most likely outcome is that this PR will operate on u8, and i will reserve network-assoc txtype for a later pr

crates/rpc-types/src/eth/transaction/request.rs
prestwich: my basic design approach here is to prefer the MOST specific (LEAST common) thing set, to try to conform to user intent in most cases. So preference order is basically as follows:

- Blobs are specific and critical to the tx, and have highest priority
- Access lists are specific and (almost always) not critical 
- legacy gas is specific and never critical
- eip1559 is the default case if no more-specific intention has been specified

> otherwise all txs with 1559 fee fields and an accesslist will be 2930

this seems like correct behavior to me, because if the user specified an access list it is a more-specific intention than specifying 1559 gas

crates/network/src/transaction/builder.rs
prestwich: okay i made a network specific tx type because it felt right >.>

prestwich:
> this flow makes sense, but I worry the preferred type detection has unexpected results
> 
> how to enforce `build` will actually result in specific type?
> 
> I think form a builder pov, I'd also like to have something like
> 
> ```rust
>  builder().
>    ....
>    .ensure_type(ty: impl AsRef<u8>)?
>    .build_preferred(, ty: impl AsRef<u8>)
> ```


@mattsse i've added the following. I copied the pattern from other builders, but i couldn't tell you which ones because it's beena minute since I saw them
```
 fn assert_preferred(&self, ty: N::TxType) -> Self {
    dbg_assert_eq!(self.preferred_type(), ty);
}

fn assert_preferred_chained(self, ty: N::TxType) -> Self { 
    self.assert_preferred(ty);
    self
}
```

prestwich:
@mattsse i failed to consider AnyTxType in the abstraction. will need to do some work there

PR #432, opened 2024-03-31
fix: infinite loop while decoding a list of transactions

Brechtpd:
In `network_decode` the decoded bytes are not correctly consumed in the buffer which makes decoding break for anything that requires more decoding afterwards, and also makes it possible for the decoding to hang.

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -392,6 +392,29 @@ mod tests {
     }
 
     #[test]
+    fn test_encode_decode_transaction_list() {
+        let signature = Signature::test_signature();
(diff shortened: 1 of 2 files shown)


prestwich:
```
    #[inline]
    fn advance(&mut self, cnt: usize) {
        if self.len() < cnt {
            panic_advance(cnt, self.len());
        }

        *self = &self[cnt..];
    }
```

https://docs.rs/bytes/latest/src/bytes/buf/buf_impl.rs.html#1412-1418

Above is the `advance` code from `impl Buf for &[u8]`. Can you explain how calling `buf.advance(1)` differs in behavior from `let buf = &mut &buf[1..]`;

Issue #387, opened 2024-03-24
[Feature] Add Http::new for reqwest::Client


mattsse:
### Component

rpc

### Describe the feature you would like

the most commonly used http transport is reqwest::Client, currently it's required to provide the type:

 `let http = Http::<Client>::new(url);`

this requires to always pull in reqwest.

I think it would be reasonable to move the `new` function to `impl Http<reqwest::Client>`, there is already `Http::with_client` that is sufficient for all other use cases.

nkysg:
I can try this.

Issue #397, opened 2024-03-26
[Bug] Missing `HttpConnect` in alloy-transport-http


prestwich:
### Component

transports

### Operating System

None

### Describe the bug

Should be a Connect struct there

PR #434, opened 2024-04-02
feat: add Http::new for reqwest::Client

nkysg:
## Motivation
close https://github.com/alloy-rs/alloy/issues/387

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/src/lib.rs
+++ b/crates/transport-http/src/lib.rs
@@ -49,14 +49,6 @@ pub struct Http<T> {
 }
 
 impl<T> Http<T> {
-    /// Create a new [`Http`] transport.
-    pub fn new(url: Url) -> Self
-    where
-        T: Default,
-    {
-        Self { client: Default::default(), url }
-    }
-
     /// Create a new [`Http`] transport with a custom client.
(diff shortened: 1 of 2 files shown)


onbjerg (approved):
lgtm

prestwich (changes requested):
this prevents us from using `new` to instantiate Http<hyper::Whatever>`. I would rather match the pattern eslewhere in the repo `new_hyper` and `new_reqwest`

could we instead take this as a chance to address #397 ?

nkysg:
> this prevents us from using `new` to instantiate Httphyper::Whatever`. I would rather match the pattern eslewhere in the repo `new_hyper`and`new_reqwest`
> 
> could we instead take this as a chance to address #397 ?

do you want to implement impl Http<HyperClient>::new ?

prestwich (approved):
> this prevents us from using new to instantiate Http<hyper::Whatever>. I would rather match the pattern eslewhere in the repo new_hyper and new_reqwest

after #462 the naming convention is `new` and `new_hyper` so I'm approving this PR

Issue #290, opened 2024-03-13
[Feature] Make provider polling interval configurable


DaniPopes:
### Component

provider, pubsub

### Describe the feature you would like

Should be a method in both `RootProvider` and `RpcClient`

https://github.com/alloy-rs/alloy/blob/512e15f23648726ddf5262804ac16eb51664cef2/crates/rpc-client/src/client.rs#L167

prestwich:
closed by #437 and #767

PR #437, opened 2024-04-02
configure polling interval

yash-atreya:
## Motivation

#290 

## Solution

Add methods to `RpcClient` and `RootProvider` to configure the polling interval. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


yash-atreya:
This wouldn't work. Making changes

crates/rpc-client/src/client.rs
```diff
@@ -81,6 +81,21 @@ impl<T> RpcClient<T> {
...
+    /// Note: This will only set the poll interval for the client if it is the only reference to the
+    /// inner client. If the reference is held by many, then it will not update the poll interval.
+    pub fn with_poll_interval(mut self, poll_interval: Duration) -> Self {
+        let inner = Arc::get_mut(&mut self.0);
+
+        if let Some(inner) = inner {
+            inner.set_poll_interval(poll_interval);
+            self
+        } else {
+            todo!("TBD: Value is already shared. Cannot be mutated");
```
prestwich, line 96: we could have the poll interval be an atomic number of seconds instead of a duration?

yash-atreya: Even if we have it as an atomic number, we would still need to use `Arc::get_mut` since `RpcClientInner` is wrapped in `Arc`.

prestwich: atomics don't require a `&mut` reference to modify

crates/rpc-client/src/client.rs
```diff
@@ -152,22 +160,33 @@ pub struct RpcClientInner<T> {
...
-            Duration::from_secs(7)
-        }
+    /// Returns the default poll interval (milliseconds) for the client.
+    pub fn poll_interval(&self) -> u64 {
+        self.poll_interval.load(Ordering::Relaxed)
+    }
+
+    /// Set the poll interval for the client in milliseconds.
+    pub fn set_poll_interval(&self, poll_interval: u64) {
+        self.poll_interval.store(poll_interval, Ordering::SeqCst);
```
prestwich, line 189: feels like it could be relaxed as well?

prestwich (approved):
looking good. only nit

crates/rpc-client/src/client.rs
```diff
@@ -152,22 +160,33 @@ pub struct RpcClientInner<T> {
...
     pub(crate) id: AtomicU64,
+    /// The poll interval for the client in milliseconds.
+    pub(crate) poll_interval: AtomicU64,
 }
 
 impl<T> RpcClientInner<T> {
     /// Create a new [`RpcClient`] with the given transport.
+    ///
+    /// Note: Sets the poll interval to 250ms for local transports and 7s for remote transports by
+    /// default.
```
mattsse, line 171: we could tweak this based in the chain, because 7s exceeds avg blocktime of a lot of chains.

the alloy_chains::Chain has a function for this.

so we could integrate this in eth_chainId call for example and track whether we've already set the poll_interval based on chain_id

mattsse (approved):

crates/rpc-client/src/client.rs
```diff
@@ -152,22 +160,33 @@ pub struct RpcClientInner<T> {
...
     #[inline]
     pub const fn new(t: T, is_local: bool) -> Self {
-        Self { transport: t, is_local, id: AtomicU64::new(0) }
+        Self {
+            transport: t,
+            is_local,
+            id: AtomicU64::new(0),
+            poll_interval: if is_local { AtomicU64::new(250) } else { AtomicU64::new(7000) },
```
gakonst, line 178: i'd make this 2s given the rise of L2s with faster blocktimes

Issue #328, opened 2024-03-15
[Feature] Multicall Support


zilayo:
### Component

contract

### Describe the feature you would like

Similar to ethers-rs, it would be great to port over multicall to alloy.

https://github.com/gakonst/ethers-rs/tree/master/ethers-contract/src/multicall

developeruche:
Would like to work on this, but I don't think this is a core implementation that should be in the `contract` crate... I think it would be more oraganised to create another crate named `contract-adapter` or `alloy-adapter`, this could house contracts like; 

1. Multicall
2. Permit2 

and so on... 

what do you think? @zilayo

DaniPopes:
I have a WIP implementation for the core ABI part here https://github.com/alloy-rs/core/pull/492 that may be of interest

developeruche:
> I have a WIP implementation for the core ABI part here [alloy-rs/core#492](https://github.com/alloy-rs/core/pull/492) that may be of interest

this is ineresting...

developeruche:
```rust 

fn basic_builder() {
        let token = Address::with_last_byte(1);

        let builder: MulticallBuilder<Aggregate, ()> = MulticallBuilder::new_aggregate();
        assert_eq!(builder.calls(), &[]);

        let call1 = ERC20::totalSupplyCall {};
        let builder: MulticallBuilder<Aggregate, (ERC20::totalSupplyCall,)> =
            builder.push(&call1, token);
        assert_eq!(
            builder.calls(),
            &[IMulticall3::Call { target: token, callData: call1.abi_encode() }]
        );

        let call2 = ERC20::balanceOfCall { owner: Address::with_last_byte(2) };
        let builder: MulticallBuilder<Aggregate, (ERC20::totalSupplyCall, ERC20::balanceOfCall)> =
            builder.push(&call2, token);
        assert_eq!(
            builder.calls(),
            &[
                IMulticall3::Call { target: token, callData: call1.abi_encode() },
                IMulticall3::Call { target: token, callData: call2.abi_encode() },
            ]
        );

        let encoded_data = builder.abi_encode();
        let expected = IMulticall3::aggregateCall {
            calls: vec![
                IMulticall3::Call {
                    target: token,
                    callData: ERC20::totalSupplyCall {}.abi_encode(),
                },
                IMulticall3::Call {
                    target: token,
                    callData: ERC20::balanceOfCall { owner: Address::with_last_byte(2) }
                        .abi_encode(),
                },
            ],
        };
[...]

PR #442, opened 2024-04-02
feat: Multicall3 Implementation

zilayo:
## Motivation

#328

Currently Alloy is missing multicall functionality. This PR aims to provide an initial implementation which can be iterated on and improved in the future.

It provides support for Multicall 1/2/3 and all of its associated read-only methods.

## Solution

The logic is heavily inspired by the logic in ethers-rs.

A Multicall instance can be instantiated via `Multicall::new(provider: Provider, Option<Address>)`.

Calls can then be added via `add_call` which simply appends to the the internal calls vector without returning anything, or via a builder pattern by chaining `with_call`.

The Multicall query is then sent by awaiting `.call()`.

When adding calls, the functions expect a `DynCallBuilder`, which it then extracts the calldata, target, and decoder from. The returned output is:
- If there is only 1 return value: A `DynSolValue` of the corresponding type.
- More than 1 return value: A `DynSolValue::Tuple` containing the corresponding `DynSolValue` for each return value.

Originally I tried to make this work with a `SolCallBuilder`, but couldn't wrap my head around how to do this nicely due to `SolCallBuilder` instances being different types for each function. I wanted to avoid boxing and/or requiring a bunch of type assertions from the end user.

Due to the usage of dynamic/json sol types, this means that the multicall feature does require the `alloy-sol-types/json` feature, and so this is enabled whenever the `multicall` feature flag in the contracts crate is enabled.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


prestwich:
Closing as stale, as we have not reviewed in a timely manner. Thanks for your contribution. We may revisit this work in the future

Issue #441, opened 2024-04-02
[Bug] Dedupe `alloy_consensus::BlockResponse<N>` and `alloy_rpc_types::Block`


prestwich:
### Component

consensus, eips, genesis, network, json-rpc

### Operating System

None

### Describe the bug

related to #440 

These types are duplicated, but block response is unused. 

cc @mattsse will all sane networks use the same pattern for block responses? flattened header + transactions + other fields?

yash-atreya:
Can be closed #444

Issue #440, opened 2024-04-02
[Bug] Dedupe `alloy_network::TransactionList` and `alloy_rpc_types::BlockTransactions`


prestwich:
### Component

consensus, eips, genesis, network, json-rpc

### Operating System

None

### Describe the bug

Related to #439 

These types are duplicated

yash-atreya:
TransactionList is in `alloy_network

prestwich:
my bad 😅

yash-atreya:
Can be closed #444

PR #444, opened 2024-04-02
dedup: TransactionList and BlockResponse

yash-atreya:
## Motivation

Ref #440 and #441

## Solution

Removed `TransactionList` from alloy_network

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/lib.rs
+++ b/crates/network/src/lib.rs
@@ -17,7 +17,8 @@
 
 use alloy_eips::eip2718::Eip2718Envelope;
 use alloy_json_rpc::RpcObject;
-use alloy_primitives::{Address, B256};
+use alloy_primitives::Address;
+use alloy_rpc_types::BlockTransactions;
 
 mod transaction;
 pub use transaction::{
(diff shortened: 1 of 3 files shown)


crates/network/src/lib.rs
```diff
@@ -33,24 +34,12 @@ pub use any::AnyNetwork;
...
-    /// Special case for uncle response
-    Uncled,
-}
-
 /// A block response
 #[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
 pub struct BlockResponse<N: Network> {
     #[serde(flatten)]
     header: N::HeaderResponse,
-    transactions: TransactionList<N::TransactionResponse>,
```
prestwich, line 53: we should probably do the following while we're here:
- delete this block response as unused and close #441 
- make `TransactionList` generic over `T` if it's not already

prestwich (changes requested):

crates/network/src/lib.rs
yash-atreya: `TransactionList` is already generic over T, I was removing it in favour of `BlockTransactions` (which does have a generic) from `rpc-types`. 

Also if we're removing `BlockResponse` we can also remove `TransactionList` as it is the only place it is used

yash-atreya: I'll leave `TransactionsList` as is but would like to change the `Hydrated` enum to `Full` to be consistent with `BlockTransactions`

prestwich: i feel like we're having a miscommunication here so i'm gonna step back

- the 2 types in alloy-network lib.rs should be deleted as they're duplicative
- the types in alloy-rpc-types should have generics added with defaults

yash-atreya: Sounds great! That's ideal.

prestwich (approved):

PR #445, opened 2024-04-03
chore: reexport receipt types

mattsse:
makes it possible to create a rpc receipt without importing additional crate


Evalir (approved):
doesn't hurt

Issue #433, opened 2024-04-01
[Bug] Numeric type audit in transaction and block structs and provider methods, remove gratuitous U256s


prestwich:
### Component

consensus, eips, genesis

### Operating System

None

### Describe the bug

Noticed discrepancy while implementing `From<TypedTransaction> for TransactionRequest` 

all types should have the following (with options as necessary)

applies to 
consensus
rpc
providers return values
etc

```
block_number: u64

transaction_index: u64

gas_price: u128

// make sure names are consistent here too. these are the same field
gas: u128
gas_limit: u128

max_fee_per_gas: u128

max_priority_fee_per_gas: u128

max_fee_per_blob_gas: u128

value: U256
```

prestwich:
closed by 454

PR #454, opened 2024-04-03
Numeric type audit: network, consensus, provider, rpc-types

yash-atreya:
## Motivation

Ref #433

## Solution

* `gas_limit` -> `u128`
* `gas` -> `u128`
* `gas_price` -> `u128`
* `max_fee_per_blob_gas` -> `u128`
* `transaction_type` -> `u8`
* `max_fee_per_gas` -> `u128`
* `max_priority_fee_per_gas` -> `u128`
* Consequently, changed fn args and return values accordingly.
* Satisfies clippy, tests will fail.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip2930.rs
+++ b/crates/consensus/src/transaction/eip2930.rs
@@ -29,8 +29,8 @@ pub struct TxEip2930 {
     /// this transaction. This is paid up-front, before any
     /// computation is done and may not be increased
     /// later; formally Tg.
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_hex_or_decimal"))]
-    pub gas_limit: u64,
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u128_hex_or_decimal"))]
+    pub gas_limit: u128,
     /// The 160-bit address of the message call’s recipient or, for a contract creation
     /// transaction, ∅, used here to denote the only member of B0 ; formally Tt.
     #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "TxKind::is_create"))]
@@ -237,12 +237,12 @@ impl Transaction for TxEip2930 {
         self.nonce
     }
 
-    fn gas_limit(&self) -> u64 {
+    fn gas_limit(&self) -> u128 {
         self.gas_limit
     }
 
-    fn gas_price(&self) -> Option<U256> {
-        Some(U256::from(self.gas_price))
+    fn gas_price(&self) -> Option<u128> {
+        Some(self.gas_price)
     }
 }
 
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -86,9 +86,7 @@ where
 
     /// Gets the gas_limit to be used in txs.
     async fn get_gas_estimate(&self, tx: &N::TransactionRequest) -> TransportResult<u128> {
-        self.inner.estimate_gas(tx, None).await.map(|gas| gas.to::<u128>()) // Temp Fix until
-                                                                            // estimate_gas returns
-                                                                            // u128
+        self.inner.estimate_gas(tx, None).await
     }
 
     /// Gets the max_fee_per_gas and max_priority_fee_per_gas to be used in EIP-1559 txs.
--- a/crates/provider/src/layers/gas.rs
+++ b/crates/provider/src/layers/gas.rs
@@ -3,7 +3,6 @@ use crate::{
 };
 use alloy_json_rpc::RpcError;
 use alloy_network::{Network, TransactionBuilder};
-use alloy_primitives::U256;
 use alloy_rpc_types::BlockNumberOrTag;
 use alloy_transport::{Transport, TransportError, TransportResult};
 use async_trait::async_trait;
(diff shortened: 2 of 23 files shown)


mattsse (approved):
I think all of these make sense, primitive types are preferable imo.
gas being u128 should make some operations easier but should not affect math for any call arguments etc

crates/provider/src/layers/gas.rs
```diff
@@ -81,13 +80,15 @@ where
...
         self.inner.get_gas_price().await
     }
 
     /// Gets the gas_limit to be used in txs.
-    async fn get_gas_estimate(&self, tx: &N::TransactionRequest) -> TransportResult<U256> {
-        self.inner.estimate_gas(tx, None).await
+    async fn get_gas_estimate(&self, tx: &N::TransactionRequest) -> TransportResult<u128> {
+        self.inner.estimate_gas(tx, None).await.map(|gas| gas.to::<u128>()) // Temp Fix until
+                                                                            // estimate_gas returns
+                                                                            // u128
```
Evalir, line 91: Should this be solved on the same PR?

crates/provider/src/provider.rs
```diff
@@ -806,15 +813,17 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
         &self,
-        block_count: U256,
+        block_count: u64,
         last_block: BlockNumberOrTag,
         reward_percentiles: &[f64],
     ) -> TransportResult<FeeHistory> {
         self.client().request("eth_feeHistory", (block_count, last_block, reward_percentiles)).await
     }
 
+    // TODO: Return TransportResult<u128> instead of U256
```
Evalir, line 826: same here

crates/serde/src/num.rs
```diff
@@ -264,6 +264,33 @@ pub mod u128_hex_or_decimal {
     }
 }
 
+/// serde functions for handling primitive optional `u128` as [U128](alloy_primitives::U128)
+pub mod u128_hex_or_decimal_opt {
+    // TODO: Add unit tests for this module
```
Evalir, line 269: Also here

Evalir (commented):
This mostly looks good to me, I'd just like to ensure we get the TODOs taken care of and add enough tests to be 100% sure we're not breaking serde

crates/rpc-types/src/eth/transaction/tx_type.rs
```diff
@@ -45,6 +45,7 @@ impl From<TxType> for u8 {
     }
 }
 
+// TODO(Numeric-Audit): Remove this conversion
```
gakonst, line 48: todo

crates/provider/src/utils.rs
```diff
@@ -46,10 +46,14 @@ pub fn eip1559_default_estimator(
     base_fee_per_gas: U256,
     rewards: &[Vec<U256>],
 ) -> Eip1559Estimation {
+    // TODO: Num type audit - return u128 here
```
gakonst, line 49: todo

gakonst (changes requested):

crates/consensus/src/transaction/eip2930.rs
```diff
@@ -29,8 +29,8 @@ pub struct TxEip2930 {
     /// this transaction. This is paid up-front, before any
     /// computation is done and may not be increased
     /// later; formally Tg.
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_hex_or_decimal"))]
-    pub gas_limit: u64,
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u128_hex_or_decimal"))]
```
prestwich, line 32: do any endpoints use decimal here? in other PRs we've avoided allowing decimal when nobody actually uses it

crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -42,29 +42,48 @@ pub struct Transaction {
...
     pub to: Option<Address>,
     /// Transferred value
     pub value: U256,
     /// Gas Price
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub gas_price: Option<U256>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "alloy_serde::num::u128_hex_or_decimal_opt"
```
prestwich, line 60: same question. pretty sure decimal would be illegal rpc response here? I don't mind tolerating it, but @mattsse asked for removal of decimal in other PRs

prestwich (approved):
just style nit for @mattsse

crates/rpc-types/src/eth/transaction/mod.rs
Evalir: + 1 to removing decimal support

mattsse: makes sense, let's enforce strict quantity rules everywhere for core eth types

prestwich: side note, the `U***` type deser DO accept decimal. so this is technically a breaking behavior change

prestwich:
followup work: #465

PR #455, opened 2024-04-03
chore: add OtsReceipt type

mattsse:
adds a new receipt type for otterscan, which is supposed to emit logs and bloom as None.

there are still a few differences, e.g. the logsBloom of the header is also none, which can get pretty ugly because we'd need to duplicate all types effectively or manually impl serde.

can track this separately in case someone wants to make this 100% right.

### DIFF
--- a/crates/rpc-types-trace/src/otterscan.rs
+++ b/crates/rpc-types-trace/src/otterscan.rs
@@ -1,9 +1,12 @@
+//! Otterscan specific types for RPC responses.
+//!
+//! <https://www.quicknode.com/docs/ethereum/ots_getBlockTransactions>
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/otterscan.rs
```diff
@@ -59,6 +62,18 @@ pub struct OtsBlock {
     pub transaction_count: usize,
 }
 
+impl From<Block> for OtsBlock {
+    fn from(block: Block) -> Self {
+        let transaction_count = match &block.transactions {
```
DaniPopes, line 67: there should be a len method

DaniPopes (changes requested):

DaniPopes (approved):

Issue #451, opened 2024-04-03
[Bug] Mark TxEnvelope and ReceiptEnvelope non-exhaustive


prestwich:
### Component

consensus, eips, genesis

### Operating System

None

### Describe the bug

Ethereum may hard fork to add new tx and receipt types

prestwich:
add the `#[non_exhaustive]` attribute

PR #456, opened 2024-04-03
consensus: mark envelopes non-exhaustive

yash-atreya:
## Motivation

Ref #451 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/receipt.rs
+++ b/crates/rpc-types/src/eth/transaction/receipt.rs
@@ -70,6 +70,7 @@ impl TransactionReceipt {
             | ReceiptEnvelope::Eip2930(receipt)
             | ReceiptEnvelope::Eip4844(receipt)
             | ReceiptEnvelope::Legacy(receipt) => receipt.receipt.status,
+            _ => false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


Evalir (approved):
agreed that these are needed—shouldn't be much of an issue downstream other than having to add the wildcard when pattern matching

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -68,6 +68,7 @@ impl TransactionReceipt {
             | ReceiptEnvelope::Eip2930(receipt)
             | ReceiptEnvelope::Eip4844(receipt)
             | ReceiptEnvelope::Legacy(receipt) => receipt.receipt.status,
+            _ => false,
```
prestwich, line 71: Followup work: 
- `trait Receipt` in alloy-network
- Receipt bound on `type Recept` in `trait Network`
- `fn status` in `trait Receipt`
- `impl Receipt for ReceiptEnvelope` and related types
- change this function to a delegate to `Receipt`

prestwich (approved):

prestwich:
rebased to resolve conflict with #461

PR #458, opened 2024-04-04
Derive arbitrary for rpc `Header` and `Transaction`

tcoratger:
## Solution

- Derive `arbitrary` for `Header` and `Transaction` rpc types.

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -15,10 +15,9 @@ use core::mem;
 /// Accesses outside the list are possible, but become more expensive.
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Default, RlpDecodable, RlpEncodable)]
 #[cfg_attr(
-    all(any(test, feature = "arbitrary"), feature = "std"),
-    derive(proptest_derive::Arbitrary)
+    any(test, feature = "arbitrary"),
+    derive(proptest_derive::Arbitrary, arbitrary::Arbitrary)
 )]
-#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct AccessListItem {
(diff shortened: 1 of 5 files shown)


crates/rpc-types/src/eth/transaction/mod.rs
```diff
@@ -32,6 +32,7 @@ pub use signature::{Parity, Signature};
 
 /// Transaction object used in RPC
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[cfg_attr(all(any(test, feature = "arbitrary"), feature = "std"), derive(arbitrary::Arbitrary))]
```
DaniPopes, line 35: arbitrary should imply std as the dep does not support no_std https://docs.rs/crate/arbitrary/latest/features

tcoratger:
@DaniPopes Is the wasm error linked to the PR or not at all, seems strange no?

crates/rpc-types/Cargo.toml
```diff
@@ -19,7 +19,7 @@ alloy-serde.workspace = true
 alloy-genesis.workspace=true
 
 alloy-consensus = { workspace = true, features = ["serde"] }
-alloy-eips = {workspace = true, features = ["std", "serde"]}
+alloy-eips = {workspace = true, features = ["std", "serde", "arbitrary"]}
```
DaniPopes, line 22: arbitrary is optional, should go in the features table

crates/eips/src/eip2930.rs
```diff
@@ -45,8 +46,9 @@ impl AccessListItem {
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Default, RlpDecodableWrapper, RlpEncodableWrapper)]
 #[cfg_attr(
     all(any(test, feature = "arbitrary"), feature = "std"),
```
DaniPopes, line 48: same here re: std feature

DaniPopes (changes requested):

crates/rpc-types/Cargo.toml
tcoratger: Fixed

crates/eips/src/eip2930.rs
tcoratger: Fixed

DaniPopes (approved):

PR #461, opened 2024-04-04
feat: `std` feature flag for `alloy-consensus`

clabby:
## Overview

Ports `alloy-consensus` to be a `no_std` compatible crate.


crates/consensus/src/transaction/eip4844/builder.rs
```diff
@@ -3,6 +3,14 @@ use alloy_eips::eip4844::Blob;
 #[cfg(feature = "kzg")]
 use c_kzg::{Blob, KzgCommitment, KzgProof};
 
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use core::cmp;
```
prestwich, line 9: no need to feature gate core ever as it is part of both the `std` prelude and the `core` prelude.

prestwich (changes requested):
meeds update to the no_std list in the readme

crates/consensus/src/transaction/eip4844/builder.rs
DaniPopes: Yes please just use `core` and `alloc` instead of feature gating imports

DaniPopes:
There's also `scripts/check_no_std.sh`

crates/consensus/src/transaction/eip4844/builder.rs
prestwich: alloc needs to be feature gated due to new redundant import lints

crates/consensus/src/receipt/mod.rs
```diff
@@ -37,6 +37,7 @@ mod tests {
     use alloy_eips::eip2718::Encodable2718;
     use alloy_primitives::{address, b256, bytes, hex, Bytes, LogData};
     use alloy_rlp::{Decodable, Encodable};
+    use std::vec;
```
prestwich, line 40: this seems out of place. does it not cause redundant import lints?

clabby: Ah good spot - not necessary, removed.

prestwich (approved):

prestwich:
rebased onto main after #454 to re-run CI with new changes. shouldn't be any conflicts, but 454 was big. will merge after CI passes

PR #468, opened 2024-04-05
feat(consensus): Optimism Deposit Tx

refcell:
**Description**

Ports the Optimism `TxDeposit` transaction type [from `reth-primitives`](https://github.com/paradigmxyz/reth/blob/3236baef5e2342e012885db84649488da58807a7/crates/primitives/src/transaction/optimism.rs) for use in down-stream users of `alloy-consensus`.

The type is hidden behind a non-default `optimism` feature flag.


prestwich:
NACK. this work belongs in a separate workspace, not as feature flags. Would recommend a crate named `alloy-network-optimism`, The recommended implementation for this is as follows:

```
pub enum OptimismTxEnvelope { ... }
pub enum OptimismReceiptEnvelope { ... }

impl Network for Optimism {
    type TxEnvelope = OptimismTxEnvelope;
    type ReceiptEnvelope = OptimismReceiptEnvelope;

    ...
}
```

gakonst:
we'll pull this to alloy-rs/op-alloy and separate maintainership, the network trait abstraction was built specifically for this as james wrote

Issue #438, opened 2024-04-02
[Feature] Typed transaction conversion traits 


prestwich:
### Component

consensus, eips, genesis

### Describe the feature you would like

blocked by #433 

add standard conversions to transaction types

- [ ] modify `Network` 
    - [ ] add a new bound: `type UnsignedTx: From<Self::TxEnvelope>`
    - [ ] add 2 new bounds: `type TransactionRequest: From<Self::TxEnvelope> + From<Self::UnsignedTx>`
- [ ] satisfy new bounds
    - [ ] `impl From<TxEnvelope> for TypedTransaction`
    - [ ] `impl From<TxEnvelope> for TransactionRequest`
    - [ ] `impl From<TypedTransaction for TransactionRequest`

mahmudsudo:
can i take on this ?

siosw:
I think this has been completed in #472 should be closed

mattsse:
yep

PR #472, opened 2024-04-05
TypeTransaction conversion trait impls

yash-atreya:
## Motivation

Ref #438 

## Solution

Add `From` bounds to `UnsignedTx` and `TransactionRequest` types in `Network`.

Satisfy bounds by implementing the following:

- `From<TypedTransaction> for TransactionRequest`
- `From<TxEnvelope> for TransactionRequest`
- `From<TypedTransaction> for WithOtherFields<TransactionRequest>`
- `From<TxEnvelope> for WithOtherFields<TransactionRequest>`

To accomplish the above, implemented

- `From<TxLegacy> for TransactionRequest`
- `From<TxEip2930> for TransactionRequest`
- `From<TxEip1559> for TransactionRequest`
- `From<TxEip4844Variant> for TransactionRequest`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -33,8 +33,8 @@ impl<T, Sig> Signed<T, Sig> {
         (self.tx, self.signature, self.hash)
     }
 
-    /// Returns the transaction.
-    pub fn owned_tx(self) -> T {
+    /// Returns the transaction without signature.
+    pub fn strip_signature(self) -> T {
         self.tx
     }
 }
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -32,6 +32,11 @@ impl<T, Sig> Signed<T, Sig> {
     pub fn into_parts(self) -> (T, Sig, B256) {
         (self.tx, self.signature, self.hash)
     }
+
+    /// Returns the transaction.
+    pub fn owned_tx(self) -> T {
+        self.tx
+    }
 }
 
 impl<T: SignableTransaction<Sig>, Sig> Signed<T, Sig> {
--- a/crates/rpc-types/src/eth/transaction/request.rs
+++ b/crates/rpc-types/src/eth/transaction/request.rs
@@ -364,10 +364,70 @@ impl From<TypedTransaction> for TransactionRequest {
 impl From<TxEnvelope> for TransactionRequest {
     fn from(envelope: TxEnvelope) -> TransactionRequest {
         match envelope {
-            TxEnvelope::Legacy(tx) => tx.strip_signature().into(),
-            TxEnvelope::Eip2930(tx) => tx.strip_signature().into(),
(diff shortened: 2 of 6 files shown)


crates/consensus/src/signed.rs
```diff
@@ -32,6 +32,11 @@ impl<T, Sig> Signed<T, Sig> {
     pub fn into_parts(self) -> (T, Sig, B256) {
         (self.tx, self.signature, self.hash)
     }
+
+    /// Returns the transaction.
+    pub fn owned_tx(self) -> T {
```
prestwich, line 37: i would prefer this to be something like `strip_signature` to make it clear that we're dropping the sig

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -238,6 +242,137 @@ impl From<Transaction> for TransactionRequest {
...
+            TypedTransaction::Eip2930(tx) => tx.into(),
+            TypedTransaction::Eip1559(tx) => tx.into(),
+            TypedTransaction::Eip4844(tx) => tx.into(),
+        }
+    }
+}
+
+impl From<TxEnvelope> for TransactionRequest {
+    fn from(envelope: TxEnvelope) -> TransactionRequest {
+        match envelope {
```
prestwich, line 366: we could attempt to extract `from` from the signature here (ignoring any error on the ecdsa recovery)

prestwich (changes requested):
minor changes

crates/rpc-types/src/eth/transaction/request.rs
prestwich: it does make the UX cumbersome, but I also don't like dropping information that we definitely have and the user is likely to want if that makes sense

I wonder if it would make sense to extend the signed struct with `signed_by: OnceLock<Address>` which can be computed if K256 enabled

prestwich: Regardless, this decision shouldn't block the PR

yash-atreya:
Ready to merge @prestwich ?

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -238,6 +242,197 @@ impl From<Transaction> for TransactionRequest {
...
+            TxEnvelope::Legacy(tx) => {
+                #[cfg(feature = "k256")]
+                {
+                    if let Ok(signer) = tx.recover_signer() {
```
prestwich, line 370:
```
let from = tx.recover_signer().ok();
let tx = tx.strip_signature().into()
if let Some(from) = from {
   tx.from(from);
}
```

bit cleaner

prestwich (changes requested):
little cleanup. 

please open a followup issue for memoizing signer address on Signed<T>

Issue #465, opened 2024-04-05
[Refactor] Remove `_or_decimal` serde encoding from integers in rpc-types


prestwich:
the decimal string and json number representations are off-spec for the vast majority of RPC types. Essentially all integers in eth json rpc are represented as Quantity, which is a minimal-nibble 0x-prefixed hex string. Ethereum nodes will always reject decimal string and json number representations for those types. there are a few random types that do permit json number for floats but they're uncommon

ruint as a separate project aims to support deser of all of decimal string, hex string, binary, and json numeric. and uses Quantity encoding for serialization only for human readable serializations

this means that alloy, where it uses U*** types will tolerantly deserialize invalid eth RPC objects. by switching to u64/128_hex_or_decimal we're actually narrowing the allowed inputs a lot while moving closer to eth RPC spec conformance. and u64/128_hex is the closest we'll come to actual spec conformance (Quantity is technical unbounded size integer but we ain't supporting that junk)

so while i don't particularly value spec conformance, we have no use case for tolerant deser, and we need a consistent policy for these so we don't keep having type divergence based on who is writing the code
which leads me to like u64/128_hex as the "obvious" choice. if we enforce the specified encoding nobody is making tolerance decisions on the fly, and therefore we end up with more consistent code 

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/454#discussion_r1552181478_

prestwich:
closing in favor of #479

PR #474, opened 2024-04-06
num-audit: extension

yash-atreya:
## Motivation

Extension of #454 

## Solution

* gas related fields to `u128`
* `nonce` and `number` to `u64` in `Header` from `B64` and `U256` respectively

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/eips/src/eip1559/basefee.rs
```diff
@@ -7,17 +7,17 @@ use crate::eip1559::constants::{
...
+    pub elasticity_multiplier: u128,
 }
 
 impl BaseFeeParams {
     /// Get the base fee parameters for Ethereum mainnet
     pub const fn ethereum() -> BaseFeeParams {
         BaseFeeParams {
-            max_change_denominator: DEFAULT_BASE_FEE_MAX_CHANGE_DENOMINATOR,
-            elasticity_multiplier: DEFAULT_ELASTICITY_MULTIPLIER,
+            max_change_denominator: DEFAULT_BASE_FEE_MAX_CHANGE_DENOMINATOR as u128,
```
prestwich, line 19: should we change the constant types?

crates/rpc-types/src/eth/block.rs
```diff
@@ -75,11 +75,11 @@ pub struct Header {
...
-    pub number: Option<U256>,
+    pub number: Option<u64>,
     /// Gas Limit
-    pub gas_limit: U256,
+    pub gas_limit: u128,
     /// Gas Used
-    pub gas_used: U256,
+    pub gas_used: u128,
     /// Timestamp
     pub timestamp: U256,
```
prestwich, line 84: maybe followup work, but should `timestamp` be a `u64` cc @mattsse?

crates/provider/src/heart.rs
```diff
@@ -445,12 +445,12 @@ impl<S> Heartbeat<S> {
             // Otherwise add it to the waiting list.
             debug!(tx=%watcher.config.tx_hash, %block_height, confirmations, "adding to waiting list");
             self.waiting_confs
-                .entry(*block_height + U256::from(confirmations) - U256::from(1))
+                .entry(U256::from(*block_height + confirmations - 1))
```
prestwich, line 448: we could change the type of `waiting_confs` to `BTreeMap<u64, Vec<TxWatcher>>,`

klkvr:
`rpc_types::TransactionReceipt` also has several `u64` gas price fields

crates/rpc-types/src/eth/block.rs
yash-atreya: `timestamp` is `u64` in certain places. I'm in favour

klkvr:
I am experiencing https://github.com/serde-rs/json/issues/625 on latest alloy while deserializing `WithOtherFields<TransactionRequest>`, so I think we should ensure that all `u128` fields have a `with = u128_hex_or_decimal`

crates/eips/src/eip1559/basefee.rs
yash-atreya: Ideally yes, but I think if users would like to use constants in their applications; It's smoother for them to use `as u128` and expand to a large type than use `try_from().into()` to shrink to a smaller.

yash-atreya: Moreover, imo, its better to expand the type in places where its absolutely needed and keep it as is where isn't required.

crates/rpc-types/src/eth/block.rs
mattsse: sg!

crates/consensus/src/header.rs
```diff
@@ -397,8 +397,8 @@ impl Decodable for Header {
             logs_bloom: Decodable::decode(buf)?,
             difficulty: Decodable::decode(buf)?,
             number: U256::decode(buf)?.to::<u64>(),
-            gas_limit: U256::decode(buf)?.to::<u64>(),
-            gas_used: U256::decode(buf)?.to::<u64>(),
+            gas_limit: U256::decode(buf)?.to::<u128>(),
+            gas_used: U256::decode(buf)?.to::<u128>(),
```
mattsse, line 401: I think we should either use u128::decode or use try_into because to:: is not safe

crates/eips/src/eip1559/basefee.rs
mattsse, line 1: after thinking about these a bit, I think these make sense, if we change fees/gas used to u128 then all of these also make sense

crates/rpc-types/src/eth/block.rs
```diff
@@ -75,11 +75,11 @@ pub struct Header {
...
     pub difficulty: U256,
     /// Block number
-    pub number: Option<U256>,
+    pub number: Option<u64>,
     /// Gas Limit
-    pub gas_limit: U256,
+    pub gas_limit: u128,
     /// Gas Used
-    pub gas_used: U256,
+    pub gas_used: u128,
```
mattsse, line 82: these need serde attributes

crates/rpc-types/src/eth/block.rs
```diff
@@ -101,19 +101,19 @@ pub struct Header {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub withdrawals_root: Option<B256>,
     /// Blob gas used
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub blob_gas_used: Option<U64>,
+    pub blob_gas_used: Option<u128>,
     /// Excess blob gas
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub excess_blob_gas: Option<U64>,
+    pub excess_blob_gas: Option<u128>,
```
mattsse, line 116: same

mattsse (changes requested):
very supportive
unifying all of them is necessary, just need to ensure we keep hex support

crates/serde/src/num.rs
```diff
@@ -243,7 +243,6 @@ where
 {
     NumberOrHexU256::deserialize(deserializer)?.try_into_u256()
 }
-
```
prestwich, line 246: nit:
was this change intentional?

prestwich (approved):

crates/serde/src/num.rs
yash-atreya: That was a mistake

crates/rpc-types/src/eth/block.rs
```diff
@@ -75,10 +75,13 @@ pub struct Header {
     /// Difficulty
     pub difficulty: U256,
     /// Block number
+    #[serde(default, with = "alloy_serde::num::u64_hex_opt")]
     pub number: Option<u64>,
     /// Gas Limit
+    #[serde(default, with = "alloy_serde::num::u128_hex_or_decimal")]
     pub gas_limit: u128,
     /// Gas Used
+    #[serde(default, with = "alloy_serde::num::u128_hex_or_decimal")]
```
mattsse: this should just be hex

crates/rpc-types/src/eth/block.rs
```diff
@@ -101,18 +104,31 @@ pub struct Header {
...
+        with = "alloy_serde::num::u128_hex_or_decimal_opt"
+    )]
     pub blob_gas_used: Option<u128>,
     /// Excess blob gas
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "alloy_serde::num::u128_hex_or_decimal_opt"
+    )]
```
mattsse: also just hex

mattsse (changes requested):
last pedantic serde nits

crates/eips/src/eip1559/basefee.rs
```diff
@@ -7,17 +7,17 @@ use crate::eip1559::constants::{
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct BaseFeeParams {
     /// The base_fee_max_change_denominator from EIP-1559
-    pub max_change_denominator: u64,
+    pub max_change_denominator: u128,
     /// The elasticity multiplier from EIP-1559
-    pub elasticity_multiplier: u64,
+    pub elasticity_multiplier: u128,
```
klkvr, line 12: This needs serde attr

crates/genesis/src/lib.rs
```diff
@@ -64,14 +66,14 @@ pub struct Genesis {
...
     /// The genesis header excess blob gas
-    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt")]
-    pub excess_blob_gas: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none", with = "u128_hex_or_decimal_opt")]
+    pub excess_blob_gas: Option<u128>,
     /// The genesis header blob gas used
-    #[serde(skip_serializing_if = "Option::is_none", with = "u64_hex_or_decimal_opt")]
-    pub blob_gas_used: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none", with = "u128_hex_or_decimal_opt")]
+    pub blob_gas_used: Option<u128>,
```
klkvr, line 76: Those need `#[serde(default)]` ref https://github.com/alloy-rs/alloy/pull/370

mattsse: good catch,
always forget that missing key is not handled by option

crates/eips/src/eip1559/basefee.rs
mattsse: it's not part of rpc, but we should do hex here yes

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -28,21 +28,29 @@ pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
     #[serde(with = "alloy_serde::u64_hex_opt")]
     pub block_number: Option<u64>,
     /// Gas used by this transaction alone.
-    #[serde(with = "alloy_serde::u64_hex_opt")]
-    pub gas_used: Option<u64>,
+    #[serde(with = "alloy_serde::u128_hex_or_decimal_opt")]
+    pub gas_used: Option<u128>,
```
klkvr, line 32: needs `default` and `skip_serializing_if`

aren't those mandatory though?

klkvr: image

crates/rpc-types/src/eth/block.rs
```diff
@@ -101,19 +105,32 @@ pub struct Header {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub mix_hash: Option<B256>,
     /// Nonce
-    pub nonce: Option<B64>,
+    #[serde(default, with = "alloy_serde::num::u64_hex_opt")]
+    pub nonce: Option<u64>,
```
klkvr, line 109: `skip_serializing_if`

crates/rpc-types/src/eth/block.rs
yash-atreya: Going to address this in a separate PR for https://github.com/alloy-rs/alloy/issues/465

crates/rpc-types/src/eth/block.rs
yash-atreya: Going to address this in a separate PR for https://github.com/alloy-rs/alloy/issues/465

crates/rpc-types/src/eth/block.rs
```diff
@@ -75,13 +75,17 @@ pub struct Header {
     /// Difficulty
     pub difficulty: U256,
     /// Block number
-    pub number: Option<U256>,
+    #[serde(default, with = "alloy_serde::num::u64_hex_opt")]
+    pub number: Option<u64>,
```
klkvr, line 79: `skip_serializing_if` and feels like this is mandatory too?

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -14,18 +14,23 @@ pub struct TransactionRequest {
...
+    #[serde(default, with = "alloy_serde::num::u128_hex_or_decimal_opt")]
     pub gas_price: Option<u128>,
     /// The max base fee per gas the sender is willing to pay.
-    #[serde(default)]
+    #[serde(default, with = "alloy_serde::num::u128_hex_or_decimal_opt")]
     pub max_fee_per_gas: Option<u128>,
     /// The max priority fee per gas the sender is willing to pay, also called the miner tip.
-    #[serde(default)]
+    #[serde(default, with = "alloy_serde::num::u128_hex_or_decimal_opt")]
     pub max_priority_fee_per_gas: Option<u128>,
```
klkvr, line 24: `skip_serializing_if`

prestwich: isnt this handled by the serde_with?

crates/rpc-types/src/eth/block.rs
mattsse: number is acutally required, yes

crates/rpc-types/src/eth/transaction/receipt.rs
mattsse: gas used is required:

```
// Receipt represents the results of a transaction.
type Receipt struct {
	// Consensus fields: These fields are defined by the Yellow Paper
	Type              uint8  `json:"type,omitempty"`
	PostState         []byte `json:"root"`
	Status            uint64 `json:"status"`
	CumulativeGasUsed uint64 `json:"cumulativeGasUsed" gencodec:"required"`
	Bloom             Bloom  `json:"logsBloom"         gencodec:"required"`
	Logs              []*Log `json:"logs"              gencodec:"required"`

	// Implementation fields: These fields are added by geth when processing a transaction.
	TxHash            common.Hash    `json:"transactionHash" gencodec:"required"`
	ContractAddress   common.Address `json:"contractAddress"`
	GasUsed           uint64         `json:"gasUsed" gencodec:"required"`
	EffectiveGasPrice *big.Int       `json:"effectiveGasPrice"` // required, but tag omitted for backwards compatibility
	BlobGasUsed       uint64         `json:"blobGasUsed,omitempty"`
	BlobGasPrice      *big.Int       `json:"blobGasPrice,omitempty"`

	// Inclusion information: These fields provide information about the inclusion of the
	// transaction corresponding to this receipt.
	BlockHash        common.Hash `json:"blockHash,omitempty"`
	BlockNumber      *big.Int    `json:"blockNumber,omitempty"`
	TransactionIndex uint        `json:"transactionIndex"`
}
```

crates/rpc-types/src/eth/transaction/request.rs
klkvr: it will be serialized as null by default

crates/rpc-types/src/eth/block.rs
yash-atreya: I think even nonce is required

mattsse: mix_hash and nonce are options because they are not supported by newer chains that otherwise are compatible with this object.
for convenience I'd like to keep them optional

crates/rpc-types/src/eth/transaction/receipt.rs
mattsse: @yash-atreya

crates/rpc-types/src/eth/transaction/receipt.rs
```diff
@@ -28,21 +28,29 @@ pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
...
     /// The price paid post-execution by the transaction (i.e. base fee + priority fee). Both
     /// fields in 1559-style transactions are maximums (max fee + max priority fee), the amount
     /// that's actually paid by users can only be determined post-execution
-    #[serde(with = "alloy_serde::u64_hex")]
-    pub effective_gas_price: u64,
+    #[serde(with = "alloy_serde::u128_hex_or_decimal")]
```
mattsse, line 36: only hex for all rpc fields

mattsse (changes requested):

crates/rpc-types/src/eth/block.rs
mattsse: there are no clear rules for the pending block,
until we figure it out, let's keep them optional

crates/rpc-types/src/eth/transaction/receipt.rs
yash-atreya: Yes, I believe there should be a separate PR for #465 as it's likely to break several tests, and I also wanted to unblock https://github.com/foundry-rs/foundry/pull/7106 quickly. I'll implement it immediately after this.

prestwich:
#465 actually somewhat superseded by the newer issue wrt serde helpers redundant code

yash-atreya:
> #465 actually somewhat superseded by the newer issue wrt serde helpers redundant code

Yes, the PR that addresses #479 will close #465

prestwich:
> > #465 actually somewhat superseded by the newer issue wrt serde helpers redundant code
> 
> 
> 
> Yes, the PR that addresses #479 will close #465

not exactly because moving to `via_ruint` canonizes the tolerant behavior, where #465 would require significant changes to reject the invalid-but-tolerated objects

Given how much surface area its add, my current preference is to leave everything tolerant

mattsse:
> Given how much surface area its add, my current preference is to leave everything tolerant

hmm, for response types, being tolerant is actually beneficial because the user doesn't really care how the server formats it

mattsse (approved):
per above, 
also, supporting decimals on deserialize is okay and perhaps even preferable, in case servers don't follow QUANTITY

PR #476, opened 2024-04-06
feat: helpers for AnyNetwork

klkvr:
## Motivation

Small changes for https://github.com/foundry-rs/foundry/pull/7106

Adds `AnyNetworkSigner` and alias for `AnyNetwork::ReceiptResponse` as it's pretty long

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/network/src/any/signer.rs
```diff
@@ -0,0 +1,66 @@
+use crate::{AnyNetwork, NetworkSigner, TxSigner};
+use alloy_consensus::{SignableTransaction, TxEnvelope, TypedTransaction};
+use alloy_signer::Signature;
+use async_trait::async_trait;
+use std::sync::Arc;
+
+/// A signer capable of signing any transaction for the [AnyNetwork] network.
+#[derive(Clone)]
+pub struct AnyNetworkSigner(Arc<dyn TxSigner<Signature> + Send + Sync>);
```
Evalir, line 9: makes sense to make this `Arc` so we can share the same signer

Evalir (approved):

DaniPopes (approved):

crates/network/src/any/mod.rs
```diff
@@ -5,6 +5,11 @@ use alloy_rpc_types::{
 };
 
 mod builder;
+mod signer;
+pub use signer::AnyNetworkSigner;
+
+/// Alias for a catch-all receipt type.
+pub type AnyReceipt = WithOtherFields<TransactionReceipt<AnyReceiptEnvelope<Log>>>;
```
mattsse, line 12: could move this to rpc types and rename `AnyTransactionReceipt`?

mattsse (changes requested):
not sure if it makes sense to duplicate the signer type,

Perhaps for forward compatibility we can do that, in case we want to support more tx types that are don't exist on ehtereum (currently there aren't any)

so I think for now we can simply use the ethereumsigner for AnyNetwork as well but add `impl NetworkSigner<AnyNetwork> for EthereumSigner {` because both networks have the same `UnsignedTx`

I believe we could even do

```rust
impl<N> NetworkSigner<N> for EthereumSigner
    where N: Network<UnsignedTx= TypedTransaction, TxEnvelope = TxEnvelope >
        {
```

wdyt @prestwich

prestwich:
supportive of type-reuse and the blanket impl

The reason that `NetworkSigner<N>` is a trait and not an assoc type is to allow many different objects to be network signers. We could even do a blanket like `impl<T, Sig> NetworkSigner<Ethereum> for T where T: Signer<Sig>` if it doesn't cause unconstrained type param issues`

prestwich:
basically the idea of `NetworkSigner<N>` is "any object that can sign any subset of transactions allowed by the network"

mattsse (approved):

PR #478, opened 2024-04-07
Implement `is_zero` method for `U64HexOrNumber`

tcoratger:
## Solution

Implement `is_zero` method for `U64HexOrNumber`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/serde/src/num.rs
+++ b/crates/serde/src/num.rs
@@ -28,6 +28,11 @@ impl U64HexOrNumber {
     pub fn to(self) -> u64 {
         self.0.to()
     }
+
+    /// Checks if the wrapped value is zero.
+    pub fn is_zero(&self) -> bool {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think this could be useful when used in manual serde impls

prestwich:
i believe this is redundant code, as ruint uints support deserializing from number without any typewrapping or decoration 

serde helpers were not examined when porting from ethers so there have been a few redundant things 😅

prestwich:
ref https://github.com/recmo/uint/blob/main/src/support/serde.rs#L127

prestwich:
on consideration, we might want to rename the helpers something like `u64_via_ruint` to clarify behavior

PR #483, opened 2024-04-08
enable default-tls for alloy-provider/reqwest feature

klkvr:
## Motivation

Currently it's not possible to connect to https endpoint via `on_http` with default feature set 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


prestwich (approved):

crates/provider/Cargo.toml
```diff
@@ -55,6 +55,7 @@ reqwest = [
     "dep:url",
     "dep:alloy-transport-http",
     "alloy-rpc-client/reqwest",
+    "alloy-transport-http/reqwest-default-tls",
```
mattsse, line 58: I think this will enable native tls, openssl by default now?

I want to avoid this because then this will now always enable default-tls for reqwest

mattsse: in order to leave this up to the user we'd need to duplicate the features, I assume:

https://github.com/alloy-rs/alloy/blob/6cc0af73786eaa4c812a6704e091dbd9a33c43de/crates/transport-http/Cargo.toml#L40-L42

klkvr: should this be rustls instead?

klkvr: so we should expose those in alloy-provider but enable default-tls by default?

mattsse: > so we should expose those in alloy-provider

yes

>  but enable default-tls by default?

since reqwest enables this by default as well, I think that'd be reasonable.
but tls must be a separate feature, so default is = ["reqwest", "reqwest-default-ts"]

Evalir (approved):
pending @mattsse

mattsse (approved):

Issue #481, opened 2024-04-08
[Feature] Add `signer` field in `Signed` tx - Use `OnceLock` 


yash-atreya:
### Component

consensus, eips, genesis

## Motivation

Ref https://github.com/alloy-rs/alloy/pull/472#discussion_r1556009874

## Solution 

Add a `signed_by: OnceLock<Address>` field in `Signed` tx

PR #484, opened 2024-04-08
feat(consensus): `signer` field in `Signed` tx.

yash-atreya:
## Motivation

Ref #481 

## Solution

Added `signer: OnceLock<Address>` in `Signed`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -1,8 +1,10 @@
+use std::sync::OnceLock;
+
 use crate::transaction::SignableTransaction;
-use alloy_primitives::{Signature, B256};
+use alloy_primitives::{Address, Signature, B256};
 
 /// A transaction with a signature and hash seal.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Signed<T, Sig = Signature> {
     #[cfg_attr(feature = "serde", serde(flatten))]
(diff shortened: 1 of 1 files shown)


crates/consensus/src/signed.rs
```diff
@@ -57,7 +75,16 @@ impl<T: SignableTransaction<Signature>> Signed<T, Signature> {
     pub fn recover_signer(
         &self,
     ) -> Result<alloy_primitives::Address, alloy_primitives::SignatureError> {
+        #[cfg(not(feature = "no_std"))]
+        if let Some(signer) = self.signer() {
+            return Ok(*signer);
+        }
```
prestwich, line 81: proposed api

- delete `recover_signer` as standard API
- add `new_with_unchecked_signer` for instantiation
- modify `signer` as follows:

```

/// get the signer if known or recoverable
fn signer(&self) -> Option<Address> {
    #[cfg(not(feature = "k256"))]
    self.signer.get().copied()

    #[cfg(feature = "k256")]
    self.signer.get_or_init(|| { /* recovery logic here */ });
}

```

prestwich: also ok if implemented as 

```
#[cfg(not(feature = "k256"))]
fn signer(&self) -> Option<Address> { ... }

#[cfg(feature = "k256")]
fn signer(&self) -> Option<Address> { ... }
```

whatever feels cleaner

prestwich:
I did a little bit of work here, and because we're attempting to be generic over Sig, this is harder than it seems without a `Recoverable` trait for the signature type. which i'm not sure we want to introduce yet. Going to close this as needing more design work

another problem specialization would solve 🤔

PR #485, opened 2024-04-09
fix: change `Header::nonce` to `B64`

klkvr:
## Motivation

It was changed to u64 by accident probably, this nonce is not the same entity as transaction nonce (it is a hash, not a number), and it is reported by RPC as a 8-byte hex string

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Evalir (commented):
hmm this is also u64 in reth https://github.com/paradigmxyz/reth/blob/main/crates/primitives/src/header.rs#L76-L78 , not sure we should change this to B64?

Also, I think the enabling default tls commit slipped in

klkvr:
> hmm this is also u64 in reth https://github.com/paradigmxyz/reth/blob/main/crates/primitives/src/header.rs#L76-L78 , not sure we should change this to B64?

yeah not sure if this should actually be u64 in consensus types

it was B64 in rpc-types and reth is currently pinned to alloy commit which has it as B64 in rpc_types::Header: https://github.com/alloy-rs/alloy/blob/6aab2164b92c3174cdd81fcb403177361338e594/crates/rpc-types/src/eth/block.rs#L104

klkvr:
The spec defines it as bytes8 so it should definitely by serialized/deserialized as B64 https://github.com/ethereum/execution-apis/blob/90a46e9137c89d58e818e62fa33a0347bba50085/src/schemas/block.yaml#L72

Evalir (approved):
gotcha, then lgtm, but pending @mattsse to double check

DaniPopes (approved):

mattsse (approved):

PR #489, opened 2024-04-09
fix: allow empty `to` field in `can_build`

klkvr:
## Motivation

`to` is required to be Some only for EIP-4844 txses

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm,

only 4844 mandates a `to` field

PR #498, opened 2024-04-10
feat(rpc): trace requests and responses

DaniPopes:
There is currently no way of knowing what raw data we are sending and receiving.
Add `trace!` calls around the `RpcClient` future for this, as well as a simpler `debug!` for just the method and ID, as the raw data could be very verbose.

Running `test_send_tx` with `alloy=debug`, I think this verbosity is fair for debug:
image

### DIFF
--- a/crates/rpc-client/src/call.rs
+++ b/crates/rpc-client/src/call.rs
@@ -71,7 +71,6 @@ where
         mut self: Pin<&mut Self>,
         cx: &mut task::Context<'_>,
     ) -> task::Poll<<Self as Future>::Output> {
-        trace!("Polling prepared");
         let fut = {
             let CallStateProj::Prepared { connection, request } = self.as_mut().project() else {
                 unreachable!("Called poll_prepared in incorrect state")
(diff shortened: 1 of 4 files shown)


crates/rpc-client/src/call.rs
```diff
@@ -81,11 +80,17 @@ where
                 self.set(CallState::Complete);
                 return Ready(RpcResult::Err(e));
             }
-            let request = request.take().expect("no request").serialize();
 
+            let request = request.take().expect("no request");
+            debug!(method=%request.meta.method, id=%request.meta.id, "sending request");
```
mattsse, line 85: should we log the entire request object?

mattsse (approved):
I definitely want this, for debugging, I think logging the entire request object would also be useful, to debug data serialization

crates/rpc-client/src/call.rs
DaniPopes: Yes I think so, added a trace! with ?request and params type name

PR #505, opened 2024-04-10
fix(rpc-types): accept `Address` in `to` builder method for TxReq

yash-atreya:
## Motivation

The `to` builder method for `TransactionRequest` takes `Option<Address>` as arg. This is inconsistent with other builder methods. 

## Solution

Use `Address` type for arg and wrap it in `Some` inside the method

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/request.rs
+++ b/crates/rpc-types/src/eth/transaction/request.rs
@@ -143,8 +143,8 @@ impl TransactionRequest {
 
     /// Sets the recipient address for the transaction.
     #[inline]
-    pub const fn to(mut self, to: Option<Address>) -> Self {
-        self.to = to;
+    pub const fn to(mut self, to: Address) -> Self {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this makes it consistent with the other setters

mattsse:
> `to` is different because an empty `to` address is a contract creation no?

yeah but the builder starts out with `None`, imo this is the right choice because this is also how the other setters work

prestwich:
not convinced we should default to create instead of erroring if not specifically set to create, as creating with value and empty data is a valid tx that burns user funds. We should probably error on building if `to` is unset, and provide a shortcut for `.create()`

prestwich:
i.e. this behavior would burn funds

```
let mut builder = Default::default().with_value(1 ether);
provider.send_transaction(
   builder
).await;
```

```

PR #506, opened 2024-04-10
feat: add missing blob versioned hashes error variant

mattsse:
4844 must have versioned hashes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/error.rs
+++ b/crates/rpc-types/src/eth/transaction/error.rs
@@ -28,6 +28,9 @@ pub enum ConversionError {
     /// Missing `chainId` field for EIP-1559 transaction.
     #[error("missing `chainId` field for EIP-155 transaction")]
     MissingChainId,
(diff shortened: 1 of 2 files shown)


crates/rpc-types/src/eth/transaction/error.rs
```diff
@@ -28,6 +28,9 @@ pub enum ConversionError {
     /// Missing `chainId` field for EIP-1559 transaction.
     #[error("missing `chainId` field for EIP-155 transaction")]
     MissingChainId,
+    /// Missing `chainId` field for EIP-1559 transaction.
+    #[error("missing `blobVersionedHashes` field for EIP-4844 transaction")]
+    MissingBlobVersionedHashes,
```
DaniPopes, line 33: Can we order these errors per EIP / tx type? :smile:

DaniPopes (approved):

crates/rpc-types/src/eth/transaction/error.rs
mattsse: good point, will do

PR #507, opened 2024-04-10
chore: reorder conversion error variants

mattsse:
make @DaniPopes happy


onbjerg (approved):
what is the desired order? for future reference

DaniPopes:
per EIP / tx

DaniPopes (approved):

PR #508, opened 2024-04-10
feat: port helpers for accesslist

mattsse:
prep for https://github.com/paradigmxyz/reth/pull/7554

adds a few helpers, could use a closer API review, perhaps some aren't even useful.

### DIFF
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -9,7 +9,7 @@ use alloc::vec::Vec;
 
 use alloy_primitives::{Address, B256, U256};
 use alloy_rlp::{RlpDecodable, RlpDecodableWrapper, RlpEncodable, RlpEncodableWrapper};
-use core::mem;
+use core::{mem, ops::Deref};
 
 /// A list of addresses and storage keys that the transaction plans to access.
 /// Accesses outside the list are possible, but become more expensive.
@@ -70,6 +70,14 @@ impl From<AccessList> for Vec<AccessListItem> {
     }
 }
 
+impl Deref for AccessList {
+    type Target = Vec<AccessListItem>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 impl AccessList {
     /// Converts the list into a vec, expected by revm
     pub fn flattened(&self) -> Vec<(Address, Vec<U256>)> {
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip2930.rs
```diff
@@ -101,6 +109,41 @@ impl AccessList {
...
 
+    /// Returns the position of the given address in the access list, if present.
+    pub fn index_of_address(&self, address: Address) -> Option<usize> {
+        self.iter().position(|item| item.address == address)
+    }
+
+    /// Checks if a specific storage slot within an account is present in the access list.
+    ///
+    /// Returns a tuple with flags for the presence of the account and the slot.
+    pub fn contains(&self, address: Address, slot: B256) -> (bool, bool) {
```
onbjerg, line 120: imo mirror other method so `contains_storage`? if not, for me the immediate expectation was that `contains` is what `contains_address` is

crates/eips/src/eip2930.rs
```diff
@@ -101,6 +109,41 @@ impl AccessList {
...
+    }
+
+    /// Checks if the access list contains the specified address.
+    pub fn contains_address(&self, address: Address) -> bool {
+        self.iter().any(|item| item.address == address)
+    }
+
+    /// Checks if the storage keys at the given index within an account are present in the access
+    /// list.
+    pub fn contains_storage_key_at_index(&self, slot: B256, index: usize) -> bool {
```
onbjerg, line 132: what is this useful for vs `contains`?

crates/eips/src/eip2930.rs
```diff
@@ -101,6 +109,41 @@ impl AccessList {
...
+        self.get(index).map_or(false, |entry| {
+            entry.storage_keys.iter().any(|storage_key| *storage_key == slot)
+        })
+    }
+
+    /// Adds an address to the access list and returns `true` if the operation results in a change,
+    /// indicating that the address was not previously present.
+    pub fn add_address(&mut self, address: Address) -> bool {
+        !self.contains_address(address) && {
+            self.0.push(AccessListItem { address, storage_keys: Vec::new() });
```
onbjerg, line 142: what is the use case for this? in my mind an access list is immutable unless it is being built by the node, which i assume would just work on a `Vec<AccessListItem>`

if it doesn't work on the `Vec`, should we add methods to add slots etc?

crates/eips/src/eip2930.rs
mattsse: made private

onbjerg (approved):

PR #509, opened 2024-04-10
fix: eip1559 estimator

klkvr:
## Motivation

ref https://github.com/foundry-rs/foundry/issues/7620

currently we are not accounting for an edge case when estimated priority fee is higher than base fee. baseFee*2 is usually enough, but in the mentioned PR anvil was started with 0 baseFee, thus baseFee was always estimated as zero

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
How can we catch this in the future in foundry ci?

klkvr:
I think we can try testing this directly: the repro for this is spinning up anvil with `--block-base-fee-per-gas 0`, then sending any tx with non-zero prio fee and then running any script

mattsse (approved):

PR #511, opened 2024-04-11
Improve `watch_pending_transaction`

### LINKED ISSUE #389: [Bug] `watch_pending_transaction` hangs for non-pending transactions
### Component

provider, pubsub

_No response_

### Operating System

None

### Describe the bug

Ref #381

Currently, `watch_pending_transaction` implementation always expected to be called with pending tx which is about to be included. Because of that, if it is invoked with the transaction hash mined in block X, while current latest block is X+1, it will wait indefinitely for a block including the transaction.

klkvr:
## Motivation

Closes #389

Now when receiving pending transaction, it is placed in `unconfirmed` map firstly, and then we start to check for it in new blocks and try to perform a sing `eth_getTransactionReceipt` request with it. That way, if tx is actually pending, we will remove it from `unconfirmed` once it is getting included into some of the future blocks, and if it is not pending, we will successfuly fetch receipt for it instead of waiting indefinitely.

This also updates `TxWatcher` to return transaction receipt, to implement this we now start polling for receipt once tx gets enough confirmations

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/lib.rs
+++ b/crates/network/src/lib.rs
@@ -17,7 +17,7 @@
 
 use alloy_eips::eip2718::{Eip2718Envelope, Eip2718Error};
 use alloy_json_rpc::RpcObject;
-use alloy_primitives::Address;
+use alloy_primitives::{Address, B256};
 use core::fmt::{Debug, Display};
 
 mod transaction;
@@ -42,6 +42,12 @@ pub use alloy_eips::eip2718;
 pub trait ReceiptResponse {
     /// Address of the created contract, or `None` if the transaction was not a deployment.
     fn contract_address(&self) -> Option<Address>;
+
+    /// Number of block in which transaction was included.
+    fn block_number(&self) -> Option<u64>;
+
+    /// Hash of the transaction.
+    fn transaction_hash(&self) -> B256;
 }
 
 /// Captures type info for network-specific RPC requests/responses.
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -1,9 +1,9 @@
 //! Block heartbeat and pending transaction watcher.
 
 use crate::{Provider, RootProvider};
-use alloy_json_rpc::RpcError;
-use alloy_network::Network;
+use alloy_network::{Network, ReceiptResponse};
 use alloy_primitives::B256;
+use alloy_rpc_client::WeakClient;
 use alloy_rpc_types::Block;
 use alloy_transport::{utils::Spawnable, Transport, TransportErrorKind, TransportResult};
 use futures::{stream::StreamExt, FutureExt, Stream};
@@ -11,6 +11,7 @@ use std::{
     collections::{BTreeMap, HashMap},
     fmt,
     future::Future,
+    marker::PhantomData,
     time::{Duration, Instant},
 };
 use tokio::{
@@ -53,7 +54,7 @@ use tokio::{
 /// ```
 #[must_use = "this type does nothing unless you call `register`, `watch` or `get_receipt`"]
 #[derive(Debug)]
-pub struct PendingTransactionBuilder<'a, T, N> {
+pub struct PendingTransactionBuilder<'a, T, N: Network> {
     config: PendingTransactionConfig,
     provider: &'a RootProvider<T, N>,
 }
@@ -153,8 +154,8 @@ impl<'a, T: Transport + Clone, N: Network> PendingTransactionBuilder<'a, T, N> {
     /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
     ///   confirmed.
     #[doc(alias = "build")]
-    pub async fn register(self) -> TransportResult<PendingTransaction> {
-        self.provider.watch_pending_transaction(self.config).await
+    pub async fn register(self) -> TransportResult<PendingTransaction<N>> {
+        self.provider.register_tx_watcher(self.config).await
     }
 
     /// Waits for the transaction to confirm with the given number of confirmations.
@@ -165,7 +166,7 @@ impl<'a, T: Transport + Clone, N: Network> PendingTransactionBuilder<'a, T, N> {
     /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
     ///   confirmed.
     pub async fn watch(self) -> TransportResult<B256> {
-        self.register().await?.await
+        Ok(self.register().await?.await?.transaction_hash())
     }
 
     /// Waits for the transaction to confirm with the given number of confirmations, and
@@ -180,11 +181,7 @@ impl<'a, T: Transport + Clone, N: Network> PendingTransactionBuilder<'a, T, N> {
     ///   confirmed.
     /// - [`watch`](Self::watch) for watching the transaction without fetching the receipt.
     pub async fn get_receipt(self) -> TransportResult<N::ReceiptResponse> {
-        let pending_tx = self.provider.watch_pending_transaction(self.config).await?;
-        let hash = pending_tx.await?;
-        let receipt = self.provider.get_transaction_receipt(hash).await?;
-
-        receipt.ok_or(RpcError::NullResp)
+        self.provider.register_tx_watcher(self.config).await?.await
     }
 }
 
@@ -272,16 +269,16 @@ impl PendingTransactionConfig {
     }
 }
 
-struct TxWatcher {
+struct TxWatcher<T> {
     config: PendingTransactionConfig,
-    tx: oneshot::Sender<()>,
+    tx: oneshot::Sender<T>,
 }
 
-impl TxWatcher {
+impl<T> TxWatcher<T> {
     /// Notify the waiter.
-    fn notify(self) {
+    fn notify(self, receipt: T) {
         debug!(tx=%self.config.tx_hash, "notifying");
-        let _ = self.tx.send(());
+        let _ = self.tx.send(receipt);
     }
 }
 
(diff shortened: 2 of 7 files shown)


crates/network/src/lib.rs
```diff
@@ -41,6 +41,12 @@ pub use alloy_eips::eip2718;
 pub trait ReceiptResponse {
     /// Address of the created contract, or `None` if the transaction was not a deployment.
     fn contract_address(&self) -> Option<Address>;
+
+    /// Number of block in which transaction was included.
+    fn block_number(&self) -> Option<u64>;
+
+    /// Hash of the transaction.
+    fn transaction_hash(&self) -> B256;
```
mattsse, line 49: these make sense

crates/provider/src/heart.rs
```diff
@@ -290,54 +291,51 @@ impl TxWatcher {
...
 /// transaction hash once the underlying transaction has been confirmed the specified number of
 /// times in the network.
-pub struct PendingTransaction {
+pub struct PendingTransaction<N: Network> {
     /// The transaction hash.
     pub(crate) tx_hash: B256,
     /// The receiver for the notification.
-    // TODO: send a receipt?
-    pub(crate) rx: oneshot::Receiver<()>,
+    pub(crate) rx: oneshot::Receiver<N::ReceiptResponse>,
```
mattsse, line 298: here we could just use T

and N::ReceiptResponse on the provider function

crates/provider/src/heart.rs
```diff
@@ -355,39 +353,50 @@ impl HeartbeatHandle {
 
 // TODO: Parameterize with `Network`
```
mattsse, line 354:
```suggestion
```

crates/provider/src/heart.rs
```diff
@@ -458,37 +504,112 @@ impl<S> Heartbeat<S> {
...
+            }
+
+            // If the stream closed before we got a receipt, it's not OK.
+            let _ = shutdown.send(()).await;
+        }));
+    }
+}
+
+impl<S: Stream<Item = Block> + Unpin + 'static, N: Network> Heartbeat<S, N> {
+    async fn into_future<T: Transport + Clone>(
```
mattsse, line 605: can we get some docs here that describe the internal logic

crates/provider/src/heart.rs
```diff
@@ -499,15 +620,43 @@ impl<S: Stream<Item = Block> + Unpin + 'static> Heartbeat<S> {
                 select! {
                     biased;
```
mattsse, line 621: the biased should not be on shutdown, so we need to change the order here

crates/provider/src/heart.rs
```diff
@@ -458,37 +504,112 @@ impl<S> Heartbeat<S> {
...
+    }
+
+    fn poll_receipt(&mut self, watcher: TxWatcher<N>) {
+        let mut poller = PollerBuilder::<_, _, Option<N::ReceiptResponse>>::new(
+            self.client.clone(),
+            "eth_getTransactionReceipt",
+            (watcher.config.tx_hash,),
+        )
+        .spawn()
+        .into_stream();
```
mattsse, line 586: hmm do we want to spawn 1 task per tx?

crates/provider/src/heart.rs
```diff
@@ -458,37 +504,112 @@ impl<S> Heartbeat<S> {
...
+        self.poll_receipt_tasks.push(Box::pin(async move {
+            while let Some(receipt) = poller.next().await {
+                if let Some(receipt) = receipt {
+                    watcher.notify(receipt);
+                    return;
+                }
+            }
+
+            // If the stream closed before we got a receipt, it's not OK.
+            let _ = shutdown.send(()).await;
```
mattsse, line 599: this shuts down the heartbeat?

mattsse (changes requested):
the main motivation for this is checking for a receipt in case the tx has already been included

It's a bit difficult to follow the heartbeat stream's logic, I'd appreciate some high-level docs that explain all of this.

but directionally, I think all of this makes sense.

@prestwich has some thoughts here probably

crates/provider/src/heart.rs
klkvr: those pollers will usually stop after first request because this task is started once tx is included, so I think this is fine?

crates/provider/src/heart.rs
klkvr: yeah, we send shutdown only when poller unexpectedly stops which usually means that there was either irrecoverable error or client was dropped

not sure how we can handle this differently, we can probably send `Result` through `TxWatcher` if something goes wrong

crates/provider/src/heart.rs
```diff
@@ -458,37 +503,112 @@ impl<S> Heartbeat<S> {
...
+        Self {
+            client,
+            shutdown_tx,
+            fetch_receipt_tasks: Default::default(),
+            poll_receipt_tasks: Default::default(),
+        }
+    }
+
+    fn fetch_receipt(&mut self, tx_hash: B256) {
+        // We don't actually poll, but use poller to retry on failures.
```
prestwich, line 557: this is a bit of a layer violation, should be handled by a tower layer on the transport stack

crates/provider/src/heart.rs
prestwich: I think the spawn-per-tx falls out of the intent to retry, which probably belongs elsewhere. using the poller to cover up transport failures is an anti-pattern

crates/provider/src/heart.rs
klkvr: should this just do a request through client? what'd be the correct behavior on request failure?

crates/provider/src/heart.rs
```diff
@@ -499,15 +619,43 @@ impl<S: Stream<Item = Block> + Unpin + 'static> Heartbeat<S> {
...
+
+                    // Watch for new tx watching requests.
                     ix_opt = ixns.recv() => match ix_opt {
-                        Some(to_watch) => self.handle_watch_ix(to_watch),
+                        Some(to_watch) => self.handle_watch_ix(to_watch, &mut receipt_tasks),
                         None => break 'shutdown, // ix channel is closed
                     },
 
+                    // Handle potential receipts for newly requested transaction.
+                    Some(Some(receipt)) = receipt_tasks.fetch_receipt_tasks.next() => {
```
prestwich, line 634: [should be flattened?](https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten)

prestwich (changes requested):
> It's a bit difficult to follow the heartbeat stream's logic, I'd appreciate some high-level docs that explain all of this.

I'm reluctant to make any changes until we have this high-level doc and a PRD for how the update would change the flow

Couple things:

- I'm not clear on the benefit of this change over adding a `get_transaction_receipt` before returning a `PendingTransactionBuilder` in `Provider::send_transaction_inner` and `Provider::send_raw_transaction` ?

- I'm not against having a receipt-retrieval task, but i'd want it to be a generalized caching task with cache invalidation detection (based on a block stream), rather than a child of the heartbeat task

klkvr:
> I'm not clear on the benefit of this change over adding a get_transaction_receipt before returning a PendingTransactionBuilder in Provider::send_transaction_inner and Provider::send_raw_transaction ?

That would not prevent user from constructing a `PendingTransactonBuilder` for a existing tx which will hang forever.

prestwich:
> That would not prevent user from constructing a PendingTransactonBuilder for an existing tx which will hang forever.

Do we want users to be doing that? Feels like we don't, right? I agree it's more complicated than the question implies tho, as the transaction isn't registered with the heartbeat until the `register()` call hits `watch_pending_transaction()` (which also seems like something we should fix, btw).

Maybe the right place to do it is in `watch_pending_transaction` i.e. immediately after the heartbeat learns of the tx?

mattsse:
yeah the issue is this step here:

https://github.com/alloy-rs/alloy/blob/7629f79e3ffb6abd0be901a06deed2ab9f695e4e/crates/provider/src/heart.rs#L182-L185

if the heartbeat has the tx's block already processed before it receives the watcher it will hang here before the receipt is fetched

so we should try fetching the tx's receipt at least once in the heartbeat

prestwich:
> if the heartbeat has the tx's block already processed before it receives the watcher it will hang here before the receipt is fetched

reminds me we also probably have some memory leaks here wrt pendings that never get confirmed. are we comfortable leaving those in memory forever or do we want to enforce some maximum timeout?

mattsse:
a timeout makes sense, the caller can then still decide whether to reregister the watcher, but should probably be combined with (pending) receipt fetching

klkvr:
> reminds me we also probably have some memory leaks here wrt pendings that never get confirmed. are we comfortable leaving those in memory forever or do we want to enforce some maximum timeout or something like an LRU?

We already have timeouts I beleive, or you mean something else?
https://github.com/alloy-rs/alloy/blob/7629f79e3ffb6abd0be901a06deed2ab9f695e4e/crates/provider/src/heart.rs#L404

prestwich:
we have user-configurable timeouts, but we don't have a "max system timeout to prevent unbounded memory growth". Consider a user that watches a tx every second and sets no timeout, but only 95% confirm. Currently I think we. grow memory by 5 txns every 100 seconds, and never reap that memory. So after a few hours memory usage will be quite large

mattsse (changes requested):
~~as discussed, integrating this in heart likely requires a bit more thinking~~

~~@klkvr as a hotfix, can we keep polling for the receipt here concurrently~~

https://github.com/alloy-rs/alloy/blob/cb95183d477024b57b27336035d10403e8ba55b8/crates/provider/src/heart.rs#L182-L187

klkvr:
with current impl heartbeat itself already polls for receipt by querying receipt-fetching task:
https://github.com/alloy-rs/alloy/blob/786a6a416ec6048206a060b4dc563363c2adae03/crates/provider/src/heart.rs#L430
https://github.com/alloy-rs/alloy/blob/786a6a416ec6048206a060b4dc563363c2adae03/crates/provider/src/heart.rs#L469

should I extract this logic? feels like we probably still want to have `TxWatcher` until the point we refactor the impl next time

crates/provider/src/heart.rs
```diff
@@ -458,37 +488,74 @@ impl<S> Heartbeat<S> {
...
-    async fn into_future(
+type TransportReceipt<R> = TransportResult<Option<R>>;
+
+/// Spawns a receipt fetching task. This task receives transaction hashes and fetches receipts for
+/// requested transactions, sending the results back.
+fn spawn_receipt_fetcher<T: Transport + Clone, N: Network>(
+    client: WeakClient<T>,
+) -> ReceiptFetcherHandle<N> {
+    let (receipt_tx, receipt_rx) = mpsc::channel(16);
+    let (hash_tx, mut hash_rx) = mpsc::channel(16);
```
mattsse, line 524: ah okay, I get how this works.

this spawns a dedicated fetcher task that receives the receipts spawns eth_getTransactionReceipt requests

crates/provider/src/heart.rs
```diff
@@ -415,37 +413,69 @@ impl<S> Heartbeat<S> {
...
+                // Edge case: receipt does not include block number, we can't count confirmations,
+                // so notify immediately.
+                watcher.notify(receipt);
             }
-            // Otherwise add it to the waiting list.
-            debug!(tx=%watcher.config.tx_hash, %block_height, confirmations, "adding to waiting list");
-            self.waiting_confs.entry(*block_height + confirmations - 1).or_default().push(watcher);
+        } else {
+            // Keep polling
+            let _ = hash_tx.send(tx_hash).await;
```
mattsse, line 469: this sends the hash back to the receipt fetcher task?

I find the naming here a bit confusing

mattsse (changes requested):
okay given that the receipt fetching logic is already implemented in heartbeat now, we should proceed with this after resolving conflicts because this is now a blocker

conceptually I believe this should work.

personally I'd rather have a special future type than doing all of this in async code

prestwich:
closing this pending a design session, as #558 patches sufficiently

PR #513, opened 2024-04-11
Revert "chore: remove outdated license"

prestwich:
Reverts alloy-rs/alloy#510

License is not outdated, it applies to code in the file which was copy-pasted from the tower project and reproducing the license text is necessary to comply with the license terms


mattsse (approved):
my bad -.-

prestwich:
no worries. do you think we should add a `// NB:`  describing the purpose of the block?

PR #515, opened 2024-04-11
Prestwich/signer multiplex

prestwich:
Resolves signer filler not setting from and failing to sign

## Motivation

Unblock foundry dev

## Solution

- `NetworkSigner<N>` is now expected to contain a store of credentials, by address.
- `NetworkSigner<N>` now selects credentials based on request `from` field
- `NetworkSigner<N>` allows user-specified sender when signing

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/network/src/transaction/signer.rs
+++ b/crates/network/src/transaction/signer.rs
@@ -22,6 +22,12 @@ pub trait NetworkSigner<N: Network>: std::fmt::Debug + Send + Sync {
     /// specified.
     fn default_signer(&self) -> Address;
 
+    /// Return true if the signer contains a credential for the given address.
+    fn is_signer_for(&self, address: &Address) -> bool;
+
+    /// Return an iterator of all signer addresses.
+    fn signers(&self) -> impl Iterator<Item = &Address>;
+
     /// Asynchronously sign an unsigned transaction, with a specified
     /// credential.
     async fn sign_transaction_from(
--- a/crates/network/src/transaction/signer.rs
+++ b/crates/network/src/transaction/signer.rs
@@ -26,7 +26,7 @@ pub trait NetworkSigner<N: Network>: std::fmt::Debug + Send + Sync {
     fn is_signer_for(&self, address: &Address) -> bool;
 
     /// Return an iterator of all signer addresses.
-    fn signers(&self) -> impl Iterator<Item = &Address>;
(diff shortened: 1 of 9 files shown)


crates/network/src/transaction/signer.rs
```diff
@@ -1,25 +1,52 @@
...
+    /// specified.
+    fn default_signer(&self) -> Address;
+
+    /// Asynchronously sign an unsigned transaction, with a specified
+    /// credential.
+    async fn sign_transaction_from(
+        &self,
+        sender: Option<Address>,
+        tx: N::UnsignedTx,
+    ) -> alloy_signer::Result<N::TxEnvelope>;
```
mattsse, line 31: this makes sense to me and is convenient to have,
it should be trivial for any, non-multiplex signers to implement by ensuring the default signer is Some(sender)

mattsse (changes requested):
I like this, this is pretty much exactly what we have here:

https://github.com/paradigmxyz/reth/blob/39dea65b63d1acac8370bac12ddea2c809ed7803/crates/rpc/rpc/src/eth/signer.rs#L44-L49

can we also add, or similar:

```rust
/// Returns `true` whether this signer can sign for this address
    fn is_signer_for(&self, addr: &Address) -> bool
```

prestwich:
> I like this, this is pretty much exactly what we have here:
> 
> https://github.com/paradigmxyz/reth/blob/39dea65b63d1acac8370bac12ddea2c809ed7803/crates/rpc/rpc/src/eth/signer.rs#L44-L49
> 
> can we also add, or similar:
> 
> ```rust
> /// Returns `true` whether this signer can sign for this address
>     fn is_signer_for(&self, addr: &Address) -> bool
> ```

Added that, and a method to get an iterator of addresses that we can sign with

crates/network/src/transaction/signer.rs
prestwich: doing a slight cleanup here as i wrote this function signature before i added the default signer

prestwich: removed the `Option` because we can rely on the caller to call `default_signer`

mattsse (approved):

DaniPopes:
What does "Preswitch/signer multiplex" mean? :smile:

prestwich:
branch name 😅

Issue #514, opened 2024-04-11
[Feature] Change all `Option<BlockId>` to `BlockId` on methods in the `Provider`


zerosnacks:
### Component

provider, pubsub

### Describe the feature you would like

`get_code_at` and `get_uncle` on the `Provider` currently require the user to define the `tag` whereas others allow for defining `tag` as `None`, where `None` means defaulting to `Latest`.

https://github.com/alloy-rs/alloy/blob/7629f79e3ffb6abd0be901a06deed2ab9f695e4e/crates/provider/src/provider.rs#L768-L771

https://github.com/alloy-rs/alloy/blob/7629f79e3ffb6abd0be901a06deed2ab9f695e4e/crates/provider/src/provider.rs#L826-L836

Whereas other methods allow for defining `tag` as `None`, 

https://github.com/alloy-rs/alloy/blob/7629f79e3ffb6abd0be901a06deed2ab9f695e4e/crates/provider/src/provider.rs#L627-L639

Where automatically defaulting means `Latest`:

https://github.com/alloy-rs/alloy/blob/987b3936f78c067baedcf811620975d3a4a26443/crates/rpc-types/src/eth/block.rs#L719-L723

Based on feedback from @mattsse on the PR of @EmperorOrokuSaki it was decided to instead of making `get_code_at` and `get_uncle` have `Option<BlockId>` to change all other methods to use `BlockId` (not optional).

EmperorOrokuSaki:
I'll work on this.

zerosnacks:
> Hey, update on this. The direction is going to be the removal of all Option params and making them required in provider methods. This ensures consistency with the JSON RPC specification.

Thanks! I'll update the description and title

EmperorOrokuSaki:
Can we close this issue? The PR was merged.

PR #516, opened 2024-04-12
refactor: make optional BlockId params required in provider functions

EmperorOrokuSaki:
## Motivation

This issue #514 

## Solution

Make the `BlockId` optional and use the default value if it's `None`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -190,7 +190,7 @@ impl CallDecoder for () {
 #[must_use = "call builders do nothing unless you `.call`, `.send`, or `.await` them"]
 pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
     request: N::TransactionRequest,
-    block: Option<BlockId>,
+    block: BlockId,
     state: Option<StateOverride>,
     /// The provider.
     // NOTE: This is public due to usage in `sol!`, please avoid changing it.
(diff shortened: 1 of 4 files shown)


mattsse (commented):
the tag is a required argument,

I think the option does not improve the API, because None is equal to `Default::default()` or `BlockId::latest()`

EmperorOrokuSaki:
> the tag is a required argument,
> 
> I think the option does not improve the API, because None is equal to `Default::default()` or `BlockId::latest()`

That is true, it is a required parameter in the RPC specification, but the other examples provided in the original issue have the same condition so it's a case of inconsistency a bit. I checked those methods and for those it's also required in the specification.

mattsse:
I see, then we should change all `Option<BlockId>` to `BlockId`

EmperorOrokuSaki:
> I see, then we should change all `Option<BlockId>` to `BlockId`
Sounds good. I'll change them in this pr.

EmperorOrokuSaki:
In some files previously `None` was supplied when the arg was of the type `Option<BlockId>`, I changed those to `BlockId::default()` since that was the final value.

prestwich:
> I think the option does not improve the API, because None is equal to Default::default() or BlockId::latest()

The only improvement is writing `None` is fewer characters than `Default::default()`. Seems kinda pointless. in favor of this change

prestwich:
alternate API designs would be to bifurcate, which seems ok-but-verbose on our end

```
fn call_with_block(..., block_id: BlockId)
fn call(...) { self.call_with_block(..., Default::default()) }
``` 

or to switch to a rpc builder pattern for return values of those types. which seems bad

prestwich (approved):

PR #522, opened 2024-04-12
fix: correctly parse IPC sockets in builtin connections

DaniPopes:
IPC sockets return `false` for all `is_*` metadata queries.


mattsse (approved):
nice, lgtm!

PR #523, opened 2024-04-12
fix: tx receipt inclusion context

mattsse:
these can be null, or missing entirely

https://github.com/ethereum/go-ethereum/blob/767b00b0b514771a663f3362dd0310fc28d40c25/core/types/receipt.go#L69-L73


DaniPopes (approved):

banteg:
this is incorrect and should be reverted. the spec says these fields are required.
https://github.com/ethereum/execution-apis/blob/c4e7da7e486f6cb6ff6281c3ff28bb37445d5415/src/schemas/transaction.yaml#L334-L339

mattsse:
sorry about this

PR #526, opened 2024-04-13
feat: add rpc-types-anvil

mattsse:
adds a crate for rpc types used by anvil

there are only a handful types, it's easier to maintain these in this workspace, and we could also an anvil provider trait for specific dev functions,
and this would also make it possible to define server API traits:

https://github.com/paradigmxyz/reth/pull/7495

### DIFF
--- a/crates/rpc-types-anvil/README.md
+++ b/crates/rpc-types-anvil/README.md
@@ -1,0 +1,3 @@
+# alloy-rpc-types-anvil
+
+RPC types for Anvil,Hardhat,Ganache dev nodes. 
--- a/crates/rpc-types-anvil/README.md
+++ b/crates/rpc-types-anvil/README.md
@@ -1,3 +1,3 @@
 # alloy-rpc-types-anvil
 
-RPC types for Anvil,Hardhat,Ganache dev nodes. 
+RPC types for Anvil, Hardhat, and Ganache dev nodes. 
--- a/crates/rpc-types-anvil/src/lib.rs
+++ b/crates/rpc-types-anvil/src/lib.rs
@@ -6,11 +6,11 @@
 #![warn(
     missing_copy_implementations,
     missing_debug_implementations,
+    missing_docs,
     unreachable_pub,
     clippy::missing_const_for_fn,
     rustdoc::all
 )]
-#![allow(missing_docs)]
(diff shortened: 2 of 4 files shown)


crates/rpc-types-anvil/README.md
```diff
@@ -0,0 +1,3 @@
+# alloy-rpc-types-anvil
+
+RPC types for Anvil,Hardhat,Ganache dev nodes.
```
onbjerg, line 3:
```suggestion
RPC types for Anvil, Hardhat, and Ganache dev nodes. 
```

crates/rpc-types-anvil/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+#![deny(unused_must_use, rust_2018_idioms)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
+use alloy_primitives::{TxHash, B256, U256};
+use serde::{Deserialize, Deserializer, Serialize};
+use std::collections::BTreeMap;
+
+/// Represents the params to set forking which can take various forms
+///  - untagged
+///  - tagged `forking`
```
onbjerg, line 24:
```suggestion
///  - untagged
///  - tagged forking
```

maybe we should clarify what tagged vs untagged means? i assume its just one is w block number the other without?

crates/rpc-types-anvil/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+    html_favicon_url = "https://raw.githubusercontent.com/alloy-rs/core/main/assets/favicon.ico"
+)]
+#![warn(
+    missing_copy_implementations,
+    missing_debug_implementations,
+    unreachable_pub,
+    clippy::missing_const_for_fn,
+    rustdoc::all
+)]
+#![allow(missing_docs)]
```
onbjerg, line 13: add follow up issue to remove this, we want docs, but don't want to block this PR on it since it's a niche-ish crate and it is blocking other work streams

onbjerg (approved):

crates/rpc-types-anvil/src/lib.rs
mattsse: yeah, could be nested object or not

PR #529, opened 2024-04-13
feat: Export inner encoding / decoding functions from `Tx*` types

clabby:
## Overview

Exports a few of the `Tx*` types' inner encoding & decoding functions for use in the `op-alloy-consensus` wrapper. Doing so allows us to re-use all of these types in the expanded transaction envelope enum.

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -117,8 +117,8 @@ impl TxEip1559 {
         })
     }
 
-    /// Encodes only the transaction's fields into the desired buffer, without a RLP header.
-    pub(crate) fn fields_len(&self) -> usize {
+    /// Outputs the length of the transaction's fields, without a RLP header.
+    pub fn fields_len(&self) -> usize {
         let mut len = 0;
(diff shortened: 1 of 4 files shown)


crates/consensus/src/transaction/eip1559.rs
```diff
@@ -118,7 +118,7 @@ impl TxEip1559 {
     }
 
     /// Encodes only the transaction's fields into the desired buffer, without a RLP header.
-    pub(crate) fn fields_len(&self) -> usize {
+    pub fn fields_len(&self) -> usize {
```
mattsse, line 121: can we also update the docs here, because they are incorrect

mattsse (changes requested):
doc nit

I think we need to make them pub so that the tx types can be reused externally, for example in a network specific enum that needs to implement rlp

prestwich:
i agree that we need to do this. i would like to open a follow up issue for improving naming

mattsse (changes requested):
can we mark these as `#[doc(hidden)]`? so that they don't show up in autosuggestions

prestwich (approved):

PR #530, opened 2024-04-13
feat: `no_std` Engine RPC types

clabby:
## Overview

Makes `alloy-rpc-types-engine` no_std compatible.

_note_: Not ready for review; no_std `alloy-rpc-types` will be stacked beneath this PR

### DIFF
--- a/crates/rpc-types-engine/Cargo.toml
+++ b/crates/rpc-types-engine/Cargo.toml
@@ -13,9 +13,9 @@ exclude.workspace = true
 
 [dependencies]
 # ethereum
-alloy-rlp = { workspace = true, features = ["arrayvec", "derive"] }
+alloy-rlp = { workspace = true, features = ["derive"] }
 alloy-primitives = { workspace = true, features = ["rlp", "serde"] }
-alloy-consensus = { workspace = true, features = ["std"] }
+alloy-consensus.workspace = true
 alloy-rpc-types.workspace = true
 alloy-serde.workspace = true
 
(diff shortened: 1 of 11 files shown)


crates/rpc-types-engine/Cargo.toml
```diff
@@ -24,15 +24,16 @@ ethereum_ssz_derive = { workspace = true, optional = true }
...
+thiserror = { workspace = true, optional = true }
 
 # jsonrpsee
 jsonrpsee-types = { version = "0.22", optional = true }
 alloy-eips = { workspace= true, optional = true }
 
 [features]
-jsonrpsee-types = ["dep:jsonrpsee-types"]
-ssz = ["dep:ethereum_ssz", "dep:ethereum_ssz_derive", "alloy-primitives/ssz", "alloy-rpc-types/ssz", "dep:alloy-eips", "alloy-eips/ssz"]
+std = ["alloy-consensus/std", "alloy-rlp/arrayvec", "dep:thiserror"]
```
mattsse, line 34: std should be default feature, so we don't need enable it manually

crates/rpc-types-engine/src/forkchoice.rs
```diff
@@ -34,19 +34,20 @@ pub struct ForkchoiceState {
 ///
 /// These are considered hard RPC errors and are _not_ returned as [PayloadStatus] or
 /// [PayloadStatusEnum::Invalid].
-#[derive(Clone, Copy, Debug, PartialEq, Eq, thiserror::Error)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
 pub enum ForkchoiceUpdateError {
```
mattsse, line 39: thiserror also takes care of Display,

I think it would be better to manually impl Display and then just feature gate the error impl

mattsse (changes requested):

clabby:
Closing for now; `alloy-rpc-types` needs some more work to get this in.

Dependencies on `itertools`, `std::net::IpAddr` (available in core, but not stable in rustc 1.7.6), etc. need to be sorted.

PR #532, opened 2024-04-14
fix: dont use fuse::select_next_some

mattsse:
closes https://github.com/foundry-rs/foundry/issues/7652

SelectNextSome over `Fuse` will panic

Fuse will return 
```rust
        if *this.done {
            return Poll::Ready(None);
        }
```

but selectsome will panic if the underlying stream already returned None once

```rust
        if let Some(item) = ready!(self.stream.poll_next_unpin(cx)) {
            Poll::Ready(item)
        } else {
            debug_assert!(self.stream.is_terminated());
            cx.waker().wake_by_ref();
            Poll::Pending
        }
```

### DIFF
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -506,7 +506,7 @@ impl<S: Stream<Item = Block> + Unpin + 'static> Heartbeat<S> {
                     },
 
                     // Wake up to handle new blocks.
-                    block = self.stream.select_next_some() => {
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

mattsse:
hmm, actually this is very weird, because the fuse stream is always marked as terminated when it returns None

which makes me think the issue is somewhere else

PR #537, opened 2024-04-15
docs: fix 404s on docs site via absolute paths

WillPapper:
## Motivation

Currently, the README uses relative paths rather than absolute paths. This leads to 404s in two places:

On https://alloy-rs.github.io/alloy/, where a link to e.g. alloy goes to https://alloy-rs.github.io/alloy/alloy/crates/alloy instead of https://github.com/alloy-rs/alloy/tree/main/crates/alloy

On the Alloy crate at https://github.com/alloy-rs/alloy/tree/main/crates/alloy, where a link to e.g. alloy-signer goes to https://github.com/alloy-rs/alloy/blob/main/crates/alloy/crates/signer instead of https://github.com/alloy-rs/alloy/tree/main/crates/signer

## Solution

This commit fixes both of these 404s by switching from relative paths to absolute paths. It also follows the same usage of absolute paths as [lib.rs](https://github.com/alloy-rs/alloy/blob/037dd4b20ec8533d6b6d5cf5e9489bbb182c18c6/crates/alloy/src/lib.rs#L197), making the README consistent with the rest of the documentation.

(In addition, a newline was added for consistency with other unordered lists, but that's not really worth commenting on)

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,30 +42,29 @@ This repository contains the following crates:
   - [`alloy-transport-ipc`] - IPC transport implementation
   - [`alloy-transport-ws`] - WS transport implementation
 
-[`alloy`]: crates/alloy
-[`alloy-consensus`]: crates/consensus
-[`alloy-contract`]: crates/contract
-[`alloy-eips`]: crates/eips
-[`alloy-genesis`]: crates/genesis
-[`alloy-json-rpc`]: crates/json-rpc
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
lgtm

prestwich (approved):

PR #542, opened 2024-04-15
Add link to docs to README

kayabaNerve:
## Motivation

While the GH repo's site is the docs page, that is off-center (and not rendered when viewing the repository via a cloned local view). I had to be explicitly told the docs site existed due to not noticing that location (nor any other which may or may not exist). Ideally, the README includes the link *somewhere*.

## Solution

This adds a link in a location with no changes to layout.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Alloy connects applications to blockchains.
 
 Alloy is a rewrite of [`ethers-rs`] from the ground up, with exciting new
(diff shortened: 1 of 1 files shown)


gakonst (approved):
tyty!

PR #545, opened 2024-04-15
add blob gas conversion error

tcoratger:
Follow up for #541.

### DIFF
--- a/crates/rpc-types/src/eth/transaction/error.rs
+++ b/crates/rpc-types/src/eth/transaction/error.rs
@@ -54,4 +54,10 @@ pub enum ConversionError {
     /// Missing block number
     #[error("missing block number")]
     MissingBlockNumber,
+    /// Blob gas used integer conversion error
+    #[error("blob gas used integer conversion error")]
+    BlobGasUsedConversion(TryFromIntError),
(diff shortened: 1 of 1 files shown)


crates/rpc-types/src/eth/transaction/error.rs
```diff
@@ -54,4 +54,10 @@ pub enum ConversionError {
     /// Missing block number
     #[error("missing block number")]
     MissingBlockNumber,
+    /// Blob gas used integer conversion error
+    #[error("blob gas used integer conversion error")]
+    BlobGasUsedConversion(TryFromIntError),
+    /// Excess blob gas integer conversion error
+    #[error("excess blob gas integer conversion error")]
+    ExcessBlobGasConversion(TryFromIntError),
```
mattsse, line 62: the wrapped error is not useful if not used in the error message, same goes for the other conversion errors

mattsse (changes requested):

crates/rpc-types/src/eth/transaction/error.rs
tcoratger: Just fixed

PR #551, opened 2024-04-16
chore: remove Sealed in Transport definition

DaniPopes:
Transport is implemented generically on any T that is a service with the required packets, the sealed trait does nothing here


mattsse (approved):
I see, this is indeed redundant

Issue #549, opened 2024-04-16
[Refactor] change `TransactionRequest` to field to `Option<TxKind>`


prestwich:
- [ ] Modify to field to make creates explicit
- [ ] Error on non-4844 build if to is unset
- [ ] Error on 4844 build if to is unset or create
- [ ] Add following to the builder trait
    - [ ] `fn as_create(self) -> Self` shortcut to set the field to `Some(Create)`
    - [ ] `fn deploy_code(self, code: Vec<u8>) -> Self` to set `data` and `to` fields at once
    - [ ] `fn with_call<T: SolCall>(self, t: T) -> Self` to set `data` and ensure `to` is NOT `Create` (set `to` to `None` if it is currently `Create`)

              not convinced we should default to create instead of erroring if not specifically set to create, as creating with `value` and empty data is a valid tx that burns user funds. We should probably error on building if `to` is unset, and provide a shortcut for `.create()`

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/issues/505#issuecomment-2050124501_

EmperorOrokuSaki:
I'll give this a try.

PR #553, opened 2024-04-16
[Blocked] Update TransactionRequest's `to` field to TxKind

EmperorOrokuSaki:
## Motivation
Issue https://github.com/alloy-rs/alloy/issues/549
## Solution
Note: This is not the full refactor, I'll mark it ready for review once it's done :)
From the issue:

[X]  Modify to field to make creates explicit
[X]  Error on non-4844 build if to is unset
[X]  Error on 4844 build if to is unset or create
[X] Add following to the builder trait
```
    - [X]  fn as_create(self) -> Self shortcut to set the field to Some(Create)

    - [X] fn deploy_code(self, code: Vec<u8>) -> Self to set data and to fields at once

    - [X] fn with_call<T: SolCall>(self, t: T) -> Self to set data and ensure to is NOT Create (set to to some if it is create)
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ X] Breaking changes


crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -460,6 +460,18 @@ impl From<TxEnvelope> for TransactionRequest {
...
+#[error("the field `to` is unset. Please populate it.")]
+#[non_exhaustive]
+pub struct TransactionRequestNon4844Error;
+
+/// Error thrown when `to` is `None` in non-EIP4844 transaction requests 
+#[derive(Debug, Default, thiserror::Error)]
+#[error("the field `to` can only be of type TxKind::Call(Account). Please change it accordingly.")]
+#[non_exhaustive]
+pub struct TransactionRequest4844Error;
+
```
prestwich, line 474: so far we have not introduced explicit error types for most tx request fields, we've been using strings for missing keys. So i'd be fine if these structs were omitted

crates/rpc-types/src/eth/transaction/request.rs
```diff
@@ -211,12 +212,14 @@ impl TransactionRequest {
     /// If required fields are missing. Use `complete_legacy` to check if the
     /// request can be built.
     fn build_legacy(self) -> TxLegacy {
+        let checked_to = self.to.expect("the `to` field should have value.");
```
prestwich, line 215: let's make these match the others `to: self.to.expect("checked in complete_legacy"),`

prestwich (changes requested):
looking good overall. need to add a `to` check to `fn check_reqd_fields` as well

prestwich:
mututally dependent with https://github.com/alloy-rs/core/pull/606

crates/contract/src/call.rs
```diff
@@ -252,12 +256,33 @@ impl<T: Transport + Clone, P: Provider<T, N>, N: Network> RawCallBuilder<T, P, N
...
+        Self::new_inner_call(provider, input, ())
+    }
+
+    /// Creates a new call builder with the provided provider and contract
+    /// deploy code.
+    ///
+    /// Will not decode the output of the call, meaning that
+    /// [`call`](Self::call) will behave the same as
+    /// [`call_raw`](Self::call_raw).
+    pub fn new_raw_deploy(provider: P, input: Bytes) -> Self {
```
DaniPopes, line 268: Can you add the sol! NOTE comment (see new_sol) ?

crates/provider/src/fillers/gas.rs
```diff
@@ -266,7 +266,7 @@ mod tests {
         let tx = TransactionRequest {
             from: Some(anvil.addresses()[0]),
             value: Some(U256::from(100)),
-            to: address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into(),
+            to: Some(address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into()),
```
gakonst, line 269: is the `.into()` required?

prestwich: yeah but its just tests so i didn't want to clutter the diff

crates/contract/src/call.rs
prestwich: 🖖

prestwich:
currently blocked by updating alloy-core to the patched version

gakonst:
Unblocked now https://github.com/alloy-rs/core/commit/9ff1e5a58be01f4846d839f558b08366f8ce7c0f

DaniPopes (approved):

PR #555, opened 2024-04-17
fix: signer fills from if unset

prestwich:
## Motivation

Unexpected behavior when both SignerFiller and NonceFiller are active causes nonces to not be filled properly, as from is unset. This kinda sucks

## Solution

- SignerFiller sets from if from is unset

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/signer.rs
+++ b/crates/provider/src/fillers/signer.rs
@@ -83,7 +83,7 @@ where
 
         if builder.from().is_none() {
             builder.set_from(self.signer.default_signer());
-            if !(builder.can_build()) {
+            if !builder.can_build() {
                 return Ok(SendableTx::Builder(builder));
             }
         }
--- a/crates/provider/src/fillers/signer.rs
+++ b/crates/provider/src/fillers/signer.rs
@@ -46,6 +46,10 @@ where
     type Fillable = ();
 
     fn status(&self, tx: &<N as Network>::TransactionRequest) -> FillerControlFlow {
+        if tx.from().is_none() {
+            return FillerControlFlow::Ready;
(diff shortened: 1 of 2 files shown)


crates/provider/src/fillers/signer.rs
```diff
@@ -72,11 +76,18 @@ where
...
     ) -> TransportResult<SendableTx<N>> {
-        let builder = match tx {
+        let mut builder = match tx {
             SendableTx::Builder(builder) => builder,
             _ => return Ok(tx),
         };
 
+        if builder.from().is_none() {
+            builder.set_from(self.signer.default_signer());
+            if !(builder.can_build()) {
```
onbjerg, line 86:
```suggestion
            if !builder.can_build() {
```
nit

onbjerg (approved):

mattsse (approved):

DaniPopes (approved):

crates/provider/src/fillers/signer.rs
```diff
@@ -46,6 +46,10 @@ where
     type Fillable = ();
 
     fn status(&self, tx: &<N as Network>::TransactionRequest) -> FillerControlFlow {
+        if tx.from().is_none() {
+            return FillerControlFlow::Ready;
+        }
+
```
gakonst, line 52: wait is this correct? shouldn't this be `is_some()`, "only skip signing by signaling ready if the from is set"

prestwich: if it is none, we are ready for the signer to fill its default

PR #558, opened 2024-04-17
fix: tmp fix for `PendingTransactionBuilder::get_receipt`

klkvr:
## Motivation

As discussed, try to fetch receipt before passing transaction to heartbeat

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
this should solve most issues,
there's still a race here if the heartbeat receives the block before the pending txs, but this is less likely, can tune this if we encounter issues

PR #560, opened 2024-04-17
chore: rename `<module>/mod.rs` into `<module>.rs`

RGafiyatullin:
## Motivation

[No more `mod.rs`](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html#no-more-modrs)

TLDR:
> ... This eliminates the special name, and if you have a bunch of files open in your editor, you can clearly see their names, instead of having a bunch of tabs named mod.rs. ...

## Solution

```shell
find . -iname mod.rs | while read f; do git mv $f $(echo $f | sed -e 's/\/mod//'); done
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

The test `rpc-types-trace::geth::tests::serde_default_frame` changed the location, so the path had to be adjusted (`../..` -> `..`).

Other than that nothing has been affected.


prestwich:
We are currently happy with our repo layout style. Closing as not planned.

PR #569, opened 2024-04-18
feature: WalletProvider

prestwich:
Adds a trait that allows users to access the `NetworkSigner` in a signing fill + provider stack.

## Motivation

Allow users to view keys and add signing credentials after building the provider stack. 

## Solution

Add a trait that exposes the underlying signer via borrows and mut borrows, and includes convenience functions for the useful `NetworkSigner` functions

Note: 
This trait is implemented only when both of the following are true:
- the signer is the `R` filler on the joinfill, 
- the fillprovider is the outermost layer

These conditions are true for all stacks constructed by the builder. This is a bit non-intuitive, but is necessary to avoid conflicting impls when both L and R are signer provider and when multiple signers are present in a stack. While neither of those cases are expected ever, the type system allows for them. Therefore we cannot neatly describe this trait without the assumptions above

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/wallet.rs
+++ b/crates/provider/src/wallet.rs
@@ -1,0 +1,72 @@
+use crate::{
+    fillers::{FillProvider, JoinFill, SignerFiller, TxFiller},
+    Provider,
+};
+use alloy_network::{Ethereum, Network, NetworkSigner};
+use alloy_primitives::Address;
+use alloy_transport::Transport;
+
+/// Trait for Providers, Fill stacks, etc, which contain [`NetworkSigner`].
+pub trait WalletProvider<N: Network = Ethereum> {
+    /// The underlying [`NetworkSigner`] type contained in this stack.
+    type Signer: NetworkSigner<N>;
+
+    /// Get a reference to the underlying signer.
+    fn signer(&self) -> &Self::Signer;
+
+    /// Get the default signer address.
+    fn default_signer(&self) -> Address {
+        self.signer().default_signer()
+    }
+
+    /// Check if the signer can sign for the given address.
+    fn is_signer_for(&self, address: &Address) -> bool {
+        self.signer().is_signer_for(address)
+    }
+
+    /// Get an iterator of all signer addresses.
+    fn signers(&self) -> impl Iterator<Item = Address> {
+        self.signer().signers()
+    }
+}
+
+impl<S, N> WalletProvider<N> for SignerFiller<S>
+where
+    S: NetworkSigner<N> + Clone,
+    N: Network,
+{
+    type Signer = S;
+
+    #[inline(always)]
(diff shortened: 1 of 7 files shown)


crates/provider/src/wallet.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Get a reference to the underlying signer.
+    fn signer(&self) -> &Self::Signer;
+
+    /// Get a mutable reference to the underlying signer.
+    fn signer_mut(&mut self) -> &mut Self::Signer;
+
+    /// Get the default signer address.
+    fn default_signer(&self) -> Address {
+        self.signer().default_signer()
+    }
```
mattsse, line 23: hmm, perhaps naming could be confusing here.

`default_signer` vs `signer` perhaps we should include `_address` in the fn name?

crates/provider/src/wallet.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Check if the signer can sign for the given address.
+    fn is_signer_for(&self, address: &Address) -> bool {
+        self.signer().is_signer_for(address)
+    }
+
+    /// Get an iterator of all signer addresses.
+    fn signers(&self) -> impl Iterator<Item = Address> {
```
mattsse, line 31: I'd expect an iterator over `Self::Signer` here, perhaps

`signer_addresses` or just addresses?

crates/provider/src/wallet.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn default_signer(&self) -> Address {
+        self.signer().default_signer()
+    }
+
+    /// Check if the signer can sign for the given address.
+    fn is_signer_for(&self, address: &Address) -> bool {
+        self.signer().is_signer_for(address)
+    }
+
+    /// Get an iterator of all signer addresses.
```
mattsse, line 30: should mention that this is non-empty because always includes the default signer address

mattsse (changes requested):
makes sense, only have some naming suggestions

crates/provider/src/wallet.rs
prestwich: I want to rename `NetworkSigner` to `NetworkWallet`, and `LocalWallet` to `LocalSigner`, and always use `Wallet` to mean "a collection of credentials" and signer to mean "a single credential". but that is a more extensive refactor

mattsse (approved):

crates/provider/src/wallet.rs
prestwich: made several renames 🖖

crates/provider/src/wallet.rs
prestwich: opened a new issue #575

PR #573, opened 2024-04-18
feat: add generics to filter, transaction, and pub_sub.

EmperorOrokuSaki:
## Motivation

In order to be able to use several structs and types from this repository for the Optimism port, they should take generics that can be replaced/used with the optimism types.

## Solution

Add generics to the necessary structs and types.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/block.rs
+++ b/crates/rpc-types/src/eth/block.rs
@@ -12,7 +12,7 @@ use std::{collections::BTreeMap, fmt, num::ParseIntError, ops::Deref, str::FromS
 /// Block representation
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Block<T = Transaction> {
+pub struct Block {
     /// Header of the block.
     #[serde(flatten)]
     pub header: Header,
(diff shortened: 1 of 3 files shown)


crates/rpc-types/src/eth/block.rs
```diff
@@ -160,11 +160,11 @@ impl Header {
 /// or if used by `eth_getUncle*`
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum BlockTransactions<T = Transaction> {
+pub enum BlockTransactions {
```
prestwich, line 163: why is this getting less generic?

EmperorOrokuSaki: Ah that was a mistake. Good catch

prestwich:
In favor of making the code reusable by other networks. At what point is this PR done? do we have a list of types we need to update?

mattsse:
> At what point is this PR done?

this process will take a few iterations because when adding new network specific types we'll discover more things that could be reused.

I think we will run into a few issues with types that have multiple network specific fields, like Transactions,Logs,Block,Header, for which we either need to add a bunch of generics or also do some kind of network trait with associated types...

perhaps, at some point it's just easier to simply duplicate the types but not sure if we want to

EmperorOrokuSaki:
The block struct also needs to take a generic for example but that means many other dependent structs will need the same thing later. Making Transaction a trait here would solve such issue for example.
So far for optimism rpc types everything needed to take a generic, takes it with this pr (except block as that depends on the trait discussion)

mattsse (changes requested):
these are reasonable,

I think we could also restrict this to `: Transaction`

mattsse (approved):

EmperorOrokuSaki:
> these are reasonable,
> 
> I think we could also restrict this to `: Transaction`

Just updated that. We can add the log generic later when needed probably

EmperorOrokuSaki:
Can we merge this? @mattsse @prestwich

PR #584, opened 2024-04-19
feat: add tx type bound to `Encodable2718` and `Decodable2718`

### LINKED ISSUE #540: [Refactor] Make En/Decodable2718 generic over a tx type
I want to incorporate a txtype enum into `En/Decodable2718`, and make something like `trait En/Decodable2718<T: Into<u8> + TryFrom<u8, Error = _>>` or similar so that we can enforce that the 2718 implementation uses the network `TxType` via a `type TxEnvelope: Encodable2718<Self::TxType>` bound 

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/431#discussion_r1563268235_

leruaa:
Closes #540


leruaa:
Adapting `AnyNetwork` lead to a few questions: I wonder if I should create  `AnyTxEnvelope` and `AnyTransaction` structs for `N::TxEnvelope` and `N::UnsignedTx` bounds?

prestwich:
> Adapting AnyNetwork lead to a few questions: I wonder if I should create AnyTxEnvelope and AnyTransaction structs for N::TxEnvelope and N::UnsignedTx bounds?

Yes, although each of those should be a separate issue + PR

prestwich:
After stewing on this for a month (sorry) I think this is currently over-abstraction. I'd want to re-approach this idea in the future, potentially as an assoc type instead of a generic trait, AFTER `AnyTxEnvelope` and `AnyTransaction`

PR #586, opened 2024-04-19
Add signature related `ConversionError` variants

tcoratger:
Add signature related `ConversionError` variants.

Related https://github.com/paradigmxyz/reth/pull/7752#discussion_r1572680180


mattsse (approved):
reasonable.

we should also consider adding something like `Other(String)`

PR #587, opened 2024-04-19
fix(provider): uncle methods for block hash

yash-atreya:
## Motivation

`get_uncle` and `get_uncle_count` failed when `BlockId::Hash` was passed as the tag as the `RpcBlockHash` was being passed to the params directly instead of only the actual `block_hash`.

## Solution

Pass actual `block_hash` and not the type.
Also, changed `idx: U64` -> `u64`. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -813,10 +813,12 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     }
 
     /// Gets an uncle block through the tag [BlockId] and index [U64].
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider.rs
```diff
@@ -813,10 +813,12 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     }
 
     /// Gets an uncle block through the tag [BlockId] and index [U64].
```
Evalir, line 815:
```suggestion
    /// Gets an uncle block through the tag [BlockId] and index [u64].
```

mattsse (approved):

PR #589, opened 2024-04-19
Temp get_uncle fix

yash-atreya:
## Motivation

Passing the primitive type for the `idx` leads to the following error:
```bash
server returned an error response: error code -32602: invalid argument 1: json: cannot unmarshal non-string into Go value of type hexutil.Uint
```
in the anvil integration test: https://github.com/yash-atreya/foundry/blob/cfeffd5f5cd7c4225ddd3f35ab1c2618648200db/crates/anvil/tests/it/fork.rs#L852

## Solution

Temp fix is to rollback to non-primitive type `U64`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider.rs
+++ b/crates/provider/src/provider.rs
@@ -814,6 +814,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
 
     /// Gets an uncle block through the tag [BlockId] and index [u64].
     async fn get_uncle(&self, tag: BlockId, idx: u64) -> TransportResult<Option<Block>> {
+        let idx = U64::from(idx);
         match tag {
             BlockId::Hash(hash) => {
                 self.client()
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider.rs
```diff
@@ -813,7 +813,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     }
 
     /// Gets an uncle block through the tag [BlockId] and index [u64].
-    async fn get_uncle(&self, tag: BlockId, idx: u64) -> TransportResult<Option<Block>> {
+    async fn get_uncle(&self, tag: BlockId, idx: U64) -> TransportResult<Option<Block>> {
```
DaniPopes, line 816: Please keep the public API the same and convert to U64 in the function

DaniPopes (approved):

Issue #583, opened 2024-04-19
[Feature] Add more builder style functions for large rpc args


mattsse:
### Component

rpc

### Describe the feature you would like

we're lacking convenience functions to configure large rpc args such as:

https://github.com/alloy-rs/alloy/blob/2e16c84df9040ffec8dcea809e5b1706e131d63a/crates/rpc-types-trace/src/geth/mod.rs#L486-L498

https://github.com/alloy-rs/alloy/blob/2e16c84df9040ffec8dcea809e5b1706e131d63a/crates/rpc-types-trace/src/filter.rs#L11-L11

## TODO

* add helper functions in builder style (`fn x(self, arg) -> Self`) that make it easier to set certain options

EmperorOrokuSaki:
I will take this.

PR #591, opened 2024-04-20
feat: add builder methods

EmperorOrokuSaki:
## Motivation
Issue #583 

## Solution
Implement builder methods for all fields

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/filter.rs
+++ b/crates/rpc-types-trace/src/filter.rs
@@ -33,6 +33,48 @@ pub struct TraceFilter {
 // === impl TraceFilter ===
 
 impl TraceFilter {
+    /// Sets the `from_block` field of the struct
+    pub fn set_from_block(&mut self, block: u64) -> &mut Self {
+        self.from_block = Some(block);
+        self
+    }
+
+    /// Sets the `to_block` field of the struct
+    pub fn set_to_block(&mut self, block: u64) -> &mut Self {
+        self.to_block = Some(block);
+        self
+    }
+
+    /// Sets the `from_address` field of the struct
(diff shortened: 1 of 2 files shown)


crates/rpc-types-trace/src/filter.rs
```diff
@@ -33,6 +33,48 @@ pub struct TraceFilter {
 // === impl TraceFilter ===
 
 impl TraceFilter {
+    /// Sets the `from_block` field of the struct
+    pub fn with_from_block(&mut self, block: u64) -> &mut Self {
```
mattsse, line 37: these should all consume `self` and return `Self`

crates/rpc-types-trace/src/filter.rs
mattsse, line 1: can drop the `with_` prefix, is cleaner imo

crates/rpc-types-trace/src/geth/mod.rs
```diff
@@ -497,6 +497,26 @@ pub struct GethDebugTracingCallOptions {
...
+    }
+
+    /// Enables block overrides
+    pub fn with_block_overrides(&mut self, overrides: BlockOverrides) -> &mut Self {
+        self.block_overrides = Some(overrides);
+        self
+    }
+
+    /// Sets the tracing options
+    pub fn with_tracing_options(&mut self, options: GethDebugTracingOptions) -> &mut Self {
```
mattsse, line 514: here we can keep the with_ but also change fro &mut self to self

mattsse (changes requested):
ty,
some nits

prestwich (approved):

Issue #595, opened 2024-04-21
[Bug] sol! generated event filters don't filter events


willmalcs:
### Component

contract

### What version of Alloy are you on?

alloy v0.1.0 (https://github.com/alloy-rs/alloy#eb0d564f)

### Operating System

macOS (Apple Silicon)

### Describe the bug

This example from https://github.com/alloy-rs/examples/blob/main/examples/subscriptions/examples/watch_contract_event.rs shows the issue quite well. It starts with an example contract put into a sol! macro:

```
sol!(
    #[allow(missing_docs)]
    #[sol(rpc, bytecode = "...")]
    contract Counter {
        int256 public counter = 0;

        event Increment(address indexed by, int256 indexed value);
        event Decrement(address indexed by, int256 indexed value);

        function increment() public {
            counter += 1;
            emit Increment(msg.sender, counter);
        }

        function decrement() public {
            counter -= 1;
            emit Decrement(msg.sender, counter);
        }
    }
);
```
generates Increment_filter() and Decrement_filter() functions:

```
// Create filters for each event.
let increment_filter = contract.Increment_filter().watch().await?;
let decrement_filter = contract.Decrement_filter().watch().await?;
```
the functions seem to imply that they should filter the watch / subscribe to only the particular event in the name of the function.

However, if you just watch one of the filters, e.g. the increment:

```
let increment_filter = contract.Increment_filter().watch().await?;
```

and make a call to emit the decrement event while monitoring the increment event:

```
// Call the decrement event
contract.decrement().send().await?;

// Listen for the events.
increment_filter
    .into_stream()
    .take(2)
    .for_each(|log| async {
        match log {
            Ok((_event, log)) => {
                println!("Received Increment: {log:?}");
            }
            Err(e) => {
                println!("Error: {e:?}");
            }
        }
    })
    .await;
```

You will hear the decrement event on the increment channel:

```
Received Increment: Log {
    inner: Log {
        address: 0x5fbdb2315678afecb367f032d93f642f64180aa3,
        data: LogData {
            topics: [
                0xdc69c403b972fc566a14058b3b18e1513da476de6ac475716e489fae0cbe4a26,
                0x000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb92266,
                0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe,
            ],
            data: 0x,
        },
    },
    block_hash: Some(
[...]

PR #600, opened 2024-04-22
fix: sol macro generated event filters were not filtering

willmalcs:
## Motivation

Fix for #595 

## Solution

changed:

```rust
#[doc(hidden)]
impl<'a, T: Transport + Clone, P: Provider<T, N>, E: SolEvent, N: Network> Event<T, &'a P, E, N> {
    // `sol!` macro constructor, see `#[sol(rpc)]`. Not public API.
    // NOTE: please avoid changing this function due to its use in the `sol!` macro.
    pub fn new_sol(provider: &'a P, address: &Address) -> Self {
        Self::new(provider, Filter::new().address(*address))
    }
}
```

to:

```rust
#[doc(hidden)]
impl<'a, T: Transport + Clone, P: Provider<T, N>, E: SolEvent, N: Network> Event<T, &'a P, E, N> {
    // `sol!` macro constructor, see `#[sol(rpc)]`. Not public API.
    // NOTE: please avoid changing this function due to its use in the `sol!` macro.
    pub fn new_sol(provider: &'a P, address: &Address) -> Self {
        // keccak256 hash of the event signature needed for the filter to actually filter by event
        // check that the event is not anonymous to include the event signature in the filter
        if E::ANONYMOUS {
            Self::new(provider, Filter::new().address(*address))
        } else {
            Self::new(provider, Filter::new().address(*address).event_signature(E::SIGNATURE_HASH))
        }
    }
}
```

This adds the event signature to the filter if the event is not anonymous meaning it will actually filter for that event.

The rest of the changes are just updates to the test at the bottom of the file so that the test will fail if filtering stops working again.

This update shouldn't introduce breaking changes — from the outside the function does not change at all, no parameter nor expected input changes; inside, the function just changes so that the event filtering does what it says on the tin.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

P.S. not sure if the breaking change checkbox is meant to be ticked if there are or aren't breaking changes

### DIFF
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -34,7 +34,13 @@ impl<'a, T: Transport + Clone, P: Provider<T, N>, E: SolEvent, N: Network> Event
     // `sol!` macro constructor, see `#[sol(rpc)]`. Not public API.
     // NOTE: please avoid changing this function due to its use in the `sol!` macro.
     pub fn new_sol(provider: &'a P, address: &Address) -> Self {
-        Self::new(provider, Filter::new().address(*address))
+        // keccak256 hash of the event signature needed for the filter to actually filter by event
+        // check that the event is not anonymous to include the event signature in the filter
+        if E::ANONYMOUS {
+            Self::new(provider, Filter::new().address(*address))
+        } else {
+            Self::new(provider, Filter::new().address(*address).event_signature(E::SIGNATURE_HASH))
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


willmalcs:
> Nice catch! Please format with `cargo +nightly fmt`.

thanks, have done now

crates/contract/src/event.rs
```diff
@@ -34,7 +34,13 @@ impl<'a, T: Transport + Clone, P: Provider<T, N>, E: SolEvent, N: Network> Event
     // `sol!` macro constructor, see `#[sol(rpc)]`. Not public API.
     // NOTE: please avoid changing this function due to its use in the `sol!` macro.
     pub fn new_sol(provider: &'a P, address: &Address) -> Self {
-        Self::new(provider, Filter::new().address(*address))
+        // keccak256 hash of the event signature needed for the filter to actually filter by event
+        // check that the event is not anonymous to include the event signature in the filter
+        if E::ANONYMOUS {
```
mattsse, line 39: nice,

mattsse (approved):
ty for the tests

PR #605, opened 2024-04-23
fix: continue reading ipc on data error

mattsse:
the json stream deserializer has issues with partial objects.

this continues the stream on data errors as long as the internal buffer does not exceed the configured capacity

### DIFF
--- a/crates/transport-ipc/src/lib.rs
+++ b/crates/transport-ipc/src/lib.rs
@@ -109,6 +109,9 @@ impl IpcBackend {
     }
 }
 
+/// Default capacity for the IPC buffer.
+const CAPACITY: usize = 4096;
+
 /// A stream of JSON-RPC items, read from an [`AsyncRead`] stream.
 #[derive(Debug)]
 #[pin_project::pin_project]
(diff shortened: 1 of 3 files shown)


prestwich (approved):

crates/transport-ipc/src/lib.rs
```diff
@@ -213,3 +230,53 @@ where
...
+
+    #[tokio::test]
+    async fn test_large_invalid() {
+        let mock = tokio_test::io::Builder::new()
+            // partial object
+            .read(b"{\"jsonrpc\":\"2.0\",\"method\":\"eth_subscription\"")
+            // trigger pending read
+            .wait(std::time::Duration::from_millis(1))
+            // complete object
+            .read(vec![b"a"[0]; CAPACITY].as_ref())
```
DaniPopes, line 269: `b'a'` :cat:

publicqi:
Is there a reason for setting CAPACITY 4096? Lots of RPC response can exceed this.

mattsse:
this is the initial, pre allocated capacity

publicqi:
Yeah but a partial response will hit 
```rust
if this.buf.len() > CAPACITY {
```

mattsse:
oh right, this is wrong of course -.-

PR #612, opened 2024-04-23
feat(engine): Add JSON Web Token (JWT) token generation and validation support

### LINKED ISSUE #599: [Feature] Add `JSON Web Token` (`JWT`) token generation and validation support
### Component

transports

To add `JWT` claims generation and validation support 

Besides having use in Foundry it is a common method of authentication between execution client / consensus client

For reference:

- https://github.com/gakonst/ethers-rs/blob/master/ethers-providers/src/rpc/transports/common.rs
- https://crates.io/crates/tower-jwt

Both implementations use: https://github.com/Keats/jsonwebtoken

### Additional context

It is possibly preferred that this is added to a separate new repo in the `alloy` org

@mattsse indicated that this is also useful for Reth

I should be able to implement it and propose a PR if that is desired

zerosnacks:
## Motivation

Closes #599 

## Solution

Implements JSON Web Token utilities in the `rpc-types-engine` crate. Only focuses on JWT token generation and validation, not the auth layer as the generated token is used as a string in bearer auth.

Largely uses the JWT secret implementation by Reth with some enhancements to documentation and using built-in `get_current_timestamp`. Notably does not implement loading from file, reading from file or writing to file as that is out of scope for Alloy I feel. 

Regarding the defined limitations on the algorithm and length, I've kept that 1:1 with the Reth implementation to make sure it can be used in that context and the limitations seem reasonable enough for a more general case.

See for reference: https://github.com/paradigmxyz/reth/blob/79452eadaf4963f1e8d78a18b1f490d7c560aa54/crates/rpc/rpc/src/layers/jwt_secret.rs

Proposed API:

```rust
let secret = JwtSecret::random(); or JwtSecret::from_hex(...)

let claims = Claims::default();
let jwt = secret.encode(&claims).unwrap();
let result = secret.validate(&jwt);

assert!(matches!(result, Ok(())));
```

Where `Claims::default()` is a claim defined as: `Claim { iat: get_current_timestamp(), exp: None }`

The `random()` syntax is consistent with other places that uses it (`Address::random()`, etc..)

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/jwt.rs
+++ b/crates/rpc-types-engine/src/jwt.rs
@@ -1,0 +1,81 @@
+use alloy_primitives::hex;
+use jsonwebtoken::{decode, errors::ErrorKind, Algorithm, DecodingKey, Validation};
+use rand::Rng;
+use std::time::Duration;
+use thiserror::Error;
+
+/// Errors returned by the [`JwtSecret`]
+#[derive(Error, Debug)]
+pub enum JwtError {
+    /// An error encountered while decoding the hexadecimal string for the JWT secret.
+    #[error(transparent)]
+    JwtSecretHexDecodeError(#[from] hex::FromHexError),
+
+    /// The JWT key length provided is invalid, expecting a specific length.
+    #[error("JWT key is expected to have a length of {0} digits. {1} digits key provided")]
+    InvalidLength(usize, usize),
+
+    /// The signature algorithm used in the JWT is not supported. Only HS256 is supported.
+    #[error("unsupported signature algorithm. Only HS256 is supported")]
+    UnsupportedSignatureAlgorithm,
+
+    /// The provided signature in the JWT is invalid.
+    #[error("provided signature is invalid")]
+    InvalidSignature,
+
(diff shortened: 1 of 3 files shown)


crates/rpc-types-engine/src/jwt.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    /// Validates a JWT token along the following rules:
+    /// - The JWT signature is valid.
+    /// - The JWT is signed with the `HMAC + SHA256 (HS256)` algorithm.
+    /// - The JWT `iat` (issued-at) claim is a timestamp within +-60 seconds from the current time.
+    /// - The JWT `exp` (expiration time) claim is validated by default if defined.
+    ///
+    /// See also: [JWT Claims - Engine API specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#jwt-claims)
+    pub fn validate(&self, jwt: String) -> Result<(), JwtError> {
```
mattsse, line 130: now that this part of a library we should use &str here

mattsse (changes requested):
smol nit

mattsse (approved):

PR #617, opened 2024-04-23
Make additional TxReceipt impls generic over T

### LINKED ISSUE #606: Make additional `TxReceipt` impls generic over T
Followup to #477

Make the following blocks generic n `consensus/src/receipt/receipt.rs`

- `impl TxReceipt for Receipt`
- `impl TxReceipt for ReceiptWithBloom`

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/477#discussion_r1576265813_

developeruche:
## Motivation
Closes #606 

## Solution
Just made the impl and extended the trait bound on `ReceiptWithBloom`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/consensus/src/receipt/receipts.rs
```diff
@@ -42,7 +42,10 @@ where
     }
 }
 
-impl TxReceipt for Receipt {
+impl<T> TxReceipt<T> for Receipt<T>
+where
+    T: Borrow<Log>,
```
gakonst, line 47: do we want this? thought we may want to be generic over this and do `T = Log` perhaps

developeruche:
```rust 

pub fn bloom_slow(&self) -> Bloom {
        self.logs.iter().map(Borrow::borrow).collect()
}

```

We made use of this here; 
We might have to change this implementation, thou `Log` works for most of the use-cases I can think of now.

DaniPopes (approved):

Evalir (approved):

onbjerg (approved):

crates/consensus/src/receipt/receipts.rs
prestwich: it's a bit more complicated and we will need more work here. Because `Log` is actually `Log<T = LogData>` we want to be able to implement generically across a wide variety of types, and may need to introduce a `trait Log` in order to do it

For now, this is an improvement over the existing state of things so we might as well

PR #621, opened 2024-04-23
fix: don't serialize nulls in tx request

DaniPopes:
Fixes https://github.com/foundry-rs/foundry/issues/7739


onbjerg (approved):
reasonable

PR #631, opened 2024-04-25
feat: add extra_fields to ChainConfig

fgimenez:
## Motivation

Towards https://github.com/paradigmxyz/reth/issues/7702

Currently fields specific to chains different from mainnet Ethereum (like those related to OP forks) are not being deserialized into ChainConfig. 

## Solution

Adds an `extra_fields` field to ChainConfig to gather the additional chain-specific information without polluting the top-level object. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -20,7 +20,7 @@
 
 extern crate alloc;
 
-use alloc::collections::BTreeMap;
+use alloc::{collections::BTreeMap, string::String};
 
 use alloy_primitives::{Address, Bytes, B256, U256};
 use alloy_serde::{
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -24,6 +24,7 @@ use alloc::collections::BTreeMap;
 
 use alloy_primitives::{Address, Bytes, B256, U256};
 use alloy_serde::{
+    json_value,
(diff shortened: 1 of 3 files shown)


crates/genesis/src/lib.rs
```diff
@@ -422,6 +423,10 @@ pub struct ChainConfig {
     /// Clique parameters.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clique: Option<CliqueConfig>,
+
+    /// Additional fields specific to each chain.
+    #[serde(flatten, default)]
+    pub extra_fields: BTreeMap<String, json_value>,
```
mattsse, line 429:
```suggestion
    pub extra_fields: BTreeMap<String, serde_json::Value>,
```
and move serde-json from dev to regular deps in cargo.toml

crates/serde/src/lib.rs
```diff
@@ -40,6 +40,8 @@ use serde::Serializer;
 
 use alloy_primitives::B256;
 
+pub use serde_json::Value as json_value;
```
mattsse, line 43:
```suggestion
```

mattsse (changes requested):
pedantic nit about re-exports

onbjerg (approved):

fgimenez:
> pedantic nit about re-exports

makes sense, done

Issue #634, opened 2024-04-25
[Bug] Panics From EIP1559 to Transaction request


loocapro:
### Component

rpc

### What version of Alloy are you on?

alloy-rpc-types v0.1.0

### Operating System

macOS (Apple Silicon)

### Describe the bug

```rust
    #[test]
    fn from_eip1559_to_tx_req() {
        let tx = TxEip1559 {
            chain_id: 1,
            nonce: 0,
            gas_limit: 21_000,
            to: TxKind::Call(Address::ZERO),
            max_priority_fee_per_gas: 20e9 as u128,
            max_fee_per_gas: 20e9 as u128,
            value: U256::from(1),
            ..Default::default()
        };
        let tx_req: TransactionRequest = tx.into();

        tx_req.build_unsigned().unwrap();
    }
```

This panics with: 

```
thread 'ethereum::builder::tests::from_eip1559_to_tx_req' panicked at crates/network/src/ethereum/builder.rs:204:33:

called `Result::unwrap()` on an `Err` value: 

(TransactionRequest { from: None, to: Some(Call(0x0000000000000000000000000000000000000000)), gas_price: None, max_fee_per_gas: Some(20000000000), max_priority_fee_per_gas: Some(20000000000), max_fee_per_blob_gas: None, gas: Some(21000), value: Some(0x0000000000000000000000000000000000000000000000000000000000000001_U256), input: TransactionInput { input: Some(0x), data: None }, nonce: Some(0), chain_id: Some(1), access_list: Some(AccessList([])), transaction_type: Some(2), blob_versioned_hashes: None, sidecar: None }, 

InvalidTransactionRequest(Eip2930, ["gas_price"]))
```

I believe the issue is here:

https://github.com/alloy-rs/alloy/blob/4d77029ba61284a6b4185f46581b4d9bce34f5fd/crates/rpc-types/src/eth/transaction/request.rs#L374-L374

This should also check that the EIP1559 gas fields are not set.

I ll open up a pr soon!

PR #635, opened 2024-04-25
fix: checking if the eip1559 gas fields are not set on eip2930 check

loocapro:
## Motivation
Fixes the bug documented in: https://github.com/alloy-rs/alloy/issues/634

## Solution
Added EIP1559 gas checks on the EIP2930 check.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/ethereum/builder.rs
+++ b/crates/network/src/ethereum/builder.rs
@@ -230,7 +230,7 @@ mod tests {
             .with_max_priority_fee_per_gas(0)
             .with_to(Address::ZERO)
             .with_gas_price(0)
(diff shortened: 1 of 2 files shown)


mattsse (commented):
the only other check that makes sense for 2930 is `gas_price.is_some && access_list.is_some`

his way we'd also fallback to 1559 if only accesslist is set

prestwich:
@loocapro I made a couple quick touchups, including fixing the `trim_conflicting_keys` function

prestwich (approved):

mattsse (approved):

Issue #613, opened 2024-04-23
[Feature] Change way we specify the `BlockId` in some methods of the provider


NicolasWent:
### Component

provider, pubsub

### Describe the feature you would like

In the following functions of the provider:

- [get_transaction_count](
https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L627)
- [get_balance](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L711)
- [get_storage_at](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L748)
- [get_code_at](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L758)
- [call](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L852)
- [call_with_overrides](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L862)
- [estimate_gas](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L884)
- [get_proof](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L935)
- [create_access_list](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L947)
- [trace_call](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L960)
- [trace_call_many](https://github.com/alloy-rs/alloy/blob/cad7935d69f433e45d190902e58b1c996b35adfa/crates/provider/src/provider.rs#L977)

I think it would more user friendly if we could change the way we specify the `BlockId` field.

For example for `get_transaction_count` in the latest block, we currently need to do:
```rust
provider.get_transaction_count(address, BlockId::latest()).await?;
```

[...]

PR #644, opened 2024-04-27
[Refactor] Move Provider into its own module

prestwich:
Simplify provider code layout, prep to address #613 

## Solution

- Break up provider file into modules 
- move extension APIs into `ext` module
- move wallet into provider

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
breaks existing imports so next release should be a minor bump, correct?

prestwich:
oh what breaks? i should fix re-exports

prestwich:
Oh the `ext` module. Yes. Those imports are broken

PR #645, opened 2024-04-27
feat: eth_call builder 

prestwich:
Depends on #644 

## Motivation

Partially addresses #613

## Solution

- Implement a basic builder returned by `Provider::call` that allows configuration of the 
- delete call_with_overrides
- remove block arg from `Provider::call`

If we like the pattern we can apply it to the other rpc reqs that take `BlockId` as optional param

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


prestwich:
cc @DaniPopes i don't love the amount of lifetime complexity i added to the call with decoder, but it should be invisible to users pretty much all the time i think?

prestwich:
Notes on why this level of complexity:

- `'client` -  we use `ClientRef` instead of `WeakClient` because `IntoFuture` impl would otherwise require upgrading inside `into_future()` in an awkward way
- `'req` - we borrow the request so that it can be reused for concurrent calls with other overrides, and for sending. this is also consistent with the provider interface
- `'state` - we borrow the state to avoid cloning it, as the clone may be expensive. this also allows it to be used with other concurrent calls. Future we may want to allow state to be a `RawValue` so that users can share it between calls and also avoid serializing it each time
- `'coder` - we borrow the coder to be consistent with behavior before this PR

prestwich:
okay I have removed the `'client` lifetime by making an intermediate future type that handles upgrading a `WeakClient`

crates/provider/src/provider/trait.rs
```diff
@@ -650,24 +650,14 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
     ///
     /// Not all client implementations support state overrides.
-    async fn call_with_overrides(
-        &self,
-        tx: &N::TransactionRequest,
-        block: BlockId,
-        state: StateOverride,
-    ) -> TransportResult<Bytes> {
-        self.client().request("eth_call", (tx, block, state)).await
+    fn call<'req>(&self, tx: &'req N::TransactionRequest) -> EthCall<'req, 'static, T, N> {
```
onbjerg, line 659: maybe add an example and also an alias for `call_with_overrides` so it comes up on docs.rs?

onbjerg (approved):
lgtm

onbjerg:
beatiful

PR #647, opened 2024-04-29
Correctly sign non legacy transaction without EIP155

alisinabh:
## Motivation

Correctly sign non-legacy transactions with y_parity value instead of EIP155 value.

## Solution

Only use with_chain_id method if transaction type is legacy.

This PR is in a PoC state and opened for discussion.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/legacy.rs
+++ b/crates/consensus/src/transaction/legacy.rs
@@ -1,4 +1,4 @@
-use crate::{SignableTransaction, Signed, Transaction};
+use crate::{SignableTransaction, Signed, Transaction, TxType};
 use alloy_primitives::{keccak256, Bytes, ChainId, Signature, TxKind, U256};
 use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable, Header, Result};
 use core::mem;
@@ -228,6 +228,10 @@ impl Transaction for TxLegacy {
 }
 
 impl SignableTransaction<Signature> for TxLegacy {
+    fn tx_type(&self) -> crate::TxType {
+        TxType::Legacy
+    }
+
     fn set_chain_id(&mut self, chain_id: ChainId) {
         self.chain_id = Some(chain_id);
     }
--- a/crates/consensus/src/transaction/legacy.rs
+++ b/crates/consensus/src/transaction/legacy.rs
@@ -1,4 +1,4 @@
-use crate::{SignableTransaction, Signed, Transaction, TxType};
+use crate::{SignableTransaction, Signed, Transaction};
 use alloy_primitives::{keccak256, Bytes, ChainId, Signature, TxKind, U256};
 use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable, Header, Result};
 use core::mem;
@@ -228,8 +228,8 @@ impl Transaction for TxLegacy {
 }
 
 impl SignableTransaction<Signature> for TxLegacy {
-    fn tx_type(&self) -> crate::TxType {
-        TxType::Legacy
+    fn use_eip155(&self) -> bool {
+        self.chain_id.is_some()
     }
 
     fn set_chain_id(&mut self, chain_id: ChainId) {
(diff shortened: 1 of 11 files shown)


DaniPopes:
cc @mattsse @prestwich , see also https://github.com/foundry-rs/foundry/pull/7804#issuecomment-2081520316

crates/consensus/src/transaction/mod.rs
```diff
@@ -84,6 +84,9 @@ pub trait SignableTransaction<Signature>: Transaction {
     /// RLP-encodes the transaction for signing.
     fn encode_for_signing(&self, out: &mut dyn alloy_rlp::BufMut);
 
+    /// Transaction Type
+    fn tx_type(&self) -> TxType;
```
prestwich, line 88: Unfortunately using `TxType` in this interface would require network paramterization. Instead I think we could make this function signature `fn use_eip155(&self) -> bool`

However, i'm also concerned about users who intentionally want a non-eip155 on a legacy tx (e.g. people using Nick's method to deploy deterministic contract addresses across networks). If eip-155 is always applied, they would be 

Maybe the long-term solution is to make an `Option<chain_id>` prop on the `LegacyTransaction`, just for signaling eip155 (non-)usage? That prop could be lost when RLPing, but that seems ok I guess? 🤔

prestwich: cc @mattsse

alisinabh: > Unfortunately using TxType in this interface would require network paramterization. Instead I think we could make this function signature fn use_eip155(&self) -> bool

That will work too. 
But generally because Alloy is meant to be used by other users (other than foundry) I think maybe adding a tx_type in the Transaction trait might even be the better option. 
Though I did not understand what is exactly meant by "network paramterization". Isn't the type already a part of transaction?

> Maybe the long-term solution is to make an Option<chain_id> prop on the LegacyTransaction, just for signaling eip155 (non-)usage? That prop could be lost when RLPing, but that seems ok I guess?

I agree. I think ganache also behaved like this. When there was no chain id value in a legacy transaction it would not do a EIP155 encoding.

alisinabh: Sorry to ping you guys. Any updates?

cc @mattsse

crates/consensus/src/transaction/legacy.rs
```diff
@@ -228,6 +228,10 @@ impl Transaction for TxLegacy {
 }
 
 impl SignableTransaction<Signature> for TxLegacy {
+    fn use_eip155(&self) -> bool {
+        self.chain_id.is_some()
```
alisinabh, line 232: This has the same issue that existed before! Having chain_id does not mean that EIP155 should be used. Having chain_id **in a LEGACY transaction** means EIP155 signature v value should be used.

alisinabh: cc @prestwich

crates/consensus/src/transaction/mod.rs
```diff
@@ -59,6 +59,11 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
 /// types, or in other networks. For example, in Optimism, the deposit transaction signature is the
 /// unit type `()`.
 pub trait SignableTransaction<Signature>: Transaction {
+    /// True if the transaction uses EIP-155 signatures.
+    fn use_eip155(&self) -> bool {
+        false
```
mattsse, line 64: should the default be true here?

prestwich: EIP155 signatures are disallowed for all types except legacy, no?

crates/consensus/src/transaction/legacy.rs
alisinabh: NVM, Didn't see that this is in legacy file.

prestwich: please read the `impl ... for TxLegacy` immediately above the line you commented on 🙃

crates/consensus/src/transaction/mod.rs
mattsse: right, misinterpreted the macro

PR #649, opened 2024-04-30
feat: add PayloadError variants

fgimenez:
Towards https://github.com/paradigmxyz/reth/pull/7993

## Motivation

During the checks for well-formed payload the existence of hardfork-specific fields is missing.

## Solution

Adds `PayloaError` variants to allow signaling failed checks for the existence of hardfork-specific fields. 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -722,9 +722,34 @@ pub enum PayloadError {
     /// Invalid payload excess blob gas.
     #[error("invalid payload excess blob gas: {0}")]
     ExcessBlobGas(U256),
+    /// Pre-shanghai Payload has withdrawals.
+    #[error("pre-Shanghai payload has withdrawals")]
+    PreShanghaiBlockWithWitdrawals,
+    /// Post-shanghai Payload does not have withdrawals.
+    #[error("post-Shanghai payload does not have withdrawals")]
+    PostShanghaiBlockWithoutWitdrawals,
     /// Pre-cancun Payload has blob transactions.
     #[error("pre-Cancun payload has blob transactions")]
     PreCancunBlockWithBlobTransactions,
+    /// Pre-cancun Payload has blob gas used.
+    #[error("pre-Cancun payload has blob gas used")]
+    PreCancunBlockWithBlobGasUsed,
+    /// Pre-cancun Payload has excess blob gas.
+    #[error("pre-Cancun payload has excess blob gas")]
+    PreCancunBlockWithExcessBlobGas,
+    /// Pre-cancun Payload has cancun fields.
+    #[error("pre-Cancun payload has Cancun fields")]
+    PreCancunWithCancunFields,
+    /// Post-cancun Payload does not have blob gas used.
+    #[error("post-Cancun payload does not have blob gas used")]
+    PostCancunBlockWithoutBlobGasUsed,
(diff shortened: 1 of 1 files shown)


mattsse:
ptal @Rjected

crates/rpc-types-engine/src/payload.rs
```diff
@@ -722,9 +722,37 @@ pub enum PayloadError {
...
+    #[error("pre-Cancun payload has excess blob gas")]
+    PreCancunBlockWithExcessBlobGas,
+    /// Pre-cancun Payload has cancun fields.
+    #[error("pre-Cancun payload has Cancun fields")]
+    PreCancunWithCancunFields,
+    /// Post-cancun Payload does not have blob transactions.
+    #[error("post-Cancun payload does not have blob transactions")]
+    PostCancunBlockWithoutBlobTransactions,
+    /// Post-cancun Payload does not have blob gas used.
+    #[error("post-Cancun payload does not have blob gas used")]
```
mattsse, line 747: hmm, these error msgs are not super clear imo, because this reads as if blob gas used does not exist post cancun.

any suggestions @Rjected ?

Rjected: Looks like the only reason why we need errors for each field is to make it more clear which field is missing, so we should note that a field is missing in the error.

maybe "blob gas used missing in post-cancun payload"

crates/rpc-types-engine/src/payload.rs
```diff
@@ -722,9 +722,37 @@ pub enum PayloadError {
...
     PreCancunBlockWithBlobTransactions,
+    /// blob gas used present in pre-cancun payload.
+    #[error("blob gas used present in pre-cancun payload")]
+    PreCancunBlockWithBlobGasUsed,
+    /// excess blob gas present in pre-cancun payload.
+    #[error("excess blob gas present in pre-cancun payload")]
+    PreCancunBlockWithExcessBlobGas,
+    /// cancun fields present in pre-cancun payload.
+    #[error("cancun fields present in pre-cancun payload")]
+    PreCancunWithCancunFields,
```
onbjerg, line 742: this seems unused in reth, do we need it?

onbjerg (changes requested):

crates/rpc-types-engine/src/payload.rs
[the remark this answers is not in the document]
onbjerg: sorry I meant to do that as a comment not a request changes:)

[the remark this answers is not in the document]
fgimenez: > what fields in here are not captured in the others?

yep, `ensure_well_formed_payload` receives a `cancun_fields` argument, of type `MaybeCancunPayloadFields`, which contains a field of type `Option<CancunPayloadFields>`,  `CancunPayloadFields` includes `parent_beacon_block_root` and `versioned_hashes`, see  https://github.com/paradigmxyz/reth/blob/04636400dae1bfb19dddbb3dcc848b163be4dfbd/crates/payload/validator/src/lib.rs#L107. 

The error here refers to `MaybeCancunPayloadFields` containing a `None` in its single field.

mattsse (approved):
lgtm

pending @onbjerg

onbjerg (approved):

PR #650, opened 2024-04-30
Add `AuthCall` variant to `CallType`

tcoratger:
Related https://eips.ethereum.org/EIPS/eip-3074


mattsse (approved):
reasonable, we definitely need this for 3074, so can already include the variant to experiment with it

PR #661, opened 2024-05-01
implement `From<B256>` for `JsonStorageKey`

tcoratger:


crates/serde/src/storage.rs
```diff
@@ -28,10 +28,16 @@ use serde::{Deserialize, Deserializer, Serialize};
...
+    fn from(value: B256) -> Self {
+        JsonStorageKey(value)
+    }
+}
+
 impl From<U256> for JsonStorageKey {
     fn from(value: U256) -> Self {
         // SAFETY: Address (B256) and U256 have the same number of bytes
-        JsonStorageKey(B256::from(value.to_be_bytes()))
+        B256::from(value.to_be_bytes()).into()
```
prestwich, line 40: there is a From<U256> for B256

drive-by would be

From<[u8;32]> for JsonStorageKey

prestwich (changes requested):

crates/serde/src/storage.rs
tcoratger: Just fixed, let me know if it fits your request

prestwich (approved):
approving as its a strict improvement. 

cc @mattsse followup wuestion is this type still necessary after geth fixed their weird spec issues?

tcoratger:
> approving as its a strict improvement.
> 
> cc @mattsse followup wuestion is this type still necessary after geth fixed their weird spec issues?

Used for stuffs like `eth_getStorageAt` to know the index or `eth_getProof` no?

prestwich:
> > approving as its a strict improvement.
> 
> > 
> 
> > cc @mattsse followup wuestion is this type still necessary after geth fixed their weird spec issues?
> 
> 
> 
> Used for stuffs like `eth_getStorageAt` to know the index or `eth_getProof` no?

it was used for geth's offspec input mirroring, which has since been fixed. geth now returns minimal quantity encoded keys, iirc

mattsse (approved):
@Rjected can we deprecate this type then?
or at least where we use it in eth_getStorage for example

PR #662, opened 2024-05-01
Implement `TryFrom<Transaction>` for `TransactionInfo`

tcoratger:


crates/rpc-types/src/eth/transaction/common.rs
```diff
@@ -17,3 +20,26 @@ pub struct TransactionInfo {
...
+
+    fn try_from(tx: Transaction) -> Result<Self, Self::Error> {
+        Ok(TransactionInfo {
+            hash: Some(tx.hash),
+            index: tx.transaction_index,
+            block_hash: tx.block_hash,
+            block_number: tx.block_number,
+            base_fee: match tx.transaction_type {
+                Some(tx_type) => match tx_type.try_into() {
+                    Ok(TxType::Legacy) | Ok(TxType::Eip2930) => tx.gas_price,
```
prestwich, line 35: is this behavior correct? wouldn't it be the block base fee regardless of tx content?

crates/rpc-types/src/eth/transaction/common.rs
```diff
@@ -1,6 +1,9 @@
 //! Commonly used additional types that are not part of the JSON RPC spec but are often required
 //! when working with RPC types, such as [Transaction](crate::Transaction)
```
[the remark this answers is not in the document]
prestwich: for what exactly is this type used? a search of the codebase only shows it re-exported

crates/rpc-types/src/eth/transaction/common.rs
tcoratger: I would say yes because it is precisely in the situation where we do not have information on the block but only on the tx. So we try to reconstruct the base fee as a function of the transaction type with the available information.

tcoratger: Maybe we can just put a `None` to say that we don't know precisely what is the base fee as we just have information about the transaction, wdyt?

prestwich: i would prefer that, as it guarantees we give no bad output. However, i don't know what this type is for so i can't say that the incorrect output is not expected 😅  What is the motivating use case of this PR?

tcoratger: In fact I realized that in tracing to trace a transaction, we quite often do the conversion `Transaction` to `TransactionInfo` and so I said to myself that it was useful to have a `try_from` here but it's true that we can set to `None` and then fill with the base fee coming from the env block.

Just pushed that, tell me if it's ok for you :)

prestwich: > In fact I realized that in tracing to trace a transaction, we quite often do the conversion Transaction to TransactionInfo and so I said to myself that it was useful to have a try_from here but it's true that we can set to None and then fill with the base fee coming from the env block.

Gotcha. makes sense. I think that now we have no errors tho, so we can make this `From<&Transaction>` for infallible conversion :)

tcoratger: ouuups just fixed

crates/rpc-types/src/eth/transaction/common.rs
```diff
@@ -17,3 +18,17 @@ pub struct TransactionInfo {
     /// Base fee of the block.
     pub base_fee: Option<u128>,
 }
+
+impl From<Transaction> for TransactionInfo {
+    fn from(tx: Transaction) -> Self {
```
prestwich, line 23: since this is small, could we also do `From<&Transaction>` or do we always want to clone?

tcoratger: Yes you are right, in this case I think `From<&Transaction>` is better and `From<Transaction>` would be useless, just fixed

prestwich (approved):

PR #666, opened 2024-05-02
fix: more abstraction for block transactions

### LINKED ISSUE #665: [Bug] Full BlockTransactions doesn't offer access to Transaction slice
### Component

rpc

_No response_

### Operating System

None

### Describe the bug

Hello sirs.

Wonderful job!

Filling the above. Basically, it's impossible to access the _Full_ transaction field from the BlockTransactionHashesInner object as the only exposed API will collapse a _Full_ transaction into its transaction hash. I would classify this as a bug and not a feature request as it is quite misleading, but happy to be corrected, mostly nomenclature to be honest. It does seem erroneous to download full payloads only to be offered access to the transaction hashes, which are returned in default mode anyways.

Obviously, something might have escaped me from the API, which might deprecate this bug report. Happy to stand corrected!

prestwich:
## Motivation

Closes #665, lays groundwork for re-use of `BlockTransactions` by other networks

## Solution

Propagate generic over T wherever possible, add fallible casting convenience methods
includes a drive-by doctest fix

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -675,7 +675,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     /// // Execute a call on the latest block, with no state overrides
     /// let output = provider.call(&tx).await?;
     /// // Execute a call with a block ID.
-    /// let output = provider.call(&tx).block(BlockId::Number(1.into())).await?;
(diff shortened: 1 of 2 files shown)


prestwich:
necessary followup work: 
- decide how to bound the `N::TransactionResponse` type to allow access to the hash

onbjerg (approved):

PR #668, opened 2024-05-02
feat: add eip-7685 requests root to header

onbjerg:
Adds EIP-7685 header fields to both the consensus and RPC `Header` types.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/header.rs
+++ b/crates/consensus/src/header.rs
@@ -89,8 +89,8 @@ pub struct Header {
     ///
     /// The beacon roots contract handles root storage, enhancing Ethereum's functionalities.
     pub parent_beacon_block_root: Option<B256>,
-    /// The Keccak 256-bit hash of the root node of the trie structure populated with each [EIP-7685]
-    /// request in the block body.
+    /// The Keccak 256-bit hash of the root node of the trie structure populated with each
+    /// [EIP-7685] request in the block body.
     ///
     /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
     pub requests_root: Option<B256>,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
all of this looks correct, this basically shifts replaces the parent beacon root as the last entry

pending @Rjected 

I assume we also need additional checks in the consensus impl?

onbjerg:
@mattsse yeah i'm opening PRs as I progress on https://github.com/paradigmxyz/reth/pull/8053, we need to add some fields to body + calc the root itself, will keep the alloy impl on par

prestwich:
should we be managing changes in e.g. an enum over versions?

mattsse:
> should we be managing changes in e.g. an enum over versions?

while this would have advantages, we'd still need this type for decoding any rlp blob or we need additional context for decoding, like the hardfork so we know how to decode

prestwich:
great i love it when field semantics change without any structural indication 🎉

crates/consensus/src/header.rs
```diff
@@ -285,15 +291,22 @@ impl Header {
...
         // need to be repeated and placeholder length added. Otherwise, it's impossible to
         // tell _which_ fields are missing. This is mainly relevant for contrived cases
         // where a header is created at random, for example:
         //  * A header is created with a withdrawals root, but no base fee. Shanghai blocks are
         //    post-London, so this is technically not valid. However, a tool like proptest would
         //    generate a block like this.
-        if let Some(parent_beacon_block_root) = self.parent_beacon_block_root {
-            length += parent_beacon_block_root.length();
+        if let Some(requests_root) = self.requests_root {
+            length += requests_root.length();
```
Rjected, line 309: note: the hack in this comment is not needed in reth any more, the arbitrary impl was changed instead, we should copy it over so the encoding / decoding impl is more sensible

Rjected (approved):
lgtm

crates/consensus/src/header.rs
onbjerg: will do

PR #674, opened 2024-05-03
RpcWithBlock

### LINKED ISSUE #613: [Feature] Change way we specify the `BlockId` in some methods of the provider

prestwich:
## Motivation

followup to #645 
closes #613 

## Solution

Generic builder struct for rpc calls that take optional block id parameters

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -394,7 +394,7 @@ impl<T: Transport + Clone, P: Provider<T, N>, D: CallDecoder, N: Network> CallBu
 
     /// Returns the estimated gas cost for the underlying transaction to be executed
     pub async fn estimate_gas(&self) -> Result<u128> {
-        self.provider.estimate_gas(&self.request, self.block).await.map_err(Into::into)
+        self.provider.estimate_gas(&self.request).block(self.block).await.map_err(Into::into)
     }
 
     /// Queries the blockchain via an `eth_call` without submitting a transaction to the network.
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -394,7 +394,7 @@ impl<T: Transport + Clone, P: Provider<T, N>, D: CallDecoder, N: Network> CallBu
 
     /// Returns the estimated gas cost for the underlying transaction to be executed
     pub async fn estimate_gas(&self) -> Result<u128> {
-        self.provider.estimate_gas(&self.request).block(self.block).await.map_err(Into::into)
+        self.provider.estimate_gas(&self.request).block_id(self.block).await.map_err(Into::into)
     }
 
     /// Queries the blockchain via an `eth_call` without submitting a transaction to the network.
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -40,6 +40,7 @@ tokio = { workspace = true, features = ["sync", "macros"] }
 tracing.workspace = true
 url = { workspace = true, optional = true }
 futures-utils-wasm.workspace = true
+pin-project.workspace = true
 
 [dev-dependencies]
 alloy-consensus = { workspace = true, features = ["std"] }
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -1,3 +1,5 @@
+use std::future::IntoFuture;
+
 use crate::{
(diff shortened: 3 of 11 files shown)


prestwich:
I can't help but feel like the bounds could be simpler

prestwich:
> I can't help but feel like the bounds could be simpler

I was right. they are now simpler. Unpin and Clone bounds are removed

PR #677, opened 2024-05-03
chore: move blob validation to sidecar

### LINKED ISSUE #675: [Feature] Move kzg validation to blobtransaction sidecar
### Component

consensus, eips, genesis

currently blob validation is part of the 4844 tx:

https://github.com/alloy-rs/alloy/blob/44b8a6da656602fd9fd2c52f28ecffa4acaed96c/crates/consensus/src/transaction/eip4844.rs#L436-L440

it would be more useful to have the impl as a function on the sidecar: 

```rust
pub fn validate(
        &self,
        versioned_hashes: &[B256],
        proof_settings: &KzgSettings,
    ) -> Result<(), BlobTransactionValidationError> {

```

this way it becomes reusable

the tx4844 impl then simply calls this will its versioned hashes and the given settings.

### Additional context

_No response_

mattsse:
closes #675

this makes it easier to reuse blob validation, for example in reth.

the `MissingSidecar` error variant no longer fits, but it's also not worth rolling an additional error type for this case imo

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -18,7 +18,6 @@ alloy-eips.workspace = true
 alloy-serde = { workspace = true, optional = true }
 
 # kzg
-thiserror = { workspace = true, optional = true }
 c-kzg = { workspace = true, features = ["serde"], optional = true }
 
 # arbitrary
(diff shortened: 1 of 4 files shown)


onbjerg (approved):
lgtm

PR #678, opened 2024-05-03
fix(provider): `get_blob_base_fee` map to primitive u128

yash-atreya:
## Motivation

Doesn't map to primitive, results in deser error

## Solution

Map to primitive `u128`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
great!

PR #684, opened 2024-05-04
move beacon API types from github.com/paradigmxyz/reth to a new crate in alloy

rupam-04:
## Motivation
https://github.com/paradigmxyz/reth/issues/8094

## Solution
Moved the ```beacon``` crate to ```rpc-types-beacon```

## PR Checklist
- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-beacon/Cargo.toml
+++ b/crates/rpc-types-beacon/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "rpc-types-beacon"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
+
+[dependencies]
+# ethereum
+alloy-rpc-types.workspace = true
+alloy-rpc-types-trace.workspace = true
+alloy-rpc-types-anvil.workspace = true
+alloy-rpc-types-engine.workspace = true
+alloy-primitives = "0.7.2"
+serde = "1.0.200"
+serde_with = "3.8.1"
+serde_json = "1.0.116"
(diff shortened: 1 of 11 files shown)


crates/rpc-types-beacon/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+[dependencies]
+# ethereum
+alloy-rpc-types.workspace = true
+alloy-rpc-types-trace.workspace = true
+alloy-rpc-types-anvil.workspace = true
+alloy-rpc-types-engine.workspace = true
+alloy-primitives = "0.7.2"
+serde = "1.0.200"
+serde_with = "3.8.1"
+serde_json = "1.0.116"
```
mattsse, line 21: these need to use .workspace

mattsse (changes requested):
smol nit

@prestwich imo beacon API types are useful and there are more coming, for example for sidecars https://github.com/paradigmxyz/reth/pull/7443

they are closely related to EL types (engines) but for some reason the CL uses different formatting rules...

I'd prefer having this crate in this repo because we want them to be compatible with engine and other types

wdyt

DaniPopes (approved):

rupam-04:
@mattsse

prestwich:
> smol nit
> 
> @prestwich imo beacon API types are useful and there are more coming, for example for sidecars [paradigmxyz/reth#7443](https://github.com/paradigmxyz/reth/pull/7443)
> 
> they are closely related to EL types (engines) but for some reason the CL uses different formatting rules...
> 
> I'd prefer having this crate in this repo because we want them to be compatible with engine and other types
> 
> wdyt

yeah, works for me

mattsse (approved):

mattsse (approved):

PR #696, opened 2024-05-06
feat: manual blob deserialize

mattsse:
should prevent stackoverflows on sidecar deserde, ref https://github.com/foundry-rs/foundry/actions/runs/8968170602/job/24627087079?pr=7868

I assume this could be improved further by implementing a custom visitor, but will followup separately

### DIFF
--- a/crates/eips/src/eip4844/sidecar.rs
+++ b/crates/eips/src/eip4844/sidecar.rs
@@ -21,6 +21,7 @@ use alloc::vec::Vec;
 )]
 pub struct BlobTransactionSidecar {
     /// The blob data.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "deserialize_blobs"))]
     pub blobs: Vec<Blob>,
     /// The blob commitments.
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip4844/sidecar.rs
```diff
@@ -21,6 +21,7 @@ use alloc::vec::Vec;
 )]
 pub struct BlobTransactionSidecar {
     /// The blob data.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "deserialize_blobs"))]
```
DaniPopes, line 24: + `debug_assertions`, this is not a problem with opt-level > 0

DaniPopes: cfg_attr(all(feature,debug_assertions)

mattsse: ah gotcha,

DaniPopes (approved):

PR #704, opened 2024-05-07
feat: add eip-7685 enc/decode traits

onbjerg:
## Motivation

[EIP-7685](https://eips.ethereum.org/EIPS/eip-7685) adds an EL → CL request list. The list items are requests, which are encoded as `type ++ request_data`, where `request_data` is an opaque byte buffer. The meaning of the bytes is up to each request, which are defined in separate EIPs.

## Solution

Add traits for encoding and decoding requests. This follows the same overall design as our traits for EIP-2718, with the difference being that EIP-2718 also has special handling for legacy transactions.

The expected usage is to define something a la `TxEnvelope`, e.g. `RequestsEnvelope`, which is an enum that internally holds each legal request type. The requests themselves are decoded and encoded according to their own EIPs, e.g. with RLP.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7685.rs
+++ b/crates/eips/src/eip7685.rs
@@ -17,6 +17,7 @@ use core::{
 ///
 /// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
 #[derive(Clone, Copy, Debug)]
+#[non_exhaustive]
 pub enum Eip7685Error {
     /// Rlp error from [`alloy_rlp`].
     RlpError(alloy_rlp::Error),
--- a/crates/eips/src/eip7685.rs
+++ b/crates/eips/src/eip7685.rs
@@ -1,0 +1,112 @@
+//! [EIP-7685]: General purpose execution layer requests
+//!
+//! Contains traits for encoding and decoding EIP-7685 requests, as well as validation functions.
+//!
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
can we add a combined trait for this as well?

https://github.com/alloy-rs/alloy/blob/77c1240533b411ed0eb5533f94396eba8d7f6ab6/crates/eips/src/eip2718.rs#L189-L190

mattsse (approved):
imo it makes sense to roll a dedicated trait for this.

crates/eips/src/eip7685.rs
```diff
@@ -0,0 +1,123 @@
...
+/// [EIP-7685]: https://eips.ethereum.org/EIPS/eip-7685
+#[derive(Clone, Copy, Debug)]
+pub enum Eip7685Error {
+    /// Rlp error from [`alloy_rlp`].
+    RlpError(alloy_rlp::Error),
+    /// Got an unexpected request type while decoding.
+    UnexpectedType(u8),
+    /// There was no request type in the buffer.
+    MissingType,
+}
```
prestwich, line 27: maybe non_exhaustive or custom variant?

onbjerg: good idea

PR #707, opened 2024-05-07
feat: add eth mainnet EL requests envelope

onbjerg:
Adds a 7685 envelope for mainnet EL request types. The current ones for Prague are EIP-6110 deposits and EIP-7002 withdrawals.

Depends on #704 and #705

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/request.rs
+++ b/crates/consensus/src/request.rs
@@ -21,6 +21,18 @@ pub enum Request {
     WithdrawalRequest(WithdrawalRequest),
 }
 
+impl From<DepositRequest> for Request {
+    fn from(v: DepositRequest) -> Self {
+        Self::DepositRequest(v)
+    }
+}
+
+impl From<WithdrawalRequest> for Request {
+    fn from(v: WithdrawalRequest) -> Self {
+        Self::WithdrawalRequest(v)
+    }
+}
+
 impl Encodable7685 for Request {
     fn request_type(&self) -> u8 {
         match self {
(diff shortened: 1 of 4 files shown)


crates/consensus/src/request.rs
```diff
@@ -0,0 +1,46 @@
...
+    eip6110::DepositRequest,
+    eip7002::WithdrawalRequest,
+    eip7685::{Decodable7685, Eip7685Error, Encodable7685},
+};
+use alloy_rlp::{Decodable, Encodable};
+
+/// Ethereum execution layer requests.
+///
+/// See also [EIP-7685](https://eips.ethereum.org/EIPS/eip-7685).
+pub enum Request {
```
mattsse, line 11: derives?

crates/consensus/src/request.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// An [EIP-6110] deposit request.
+    ///
+    /// [EIP-6110]: https://eips.ethereum.org/EIPS/eip-6110
+    DepositRequest(DepositRequest),
+    /// An [EIP-7002] withdrawal request.
+    ///
+    /// [EIP-7002]: https://eips.ethereum.org/EIPS/eip-7002
+    WithdrawalRequest(WithdrawalRequest),
```
mattsse, line 19: do we need helper functions like `is_` `as_`?

prestwich: i like that as library convention

crates/consensus/src/request.rs
onbjerg: yeah sorry i should've marked this as a draft, not done with it

mattsse (approved):

PR #709, opened 2024-05-07
feature: beacon sidecar types

JackG-eth:
New beacon type for EIP4844 Transactions.

## Motivation

Beacon types have been moved from reth to alloy, this is a new type that will be needed for users to be able to subscribe to a stream of blob transactions.

it will also unblock https://github.com/paradigmxyz/reth/pull/7443


crates/rpc-types-beacon/src/beacon/sidecar.rs
```diff
@@ -0,0 +1,169 @@
...
+}
+
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct BlobSidecar {
+    #[serde_as(as = "DisplayFromStr")]
+    pub index: u64,
+    pub blob: String,
+    pub kzg_commitment: Bytes,
+    pub kzg_proof: Bytes,
```
mattsse, line 18: these should be Bytes48

crates/rpc-types-beacon/src/beacon/sidecar.rs
```diff
@@ -0,0 +1,169 @@
...
+pub struct BeaconBlobBundle {
+    pub data: Vec<BlobSidecar>,
+}
+
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct BlobSidecar {
+    #[serde_as(as = "DisplayFromStr")]
+    pub index: u64,
+    pub blob: String,
```
mattsse, line 16:
```suggestion
    pub blob: Blob,
```

mattsse (changes requested):
smol nits

crates/rpc-types-beacon/src/beacon/sidecar.rs
```diff
@@ -0,0 +1,90 @@
...
+#[cfg(test)]
+mod tests {
+    use std::{fs, path::PathBuf};
+
+    use super::*;
+
+    fn read_blob_json() -> String {
+        let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        path.push("examples/sidecar.json");
+        fs::read_to_string(path).expect("Unable to read the JSON file")
```
mattsse, line 77: we can replace this with include_str!

mattsse (changes requested):

mattsse (approved):

PR #719, opened 2024-05-10
fix: required fields for transactions and receipts

banteg:
## Motivation

this reverts incorrect changes introduced in https://github.com/alloy-rs/alloy/pull/523

in transaction info (for example `eth_getTransactionByHash`), these fields [are required](https://github.com/ethereum/execution-apis/blob/c4e7da7e486f6cb6ff6281c3ff28bb37445d5415/src/schemas/transaction.yaml#L334-L339)

```yaml
required:
  - blockHash
  - blockNumber
  - from
  - hash
  - transactionIndex
```

in transaction receipts, there fields [are required](https://github.com/ethereum/execution-apis/blob/c4e7da7e486f6cb6ff6281c3ff28bb37445d5415/src/schemas/receipt.yaml#L40-L50).

```yaml
required:
  - blockHash
  - blockNumber
  - from
  - cumulativeGasUsed
  - gasUsed
  - logs
  - logsBloom
  - transactionHash
  - transactionIndex
  - effectiveGasPrice
```

moreover, the original justification is also incorrect, as the author has misunderstood how geth [does the serialization](https://github.com/ethereum/go-ethereum/blob/74edc9386495f0ff44d68bf6e237a69f27021144/internal/ethapi/api.go#L1704-L1721). this screencap is freshly out of `geth --dev`. as you can see, the optional fields are present, and the linked code hasn't changed since 2019.

image

## Solution

i removed `skip_serializing_if` for required fields.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/transaction/mod.rs
+++ b/crates/rpc-types/src/eth/transaction/mod.rs
@@ -41,21 +41,13 @@ pub struct Transaction {
     #[serde(with = "alloy_serde::num::u64_via_ruint")]
     pub nonce: u64,
     /// Block hash
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(default)]
     pub block_hash: Option<B256>,
     /// Block number
-    #[serde(
-        default,
-        skip_serializing_if = "Option::is_none",
-        with = "alloy_serde::num::u64_opt_via_ruint"
-    )]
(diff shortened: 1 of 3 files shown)


mattsse (approved):

prestwich:
if the fees are required, then removing the serializing, if is not sufficient. Need to also make the prop not an Option type

@banteg

PR #720, opened 2024-05-10
fix: dont skip block context

mattsse:
ref https://github.com/alloy-rs/alloy/pull/523#issuecomment-2103693743

cc @banteg


banteg:
already made a pr here https://github.com/alloy-rs/alloy/pull/719

mattsse:
sweet, ty

PR #721, opened 2024-05-10
add `with_base_fee` for `TransactionInfo`

tcoratger:
For situation where `TransactionInfo` is built from `Transaction`, `base_fee` field is set to `None`. As a result, it is interesting to have a `with_base_fee` method implemented in order to consume the created `TransactionInfo` instance and populate its `base_fee` field.

### DIFF
--- a/crates/rpc-types/src/eth/transaction/common.rs
+++ b/crates/rpc-types/src/eth/transaction/common.rs
@@ -19,6 +19,14 @@ pub struct TransactionInfo {
     pub base_fee: Option<u128>,
 }
 
+impl TransactionInfo {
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
supportive, ~~while we're at it, can we also add `excess_blob_gas: Option<u64>` here. this is needed to calc the blob gas price~~

pedantic clippy

mattsse (approved):

tcoratger:
> supportive, ~while we're at it, can we also add `excess_blob_gas: Option<u64>` here. this is needed to calc the blob gas price~
> 
> pedantic clippy

Fixed :)

PR #725, opened 2024-05-11
refactor: rm `into_request`

tcoratger:
The `into_request` function that transforms a rpc `Transaction` to a rpc `TransactionRequest` was redundant and less Rust idiomatic than the `From<Transaction> for TransactionRequest` implementation. 

As a result, the `into_request` function is removed and its content is migrated to `From<Transaction> for TransactionRequest` implementation.


prestwich (approved):

prestwich:
rebase after #724

Evalir (approved):

DaniPopes (commented):
I don't really like this type of changes, since it requires importing the other type to convert to it either with .into or ::from.

prestwich:
i'm fine closing it if we don't like it

mattsse:
agree with @DaniPopes 

I prefer separate into calls,

PR #740, opened 2024-05-13
feat: add Parlia genesis config for BSC

forcodedancing:
## Motivation

BSC uses Parlia consensus engine. To support BSC genesis, Parlia config is added.

## Solution
Update genesis config to support Parlia.

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -423,6 +423,10 @@ pub struct ChainConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clique: Option<CliqueConfig>,
 
+    /// Parlia parameters.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub parlia: Option<ParliaConfig>,
+
     /// Additional fields specific to each chain.
     #[serde(flatten, default)]
     pub extra_fields: BTreeMap<String, serde_json::Value>,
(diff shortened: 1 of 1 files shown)


crates/genesis/src/lib.rs
```diff
@@ -423,6 +423,10 @@ pub struct ChainConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub clique: Option<CliqueConfig>,
 
+    /// Parlia parameters.
```
mattsse, line 426: can we add an url to to a spec if that exists?

forcodedancing: thanks a lot @mattsse 
The introduction and code specification are added.

prestwich:
given that the config props are identical to clique config, should we reuse the struct?

forcodedancing:
> given that the config props are identical to clique config, should we reuse the struct?

You are right. The fields are the same. However, the meanings of the fields are different for clique and parlia. 

I am ok if you think reuse is better. thanks a lot.

prestwich:
> You are right. The fields are the same. However, the meanings of the fields are different for clique and parlia.

Ya, I understand that in Parlia the fields are used differently than in clique. But given that the names and comments are the same it seems safe to reuse. I don't have a strong opinion here, so I'm fine leaving it as 2 identical structs

crates/genesis/src/lib.rs
```diff
@@ -541,6 +545,21 @@ pub struct CliqueConfig {
     pub epoch: Option<u64>,
 }
 
+/// Consensus configuration for Parlia.
+/// Parlia is the consensus engine for BNB Smart Chain.
+/// For the general introduction: https://docs.bnbchain.org/docs/learn/consensus/
```
prestwich, line 550: doc nit again: 
for links you should follow recommendation here:

https://doc.rust-lang.org/rustdoc/lints.html?highlight=http%3A%2F%2F#bare_urls

DaniPopes: https://github.com/alloy-rs/alloy/actions/runs/9062315958/job/24951891283?pr=740

forcodedancing: fixed, thanks a lot

mattsse (approved):
this doesn't have any impact on how genesis is used for eth and alike so I'm fine with adding this field out of convenience

prestwich (approved):

PR #743, opened 2024-05-13
fix: make eip-7685 req untagged

onbjerg:
Self explanatory I think


crates/consensus/src/request.rs
```diff
@@ -15,6 +15,7 @@ use alloy_rlp::{Decodable, Encodable, Header};
     derive(proptest_derive::Arbitrary, arbitrary::Arbitrary)
 )]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[serde(untagged)]
```
mattsse, line 18: needs cfg

gakonst, line 18:
```suggestion
#[cfg_attr(feature = "serde", serde(untagged))]
```

Issue #752, opened 2024-05-15
[Feature] Unify engine rpc types fee fields with rpc-types


mattsse:
### Component

rpc

### Describe the feature you would like

the fee fields in the engine types (payload) should also use u128 just as the header and block fields in rpc-types

@fgimenez

mattsse:
actually this goes against the engine specs which requires 64bits for these types, we should stick to the spec even if this requires some ugly conversions when dealing with eth blocks.

this would however mean that eth types also are just u64

PR #753, opened 2024-05-15
unify engine types to u128

yash-atreya:
## Motivation

#752 

## Solution

Use `u128` for gas related fields in `ExecutionPayloads`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (changes requested):
actually, we shouldn't do this, because this goes against the spec which requires 64bits

https://github.com/ethereum/execution-apis/blob/6709c2a795b707202e93c4f2867fa0bf2640a84f/src/engine/paris.md#executionpayloadv1

so I think we should just stick with the proper spec for the engine types here @prestwich ? the conversion payload <-> block should be manageable

prestwich:
per discussion in discord, we want to keep u128 for gas fields everywhere except the engine API

PR #755, opened 2024-05-16
feat: Add methods to JwtSecret to read and write from filesystem

quentinv72:
## Motivation

Adding `from_file` and `try_create` methods to `JwtSecret`. These method allow the caller to read and write from the filesystem. This change should help fully deprecate and remove the `JwtSecret` struct defined in Reth (see https://github.com/paradigmxyz/reth/issues/8276). 
## Solution

The implementation and tests are ported over from Reth into Alloy.
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/jwt.rs
+++ b/crates/rpc-types-engine/src/jwt.rs
@@ -4,7 +4,7 @@ use alloy_primitives::hex;
 use jsonwebtoken::{
     decode, errors::ErrorKind, get_current_timestamp, Algorithm, DecodingKey, Validation,
 };
-use rand::random;
+use rand::{random, Rng};
 use serde::{Deserialize, Serialize};
 use std::{fs, path::Path, str::FromStr, time::Duration};
 use thiserror::Error;
@@ -135,7 +135,7 @@ impl JwtSecret {
 
     /// Creates a random [`JwtSecret`] and tries to store it at the specified path. I/O errors might
     /// occur during write operations in the form of a [`JwtError`]
-    pub fn try_create(fpath: &Path) -> Result<Self, JwtError> {
+    pub fn try_create_random(fpath: &Path) -> Result<Self, JwtError> {
         if let Some(dir) = fpath.parent() {
             // Create parent directory
             fs::create_dir_all(dir)?
@@ -183,7 +183,7 @@ impl JwtSecret {
 
     /// Generates a random [`JwtSecret`] containing a hex-encoded 256 bit secret key.
     pub fn random() -> Self {
-        let random_bytes: [u8; 32] = random();
+        let random_bytes: [u8; 32] = rand::thread_rng().gen();
         let secret = hex::encode(random_bytes);
         JwtSecret::from_hex(secret).unwrap()
     }
--- a/crates/rpc-types-engine/src/jwt.rs
+++ b/crates/rpc-types-engine/src/jwt.rs
@@ -46,7 +46,7 @@ pub enum JwtError {
     #[error("JWT decoding error: {0}")]
     JwtDecodingError(String),
 
-    /// Error variant for failed directory creation operation with additional path context.
+    /// An error occurred while creating a directory to store the JWT.
     #[error("failed to create dir {path:?}: {source}")]
     CreateDir {
         /// The source `io::Error`.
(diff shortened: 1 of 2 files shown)


crates/rpc-types-engine/src/jwt.rs
```diff
@@ -155,7 +183,7 @@ impl JwtSecret {
 
     /// Generates a random [`JwtSecret`] containing a hex-encoded 256 bit secret key.
     pub fn random() -> Self {
-        let random_bytes: [u8; 32] = rand::thread_rng().gen();
+        let random_bytes: [u8; 32] = random();
```
prestwich, line 186: if we're changing this, it should be generic over the RNG trait

crates/rpc-types-engine/src/jwt.rs
```diff
@@ -120,6 +124,30 @@ impl JwtSecret {
...
+    /// a [`JwtError`].
+    pub fn from_file(fpath: &Path) -> Result<Self, JwtError> {
+        let hex = fs::read_to_string(fpath)?;
+        let secret = JwtSecret::from_hex(hex)?;
+        Ok(secret)
+    }
+
+    /// Creates a random [`JwtSecret`] and tries to store it at the specified path. I/O errors might
+    /// occur during write operations in the form of a [`JwtError`]
+    pub fn try_create(fpath: &Path) -> Result<Self, JwtError> {
```
prestwich, line 138: function name should indicate that the created token is random. function should be generic over `Rng` trait

prestwich (changes requested):

crates/rpc-types-engine/src/jwt.rs
quentinv72: updated the function name. 

> function should be generic over Rng trait

Do you still need me to update the `random` method to be more generic? I just ported the code over from Reth and it was not something that was yet supported there, so not sure if this is in scope for this change.

prestwich: no, i think this is fine. if we're not changing the internals of `random()` let's just leave this as is

crates/rpc-types-engine/src/jwt.rs
```diff
@@ -36,6 +36,10 @@ pub enum JwtError {
     /// An error occurred during JWT decoding.
     #[error("JWT decoding error: {0}")]
     JwtDecodingError(String),
+
+    /// An I/O error occurred during JWT operations.
+    #[error(transparent)]
+    IOError(#[from] std::io::Error),
```
mattsse, line 42: wrapping IO errors for fs read transparently, does not include sufficient context because this does not provide enough context because the path is not included in the error message.

so this error should include the Pathbuf and modify the error message.

crates/rpc-types-engine/src/jwt.rs
```diff
@@ -336,7 +366,79 @@ mod tests {
...
+        fs::metadata(path).is_ok()
+    }
+
+    fn hex(secret: &JwtSecret) -> String {
+        hex::encode(secret.0)
+    }
+
+    fn delete(path: &Path) {
+        fs::remove_file(path).unwrap();
+    }
```
mattsse, line 443: these aren't much useful, can be removed

mattsse (changes requested):
I agree that these are nice to have for the type, because usually the jwt is stored on disk.

few nits and suggestions

crates/rpc-types-engine/src/jwt.rs
quentinv72: I've updated this by adding 3 new enum values to capture the possible errors from reading, writing, and creating a directory. 
Should I create a follow up issue to create something akin to the [fs primitive](https://github.com/paradigmxyz/reth/blob/0edf3509a96d1b1eef5108204c3ebf810b1b5058/crates/primitives/src/fs.rs) in Reth? That will allow us to remove the enum variants I've created and replace them all with an FsPathErrror instead.

prestwich (approved):

crates/rpc-types-engine/src/jwt.rs
```diff
@@ -33,9 +38,40 @@ pub enum JwtError {
...
+    },
+
+    /// Error variant for failed write operation with additional path context.
+    #[error("failed to write to {path:?}: {source}")]
+    Write {
+        /// The source `io::Error`.
+        source: io::Error,
+        /// The path related to the operation.
+        path: PathBuf,
+    },
```
Evalir, line 74: Instead of having "Error variant..." on all, can we follow the style of the other variants?

quentinv72: Updated

PR #760, opened 2024-05-20
feat(`provider`): `eth_getAccount` support

Evalir:
## Motivation

See https://t.me/ethers_rs/37923.

## Solution

Implement `eth_getAccount`.

Haven't added a test, as Anvil doesn't implement the method—needs an anvil companion PR

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types/src/eth/account.rs
```diff
@@ -41,6 +41,20 @@ pub struct EIP1186AccountProofResponse {
     pub storage_proof: Vec<EIP1186StorageProof>,
 }
 
+/// Response for `eth_getAccount`
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
```
prestwich, line 46: do we have an Account struct in alloy-consensus? if not i would prefer to push this down and ensure it has RLP encoding

crates/provider/src/provider/trait.rs
```diff
@@ -574,6 +574,16 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         self.client().request("eth_getLogs", (filter,)).await
     }
 
+    /// Retrieves account information ([AccountResponse]) for the given [Address] at the particular
+    /// [BlockId].
+    async fn get_account(
+        &self,
+        address: Address,
+        block: BlockId,
```
mattsse, line 582: can we reuse the new call struct here and only accept address?

crates/rpc-types/src/eth/account.rs
mattsse: yeah we should either move this to consensus or eips, maybe eip1186 because more or less related

mattsse (changes requested):
one note here, the eth api proposal for this was closed because this no longer makes sense post verkle, but rpc providers currently support this, so reasonable to add

mattsse (approved):
lgtm

prestwich (changes requested):
still needs struct pushed down

mattsse:
yeah, I realized we don't have eip1186 in the eips yet, so I think we should move them separately.

prestwich:
the account struct should be in consensus, as it's an member of the state trie, no?

mattsse:
where do you want me to put this, consensus/account.rs ?

prestwich:
that'd be great. I can get to it in a couple hours if you don't. if you want to merge this before adding the RLP impl, please open an issue to follow up 🫡

mattsse:
this would introduce alloy-consensus dep in provider, not sure if we want this

prestwich:
don't we already have that indirectly via receipts and txenvelopes?

mattsse:
ah we indeed have this already

prestwich (approved):

PR #763, opened 2024-05-21
refactor: improve eth_call internals

### LINKED ISSUE #761: [Bug] state changeset arg breaks compatibility with certain endpoints
### Component

network, json-rpc, rpc

_No response_

### Operating System

None

### Describe the bug

State changeset arg is not a part of specification, and is not supported by all client implementations. currently it is always being included as a third argument in `eth_call` which is incorrect for some of the endpoints

ref https://github.com/foundry-rs/foundry/issues/7954 and https://github.com/foundry-rs/foundry/issues/7943

we should probably only include it into the request payload if it's not empty

prestwich:
## Motivation

Closes #761 

## Solution

- Ensure the option is propagated to the internal `RpcCall` future params
- drive-by improvements
    - remove redundant inner type
    - add polling state to simplify taking ownership of data
    - remove unnecessary nesting in future creation
    - `is_*` functions to remove weird inline `matches!`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/call.rs
+++ b/crates/provider/src/provider/call.rs
@@ -70,16 +70,9 @@ where
     }
 
     fn poll_running(&mut self, cx: &mut std::task::Context<'_>) -> Poll<TransportResult<Bytes>> {
-        let Self::Running(mut call) = std::mem::replace(self, Self::Polling) else {
-            unreachable!("bad state")
-        };
+        let Self::Running(ref mut call) = self else { unreachable!("bad state") };
 
-        if let Ready(res) = call.poll_unpin(cx) {
-            Ready(res)
-        } else {
-            *self = Self::Running(call);
-            Poll::Pending
-        }
+        call.poll_unpin(cx)
     }
 }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):

crates/provider/src/provider/call.rs
```diff
@@ -5,11 +5,21 @@ use alloy_rpc_client::{RpcCall, WeakClient};
...
+type RunningFut<'req, 'state, T, N> = RpcCall<
+    T,
+    (&'req <N as Network>::TransactionRequest, BlockId, Option<Cow<'state, StateOverride>>),
+    Bytes,
+>;
+
+/// The [`EthCallFut`] future is the future type for an `eth_call` RPC request.
 #[derive(Debug, Clone)]
-enum States<'req, 'state, T, N>
+pub enum EthCallFut<'req, 'state, T, N>
```
DaniPopes, line 22: don't think this is very useful as public

Issue #604, opened 2024-04-23
[Feature] Set default poll interval based on connected chain


mattsse:
### Component

network, json-rpc, provider, pubsub

### Describe the feature you would like

ref 

https://github.com/alloy-rs/alloy/pull/437#discussion_r1575965143

if not local

DaniPopes:
Related https://github.com/alloy-rs/alloy/pull/437#discussion_r1576824310
> i'd make this 2s given the rise of L2s with faster blocktimes

PR #767, opened 2024-05-21
feat: set poll interval based on connected chain

RexCloud:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/604

## Solution

Add `with_chain()` method to `ProviderBuilder`. The `ChainLayer` will fetch the average block time for the chain and set client's poll interval

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/chain.rs
+++ b/crates/provider/src/layers/chain.rs
@@ -1,0 +1,44 @@
+use alloy_chains::NamedChain;
+use alloy_network::Ethereum;
+use alloy_transport::Transport;
+use std::time::Duration;
+
+use crate::{Provider, ProviderLayer};
+
+/// A layer that wraps a [`NamedChain`]. The layer will be used to set
+/// the client's poll interval based on the average block time for this chain.
+///
+/// Does nothing to the client with a local transport.
+#[derive(Debug, Clone, Copy)]
+pub struct ChainLayer(NamedChain);
+
+impl ChainLayer {
+    /// Get the chain's average blocktime, if applicable.
+    pub fn average_blocktime_hint(&self) -> Option<Duration> {
+        self.0.average_blocktime_hint()
+    }
+}
+
+impl From<NamedChain> for ChainLayer {
+    fn from(chain: NamedChain) -> Self {
+        Self(chain)
+    }
+}
+
+impl<P, T> ProviderLayer<P, T, Ethereum> for ChainLayer
+where
+    P: Provider<T>,
+    T: Transport + Clone,
+{
+    type Provider = P;
+
+    fn layer(&self, inner: P) -> Self::Provider {
+        if !inner.client().is_local() {
+            if let Some(avg_block_time) = self.average_blocktime_hint() {
+                let poll_interval = avg_block_time.mul_f32(0.6);
+                inner.client().set_poll_interval(poll_interval);
+            }
+        }
+        inner
+    }
+}
--- a/crates/provider/src/layers/chain.rs
+++ b/crates/provider/src/layers/chain.rs
@@ -14,7 +14,7 @@ pub struct ChainLayer(NamedChain);
 
 impl ChainLayer {
     /// Get the chain's average blocktime, if applicable.
-    pub fn average_blocktime_hint(&self) -> Option<Duration> {
+    pub const fn average_blocktime_hint(&self) -> Option<Duration> {
         self.0.average_blocktime_hint()
     }
 }
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -5,6 +5,7 @@ use crate::{
     EthCall, PendingTransaction, PendingTransactionBuilder, PendingTransactionConfig, RootProvider,
     RpcWithBlock, SendableTx,
 };
+use alloy_chains::Chain;
 use alloy_eips::eip2718::Encodable2718;
 use alloy_json_rpc::{RpcError, RpcParam, RpcReturn};
 use alloy_network::{Ethereum, Network};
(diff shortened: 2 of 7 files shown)


crates/rpc-client/src/client.rs
```diff
@@ -163,6 +163,8 @@ pub struct RpcClientInner<T> {
     pub(crate) id: AtomicU64,
     /// The poll interval for the client in milliseconds.
     pub(crate) poll_interval: AtomicU64,
+    /// `true` if the poll interval is custom.
+    pub(crate) is_custom_poll_interval: AtomicBool,
```
mattsse, line 167: don't really like the `custom` part because we now also set this based on the chainid, which isn't really custom imo.

but can't come up with better alternatives rn.

crates/provider/src/provider/trait.rs
```diff
@@ -537,9 +538,22 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
+    async fn get_chain_id(&self) -> TransportResult<u64> {
+        self.client().request("eth_chainId", ()).map_resp(crate::utils::convert_u64).await.and_then(
+            |chain_id| {
+                let client = self.client();
+                if !client.is_custom_poll_interval() && !client.is_local() {
+                    let poll_interval = match Chain::from_id(chain_id).average_blocktime_hint() {
+                        Some(avg_block_time) => avg_block_time,
+                        None => client.poll_interval(),
+                    };
+                    client.set_poll_interval(poll_interval.as_millis() as u64);
```
mattsse, line 552: this should be if let some

crates/rpc-client/src/client.rs
```diff
@@ -177,17 +179,24 @@ impl<T> RpcClientInner<T> {
...
 
-    /// Returns the default poll interval (milliseconds) for the client.
+    /// Returns the poll interval (milliseconds) for the client.
     pub fn poll_interval(&self) -> Duration {
         Duration::from_millis(self.poll_interval.load(Ordering::Relaxed))
     }
 
-    /// Set the poll interval for the client in milliseconds.
+    /// Set custom poll interval for the client in milliseconds.
     pub fn set_poll_interval(&self, poll_interval: u64) {
```
mattsse, line 192: while we're at it, I think we should consider changing the arg type to duration

mattsse (changes requested):
imo this makes sense, and isn't very complex

crates/provider/src/provider/trait.rs
```diff
@@ -537,9 +538,22 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
-    fn get_chain_id(&self) -> RpcCall<T, (), U64, u64> {
-        self.client().request("eth_chainId", ()).map_resp(crate::utils::convert_u64)
+    /// Gets the chain ID. Sets the client's poll interval (if default)
+    /// to match the average block time for this chain.
+    async fn get_chain_id(&self) -> TransportResult<u64> {
+        self.client().request("eth_chainId", ()).map_resp(crate::utils::convert_u64).await.and_then(
+            |chain_id| {
+                let client = self.client();
+                if !client.is_custom_poll_interval() && !client.is_local() {
+                    let poll_interval = match Chain::from_id(chain_id).average_blocktime_hint() {
```
mattsse, line 548: this should probably do something like / 2 or `* 0.4 or 0.6` because we should use something smaller than the actual block time

crates/provider/src/provider/trait.rs
DaniPopes: only set if some

prestwich (changes requested):
Approach is a hard NACK from me

- Behavior is non-obvious. Users would not expect that calling an RPC method would change provider behavior
- Not all users will call `get_chain_id`, leading to different behavior in a hard-to-diagnose setting
- The same program may not call `get_chain_id` on different runs, leading to inconsistent configuration for the same code


This could be done idiomatically by adding `ProviderBuilder` methods to provide or fetch this at instantiation time

```
fn with_chain()
async fn fetch_chain()
```

crates/rpc-client/src/client.rs
RexCloud: changed to duration

crates/provider/src/provider/trait.rs
RexCloud: should mul by 0.6 now

crates/rpc-client/src/client.rs
```diff
@@ -86,7 +86,7 @@ impl<T> RpcClient<T> {
     ///
     /// Note: This will only set the poll interval for the client if it is the only reference to the
     /// inner client. If the reference is held by many, then it will not update the poll interval.
-    pub fn with_poll_interval(self, poll_interval: u64) -> Self {
+    pub fn with_poll_interval(self, poll_interval: Duration) -> Self {
```
prestwich, line 89: @yash-atreya I seem to remember us specifically avoiding changing this to `Duration` in another PR. Do you remember why?

crates/provider/src/layers/chain.rs
```diff
@@ -0,0 +1,44 @@
...
+}
+
+impl<P, T> ProviderLayer<P, T, Ethereum> for ChainLayer
+where
+    P: Provider<T>,
+    T: Transport + Clone,
+{
+    type Provider = P;
+
+    fn layer(&self, inner: P) -> Self::Provider {
```
prestwich, line 35: Are there any other properties of a named chain that should affect provider behavior? e.g. do we do any hardfork support tracking in `alloy-chains`?

prestwich (changes requested):
i like the approach, thanks :) couple quick questions in the review comments

crates/provider/src/layers/chain.rs
RexCloud: I can only think of `is_legacy()` that should affect eip1599 methods: `eth_maxPriorityFeePerGas`, `eth_feeHistory`

prestwich: could also use `is_legacy` to set some sort of provider or txfiller behavior I suppose. So far we have special-cased 1559 and generally assume it to be active, so I am happy not doing that right now

crates/rpc-client/src/client.rs
prestwich: I believe I'm thinking of #437 , which chose to represent duration as atomic internally to allow sharing between threads and setting via the `Arc<_>`. because we still represent as atomic internally, I'm fine with this api change, although it feels weird to do the conversion both ways

prestwich (approved):

PR #772, opened 2024-05-22
fix: include auth token in display

### LINKED ISSUE #771: [Bug] Ws auth headers are broken
### Component

transports

_No response_

### Operating System

None

### Describe the bug

Currently we use `to_string()` to insert the authorization header for the ws transport:
https://github.com/alloy-rs/alloy/blob/b9f9f9f08db3cba4d3ac9d79d88c18005763bd57/crates/transport-ws/src/native.rs#L40-L44

This will not work because the `Display` implementation for `Authorization` does not print secrets, making the entire header just `Bearer`:
https://github.com/alloy-rs/alloy/blob/b9f9f9f08db3cba4d3ac9d79d88c18005763bd57/crates/transport/src/common.rs#L48-L55

mattsse:
closes #771

include the auth token in the display impl

### DIFF
--- a/crates/transport/src/common.rs
+++ b/crates/transport/src/common.rs
@@ -6,9 +6,9 @@ use std::{fmt, net::SocketAddr};
 /// Use to inject username and password or an auth token into requests
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Authorization {
-    /// HTTP Basic Auth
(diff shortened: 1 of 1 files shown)


Rjected (approved):
lgtm

mattsse:
reasonable to leak on display, imo
the hyper types also do this and we also leak on debug

PR #779, opened 2024-05-25
style: use poll loop for CallState

mattsse:
transforms the separate CallState::poll_ functions into a single poll loop.

this effectively replaces a waker call with a loop so the response is polled (started) right await without going through the scheduler

### DIFF
--- a/crates/rpc-client/src/call.rs
+++ b/crates/rpc-client/src/call.rs
@@ -62,64 +62,6 @@ where
     }
 }
 
-impl<Params, Conn> CallState<Params, Conn>
-where
-    Conn: Transport + Clone,
-    Params: RpcParam,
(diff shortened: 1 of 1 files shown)


prestwich:
the other idiom i've been moving towards is an explicit continuation:

https://github.com/alloy-rs/alloy/blob/5940871a103ed9332956959a9f2efcc2caa73bf0/crates/provider/src/provider/call.rs#L88-L90

mattsse:
this also works, but this becomes more tricky if the state machine is >2 states or has more complex transitions

prestwich (approved):

smhwangd73 (approved):

smhwangd73 (approved):

PR #788, opened 2024-05-27
feature: ProviderCall

### LINKED ISSUE #736: [Feature] Allow custom implementation for Provider methods
### Component

provider, pubsub, transports

Looking at the new `eth_call` implementation, it looks like `Provider` functions will now return `RpcCall` or a wrapper around it (like `EthCall`). 
The previous implementation returned raw values, making it possible to implement the `Provider` trait on your own provider and override methods to execute your own logic (reading from database instead of using JSON-RPC transport in my case) because the returned values had no ties to `Transport`.

What I ask is to maintain the possibility of executing arbitrary code in `Provider` methods. This can maybe be achieved by making the `Provider` methods return an enum which can either hold an RpcCall or a CustomCall, the later containing a future not depending on the `Transport` trait. 

### Additional context

_No response_

prestwich:
Closes #736 

## Motivation

Allow providers to get outputs from arbitrary sources, including synchronous

## Solution

- [x] Add a `ProviderCall` future that wraps 4 data sources (rpc, batch rpc, other future, ready)
- [x] change provider return types
- [x] change inner type of eth call and with_block

drive-by:
- [x] generalize `into_owned_params`
- [x] alphabetize some mod defs
- [x] improve mapping internal to rpccall
- [x] add map to waiter

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


mattsse (approved):
makes sense,

this is still easy to follow imo

prestwich:
the main problem right now is figuring out how to integrate it into `WithBlock` and `EthCall` 🤔

yash-atreya:
hey @prestwich 

Taking this over, need it for #954

yash-atreya:
> hey @prestwich
> 
> Taking this over, need it for #954

Would appreciate it if you have any pointers to move forward from here. 

I think only integrating `ProviderCall` with `RpcWithBlock` and `EthCall` remains?

crates/provider/src/provider/trait.rs
```diff
@@ -134,26 +137,58 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
+    /// ```
+    #[doc(alias = "eth_call")]
+    #[doc(alias = "call_with_overrides")]
+    fn call<'req>(
+        &'req self,
+        tx: &'req N::TransactionRequest,
+    ) -> ProviderCall<T, EthCallParams<'req, 'req, N>, Bytes, Bytes> {
+        let call = self.call_internal(tx);
+        call.into_provider_call()
+    }
```
mattsse, line 151: this is a breaking api change, that forces usage of call_internal if you want to use EthCall functions. I don't think this is what we want here, because this is very counter-intuitive.

crates/provider/src/provider/trait.rs
```diff
@@ -251,8 +300,10 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     /// Gets the balance of the account.
     ///
     /// Defaults to the latest block. See also [`RpcWithBlock::block_id`].
-    fn get_balance(&self, address: Address) -> RpcWithBlock<T, Address, U256> {
-        RpcWithBlock::new(self.weak_client(), "eth_getBalance", address)
+    fn get_balance(&self, address: Address) -> ProviderCall<T, Address, U256, U256> {
+        let call = RpcWithBlock::new(self.weak_client(), "eth_getBalance", address);
+
+        ProviderCall::from(call)
```
mattsse, line 306: this does not really work, because this now erases the RpcWithBlock functions making this type a bit redundant.

mattsse (changes requested):
hmm, this solution forces us to duplicate the trait functions and therefor doesn't solve the problem because ethcall still only supports rpccall.

the critical part is this:

https://github.com/alloy-rs/alloy/blob/095ed893f73487ed5cba22b7e14b429c0873e102/crates/provider/src/provider/eth_call.rs#L261-L261

basically a transformer that takes the ethcall and converts it into a providercall, currently this is hardcoded to the rpc client, but ideally this is anything that can `fn(ethcallparams) -> ProviderCall`

this needs a different approach so that the ethcall and rpcwith block can be converted to any providercall.

crates/provider/src/provider/prov_call.rs
```diff
@@ -33,12 +33,12 @@ where
...
     Resp: RpcReturn,
-    Map: Fn(Resp) -> Output,
+    Map: Fn(Resp) -> Output + Clone,
 {
     /// An underlying call to an RPC server.
     RpcCall(RpcCall<Conn, Params, Resp, Output, Map>),
-    /// Call With Block
-    RpcCallWithBlock(RpcWithBlockFut<Conn, Params, Resp, Output, Map>),
+    /// RpcWithBlock
+    RpcWithBlock(RpcWithBlock<Conn, Params, Resp, Output, Map>),
```
mattsse, line 41: this looks like the same approach, just slightly modified.

all of this is a bit weird, because imo we need the inverse here, because rpcwithblock is just a helper type that drives something to completion. otherwise this doesn't solve the problem described in the issue. because this is restricted to rpcall:

https://github.com/alloy-rs/alloy/pull/788#pullrequestreview-2153603587

crates/provider/src/provider/prov_call.rs
```diff
@@ -189,14 +189,56 @@ where
...
-            return Self::RpcCallWithBlock(call.block_id(block_id));
+            return Self::RpcWithBlock(call.block_id(block_id));
         }
         self
     }
+
+    /// Set the block id to "pending".
+    pub fn pending(self) -> Self {
+        self.block_id(BlockId::pending())
+    }
```
mattsse, line 204: this leaks rpcwithblock features to provider call.

crates/provider/src/provider/eth_call.rs
```diff
@@ -37,40 +41,69 @@ impl<N: Network> serde::Serialize for EthCallParams<'_, '_, N> {
...
 where
     T: Transport + Clone,
     N: Network,
     Resp: RpcReturn,
-    Map: Fn(Resp) -> Output;
+    Output: 'static,
+    Map: Fn(Resp) -> Output,
+{
+    inner: EthCallFutInner<'req, T, N, Resp, Output, Map>,
+    fut: OnceLock<ProviderCall<T, serde_json::Value, Resp>>,
```
klkvr, line 57: wondering if moving `fut` into `EthCallFutInner` would make more sense?

crates/provider/src/provider/caller.rs
```diff
@@ -0,0 +1,124 @@
+use crate::ProviderCall;
+use alloy_json_rpc::{RpcParam, RpcReturn};
+use alloy_rpc_client::WeakClient;
+use alloy_rpc_types_eth::BlockId;
+use alloy_transport::{RpcError, Transport, TransportErrorKind, TransportResult};
+use std::borrow::Cow;
+
+/// A caller that helper convert `RpcWithBlock` and `EthCall` into a `ProviderCall`.
+pub trait Caller<T, Params, Resp>: Send + Sync
```
klkvr, line 9: so do we expect people to write custom `Caller` implementations which can customize `EthCall` and `WithBlock` behavior based on final `method` and `params`?

crates/provider/src/provider/caller.rs
```diff
@@ -0,0 +1,124 @@
...
+where
+    T: Transport + Clone,
+    Params: RpcParam,
+    Resp: RpcReturn,
+{
+    fn call(
+        &self,
+        method: Cow<'static, str>,
+        params: Params,
+        _block_id: BlockId,
```
klkvr, line 110: given that `block_id` does not always represent the actual `block_id` passed into response, I think we should just keep `method` and `params` here

`WithBlockFut` can just use `ParamsWithBlock` similar to `EthCallParams`

crates/provider/src/provider/caller.rs
yash-atreya: Yes exactly!

crates/provider/src/provider/eth_call.rs
yash-atreya: We can do that, but it would make `EthCallFutInner` non-cloneable. `ProviderCall` cannot be cloned due to the `BoxedFuture` variant.

yash-atreya: Don't see a scenario where cloning this would be necessary, so +1. Can apply this to `States` and remove `WithBlockFut` as well.

klkvr: yep, it's a private type anyway

klkvr:
I think we can now unify `Caller` impls for `WithBlock` and `EthCall`? or maybe just do `impl Caller for WeakClient<T>`?

yash-atreya:
> I think we can now unify `Caller` impls for `WithBlock` and `EthCall`? or maybe just do `impl Caller for WeakClient<T>`?

`impl Caller for WeakClient<T>` makes sense

klkvr:
@yash-atreya one thing I find a bit confusing here is that by design of `Caller` it does not receive more context than a `RequestPacket` we are working with when overriding transport layers.

I think one valid usecase for this PR is for example changing `Provider` implementation to reading storage slots from some other source for `get_storage_at` requests. Currently this would require implementing a custom `Caller` which would manually deserialize `eth_getStorageAt` params to determine address, slot and block which adds an additional fallible step which wasn't required earlier when those methods returned just `TransportResult`

I think what we want for `WithBlock` is something that can produce a `ProviderCall` given a `block_id: BlockId`, and also is able to capture context such as `Params`. This is essentially `fn(BlockId) -> ProviderCall`, though we'd likely need more concrete type here to handle lifetimes in cases of borrowed rpc call params

yash-atreya:
@mattsse I believe this can be merged now

mattsse (commented):
I think this should work now,

mind giving this another close look @klkvr ?

klkvr:
@yash-atreya wondering how do we decide which `Provider` methods return `ProviderCall` vs `async fn -> TransportResult<...>`?

yash-atreya:
> @yash-atreya wondering how do we decide which `Provider` methods return `ProviderCall` vs do `async fn -> TransportResult<...>`?

Sorry for the delayed response. 

`TransportResult` is also overridable like `ProviderCall`. 

I believe subscribe/watch methods should be `TransportResult` since they return `PollerBuilder`. 

Rest can be made to return `ProviderCall`. 

Just about what we want to follow as convention moving forward even for extension traits such as `AdminApi`, `TraceApi` etc. 

@mattsse wdyt?

klkvr:
> Also, any RPC method where we accept a parameter by reference, should be TransportResult since replacing it with ProviderCall will introduce lifetimes. eg. get_fee_history

Lifetimes should be fine, we already have them in e.g. `estimate_gas`

> Should extension traits such as AdminApi, TraceApi etc. be moved to ProviderCall as well?

We are already using `RpcWithBlock` in `TraceApi`, so I wouldn't say that those aren't tied to provider already. Though I can see how changing their returns types to `ProviderCall` might break usecases unrelated to provider

yash-atreya:
> Lifetimes should be fine, we already have them in e.g. `estimate_gas`

We want to avoid introducing lifetimes in the trait as much as possible, `eth_call` and `estimate_gas` are exceptions since the addition of `EthCall`.

> how changing their returns types to ProviderCall might break usecases unrelated to provider

 +1

crates/provider/src/provider/caller.rs
```diff
@@ -0,0 +1,42 @@
+use crate::ProviderCall;
+use alloy_json_rpc::{RpcParam, RpcReturn};
+use alloy_rpc_client::WeakClient;
+use alloy_transport::{RpcError, Transport, TransportErrorKind, TransportResult};
+use std::borrow::Cow;
+
+/// A caller that helper convert `RpcWithBlock` and `EthCall` into a `ProviderCall`.
+pub trait Caller<T, Params, Resp>: Send + Sync
```
klkvr, line 8: given we're now wonly using this for `EthCall`, I think we should make it more call-specific, eg make it accept `TransactionRequest`, block and overrides directly to allow implementations to directly call underlying evm with them.

klkvr (approved):
lgtm overall, just a note on `Caller` trait, we can address this in a follow-up

mattsse (approved):

Issue #780, opened 2024-05-25
[Feature] Implement RLP for new account type


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

ref #760

# TODO

* add RLP support for Account type

PR #789, opened 2024-05-27
feat(consensus) implement RLP for Account information

DoTheBestToGetTheBest:
Closes https://github.com/alloy-rs/alloy/issues/780

### DIFF
--- a/crates/consensus/src/account.rs
+++ b/crates/consensus/src/account.rs
@@ -6,7 +6,6 @@ use std::sync::OnceLock;
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-#[cfg_attr(feature = "ssz", derive(ssz_derive::Encode, ssz_derive::Decode))]
 pub struct Account {
     /// The account's nonce.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u64_via_ruint"))]
--- a/crates/consensus/src/account.rs
+++ b/crates/consensus/src/account.rs
@@ -76,8 +76,7 @@ impl Account {
             let encoded = self.encode_to_vec();
             let mut hasher = Keccak256::new();
             hasher.update(&encoded);
-            let result = hasher.finalize();
-            result
+            hasher.finalize()
         })
     }
 
@@ -87,11 +86,10 @@ impl Account {
         self.encode(&mut buf);
         let mut hasher = Keccak256::new();
         hasher.update(&buf);
-        let result = hasher.finalize();
-        result
+        hasher.finalize()
     }
 
-    /// encode  to a Vec<u8>
+    /// encode  to a `Vec<u8>`
     fn encode_to_vec(&self) -> Vec<u8> {
         let mut out = vec![];
         self.encode(&mut out);
--- a/crates/consensus/src/account.rs
+++ b/crates/consensus/src/account.rs
@@ -1,4 +1,5 @@
 use alloy_primitives::{B256, U256};
+use alloy_rlp::{Decodable, Encodable, Header as RlpHeader};
 
 /// Represents an Account in the account trie.
 #[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


prestwich (changes requested):
Based on the [reth struct here](https://reth.rs/docs/src/reth_primitives/trie/account.rs.html#10-19), we should be able to derive this implementation. However, we will need to re-order the fields. That seems fine to me

while we're here, I would also like to see a convenience function for `trie_hash` (similar to in [`Encodable2718`](https://alloy-rs.github.io/alloy/alloy/eips/eip2718/trait.Encodable2718.html#method.trie_hash)

```
pub fn trie_hash_slow(&self) -> B256 {
    let mut buf = vec![];
    self.encode(&mut buf);
    keccak256(buf)
}
```

bonus points for memoizing the hash in a `OnceLock`

@Rjected @mattsse do you have any RLP encoding/trie hashing test vectors handy?

mattsse (commented):
not sure, I don't think we use this type in reth because we don't store the account's storage root

maybe @rkrasiuk knows if we have some test vectors

but I wonder if we can use derive here?

prestwich:
> not sure, I don't think we use this type in reth because we don't store the account's storage root

hmmm, this type is committed in the MPT trie right? or am i misunderstanding?

DoTheBestToGetTheBest:
> Based on the [reth struct here](https://reth.rs/docs/src/reth_primitives/trie/account.rs.html#10-19), we should be able to derive this implementation. However, we will need to re-order the fields. That seems fine to me
> 
> while we're here, I would also like to see a convenience function for `trie_hash` (similar to in [`Encodable2718`](https://alloy-rs.github.io/alloy/alloy/eips/eip2718/trait.Encodable2718.html#method.trie_hash)
> 
> ```
> pub fn trie_hash_slow(&self) -> B256 {
>     let mut buf = vec![];
>     self.encode(&mut buf);
>     keccak256(buf)
> }
> ```
> 
> bonus points for memoizing the hash in a `OnceLock`
> 
> @Rjected @mattsse do you have any RLP encoding/trie hashing test vectors handy?

Thank you very much for your review i added the trie_hash  and trie_hash_slow methods

DoTheBestToGetTheBest:
> but I wonder if we can use derive here?

should i directly derive  ?

prestwich:
yes, please derive and reorder the fields to match reth as linked above

DoTheBestToGetTheBest:
Thank you guys for the review of the PR.

I reorder the field similar to reth struct and i derived instead of manual impl

prestwich:
thank you. last few things:

- after some thought, I think we're not clear on our memoization strategy yet. So let's go ahead and remove the trie_hash fn, and leave trie_hash_slow. This will give us some time to think about how we want to do this across the code base.

- I would like a couple tests, because this is consensus critical code, and used in reth consensus at some point. You don't have to provide tests right now. I can open a follow up issue.

prestwich:
oh and for no_std support in CI, need to add a 

```
#[cfg(not(feature = "std"))]
use alloc::{vec, vec::Vec};
```

DoTheBestToGetTheBest:
> thank you. last few things:

Thank you very much for you review & your patience !
> 
> * after some thought, I think we're not clear on our memoization strategy yet. So let's go ahead and remove the trie_hash fn, and leave trie_hash_slow. This will give us some time to think about how we want to do this across the code base.

No problem !

> * I would like a couple tests, because this is consensus critical code, and used in reth consensus at some point. You don't have to provide tests right now. I can open a follow up issue.

Thank you, when you open issue about it if it's possible to link any example of test like this i'd like to take that new issue :)

crates/consensus/src/account.rs
```diff
@@ -1,18 +1,32 @@
...
+    /// The hash of the code of the account.
+    pub code_hash: B256,
+}
+
+impl Account {
+    /// compute  hash as committed to in the MPT trie without memoizing
+    pub fn trie_hash_slow(&self) -> B256 {
+        let mut buf = vec![];
+        self.encode(&mut buf);
+        let mut hasher = Keccak256::new();
```
prestwich, line 28: nit:
we usually write this using the convenience`alloy_primitives::keccak256` instead of invoking `Keccak256::new()` and `update`

prestwich (approved):

crates/consensus/src/account.rs
DoTheBestToGetTheBest: > nit: we usually write this using the convenience`alloy_primitives::keccak256` instead of invoking `Keccak256::new()` and `update`

Thank you very much, i changed !

DoTheBestToGetTheBest:
> > Thank you, when you open issue about it if it's possible to link any example of test like this i'd like to take that new issue :)
> 
> this is what i was asking mattsse and rjected about earlier :) I am not sure where to link for testvectors for account encoding

Yep gotcha, ty very much for all support on this !

prestwich (approved):

PR #794, opened 2024-05-28
chore: expose Claims is_within_time_window as pub

Rjected:
## Motivation

Re-creating a `Client` when the claims are out of date is necessary if a custom transport using jwt will be used for more than 60 seconds. To do this, the transport could re-create only if the current claims are within the time window. However, this method is private, so no such custom transport can be created without pasting the existing methods and constants.

## Solution

Expose `is_within_time_window`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/jwt.rs
+++ b/crates/rpc-types-engine/src/jwt.rs
@@ -114,7 +114,7 @@ impl Claims {
     }
 
     /// Checks if the `iat` claim is within the allowed range from the current time.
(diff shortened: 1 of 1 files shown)


DaniPopes:
`with_current_timestamp` too?

Rjected:
> `with_current_timestamp` too?

sure, just exposed that too

mattsse (approved):

DaniPopes (approved):

PR #799, opened 2024-05-29
feat(rpc-types): add topic0 (alias `event_signature`) getter to `Log`

zerosnacks:
## Motivation

Currently there is not an elegant way of accessing the event signature in the `Log` as returned by the `SubscriptionStream`, which is useful for matching against when one subscribes to all (or multiple) contract events.

For context: https://github.com/alloy-rs/examples/blob/dbe65d8ca96348ffa64720ea566079e7c1e87fbf/examples/subscriptions/examples/subscribe_all_logs.rs#L45-L62

## Solution

~~Adds simple getter for 0-indexing into `topics`, mirrors the `event_signature` syntax used for setting `topic0` on the `Filter` instance. I don't see a good reason to add getters for the other topics.~~

Adds simple getter for `topic0`, checking for existence.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/eth/log.rs
+++ b/crates/rpc-types/src/eth/log.rs
@@ -57,6 +57,11 @@ impl Log<LogData> {
         self.inner.topics()
     }
 
+    /// Getter for the topic0 field, the event signature.
+    pub fn event_signature(&self) -> B256 {
+        self.inner.topics()[0]
+    }
+
     /// Get the topic list, mutably. This gives access to the internal
     /// array, without allowing extension of that array. Shortcut for
     /// [`LogData::topics_mut`]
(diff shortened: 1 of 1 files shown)


prestwich:
2 things

- this will panic. Should check that topics is non-empty
- not all events have a signature. anonymous events have none. should name this topic0

zerosnacks:
> 2 things
> 
> * this will panic. Should check that topics is non-empty
> * not all events have a signature. anonymous events have none. should name this topic0

Thanks! Good catch

prestwich:
one more nit and then lfg

we could add a doc alias for event_signature so it shows up in searches

prestwich (approved):

Issue #785, opened 2024-05-27
[Feature] Add overrides to `eth_estimateGas`


prestwich:
### Component

provider, pubsub

### Describe the feature you would like

some endpoints accept overrides on this endpoint. Modify the EthCall future to support returning U256 so we can reuse it for estimateGas

PR #802, opened 2024-05-29
feat: Add overrides to eth_estimateGas

quentinv72:
## Motivation

Close https://github.com/alloy-rs/alloy/issues/785

## Solution

Add `Resp`, `Output`,  and `Map` generic type parameters to EthCall to allow it to return `u128`. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/call.rs
+++ b/crates/provider/src/provider/call.rs
@@ -6,10 +6,10 @@ use alloy_rpc_types::state::StateOverride;
 use alloy_transport::{Transport, TransportErrorKind, TransportResult};
 use futures::FutureExt;
 use serde::ser::SerializeSeq;
-use std::{future::Future, task::Poll};
-use std::marker::PhantomData;
+use std::{future::Future, marker::PhantomData, task::Poll};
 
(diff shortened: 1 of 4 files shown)


crates/provider/src/provider/call.rs
```diff
@@ -124,34 +141,79 @@ where
...
-impl<'req, 'state, T, N> EthCall<'req, 'state, T, N>
+impl<'req, 'state, T, N, Resp, Output, Map> EthCall<'req, 'state, T, N, Resp, Output, Map>
 where
     T: Transport + Clone,
     N: Network,
+    Resp: RpcReturn,
+    Map: Fn(Resp) -> Output,
 {
+    /// Map the response to a different type.
+    pub fn map_resp<NewOutput, NewMap: Fn(Resp) -> NewOutput>(
```
prestwich, line 202: nit:
idiom elsewhere is to use a where block. i feel like it's easier to read for complex bounds

prestwich (changes requested):
single nit, then feels good to me

crates/provider/src/provider/call.rs
quentinv72: fixed!

PR #806, opened 2024-05-31
Impl Eq, PartialEq for WithOtherFields<T: PartialEq | Eq>

kpp:
## Motivation

It's not possible to compare `WithOtherFields<T>` even if T: PartialEq.

## Solution

1. Implement PartialEq for `WithOtherFields<T: PartialEq>`
2. Implement Eq for `WithOtherFields<T: Eq>`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types/src/with_other.rs
+++ b/crates/rpc-types/src/with_other.rs
@@ -55,6 +55,14 @@ impl<T: Default> Default for WithOtherFields<T> {
     }
 }
 
+impl<T: PartialEq> PartialEq for WithOtherFields<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.inner == other.inner && self.other == other.other
+    }
+}
+
+impl<T: Eq> Eq for WithOtherFields<T> {}
+
 impl<'de, T> Deserialize<'de> for WithOtherFields<T>
 where
     T: Deserialize<'de> + Serialize,


crates/rpc-types/src/with_other.rs
```diff
@@ -55,6 +55,14 @@ impl<T: Default> Default for WithOtherFields<T> {
     }
 }
 
+impl<T: PartialEq> PartialEq for WithOtherFields<T> {
+    fn eq(&self, other: &Self) -> bool {
+        self.inner == other.inner && self.other == other.other
+    }
+}
+
+impl<T: Eq> Eq for WithOtherFields<T> {}
```
mattsse, line 64: I think this can be derived which should result in the same code

mattsse (changes requested):

crates/rpc-types/src/with_other.rs
kpp: You are right:

```
struct Wrap<T> {
    inner: T,
}
#[automatically_derived]
impl<T: ::core::cmp::Eq> ::core::cmp::Eq for Wrap<T> {
    #[inline]
    #[doc(hidden)]
    #[coverage(off)]
    fn assert_receiver_is_total_eq(&self) -> () {
        let _: ::core::cmp::AssertParamIsEq<T>;
    }
}
#[automatically_derived]
impl<T> ::core::marker::StructuralPartialEq for Wrap<T> {}
#[automatically_derived]
impl<T: ::core::cmp::PartialEq> ::core::cmp::PartialEq for Wrap<T> {
    #[inline]
    fn eq(&self, other: &Wrap<T>) -> bool {
        self.inner == other.inner
    }
}
```

DaniPopes (approved):
Thanks!

PR #813, opened 2024-06-03
chore: add Into for WithOtherFields in rpc types

teddav:
Not really useful PR... but I need this in order to close https://github.com/foundry-rs/foundry/pull/4931.  
I made the code modifications as minimal as possible :)

### DIFF
--- a/crates/eips/src/eip1898.rs
+++ b/crates/eips/src/eip1898.rs
@@ -721,7 +721,6 @@ impl FromStr for BlockHashOrNumber {
     type Err = ParseBlockHashOrNumberError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        use alloc::string::ToString;
         match u64::from_str(s) {
             Ok(val) => Ok(val.into()),
             Err(parse_int_error) => match B256::from_str(s) {
--- a/crates/rpc-types/src/with_other.rs
+++ b/crates/rpc-types/src/with_other.rs
@@ -35,6 +35,21 @@ impl From<TxEnvelope> for WithOtherFields<TransactionRequest> {
     }
 }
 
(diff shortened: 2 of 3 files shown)


crates/rpc-types/src/with_other.rs
```diff
@@ -35,6 +35,21 @@ impl From<TxEnvelope> for WithOtherFields<TransactionRequest> {
     }
 }
 
+/// Convert a generic type WithOtherFields<T> to a WithOtherFields<TransactionRequest>
+pub trait IntoTransactionRequest {
+    /// Convert a generic type WithOtherFields<T> to a WithOtherFields<TransactionRequest>
+    fn into_transaction_request(self) -> WithOtherFields<TransactionRequest>;
+}
```
mattsse, line 42: hmm, do we really need this trait for implementing Into<TransactionRequest> ?

crates/eips/src/eip1898.rs
```diff
@@ -721,7 +721,6 @@ impl FromStr for BlockHashOrNumber {
     type Err = ParseBlockHashOrNumberError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        use alloc::string::ToString;
```
DaniPopes, line 724: this is necessary for no_std, maybe needs `#[cfg(not(feature = "std"))]`

DaniPopes (changes requested):
pls make it just a method, otherwise lgtm

teddav:
thanks for the review. you're right, this is much cleaner :)

mattsse (approved):

prestwich:
just the doc nit and we're good to go

DaniPopes (approved):

PR #815, opened 2024-06-03
feat(rpc): Add more helpers for `TraceResult`

tcoratger:
## Description
This pull request enhances the `TraceResult` enum by adding more helper methods to improve usability and readability.

## Changes Made
- Added `is_success()` and `is_error()` methods to check the result type easily.
- Added `result()` and `error()` methods to get the result or error payload respectively.
- Updated the unit tests to cover the new helper methods.


crates/rpc-types-trace/src/common.rs
```diff
@@ -33,4 +33,78 @@ impl<Ok, Err> TraceResult<Ok, Err> {
             Self::Error { tx_hash, .. } => tx_hash,
         }
     }
+
+    /// Returns a reference to the result if it is a success variant.
+    pub const fn result(&self) -> Option<&Ok> {
```
mattsse, line 38:
```suggestion
    pub const fn success(&self) -> Option<&Ok> {
```

mattsse (approved):

PR #816, opened 2024-06-03
implement `Default` for `TransactionTrace`

tcoratger:


mattsse (approved):
could be useful for testing and initialization via ..Default::default

tcoratger:
> could be useful for testing and initialization via ..Default::default

Yes this is my use case ;)

PR #818, opened 2024-06-03
chore: move trace methods from `Provider` to `TraceApi`

RexCloud:
## Motivation

This comment above some trace methods
`// todo: move to extension trait`

## Solution

Moved trace related methods, types and tests from `provider/trait.rs` to `ext/trace.rs`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -1,0 +1,117 @@
+//! This module extends the Ethereum JSON-RPC provider with the Trace namespace's RPC methods.
+use crate::{Provider, RpcWithBlock};
+use alloy_eips::BlockNumberOrTag;
+use alloy_network::Network;
+use alloy_primitives::TxHash;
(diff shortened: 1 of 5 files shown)


crates/provider/src/ext/trace.rs
```diff
@@ -0,0 +1,117 @@
...
+{
+    /// Executes the given transaction and returns a number of possible traces.
+    ///
+    /// # Note
+    ///
+    /// Not all nodes support this call.
+    fn trace_call<'a, 'b>(
+        &self,
+        request: &'a N::TransactionRequest,
+        trace_type: &'b [TraceType],
```
gakonst, line 28: cc @mattsse my guess is we could also return a builder that's called `TraceRpcWithBlock` or similar? since we default to `["trace"]` most times?

mattsse (approved):
moving them makes sense,

+1 on having a call type for endpoints that accept a list of tracetype for TraceCallList

and perhaps even a helper arg type for

PR #827, opened 2024-06-04
Refactor: align naming of `transaction_hash` / `transaction_type`

### LINKED ISSUE #814: [Refactor] Nit: align naming of  `tx_hash` / `transaction_hash`, `tx_type` / `transaction_type`
### Component

consensus, eips, genesis, network, json-rpc, rpc

A nit, but I noticed an inconsistency when switching between a pending transaction (`.tx_hash()`) and receipt (`.transaction_hash`):

For `Log` and `Receipt`: `transaction_hash` is used.
For `TxEnvelope`, `Heartbeat`, `TraceResult`: `tx_hash` is used.

Another case is `tx_type` and `transaction_type`:

For `Transaction`, `TransactionReceipt`, `TransactionRequest`: `transaction_type` is used.
For `ReceiptEnvelope` and : `tx_type` is used.

I propose we align the naming. My preference is `transaction_hash` and `transaction_type`, this is in line with `signature_hash`.

### Additional context

_No response_

zerosnacks:
## Motivation

Closes #814 

## Solution

Renaming, no behavioral changes

Consistently apply:

- `tx_hash` -> `transaction_hash`
- `tx_type` -> `transaction_type`

Originally planned a much larger PR (see https://github.com/alloy-rs/alloy/compare/main...zerosnacks/refactor-align-naming-tx) but I feel that it is better to break (possibly contentious) proposed changes into different PRs with accompanying tickets.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/typed.rs
+++ b/crates/consensus/src/transaction/typed.rs
@@ -78,7 +78,7 @@ impl From<TxEnvelope> for TypedTransaction {
 
 impl TypedTransaction {
     /// Return the [`TxType`] of the inner txn.
-    pub const fn tx_type(&self) -> TxType {
+    pub const fn transaction_type(&self) -> TxType {
         match self {
             Self::Legacy(_) => TxType::Legacy,
             Self::Eip2930(_) => TxType::Eip2930,
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -40,7 +40,7 @@ pub enum ReceiptEnvelope<T = Log> {
 
 impl<T> ReceiptEnvelope<T> {
     /// Return the [`TxType`] of the inner receipt.
-    pub const fn tx_type(&self) -> TxType {
+    pub const fn transaction_type(&self) -> TxType {
         match self {
             Self::Legacy(_) => TxType::Legacy,
             Self::Eip2930(_) => TxType::Eip2930,
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -192,7 +192,7 @@ impl TxEip1559 {
             }
             .encode(out);
         }
-        out.put_u8(self.tx_type() as u8);
+        out.put_u8(self.transaction_type() as u8);
         self.encode_with_signature_fields(signature, out);
     }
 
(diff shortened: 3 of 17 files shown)


crates/consensus/src/transaction/typed.rs
```diff
@@ -78,7 +78,7 @@ impl From<TxEnvelope> for TypedTransaction {
 
 impl TypedTransaction {
     /// Return the [`TxType`] of the inner txn.
-    pub const fn tx_type(&self) -> TxType {
+    pub const fn transaction_type(&self) -> TxType {
```
prestwich, line 81: should the type `TxType` also be renamed?

zerosnacks: I tend to agree, it is however one of the rabbit holes I fell down when drafting the PR

There are quite a few structs following that naming pattern that I suggest we change:

- `TxEip1559` -> `Eip1559Transaction`
- `TxLegacy` -> `LegacyTransaction`
- `TxEnvelope` -> `TransactionEnvelope`
- `TxType` -> `TransactionType`
- `AnyTxType` -> `AnyTransactionType`
- `TxReceipt` -> `TransactionReceipt`
- `TxSigner` -> `TransactionSigner`
- `TxSignerSync` -> `TransactionSignerSync`
- `SendableTx` -> `SendableTransaction`
- `TxFiller` -> `TransactionFiller`
- `TxGasAndReward` -> `TransactionGasAndReward`
- `UnsignedTx` -> `UnsignedTransaction`
- `Signed` -> `SignedTransaction`

In line with the proposed changes in the PR I would suggest we change in `core`:

- `TxKind` -> `TransactionKind`
- `TxHash` -> `TransactionHash`
- `TxIndex` -> TransactionIndex`
- `TxNumber` -> `TransactionNumber`

prestwich: i don't love making every type 9 characters longer 😮‍💨

mattsse: I think so.

DaniPopes: If everything is already on `Tx` instead of `Transaction` why are we changing it to the more verbose version?

mattsse: meh, so maybe not rename anything.
don't love `tx_` on functions but it's better than renaming everything

Issue #809, opened 2024-06-03
[Feature] add getter method `request()` for `CallBuilder`


StackOverflowExcept1on:
### Component

contract

### Describe the feature you would like

For example, you want to send an offline transaction that interacts with a smart contract. You need `TransactionRequest`, but now it can only be obtained through `.map()`:
```rust
let contract = Contract::new(CONTRACT_ADDRESS, provider);
let builder = contract.method(...);

let mut transaction_request = <_>::default();
let _ = builder.map(|request| {
    transaction_request = request;
    <_>::default()
});
```

It would be possible to simply provide a cloning getter.

PR #828, opened 2024-06-04
Feat(contract) : add reference to TransactionRequest type

DoTheBestToGetTheBest:
Closes https://github.com/alloy-rs/alloy/issues/809

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -140,7 +140,10 @@ where
     Self: 'static,
 {
     /// Retrieves a reference to the transaction request.
-    pub fn request(&self) -> &N::TransactionRequest {
+    pub const fn request(&self) -> impl AsRef<N::TransactionRequest> + '_
+    where
+        <N as Network>::TransactionRequest: AsRef<<N as Network>::TransactionRequest>,
+    {
         &self.request
     }
 }
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -131,6 +131,20 @@ pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
     transport: PhantomData<T>,
 }
 
+impl<T, P, D, N> CallBuilder<T, P, D, N>
+where
+    T: Transport + Clone,
+    P: Provider<T, N>,
+    D: CallDecoder + Send + Sync + Unpin,
+    N: Network,
+    Self: 'static,
+{
+    /// Retrieves a reference to the transaction request.
+    pub fn request(&self) -> &N::TransactionRequest {
+        &self.request
+    }
+}
+
 // See [`ContractInstance`].
 impl<T: Transport + Clone, P: Provider<T, N>, N: Network> DynCallBuilder<T, P, N> {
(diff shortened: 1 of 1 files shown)


crates/contract/src/call.rs
```diff
@@ -131,6 +131,20 @@ pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
...
+impl<T, P, D, N> CallBuilder<T, P, D, N>
+where
+    T: Transport + Clone,
+    P: Provider<T, N>,
+    D: CallDecoder + Send + Sync + Unpin,
+    N: Network,
+    Self: 'static,
+{
+    /// Retrieves a reference to the transaction request.
+    pub fn request(&self) -> &N::TransactionRequest {
```
prestwich, line 143: `impl AsRef<N::TransactionRequest>`

DoTheBestToGetTheBest: > `impl AsRef<N::TransactionRequest>`

sorry for the question but you want me to return impl AsRef<N::TransactionRequest> or you want me to implement impl AsRef<N::TransactionRequest> for CallBuilder ? so we can use as_ref ?

DoTheBestToGetTheBest: i assum we want to return impl AsRef<N::TransactionRequest>

crates/contract/src/call.rs
```diff
@@ -131,6 +131,23 @@ pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
...
+    D: CallDecoder + Send + Sync + Unpin,
+    N: Network,
+    Self: 'static,
+{
+    /// Retrieves a reference to the transaction request.
+    pub const fn request(&self) -> impl AsRef<N::TransactionRequest> + '_
+    where
+        <N as Network>::TransactionRequest: AsRef<<N as Network>::TransactionRequest>,
+    {
+        &self.request
```
prestwich, line 147: sorry, I mean that instead of `fn request() -> &N::TransactionRequest` you should write the function as an `impl AsRef<N::TransactionRequest>` block. that way it conforms to standard rust patterns

https://doc.rust-lang.org/std/convert/trait.AsRef.html

```
impl<T,P,D,N> AsRef<N::TransactionRequest> for CallBuilder<T, P, D, N>
where
    T: Transport + Clone,
    P: Provider<T, N>,
    D: CallDecoder + Send + Sync + Unpin,
    N: Network,
{
    ... 
}

DoTheBestToGetTheBest: > sorry, I mean that instead of `fn request() -> &N::TransactionRequest` you should write the function as an `impl AsRef<N::TransactionRequest>` block. that way it conforms to standard rust patterns
> 
> https://doc.rust-lang.org/std/convert/trait.AsRef.html
> 
> ```
> impl<T,P,D,N> AsRef<N::TransactionRequest> for CallBuilder<T, P, D, N>
> where
>     T: Transport + Clone,
>     P: Provider<T, N>,
>     D: CallDecoder + Send + Sync + Unpin,
>     N: Network,
> {
>     ... 
> }
> ```

ty for your fast review! appreciate that lot

prestwich (approved):

PR #831, opened 2024-06-05
feature(rpc): add remaining anvil rpc methods to provider

### LINKED ISSUE #271: [Feature] Add remaining Anvil RPC methods to `AnvilProvider`
### Component

providers, pubsub

See https://github.com/alloy-rs/alloy/blob/81c98a572c037eb438c3a328d5d5a3aa8901bf48/crates/providers/src/new.rs#L423-L430

### Additional context

_No response_

zerosnacks:
## Motivation

Closes #271 

## Solution

Implements Anvil API mirroring implementation in Foundry: https://github.com/foundry-rs/foundry/blob/00854b602ef0e67379a2027ccc5d0aad553e5333/crates/anvil/src/eth/api.rs#L1492-L2025

Found a couple of issues in Anvil, very likely unrelated to this API implementation that will requires follow-ups in Foundry:

- `evm_mine`, `evm_mine_detailed` only mines a single block regardless of the configuration defined by MineOptions. Works OK with `None`.
- `anvil_setBlock` logic is implemented in Anvil but not available on the API (dead code).
- `anvil_enable_traces` is disabled in Anvil.
- `anvil_reset` does not update the `chain_id` correctly.

Goal of the tests are mainly as a smoke test to make sure the API is correct.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/ext/mod.rs
```diff
@@ -3,6 +3,11 @@
 mod admin;
 pub use admin::AdminApi;
 
+#[cfg(feature = "anvil")]
+mod anvil;
+#[cfg(feature = "anvil")]
```
prestwich, line 8: this feature pulls in node-bindings, which i don't love doing just for using the anvil API. Should we rename this flag to `anvil-api` and the other flag to `anvil-node`

Cargo.toml
```diff
@@ -119,7 +119,10 @@ url = "2.5"
 derive_more = "0.99.17"
 
 ## serde
-serde = { version = "1.0", default-features = false, features = ["derive", "alloc"] }
+serde = { version = "1.0", default-features = false, features = [
```
prestwich, line 122: unrelated, format only?

crates/provider/Cargo.toml
```diff
@@ -77,5 +78,11 @@ ipc = ["pubsub", "alloy-rpc-client/ipc", "alloy-transport-ipc"]
 reqwest-default-tls = ["alloy-transport-http?/reqwest-default-tls"]
 reqwest-rustls-tls = ["alloy-transport-http?/reqwest-rustls-tls"]
 reqwest-native-tls = ["alloy-transport-http?/reqwest-native-tls"]
-anvil = ["reqwest", "dep:alloy-node-bindings", "dep:alloy-signer-wallet"]
+anvil-api = ["dep:alloy-rpc-types-anvil"]
```
prestwich, line 81: unsure if newlines are from your formatter, or if other formatters would also produce them.

prestwich (changes requested):
couple nits

crates/provider/Cargo.toml
zerosnacks: its my formatter (`Even Better TOML` extension)

Cargo.toml
zerosnacks: Reverted the formatting to not include unrelated diff

crates/rpc-types-anvil/src/lib.rs
```diff
@@ -148,6 +148,7 @@ pub enum MineOptions {
         /// The timestamp the block should be mined with
         #[serde(with = "alloy_serde::u64_opt_via_ruint")]
         timestamp: Option<u64>,
+        #[serde(with = "alloy_serde::u64_opt_via_ruint")]
```
DaniPopes, line 151: Move to below doc-comments

crates/rpc-types/src/eth/admin.rs
```diff
@@ -1,4 +1,5 @@
-//! Types for the admin api
+//! Types for the `admin` api
```
DaniPopes, line 1:
```suggestion
//! Types for the `admin` API.
```

crates/provider/src/ext/anvil.rs
```diff
@@ -0,0 +1,949 @@
...
+    async fn anvil_revert(&self, id: U256) -> TransportResult<bool> {
+        self.client().request("evm_revert", (id,)).await
+    }
+
+    async fn anvil_increase_time(&self, seconds: U256) -> TransportResult<i64> {
+        self.client().request("evm_increaseTime", (seconds,)).await
+    }
+
+    async fn anvil_set_next_block_timestamp(&self, seconds: u64) -> TransportResult<()> {
+        self.client().request("evm_setNextBlockTimestamp", (seconds,)).await
```
DaniPopes, line 273: some of these are declared as fn anvil but are evm_ ?

DaniPopes (commented):
nits

crates/provider/src/ext/anvil.rs
mattsse: yeah, these were originally introduced by ganache which is deprecated now, imo we can keep the anvil prefix because these are also aliased in anvil:

https://github.com/foundry-rs/foundry/blob/e764c316c96c8062e9bc52f78f7ee45ea359bc60/crates/anvil/core/src/eth/mod.rs#L559-L559

but these also exist on hardhat iirc

mattsse (approved):

crates/provider/src/ext/anvil.rs
zerosnacks: Yeah I've used the anvil prefix if it is available. This is the case for most methods but unfortunately not all.

crates/rpc-types/src/eth/mod.rs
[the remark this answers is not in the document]
zerosnacks: Moving these changes to a separate PR

I originally implemented the Anvil changes in the location where they were prior to the `ext` update hence it was related at one point but became unrelated - will pay attention to this going forward

crates/rpc-types-anvil/src/lib.rs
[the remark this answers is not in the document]
zerosnacks: Added two followup tickets assigned to me, one to move Foundry over to use the Alloy types, another to ensure correctness.

- https://github.com/foundry-rs/foundry/issues/8084
- https://github.com/alloy-rs/alloy/issues/838

zerosnacks: Opted to remove this addition for now, going to handle any changes to the types in https://github.com/alloy-rs/alloy/issues/838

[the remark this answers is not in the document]
zerosnacks: Undid this change, will be handled in https://github.com/alloy-rs/alloy/issues/838

mattsse (approved):

DaniPopes (approved):

PR #832, opened 2024-06-05
feat: full block ambiguity

### LINKED ISSUE #805: [Bug] Confusing `BlockTransactions::Uncle` variant
### Component

rpc

_No response_

### Operating System

None

### Describe the bug

Just noting something confusing I ran into - I've been using alloy for getting blocks with full txs and came across this block:
https://holesky.etherscan.io/block/1406078

which has zero transactions, so `transactions` deserializes to `BlockTransactions::Uncle`. This one is clearly not an uncle so it was kind of confusing to see it deserialized like that. Although it seems ambiguous to deserialize into any of the variants.

I was using some reth conversion code, which assumed that `Uncle` meant that full transactions were not present. Had to make this fix:
https://github.com/paradigmxyz/reth/pull/8284/commits/47935c6b35c1c886d4bdce65168ea343d913a9da#diff-577822db25be7f22809b10329e82f83cbf6887190248af08e008d28a2eebca1cR40-R46

mattsse:
closes #805

this changes the deserialization order, so empty response is deserialized as `Full`

this adds an additional check to get_block_by calls to enforce hashes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -437,7 +437,19 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         number: BlockNumberOrTag,
         hydrate: bool,
     ) -> TransportResult<Option<Block>> {
-        self.client().request("eth_getBlockByNumber", (number, hydrate)).await
+        let block = self
+            .client()
+            .request::<_, Option<Block>>("eth_getBlockByNumber", (number, hydrate))
+            .await?
+            .map(|mut block| {
+                if !hydrate {
+                    // this ensures an empty response for `Hashes` has the expected form
+                    // this is required because deserializing [] is ambiguous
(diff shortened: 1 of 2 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -437,7 +437,19 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
...
+        let block = self
+            .client()
+            .request::<_, Option<Block>>("eth_getBlockByNumber", (number, hydrate))
+            .await?
+            .map(|mut block| {
+                if !hydrate {
+                    // this ensures an empty response for `Hashes` has the expected form
+                    // this is required because deserializing [] is ambiguous
+                    block.transactions.convert_to_hashes();
+                }
```
Rjected, line 449: nice, this pattern makes sense

Rjected (approved):
lgtm

Issue #667, opened 2024-05-02
[Feature] Constrain `N::TransactionResponse` type to allow access to the hash or other tx details


prestwich:
necessary followup work: 
              - decide how to bound the `N::TransactionResponse` type to allow access to the hash

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/issues/666#issuecomment-2090574461_

gakonst:
```
trait TransactionResponse {
    fn hash(&self) -> B256
}
```

similar to Receipt, needs to be implemented for AnyNetwork / EthereumNetwork

https://github.com/alloy-rs/alloy/blob/main/crates/network/src/any/mod.rs#L74
https://github.com/alloy-rs/alloy/blob/main/crates/network/src/ethereum/mod.rs#L34

yash-atreya:
Closed by #835

PR #835, opened 2024-06-06
feat(network): constrain `TransactionResponse`

yash-atreya:
## Motivation

#667 

## Solution

Add `TransactionResponse` trait. 

Exposes `hash(&self) -> B256` method. 

@prestwich @mattsse I believe we can add more fields that to this that we know will remain consistent across networks such as `from`, `to`, `value`??

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/lib.rs
+++ b/crates/network/src/lib.rs
@@ -9,7 +9,7 @@
 use alloy_consensus::TxReceipt;
 use alloy_eips::eip2718::{Eip2718Envelope, Eip2718Error};
 use alloy_json_rpc::RpcObject;
-use alloy_primitives::Address;
+use alloy_primitives::{Address, B256};
 use core::fmt::{Debug, Display};
 
 mod transaction;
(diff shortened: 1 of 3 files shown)


crates/network/src/lib.rs
```diff
@@ -36,6 +36,12 @@ pub trait ReceiptResponse {
     fn contract_address(&self) -> Option<Address>;
 }
 
+/// Transaction Response
```
mattsse, line 39: we should add similar docs to what we have above for ReceiptResponse

mattsse (changes requested):
pedantic doc nit

mattsse (approved):

crates/network/src/lib.rs
```diff
@@ -36,6 +36,16 @@ pub trait ReceiptResponse {
...
+///
+/// This is distinct from [`Transaction`], since this is a JSON-RPC response.
+///
+/// [`Transaction`]: alloy_consensus::Transaction
+pub trait TransactionResponse {
+    /// Hash of the transaction
+    fn hash(&self) -> B256;
```
prestwich, line 46: while we're talking renaming in the other issue, should this be `tx_hash` or something similar? will `hash` cause collisions with `Hash::hash`

yash-atreya: Supportive of `tx_hash` since `PendingTransactionBuilder` also uses `tx_hash`

prestwich:
> @prestwich @mattsse I believe we can add more fields that to this that we know will remain consistent across networks such as from, to, value??

the optimism deposit tx has
from
to
value
gas

this seems like a safe set?

mattsse (approved):
lgtm

DaniPopes (approved):

PR #841, opened 2024-06-06
[feat] Synchronous filling

### LINKED ISSUE #830: [Bug] Contract interactions through `#[sol(rpc)]` do not set tx's `from` field
### Component

contract, provider, pubsub, rpc

2bff1dd

### Operating System

macOS (Apple Silicon)

### Describe the bug

Some transactions submitted via `call()/send()` fail because the `from` field of the tx is not set. Submitting the same tx via `send_transaction()` succeeds.

```rust
let contract = Erc721::new(contract_addr, &alice.signer);
let input = Erc721::transferFromCall {
    from: alice.address(),
    to: bob.address(),
    tokenId: token_id,
}
.abi_encode();

    .with_from(alice.address())
    .with_to(contract_addr)
    .with_input(input);

// receipt.status = true
let receipt =
alice.signer.send_transaction(tx).await?.get_receipt().await?;

// receipt.status = false
let receipt = contract
    .transferFrom(alice.address(), bob.address(), token_id)
    .send()
    .await?
    .get_receipt()
    .await?;
```

We should infer the `from` field if the provider contains a `SignerFiller` (I think this is what `send_transaction()` does).

prestwich:
Closes #830 

## Motivation

Ensure that `from` is set before gas estimation occurs

## Solution

Ensure that all properties that can be filled entirely locally are filled BEFORE any network requests are made
This will set `from` and `chain_id` synchronously locally

Potential followup:
- change to sync mutex in noncefiller

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -160,8 +160,9 @@ pub trait TxFiller<N: Network = Ethereum>: Clone + Send + Sync + std::fmt::Debug
         self.status(tx).is_finished()
     }
 
-    /// Performs any synchoronous filling. This will always be called before
-    /// `prepare`.
+    /// Performs any synchoronous filling. This should be called before
+    /// [`TxFiller::prepare`] and [`TxFiller::fill`] to fill in any properties
+    /// that can be filled synchronously.
     fn fill_sync(&self, tx: &mut SendableTx<N>);
 
     /// Prepares fillable properties, potentially by making an RPC request.
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -20,6 +20,7 @@ pub use gas::GasFiller;
 
 mod join_fill;
 pub use join_fill::JoinFill;
+use tracing::error;
 
 use crate::{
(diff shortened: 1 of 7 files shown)


crates/provider/src/fillers/mod.rs
```diff
@@ -244,12 +249,26 @@ where
...
+        let mut count = 0;
+
+        while self.filler.continue_filling(&tx) {
+            self.filler.fill_sync(&mut tx);
+            tx = self.filler.prepare_and_fill(&self.inner, tx).await?;
+
+            count += 1;
+            if count >= 20 {
+                panic!(
+                    "Tx filler loop detected. This indicates a bug in some filler implementation. Please file an issue containing your tx filler set."
```
DaniPopes, line 262: this could log the tx?

DaniPopes (approved):

mattsse (approved):
makes sense

crates/provider/src/fillers/mod.rs
prestwich: done 👍

Issue #845, opened 2024-06-07
[Feature] Add a method `status` in `ReceiptResponse` trait


moricho:
### Component

network, json-rpc

### Describe the feature you would like

It would be helpful to know the tx status from the receipt gotten by `get_receipt` method in `PendingTransactionBuilder` or by `get_transaction_receipt` in `Provider`.

(There might already be a way to understand the status from ReceiptResponse, but I couldn't find it.)

### Additional context

The solution of this issue would be like: https://github.com/alloy-rs/alloy/pull/846

prestwich:
related to https://github.com/alloy-rs/alloy/issues/630

moricho:
Closed by https://github.com/alloy-rs/alloy/pull/846

PR #846, opened 2024-06-07
network: Add `status` method to `ReceiptResponse` trait

moricho:
## Motivation

from: https://github.com/alloy-rs/alloy/issues/845

It would be helpful to know the tx status from the receipt gotten by `get_receipt` method in `PendingTransactionBuilder` or by `get_transaction_receipt` in `Provider`.

## Solution

Define `status` method in `ReceiptResponse` trait and implement it in `TransactionReceipt` for each network

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/ethereum/mod.rs
+++ b/crates/network/src/ethereum/mod.rs
@@ -36,17 +36,6 @@ impl ReceiptResponse for alloy_rpc_types_eth::TransactionReceipt {
         self.contract_address
     }
 
-    /// ## Note
-    ///
-    /// Caution must be taken when using this method for deep-historical
-    /// receipts, as it may not accurately reflect the status of the
-    /// transaction. The transaction status is not knowable from the receipt
-    /// for transactions before [EIP-658].
-    ///
-    /// This can be handled using [`TxReceipt::status_or_post_state`].
-    ///
-    /// [EIP-658]: https://eips.ethereum.org/EIPS/eip-658
-    /// [`TxReceipt::status_or_post_state`]: alloy_consensus::TxReceipt::status_or_post_state
     fn status(&self) -> bool {
         self.inner.status()
     }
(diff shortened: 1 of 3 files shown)


prestwich:
directionally approve. can you copy the doc warnings from #848 ?

prestwich:
cc @mattsse should `type ReceiptResponse: ReceiptResponse` also be constrained to ` + AsRef<Self::ReceiptEnvelope>` to ensure that network implementers ALWAYS embed the envelope in the response?

mattsse:
ah, right this way we could enforce embedding the receipt type in rpc

this makes sense, then we could also add `Into<Self::ReceiptEnvelope>`

prestwich:
this is more work than belongs in this PR. captured in #854

moricho:
@prestwich Added the doc warnings ✅

crates/network/src/ethereum/mod.rs
```diff
@@ -35,6 +35,21 @@ impl ReceiptResponse for alloy_rpc_types_eth::TransactionReceipt {
...
+    ///
+    /// Caution must be taken when using this method for deep-historical
+    /// receipts, as it may not accurately reflect the status of the
+    /// transaction. The transaction status is not knowable from the receipt
+    /// for transactions before [EIP-658].
+    ///
+    /// This can be handled using [`TxReceipt::status_or_post_state`].
+    ///
+    /// [EIP-658]: https://eips.ethereum.org/EIPS/eip-658
+    /// [`TxReceipt::status_or_post_state`]: alloy_consensus::TxReceipt::status_or_post_state
```
mattsse, line 49: duplicating the docs is not necessary imo

prestwich: i'd like the docs to be on every place the user may encounter this (potentially confusing) behavior

prestwich: trait impls never need docstrings tho :)

moricho: I got it, removed the docs from the trait impls: https://github.com/alloy-rs/alloy/pull/846/commits/b42dce55848adf6e5e5342d7e50083087729c5da

DaniPopes (approved):

prestwich (approved):

PR #847, opened 2024-06-07
feat(rpc): Split off `eth` namespace in `alloy-rpc-types` to `alloy-rpc-types-eth`

### LINKED ISSUE #198: [Feature] Split off `eth` namespace in `alloy-rpc-types` to `alloy-rpc-types-eth`
### Component

network, json-rpc

Since it only implements the `eth` namespace types.
Utilities should remain in `rpc-types`.
Other namespaces should have their own crates.

### Additional context

_No response_

zerosnacks:
## Motivation

Closes #198 

## Solution

- Converts `alloy-rpc-types` into a metacrate linking the individual type crates
- Split off `eth` namespace in `alloy-rpc-types` to `alloy-rpc-types-eth`

Given that https://github.com/alloy-rs/alloy/blob/31054e2ecbe5f1720829d30124a1e3ee74c88c5b/crates/rpc-types-eth/src/transaction/request.rs#L803 and https://github.com/alloy-rs/alloy/blob/31054e2ecbe5f1720829d30124a1e3ee74c88c5b/crates/rpc-types-eth/src/transaction/receipt.rs#L3 rely on `WithOtherFields` moving this to the `alloy-rpc-types` metacrate creates a circular dependency. Therefore I've included it in the `alloy-rpc-types-eth` crate for now. If we do want to move it to `core` I'm not sure where we should put it.

Given that there are some open questions: https://github.com/alloy-rs/alloy/issues/198#issuecomment-2144658831 I think it is best to see this as a draft - open to feedback

Tested against `alloy/examples`, no changes required there

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types/Cargo.toml
```diff
@@ -15,53 +15,16 @@ exclude.workspace = true
...
-arbitrary = { workspace = true, features = ["derive"] }
-proptest.workspace = true
-proptest-derive.workspace = true
-rand.workspace = true
-similar-asserts.workspace = true
+rpc-types-anvil = ["alloy-rpc-types-anvil"]
+rpc-types-beacon = ["alloy-rpc-types-beacon"]
+rpc-types-engine = ["alloy-rpc-types-engine"]
+rpc-types-eth = ["alloy-rpc-types-eth"]
+rpc-types-trace = ["alloy-rpc-types-trace"]
```
mattsse, line 30: this removes all the existing features, so if you need arbitrary for example, you'd need to import the individual crates instead.

which means we'd need to keep all previous features for convenience

crates/rpc-types/src/lib.rs
```diff
@@ -10,10 +10,19 @@
...
+pub use alloy_rpc_types_anvil as anvil;
+
+#[cfg(feature = "rpc-types-beacon")]
+pub use alloy_rpc_types_beacon as beacon;
+
+#[cfg(feature = "rpc-types-engine")]
+pub use alloy_rpc_types_engine as engine;
+
+#[cfg(feature = "rpc-types-eth")]
+pub use alloy_rpc_types_eth as eth;
```
zerosnacks, line 25: It is currently available as:

https://github.com/alloy-rs/alloy/blob/f6ebef272c5bd2783bae3fe7ef6ebc1dc3cee015/crates/rpc-types/src/lib.rs#L16

That is likely preferred but inconsistent

crates/rpc-types/Cargo.toml
```diff
@@ -15,53 +15,30 @@ exclude.workspace = true
...
-    "dep:proptest-derive",
-    "dep:proptest",
-    "alloy-primitives/arbitrary",
-    "alloy-eips/arbitrary",
+default = ["rpc-types-eth"]
+rpc-types-anvil = ["alloy-rpc-types-anvil"]
+rpc-types-beacon = ["alloy-rpc-types-beacon"]
+rpc-types-engine = ["alloy-rpc-types-engine"]
+rpc-types-eth = ["alloy-rpc-types-eth"]
+rpc-types-trace = ["alloy-rpc-types-trace"]
```
mattsse, line 31: perhaps we drop the rpc-types- prefix?

zerosnacks: Done :+1:

prestwich:
needs rebase onto main then g2g i think

PR #848, opened 2024-06-07
Fix: Support pre-658 status codes

### LINKED ISSUE #630: [Bug] Invalid encoding for pre-eip658 tx receipts
### Component

consensus, eips, genesis

_No response_

### Operating System

None

### Describe the bug

In `alloy_consensus`, the `Receipt` struct contains the status parameter as a boolean. This is a problem when encoding pre-eip658 (added with Byzantium) receipts, as this parameter was introduced with this hardfork. In `alloy-rpc-types` the `TransactionReceipt` implements this correctly, wrapping the status as an `Option<>`.

prestwich:
Closes #630 

## Motivation

Support receipts from 2017 and earlier

## Solution

- Add an enum that covers the pre- and post-Eip658 values
- impl serde and rlp for that enum
- change receipt to manual serialize impl to allow for naming `root` or `status`
- add docs warning of incorrect status behavior for old receipts
- 

Open question
- when serializing, should the unused field be set to `null` by serializing a `None`?

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -16,7 +16,7 @@ pub struct Receipt<T = Log> {
     /// If transaction is executed successfully.
     ///
     /// This is the `statusCode`
-    #[cfg_attr(feature = "serde", serde(alias = "name"))]
+    #[cfg_attr(feature = "serde", serde(alias = "root"))]
     pub status: Eip658Value,
     /// Gas used
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::u128_via_ruint"))]
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -1,15 +1,14 @@
-use core::borrow::Borrow;
-
-use super::TxReceipt;
-use alloy_primitives::{Bloom, Log};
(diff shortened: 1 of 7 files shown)


crates/consensus/src/receipt/receipts.rs
```diff
@@ -1,31 +1,55 @@
...
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 #[doc(alias = "TransactionReceipt", alias = "TxReceipt")]
 pub struct Receipt<T = Log> {
     /// If transaction is executed successfully.
     ///
     /// This is the `statusCode`
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity_bool"))]
-    pub status: bool,
+    #[cfg_attr(feature = "serde", serde(alias = "name"))]
+    pub status: Eip658Value,
```
mattsse, line 20: this is a clever workaround

mattsse (approved):
eips 👍

crates/consensus/src/receipt/receipts.rs
prestwich: oh i did the alias wrong lol sec

PR #849, opened 2024-06-07
feat(transport): retry layer

### LINKED ISSUE #140: Document usage of tower middleware like retry
### Component

providers, pubsub, transports

We should explain that the transports are tower services and how to layer things like retries/backoff (with a simple example) in either the provider or transport docs (or both).

### Additional context

_No response_

### LINKED ISSUE #288: [Feature] Add standalone HttpError response error type
### Component

rpc

https://github.com/alloy-rs/alloy/blob/09e90e8a76542aaffb3ac0c551a678f2e4fa2890/crates/transport-http/src/hyper.rs#L43-L47

this makes it very hard to check for HTTP errors for things like retry:

https://github.com/foundry-rs/foundry/pull/7384

## TODO
add a standalone `Error` struct so we can try downcasting this from the Custom variant, though perhaps this should even be its own variant

FYI: @zerosnacks

### Additional context

_No response_

### LINKED ISSUE #717: [Feature] Support `tower::retry::RetryLayer`
### Component

transports

I'm hoping to add a retry middleware to the http transport. Telegram discussion pointed me to https://docs.rs/tower/latest/tower/retry/struct.RetryLayer.html. However, experiment with this crate proves unsuccessful so far. Could you help me on how to make this work?

```rust
use alloy::providers::{Provider, ProviderBuilder, ReqwestProvider};
use alloy::rpc::client::ClientBuilder;
use alloy::rpc::json_rpc::{RequestPacket, ResponsePacket};
use alloy::transports::{RpcError, TransportErrorKind};
use futures_util::future;
use tower::retry::{Policy, RetryLayer};

#[derive(Clone, Debug, Default)]
pub struct RetryPolicy(usize);

// Simple retry policy that retries a fixed number of times...
// from https://github.com/tower-rs/tower/blob/master/tower/src/retry/policy.rs#L11-L44
type Req = RequestPacket;
type Res = ResponsePacket;
type Err = RpcError<TransportErrorKind>;
impl Policy<Req, Res, Err> for RetryPolicy {
    type Future = future::Ready<Self>;

    fn retry(&self, req: &Req, result: Result<&Res, &Err>) -> Option<Self::Future> {
        match result {
            Ok(_) => {
                // Treat all `Response`s as success,
                // so don't retry...
                None
            }
            Err(_) => {
                // Treat all errors as failures...
                // But we limit the number of attempts...
                if self.0 > 0 {
                    // Try again!
                    Some(future::ready(RetryPolicy(self.0 - 1)))
                } else {
                    // Used all our attempts, no retry...
                    None
                }
            }
        }
    }

    fn clone_request(&self, req: &Req) -> Option<Req> {
        Some(req.clone())
    }
}

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt().init();
    tracing::info!("Hello, world!");

    let rpc_url = reqwest::Url::parse("https://eth.merkle.io").unwrap();
    // Below is a basic, working version without the retry layer
    // let http_provider: ReqwestProvider = ProviderBuilder::default().on_http(rpc_url);
    //
    // Now we add a simple retry layer. This does not compile
    let MAX_RETRIES = 3;
    let retry_policy = RetryPolicy(MAX_RETRIES);
    let client = ClientBuilder::default()
        .layer(RetryLayer::new(retry_policy))
        .http(rpc_url);
    let http_provider: ReqwestProvider = ProviderBuilder::default().on_client(client);
    let latest_block = http_provider.get_block_number().await.unwrap();
    tracing::info!("Latest block: {:?}", latest_block);
}

```

### Additional context

compiler diagnostics:
```
error[E0271]: type mismatch resolving `<Retry<RetryPolicy, Http<Client>> as Service<RequestPacket>>::Future == Pin<Box<dyn Future<Output = Result<ResponsePacket, RpcError<TransportErrorKind>>> + Send>>`
  --> crates/cli/src/examples/retryhttp.rs:59:10
   |
59 |         .http(rpc_url);
   |          ^^^^ expected `Pin<Box<...>>`, found `ResponseFuture<RetryPolicy, ..., ...>`
   |
   = note: expected struct `Pin<Box<(dyn futures_util::Future<Output = Result<ResponsePacket, RpcError<TransportErrorKind>>> + std::marker::Send + 'static)>>`
              found struct `tower::retry::future::ResponseFuture<RetryPolicy, alloy::transports::alloy_transport_http::Http<Client>, RequestPacket>`
   = note: required for `Retry<RetryPolicy, alloy::transports::alloy_transport_http::Http<Client>>` to implement `alloy::transports::Transport`
```
and
```
error[E0308]: mismatched types
   --> crates/cli/src/examples/retryhttp.rs:60:79
    |
60  |     let http_provider: ReqwestProvider = ProviderBuilder::default().on_client(client);
    |                                                                     --------- ^^^^^^ expected `RpcClient<Http<Client>>`, found `RpcClient<Retry<RetryPolicy, ...>>`
    |                                                                     |
    |                                                                     arguments to this method are incorrect
    |
    = note: expected struct `RpcClient<alloy::transports::alloy_transport_http::Http<_>>`
               found struct `RpcClient<Retry<RetryPolicy, alloy::transports::alloy_transport_http::Http<_>>>`
```

yash-atreya:
## Motivation

Port foundry's layer into the transport crate. 
Had quite a few requests for a retry layer from users, and also, this would remove a bunch of code from the Foundry. 

Closes #717
Closes #140

## Solution

Port from Foundry, still WIP.

Closes #288

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,340 @@
...
+            TransportError::Transport(err) => should_retry_transport_level_error(err),
+            // The transport could not serialize the error itself. The request was malformed from
+            // the start.
+            TransportError::SerError(_) => false,
+            TransportError::DeserError { text, .. } => {
+                if let Ok(resp) = serde_json::from_str::<ErrorPayload>(text) {
+                    return should_retry_json_rpc_error(&resp);
+                }
+
+                // some providers send invalid JSON RPC in the error case (no `id:u64`), but the
```
prestwich, line 77: how very kind of them

mattsse (commented):
this is what we've been in using in foundry

we could rethink some of the decisions here, but overall this has been working well.

there are a few native retry types in tower like https://tower-rs.github.io/tower/tower/retry/budget/index.html#reexport.TpsBudget

but they don't really fit our use case

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,340 @@
...
+        _ => false,
+    }
+}
+
+/// Analyzes the [ErrorPayload] and decides if the request should be retried based on the
+/// error code or the message.
+fn should_retry_json_rpc_error(error: &ErrorPayload) -> bool {
+    let ErrorPayload { code, message, .. } = error;
+    // alchemy throws it this way
+    if *code == 429 {
```
prestwich, line 301: excuse me this is insane

prestwich:
this is a good inclusion. the amount of stupid wrangling makes me sad

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,302 @@
...
+                        ahead_in_queue,
+                    );
+                    let total_backoff = next_backoff
+                        + std::time::Duration::from_secs(seconds_to_wait_for_compute_budget);
+
+                    trace!(?total_backoff, budget_backoff = ?seconds_to_wait_for_compute_budget,
+                    default_backoff = ?next_backoff, ?backoff_hint, "backing off due to rate
+                    limit");
+
+                    std::thread::sleep(total_backoff);
```
prestwich, line 238: can you say more about this usage? this will block the executor, no?

crates/transport/src/error.rs
```diff
@@ -67,4 +71,69 @@ impl TransportErrorKind {
...
+    /// Instantiate a new `TrasnportError::HttpError`.
+    pub const fn http_error(status: i64, body: String) -> TransportError {
+        RpcError::Transport(Self::HttpError(HTTPError { status, body }))
+    }
+}
+
+/// Type for holding HTTP errors such as 429 or -32005 by the RPC provider.
+#[derive(Debug, thiserror::Error)]
+#[error("HTTP error {status} with body: {body}")]
+pub struct HTTPError {
```
prestwich, line 84: style: HttpError

crates/transport/src/error.rs
```diff
@@ -67,4 +71,69 @@ impl TransportErrorKind {
...
+                    || msg.contains("rate exceeded")
+                    || msg.contains("too many requests")
+                    || msg.contains("credits limited")
+                    || msg.contains("request limit")
+            }
+        }
+    }
+}
+
+impl From<&ErrorPayload> for HTTPError {
```
prestwich, line 135: this seems like an inappropiate conversion, as it will produce invalid status codes on the HttpError

crates/transport/src/error.rs
```diff
@@ -67,4 +71,69 @@ impl TransportErrorKind {
...
+    pub const fn http_error(status: i64, body: String) -> TransportError {
+        RpcError::Transport(Self::HttpError(HTTPError { status, body }))
+    }
+}
+
+/// Type for holding HTTP errors such as 429 or -32005 by the RPC provider.
+#[derive(Debug, thiserror::Error)]
+#[error("HTTP error {status} with body: {body}")]
+pub struct HTTPError {
+    pub status: i64,
```
prestwich, line 85: Http status codes are limited to `u16` no?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,302 @@
...
+    if current_queued_requests > request_capacity_per_second {
+        current_queued_requests.min(ahead_in_queue).saturating_div(request_capacity_per_second)
+    } else {
+        0
+    }
+}
+
+/// Analyzes the [TransportErrorKind] and decides if the request should be retried based on the
+/// variant.
+fn should_retry_transport_level_error(error: &TransportErrorKind) -> bool {
```
prestwich, line 280: this should be on the `TransportErrorKind` and on `RpcError<TransportErrorKind, T>` instead of a free function

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        // If the backend is gone, or there's a completely custom error, we should assume it's not
+        // retryable.
+        _ => false,
+    }
+}
+
+/// Analyzes the [ErrorPayload] and decides if the request should be retried based on the
+/// error code or the message.
+fn should_retry_json_rpc_error(error: &ErrorPayload) -> bool {
```
prestwich, line 299: should not be a free function. should not convert to HttpError

crates/transport/src/layers/retry.rs
prestwich: this would be true ONLY of http, right? as in WS we would REQUIRE the ID in order to determine which response got rate limited?

does that imply that we should resolve this in the HTTP transports by writing the ID into the resp if it's missing?

prestwich (changes requested):

crates/transport/src/layers/retry.rs
mattsse: unclear why this is using thread::sleep, it must not

mattsse (changes requested):

crates/transport/src/error.rs
```diff
@@ -67,4 +71,77 @@ impl TransportErrorKind {
...
+    }
+
+    /// Analyzes the [TransportErrorKind] and decides if the request should be retried based on the
+    /// variant.
+    pub fn is_retry_err(&self) -> bool {
+        match self {
+            // Missing batch response errors can be retried.
+            Self::MissingBatchResponse(_) => true,
+            Self::Custom(err) => {
+                // currently http error responses are not standard in alloy
```
prestwich, line 87: this comment becomes untrue after this PR?

crates/transport/src/error.rs
```diff
@@ -67,4 +71,77 @@ impl TransportErrorKind {
...
+                let msg = err.to_string();
+                msg.contains("429 Too Many Requests")
+            }
+            Self::HttpError(http_err) => http_err.is_rate_limit_err(),
```
prestwich, line 91: reorder above custom?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,276 @@
...
+                if let Ok(resp) = serde_json::from_str::<Resp>(text) {
+                    return resp.error.is_retry_err();
+                }
+
+                false
+            }
+            TransportError::ErrorResp(err) => err.is_retry_err(),
+            TransportError::NullResp => true,
+            TransportError::UnsupportedFeature(_) => false,
+            TransportError::LocalUsageError(_) => false,
```
prestwich, line 95: the falses can be captured in a `_ => false`

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,276 @@
...
+pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
+    /// Whether to retry the request based on the given `error`
+    fn should_retry(&self, error: &TransportError) -> bool;
+
+    /// Providers may include the `backoff` in the error response directly
+    fn backoff_hint(&self, error: &TransportError) -> Option<std::time::Duration>;
+}
+
+impl RetryPolicy for RateLimitRetryPolicy {
+    fn should_retry(&self, error: &TransportError) -> bool {
```
prestwich, line 66: this method takes self, but doesn't use it. Instead, this entire method could be implemented on `TransportError`

prestwich: if we wanted to preserve the option for other implementers to change the retry policy, we could move all THIS logic to the `TransportError` and still keep this method here as a simple 

```
    fn should_retry(&self, error: &TransportError) -> bool {
        error.is_retryable()
    }
```

prestwich (changes requested):

crates/transport/src/error.rs
```diff
@@ -67,4 +71,77 @@ impl TransportErrorKind {
...
+    pub fn is_rate_limit_err(&self) -> bool {
+        // alchemy throws it this way
+        if self.status == 429 {
+            return true;
+        }
+
+        // alternative alchemy error for specific IPs
+        if self.body.contains("rate limit") {
+            return true;
+        }
```
mattsse, line 119: all of this is needlessly duplicated

mattsse (changes requested):

crates/transport/src/layers/retry.rs
prestwich: actually i like this idea more now. basically try this as a fallback deser in the http transports, and inser the ID if it works

crates/transport-http/src/hyper_transport.rs
```diff
@@ -87,10 +87,10 @@ where
                 trace!(body = %String::from_utf8_lossy(&body), "response body");
 
                 if status != hyper::StatusCode::OK {
-                    return Err(TransportErrorKind::custom_str(&format!(
-                        "HTTP error {status} with body: {}",
-                        String::from_utf8_lossy(&body)
-                    )));
+                    return Err(TransportErrorKind::http_error(
+                        status.as_u16(),
+                        String::from_utf8_lossy(&body).to_string(),
```
mattsse, line 92:
```suggestion
                        String::from_utf8_lossy(&body).into_owned(),
```

crates/transport-http/src/reqwest_transport.rs
```diff
@@ -63,10 +63,10 @@ impl Http<Client> {
                 trace!(body = %String::from_utf8_lossy(&body), "response body");
 
                 if status != reqwest::StatusCode::OK {
-                    return Err(TransportErrorKind::custom_str(&format!(
-                        "HTTP error {status} with body: {}",
-                        String::from_utf8_lossy(&body)
-                    )));
+                    return Err(TransportErrorKind::http_error(
+                        status.as_u16(),
+                        String::from_utf8_lossy(&body).to_string(),
```
mattsse, line 68:
```suggestion
                        String::from_utf8_lossy(&body).into_owned(),
```

crates/transport/src/error.rs
```diff
@@ -9,6 +10,71 @@ pub type TransportError<ErrResp = Box<RawValue>> = RpcError<TransportErrorKind,
...
 pub type TransportResult<T, ErrResp = Box<RawValue>> = RpcResult<T, TransportErrorKind, ErrResp>;
 
+/// Extension trait to implement methods for [`RpcError<TransportErrorKind, E>`].
+pub trait RpcErrorExt<E> {
+    /// Analyzes whether to retry the request depending on the error.
+    fn is_retryable(&self) -> bool;
+
+    /// Fetches the backoff hint from the error message if present
+    fn backoff_hint(&self) -> Option<std::time::Duration>;
+}
```
mattsse, line 20: this should not be top of file, please move below error defs

crates/transport/src/error.rs
```diff
@@ -67,4 +137,44 @@ impl TransportErrorKind {
...
+    /// Analyzes the [TransportErrorKind] and decides if the request should be retried based on the
+    /// variant.
+    pub fn is_retry_err(&self) -> bool {
+        match self {
+            // Missing batch response errors can be retried.
+            Self::MissingBatchResponse(_) => true,
+            Self::HttpError(http_err) => http_err.is_rate_limit_err(),
+            Self::Custom(err) => {
+                let msg = err.to_string();
+                msg.contains("429 Too Many Requests")
```
mattsse, line 155: I assume this could happen for IPC or any non HTTP error?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,231 @@
...
+            initial_backoff,
+            compute_units_per_second,
+        }
+    }
+}
+
+/// [RateLimitRetryPolicy] implements [RetryPolicy] to determine whether to retry depending on the
+/// err.
+#[derive(Debug, Clone)]
+pub struct RateLimitRetryPolicy;
```
mattsse, line 52:
```suggestion
#[non_exhaustive]
pub struct RateLimitRetryPolicy;
```

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,231 @@
...
+            max_timeout_retries,
+            initial_backoff,
+            compute_units_per_second,
+        }
+    }
+}
+
+/// [RateLimitRetryPolicy] implements [RetryPolicy] to determine whether to retry depending on the
+/// err.
+#[derive(Debug, Clone)]
```
mattsse, line 51:
```suggestion
#[derive(Debug, Copy, Clone, Default)]
```

crates/transport/src/error.rs
yash-atreya: Yes and that is covered in `Custom` and `ErrorPayload.is_retry_err`

crates/transport/src/error.rs
```diff
@@ -67,4 +72,109 @@ impl TransportErrorKind {
...
+    pub const fn is_rate_limit_err(&self) -> bool {
+        if self.status == 429 {
+            return true;
+        }
+        false
+    }
+}
+
+/// Extension trait to implement methods for [`RpcError<TransportErrorKind, E>`].
+pub trait RpcErrorExt<E> {
```
gakonst, line 118: what's this generic for?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+pub struct RateLimitRetryPolicy;
+
+/// [RetryPolicy] defines logic for which [TransportError] instances should
+/// the client retry the request and try to recover from.
+pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
+    /// Whether to retry the request based on the given `error`
+    fn should_retry(&self, error: &TransportError) -> bool;
+
+    /// Providers may include the `backoff` in the error response directly
+    fn backoff_hint(&self, error: &TransportError) -> Option<std::time::Duration>;
```
gakonst, line 62: this trait seems redundant/ should be merged with the other error trait

gakonst (approved):

crates/transport/src/error.rs
mattsse: good question, what is this for?
or in general, why do we need this?

crates/transport/src/layers/retry.rs
mattsse: this is indeed redundant and was already redundant in the foundry code this is based on, the original idea was to plug this into the retrybackofflayer but we don't use this trait, I'd like to keep this because we could replace the RateLimitRetryPolicy with a box<dyn or smth

mattsse (changes requested):

crates/transport/src/error.rs
yash-atreya: Yeah my bad! It isn't required. I thought E i.e `ErrorPayload<ErrData>` would need a Borrow bound.

crates/transport/Cargo.toml
```diff
@@ -25,12 +25,12 @@ serde.workspace = true
 thiserror.workspace = true
 tower.workspace = true
 url.workspace = true
+tracing.workspace = true
+tokio = { workspace = true, features = ["rt", "time"] }
```
prestwich, line 29: rt feature is only used in tests. should be enabled under dev deps?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+                        }
+                    }
+                    Err(e) => err = e,
+                }
+
+                let should_retry = this.policy.should_retry(&err);
+                if should_retry {
+                    rate_limit_retry_number += 1;
+                    if rate_limit_retry_number > this.max_rate_limit_retries {
+                        return Err(TransportErrorKind::custom_str("Max retries exceeded"));
```
prestwich, line 158: this should probably contain information about the latest error

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+                    }
+                    trace!("retrying request due to {:?}", err);
```
prestwich, line 160: - don't debug an arbitrary-structured object in trace.
- tracing should never use string interpolation

this line should be:
`trace!(%err, "retrying request");`

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+        let inner = self.inner.clone();
+        let this = self.clone();
+        let mut inner = std::mem::replace(&mut self.inner, inner);
+        Box::pin(async move {
+            let ahead_in_queue = this.requests_enqueued.fetch_add(1, Ordering::SeqCst) as u64;
+            let mut rate_limit_retry_number: u32 = 0;
+            let mut timeout_retries: u32 = 0;
+            loop {
+                let err;
+                let fut = inner.call(request.clone()).await;
```
prestwich, line 140: this is not a `fut` it's a `res`

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+
+                match fut {
+                    Ok(res) => {
+                        if let Some(e) = res.as_error() {
```
prestwich, line 144: logic in this function should resolve any requests in the ResponsePacket that are resolvable, and should not re-try those requests

prestwich: this causes some problems with our batch model that need to be thought through carefully 🤔

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+                    }
+                    trace!("retrying request due to {:?}", err);
+
+                    let current_queued_reqs = this.requests_enqueued.load(Ordering::SeqCst) as u64;
+
+                    // try to extract the requested backoff from the error or compute the next
+                    // backoff based on retry count
+                    let backoff_hint = this.policy.backoff_hint(&err);
+                    let next_backoff = backoff_hint
+                        .unwrap_or_else(|| std::time::Duration::from_millis(this.initial_backoff));
```
prestwich, line 168: should make a getter for initial_backoff that returns a duration

crates/transport/src/layers/retry.rs
prestwich: so i suppose when doing packets, the retry loop should 
- store successful responses
- retry ONLY the errors
- if retries fail, deliver all responses (successes and failures both)

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,232 @@
...
+                    let seconds_to_wait_for_compute_budget = compute_unit_offset_in_secs(
+                        AVG_COST,
+                        this.compute_units_per_second,
+                        current_queued_reqs,
+                        ahead_in_queue,
+                    );
+                    let total_backoff = next_backoff
+                        + std::time::Duration::from_secs(seconds_to_wait_for_compute_budget);
+
+                    trace!(?total_backoff, budget_backoff = ?seconds_to_wait_for_compute_budget,
```
prestwich, line 190: duration debug is non-ASCII and should not be used. convert to u64 millis and mark tracing props as millis

crates/transport/src/lib.rs
```diff
@@ -18,14 +18,15 @@ pub use common::Authorization;
...
-pub use error::{TransportError, TransportResult};
+pub use error::{RpcErrorExt, TransportError, TransportResult};
 
 mod r#trait;
 pub use r#trait::Transport;
 
 pub use alloy_json_rpc::{RpcError, RpcResult};
 pub use futures_utils_wasm::{impl_future, BoxFuture};
 
+pub mod layers;
```
prestwich, line 29: newline afterwards

crates/transport/src/layers/retry.rs
yash-atreya: > logic in this function should resolve any requests in the ResponsePacket that are resolvable, and should not re-try those requests. As written, potentially fails to resolve some resolvable response

@prestwich Can you clarify here wdym by resolvable error?

yash-atreya: > so i suppose when doing packets, the retry loop should
> 
> * store successful responses
> * retry ONLY the errors
> * if retries fail, deliver all responses (successes and failures both)

Yeah, this works for batch responses.

crates/transport/src/error.rs
```diff
@@ -110,3 +111,65 @@ impl HttpError {
         false
     }
 }
+
+/// Extension trait to implement methods for [`RpcError<TransportErrorKind, E>`].
+pub trait RpcErrorExt {
```
mattsse, line 116: why do we need this?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,296 @@
...
+                        backoff_hint_millis = backoff_hint.map(|d| d.as_millis()),
+                        "(all in ms) backing off due to rate limit"
+                    );
+
+                    tokio::time::sleep(total_backoff).await;
+                } else {
+                    if timeout_retries < this.max_timeout_retries {
+                        timeout_retries += 1;
+                        continue;
+                    }
```
mattsse, line 250: this will always retry errors, such as reverts which are valid responses

we should remove this together with this setting

see also https://github.com/foundry-rs/foundry/pull/8141

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,296 @@
...
+                                        ResponsePacket::from(batch_success.clone());
+                                    batch_success.clear();
+                                    batch_errs.clear();
+
+                                    return Ok(response_packet);
+                                }
+                            }
+                        }
+
+                        let e = batch_errs.first().unwrap().payload.as_error().unwrap().clone();
```
mattsse, line 180: what does this do?

batch handling looks very complex,

this also only uses the first batch_err, but there could be many

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,296 @@
...
+                if should_retry {
+                    rate_limit_retry_number += 1;
+                    if rate_limit_retry_number > this.max_rate_limit_retries {
+                        // If it's a batch request then we combine the successful and failed
+                        // responses and return them as a batch
+                        if !batch_success.is_empty() {
+                            // Join batch_success and batch_errs
+                            let mut batch_calls = batch_success;
+                            batch_calls.append(&mut batch_errs);
+                            return Ok(ResponsePacket::from(batch_calls));
```
mattsse, line 196: doesn't this mess with the request ordering?

mattsse (changes requested):
the batch retry logic is very hard to follow, not immediately obvious what this does.

I suggest we don't do any of this at first, just to make some progress on the pr and retry the entire call

crates/transport/src/error.rs
yash-atreya: Created this to implement `is_retryable_err` and `backoff_hint` methods specifically on `RpcError<TransportErrorKind>` and not the `RpcError<E, ErrResp>` generic type.

yash-atreya:
@mattsse ptal.

I've made some changes to maintain the ordering of the responses corresponding to the incoming requests.

Few highlights: 

1. We store initial incoming requests into `requets_order`. This is used to reference the original order and arrange the responses accordingly. 

2.  We collect all responses (regardless of whether it is a success or err) into `batch_responses: HashMap<Id, Response>` where key is the request `Id`. 

3. As the requests are retried the new response replaces the old one in the `batch_responses` HashMap. 

4. Requests are churned/removed from the next retry batch using `req.remove_by_id(&id)`, if a success response is returned or when non-retryable is encountered.

5. The `batch_responses` is returned as a `ResponsePacket` when:
   - No more requests are left to be retried i.e `req.is_empty()`.
   - Max retries available retries have been exhausted. 
   - Non retryable errors are encountered.

crates/transport/src/error.rs
mattsse: but this is only implemented for RpcError?

could we at least make this pub(crate)?

crates/transport/src/layers/retry.rs
```diff
@@ -0,0 +1,320 @@
...
+                    Ok(ResponsePacket::Batch(batch_res)) => {
+                        // On the first iteration of the loop, populate all the initial
+                        // responses by Id into the HashMap. These responses get replaced by their
+                        // retried response.
+                        batch_responses =
+                            batch_res.iter().map(|res| (res.id.clone(), res.clone())).collect();
+                        for r in batch_res {
+                            if r.is_error() {
+                                batch_responses.insert(r.id.clone(), r.clone());
+
```
mattsse, line 160: this looks redundant?

mattsse (changes requested):
the batch retry logic is a bit complex and not easy to follow, imo this feature isn't super important. I'd rather make some progress on this pr and exclude this entirely, because this has been open for a while now.

it turns out batch requests don't necessarily have ordering guarantees:

https://docs.alchemy.com/reference/batch-requests#what-is-a-batch-request

so we can actually drastically simplify this, by doing a partition over success/errors and only retry.

but I'd like to do that separately and get this included without batch retry support

mattsse (approved):
lgtm

Issue #371, opened 2024-03-22
[Doc] Document special treatment of 1559 in provider and layers


prestwich:
we need to write down somewhere that 1559 is somewhat privileged by the library in ways that other tx types and HF features are not. I.e. we generally assume that 1559 is active and proactively attempt to use it in gas estimation even tho we acknowledge that there may be chains that don't support it

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/363#discussion_r1535775698_

developeruche:
Do you think it is cool to drop this context [here](https://github.com/alloy-rs/alloy/blob/main/README.md)

onbjerg:
@developeruche we want to document it in the code where this is the case so it shows up on docs.rs, so e.g. in the call builder

PR #850, opened 2024-06-07
chore(provider): document privileged status of EIP-1559

zerosnacks:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/371

## Solution

Adds notice on `Provider` trait

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -33,22 +33,34 @@ pub type FilterPollerBuilder<T, R> = PollerBuilder<T, (U256,), Vec<R>>;
 ///
 /// # Subscriptions
 ///
-/// **IMPORTANT:** this is currently only available when `T` is `PubSubFrontend` or `BoxedClient`
-/// over `PubSubFrontend` due to an internal limitation. This means that layering transports will
-/// always disable subscription support. See [issue #296](https://github.com/alloy-rs/alloy/issues/296).
+/// **IMPORTANT:** this is currently only available when `T` is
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -42,6 +42,13 @@ pub type FilterPollerBuilder<T, R> = PollerBuilder<T, (U256,), Vec<R>>;
 ///
 /// For a polling alternatives available over HTTP, use the `watch_*` methods. However, be aware
 /// that polling increases RPC usage drastically.
+///
+/// # EIP-1559
+///
+/// The provider supports EIP-1559 fee estimation and transaction building. We generally assume
+/// that EIP-1559 is supported by the client and will proactively use it by default. If the client
+/// does not support EIP-1559, we fall back to legacy transaction building. We acknowledge that this
+/// means EIP-1559 has a privileged status in comparison to other hardforks.
```
mattsse, line 51: other transaction types?

zerosnacks: yeah makes more sense, James mentioned `HF` but transaction types works better

prestwich:
there's a bit of nuance here. The thing I'm trying to express is that we assume that any chain we run the network abstraction over has the 1559 estimation end points and transactions will have 1559 fields. Which is to say that the network abstraction allows, encapsulating a lot of different behavior behind different traits, but it doesn't encapsulate 1559 at all

it's complicated to provide another example 🤔

prestwich (approved):

PR #855, opened 2024-06-08
feat(serde): deprecate individual num::* for a generic `quantity` module

DaniPopes:
Highlights that this is Ethereum JSON-RPC `quantity` encoding, removes `ruint` from the name, and makes it generic

Unfortunately `#[deprecated]` does not emit any warnings for `serde(with = "...")` invocations

### DIFF
--- a/crates/serde/src/quantity.rs
+++ b/crates/serde/src/quantity.rs
@@ -6,7 +6,7 @@
 //!
 //! This is only valid for human-readable [`serde`] implementations.
 //! For non-human-readable implementations, the format is unspecified.
-//! Currently, it is uses a fixed-width big-endian byte-array.
+//! Currently, it uses a fixed-width big-endian byte-array.
 
 use private::ConvertRuint;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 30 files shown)


prestwich:
I don't mind the name `quantity` but it should be documented that this is only quantity encoding when in Json (or another human-readable format). For non-human readable formats, it still produces fixed-width binary

prestwich (approved):

crates/serde/src/quantity.rs
```diff
@@ -0,0 +1,287 @@
+//! Serde functions for encoding primitive numbers using the Ethereum JSON-RPC "quantity" format.
+//!
+//! This is defined as a "hex encoded unsigned integer", with a special case of 0 being `0x0`.
+//!
+//! A regex for this format is: `^0x([1-9a-f]+[0-9a-f]*|0)$`.
+//!
+//! This is only valid for human-readable [`serde`] implementations.
+//! For non-human-readable implementations, the format is unspecified.
+//! Currently, it is uses a fixed-width big-endian byte-array.
```
Evalir, line 9:
```suggestion
//! Currently, it uses a fixed-width big-endian byte-array.
```

Evalir (approved):

mattsse (approved):

PR #858, opened 2024-06-10
feat(provider): expose `ProviderBuilder` via `fn builder()`

### LINKED ISSUE #450: [Feature] Add `builder` method to `Provider`
### Component

provider, pubsub

Considering users are more familiar with `Provider` than `ProviderBuilder` I would propose to add a `builder` method on the `Provider` giving access to the builder pattern offered by `ProviderBuilder`.

We will have to align on what to do with the `ProviderBuilder` after this method has been added, be it hidden as an internal or kept around as an alternative way of accessing the builder.

Important note is to not use the `Provider::new` syntax, we may even want to apply a `#[doc(hidden)] ` on `new` to steer users towards using the builder.

### Additional context

Previously aligned in a conversation with @prestwich + @gakonst

yash-atreya:
## Motivation

Closes #450 

## Solution

Add `fn builder() -> ProviderBuilder` in the `Provider` trait. 

@prestwich @mattsse ptal

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/root.rs
+++ b/crates/provider/src/provider/root.rs
@@ -37,7 +37,7 @@ impl<T: fmt::Debug, N> fmt::Debug for RootProvider<T, N> {
     }
 }
 
-/// Helper function to directly access [`ProviderBuilder`](crate::ProviderBuilder) with minimal
+/// Helper function to directly access [`ProviderBuilder`] with minimal
 /// generics.
 pub fn builder<N: Network>() -> ProviderBuilder<Identity, Identity, N> {
     ProviderBuilder::default()
--- a/crates/provider/src/provider/root.rs
+++ b/crates/provider/src/provider/root.rs
@@ -1,6 +1,7 @@
 use crate::{
     chain::ChainStreamPoller,
     heart::{Heartbeat, HeartbeatHandle},
+    Identity, ProviderBuilder,
 };
 use alloy_network::{Ethereum, Network};
 use alloy_rpc_client::{BuiltInConnectionString, ClientBuilder, ClientRef, RpcClient, WeakClient};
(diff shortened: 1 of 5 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -51,6 +51,14 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     /// Returns the root provider.
     fn root(&self) -> &RootProvider<T, N>;
 
+    /// Returns the [`ProviderBuilder`](crate::ProviderBuilder) to build on.
+    fn builder() -> ProviderBuilder<Identity, Identity, N>
+    where
+        Self: Sized,
+    {
+        ProviderBuilder::default()
+    }
```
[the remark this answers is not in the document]
yash-atreya: That would create a cyclic dep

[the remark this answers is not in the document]
prestwich: adding a Default bound to the Network::Transaction requested assoc type would allow you to write the impl Ethereum block above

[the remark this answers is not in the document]
yash-atreya: > adding a Default bound to the Network::TransactionRequest assoc type would allow you to write the impl Ethereum block above

Like so?

```rust
/// The JSON body of a transaction request.
    #[doc(alias = "TxRequest")]
    type TransactionRequest: RpcObject
        + TransactionBuilder<Self>
        + Debug
        + Default
        + From<Self::TxEnvelope>
        + From<Self::UnsignedTx>;
```

How would this avoid a cyclic dep, I'm probably making a noob mistake here.

crates/provider/src/provider/root.rs
```diff
@@ -36,6 +37,12 @@ impl<T: fmt::Debug, N> fmt::Debug for RootProvider<T, N> {
     }
 }
 
+/// Helper function to directly access [`ProviderBuilder`] with minimal
+/// generics.
+pub fn builder<N: Network>() -> ProviderBuilder<Identity, Identity, N> {
```
DaniPopes, line 42: I think this should be ethereum network like in `ProviderBuilder::new`?

DaniPopes (approved):

crates/provider/src/provider/root.rs
yash-atreya: That would create cyclic deps

DaniPopes:
pending @mattsse or @prestwich

crates/provider/src/provider/root.rs
DaniPopes: How? It's right there in the file

yash-atreya: Oh wait, I read that as it should be "in ethereum network", as in the network crate.

yash-atreya: > I think this should be ethereum network like in `ProviderBuilder::new`?

Cannot set defaults in a free function.

prestwich (approved):

PR #859, opened 2024-06-10
chore(other): use type aliases where possible to improve clarity 

zerosnacks:
## Motivation

Reduces magic values, reduces chance of error and makes it easier for end-users to see what value they are getting e.g. `TxHash` instead of `FixedBytes<32>` when `watch()`-ing a transaction. Main goal is to limit showing internal types and help with consistency / standardize fields / encourage native types (`u64` vs `U64`) and sizes.

Related PR in `core`: https://github.com/alloy-rs/core/pull/654 + https://github.com/alloy-rs/core/pull/655

## Solution

Proposed changes:

- When in context of:
  - Transaction hash: B256 - Use `TxHash`
  - Block hash: B256 - Use `BlockHash`
  - Chain id: `u64` - Use `ChainId`

Not used:
  - ~~Transaction index: `u64` - Use `TxIndex`~~
  - ~~Block number: `u64` - Use `BlockNumber`~~
  - ~~Transaction nonce: `u64` - Use `TxNonce`~~
  - ~~Block timestamp: `u64` - Use `BlockTimestamp`~~

Should have no impact and no logic changes should be present.

From testing it appears that `rust-analyzer` does not yet correctly deal with type aliases (long standing issue), as in it forwards the https://github.com/rust-lang/rust-analyzer/issues/1666

ezgif-4-4510c3a5f6

Fortunately it does appear to be supported in RustRover:

ezgif-4-acc41ffdf0

PR also includes a number of QOL fixes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,6 +1,6 @@
 use crate::{SignableTransaction, Signed, Transaction, TxType};
 use alloy_eips::eip2930::AccessList;
-use alloy_primitives::{keccak256, Bytes, ChainId, Signature, TxKind, U256};
+use alloy_primitives::{aliases::TxNonce, keccak256, Bytes, ChainId, Signature, TxKind, U256};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use core::mem;
 
@@ -18,7 +18,7 @@ pub struct TxEip1559 {
     pub chain_id: ChainId,
     /// A scalar value equal to the number of transactions sent by the sender; formally Tn.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
-    pub nonce: u64,
+    pub nonce: TxNonce,
     /// A scalar value equal to the maximum
     /// amount of gas that should be used in executing
     /// this transaction. This is paid up-front, before any
@@ -249,7 +249,7 @@ impl TxEip1559 {
     #[inline]
     pub fn size(&self) -> usize {
         mem::size_of::<ChainId>() + // chain_id
-        mem::size_of::<u64>() + // nonce
+        mem::size_of::<TxNonce>() + // nonce
         mem::size_of::<u64>() + // gas_limit
         mem::size_of::<u128>() + // max_fee_per_gas
         mem::size_of::<u128>() + // max_priority_fee_per_gas
@@ -261,8 +261,20 @@ impl TxEip1559 {
 }
 
 impl Transaction for TxEip1559 {
-    fn input(&self) -> &[u8] {
-        &self.input
+    fn chain_id(&self) -> Option<ChainId> {
+        Some(self.chain_id)
+    }
+
+    fn nonce(&self) -> TxNonce {
+        self.nonce
+    }
+
+    fn gas_limit(&self) -> u128 {
+        self.gas_limit
+    }
+
+    fn gas_price(&self) -> Option<u128> {
+        None
     }
 
     fn to(&self) -> TxKind {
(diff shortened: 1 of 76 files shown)


crates/eips/src/eip1898.rs
```diff
@@ -379,14 +379,14 @@ impl From<BlockNumberOrTag> for BlockId {
...
         Self::Hash(RpcBlockHash { block_hash, require_canonical: None })
     }
 }
 
-impl From<(B256, Option<bool>)> for BlockId {
-    fn from(hash_can: (B256, Option<bool>)) -> Self {
+impl From<(BlockHash, Option<bool>)> for BlockId {
+    fn from(hash_can: (BlockHash, Option<bool>)) -> Self {
         Self::Hash(RpcBlockHash { block_hash: hash_can.0, require_canonical: hash_can.1 })
     }
```
zerosnacks, line 391: Note, left this as changed to open discussion - I've generally opted for not using the alias for `From` impl's, curious what is preferred in this case:

- Change all where useful
- Do not change any

crates/genesis/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub struct Genesis {
     pub nonce: u64,
     /// The genesis header timestamp.
     #[serde(with = "alloy_serde::quantity")]
-    pub timestamp: u64,
+    pub timestamp: BlockTimestamp,
```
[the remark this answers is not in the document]
zerosnacks: Reverted this for `genesis`, considering it is almost always just read from a JSON config, rarely dynamically.

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{SignableTransaction, Signed, Transaction, TxType};
 use alloy_eips::eip2930::AccessList;
-use alloy_primitives::{keccak256, Bytes, ChainId, Signature, TxKind, U256};
+use alloy_primitives::{aliases::TxNonce, keccak256, Bytes, ChainId, Signature, TxKind, U256};
```
zerosnacks, line 3: `TxNonce` is now exported correctly see but unfortunately not part of the `0.7.6` release: https://github.com/alloy-rs/core/pull/655

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -288,6 +276,18 @@ impl Transaction for TxEip1559 {
...
+        self.to
+    }
+
+    fn value(&self) -> U256 {
+        self.value
+    }
+
+    fn input(&self) -> &[u8] {
+        &self.input
+    }
```
zerosnacks, line 290: Goal of these changes is cosmetic, simply enforcing consistent ordering for clarity

crates/provider/src/ext/anvil.rs
```diff
@@ -111,11 +111,14 @@ pub trait AnvilApi<N, T>: Send + Sync {
     async fn anvil_increase_time(&self, seconds: U256) -> TransportResult<i64>;
 
     /// Similar to `evm_increaseTime` but takes the exact timestamp that you want in the next block.
-    async fn anvil_set_next_block_timestamp(&self, seconds: u64) -> TransportResult<()>;
+    async fn anvil_set_next_block_timestamp(
+        &self,
+        timestamp: BlockTimestamp,
```
zerosnacks, line 116: Fixes incorrect parameter name

mattsse (changes requested):
I'm very much against using aliases for nonce and timestamp, this is just confusing.

zerosnacks:
@mattsse 

> I'm very much against using aliases for nonce and timestamp, this is just confusing.

I've removed the number aliases, kept the hash aliases as discussed

Any existing references using the number-type aliases I've kept around, just avoided introducing any new ones.

crates/consensus/src/transaction/eip1559.rs
zerosnacks: Not relevant anymore

mattsse (approved):

Issue #109, opened 2024-01-13
[Feature] Replace rpc types Signature with `alloy_primitives::Signature`


DaniPopes:
### Component

rpc-types

### Describe the feature you would like

cc @Evalir

PR #865, opened 2024-06-11
chore(rpc-types-eth): use `Signature` from primitives

yash-atreya:
## Motivation

Ref #109 

## Solution

Use `alloy_primitives::Signature`. 

This required implementing `Arbitrary` manually for `Transaction` since `k256::ecdsa::Signature` does not implement it. In the implementation setting the `signature` field in `Transaction` to `None`.

Moreover, while deserializing a `Transaction`, the signature fields such as `r`, `s`, `v`, and `yParity` are deserialized in the `other` fields map along with the actual `signature` field. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -26,14 +26,13 @@ pub use receipt::{AnyTransactionReceipt, TransactionReceipt};
 pub mod request;
 pub use request::{TransactionInput, TransactionRequest};
 
-mod signature;
-pub use signature::{Parity, Signature};
-
 pub use alloy_consensus::{AnyReceiptEnvelope, Receipt, ReceiptEnvelope, ReceiptWithBloom};
 
+#[cfg(any(test, feature = "arbitrary"))]
+use arbitrary::{Arbitrary, Unstructured};
```
mattsse, line 32: I'd prefer using full paths instead of imports

mattsse (changes requested):
pedantic nit

mattsse (approved):
lgtm

crates/rpc-types-eth/src/txpool.rs
```diff
@@ -250,7 +252,8 @@ mod tests {
         "chainId": "0x1",
         "v": "0x0",
         "r": "0x7b717e689d1bd045ee7afd79b97219f2e36bd22a6a14e07023902194bca96fbf",
-        "s": "0x7b0ba462c98e7b0f95a53f047cf568ee0443839628dfe4ab294bfab88fa8e251"
+        "s": "0x7b0ba462c98e7b0f95a53f047cf568ee0443839628dfe4ab294bfab88fa8e251",
+        "yParity": "0x0"
```
Evalir, line 256: is this a breaking change for deserialization?

Evalir (approved):

crates/rpc-types-eth/src/txpool.rs
yash-atreya: yes

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -313,11 +335,15 @@ mod tests {
...
+            r#"{"hash":"0x0000000000000000000000000000000000000000000000000000000000000001","nonce":"0x2","blockHash":"0x0000000000000000000000000000000000000000000000000000000000000003","blockNumber":"0x4","transactionIndex":"0x5","from":"0x0000000000000000000000000000000000000006","to":"0x0000000000000000000000000000000000000007","value":"0x8","gasPrice":"0x9","gas":"0xa","maxFeePerGas":"0x15","maxPriorityFeePerGas":"0x16","input":"0x0b0c0d","r":"0x23","s":"0x23","v":"0x23","chainId":"0x11","type":"0x14"}"#
         );
-        let deserialized: Transaction = serde_json::from_str(&serialized).unwrap();
+        let mut deserialized: Transaction = serde_json::from_str(&serialized).unwrap();
+        deserialized.other.remove_entry("r");
+        deserialized.other.remove_entry("s");
+        deserialized.other.remove_entry("v");
```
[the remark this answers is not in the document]
mattsse: this does not work due to how flatten works, the only way to make this work would be manual deserialize without flatten and then construct the signature from the individual fields and I'm not sure this is supported

crates/rpc-types-eth/src/txpool.rs
DaniPopes: this is not right, the input should not be modified to make the test pass

prestwich (changes requested):
test vectors should not change, as they represent the json-rpc output. do we need an upstream PR to alloy-core?

yash-atreya:
> test vectors should not change, as they represent the json-rpc output. do we need an upstream PR to alloy-core?

Had to change the test vectors because previously, `v`, `r`, `s`, were being to `U256::from(14)` manually. However, now that we're using primitive Signature, `Parity` value of 14 is invalid when using `from_rs_and_parity()

> test vectors should not change, as they represent the json-rpc output. do we need an upstream PR to alloy-core?

@prestwich @DaniPopes 
I would assume so, because currently the way primitive `Signature` is being deserialized is not the same as our previously implemented rpc-types `Signature`

yash-atreya:
> > test vectors should not change, as they represent the json-rpc output. do we need an upstream PR to alloy-core?
> 
> @prestwich @DaniPopes I would assume so, because currently the way primitive `Signature` is being deserialized is not the same as our previously implemented rpc-types `Signature`

Upon investigating more carefully, it is due to the `other` field being populated with the signature fields creating the dissimilarity.

crates/rpc-types-eth/src/txpool.rs
```diff
@@ -205,7 +205,7 @@ mod tests {
         "type": "0x2",
         "accessList": [],
         "chainId": "0x1",
-        "v": "0x0",
```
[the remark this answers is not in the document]
yash-atreya: @prestwich @DaniPopes ptal
https://github.com/alloy-rs/core/pull/661

onbjerg (changes requested):
needs a changelog update as well

PR #867, opened 2024-06-11
feat: derive `Default` for `WithdrawalRequest` and `DepositRequest`

tcoratger:


mattsse (approved):
useful for testing imo

Issue #424, opened 2024-03-29
[Feature] Tests for `AnyNetwork'


prestwich:
### Component

network, json-rpc, provider, pubsub

### Describe the feature you would like

Add tests to `AnyNetwork` data types to deserialize receipts etc from common networks like Arbitrum, Optimism, etc

nadtech-hub:
If it is still needed I can do it.

PR #868, opened 2024-06-11
rpc-types-eth: Add receipt deserialize tests for `AnyTransactionReceipt`

moricho:
## Motivation
To ensure that the fields present in the tx receipt for Arbitrum and Optimism are captured correctly

from: https://github.com/alloy-rs/alloy/issues/424

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -960,34 +960,7 @@ mod tests {
     use alloy_network::AnyNetwork;
     use alloy_node_bindings::Anvil;
     use alloy_primitives::{address, b256, bytes};
-    use alloy_rpc_types_eth::{request::TransactionRequest, WithOtherFields};
-    use serde::Deserialize;
-
-    // OtherFields for Optimism
-    #[derive(Debug, Deserialize)]
-    struct OptOtherFields {
-        #[serde(rename = "l1BaseFeeScalar")]
-        l1_base_fee_scalar: String,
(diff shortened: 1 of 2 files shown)


moricho:
@prestwich I am currently working on tests to deserialize Optimism and Arbitrum transaction receipts. What should I set as the RPC URL for these networks? I noticed that Ethereum utilizes Alchemy's API endpoint (https://eth-mainnet.alchemyapi.io/v2/jGiK5vwDfC3F4r0bqukm-W2GqgdrxdSr), is this provided by the Alloy organization?
I set `https://opt-mainnet.g.alchemy.com/v2/demo` and `https://arb-mainnet.g.alchemy.com/v2/demo` temporarily.

crates/provider/src/provider/trait.rs
```diff
@@ -1269,6 +1297,58 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    #[ignore]
+    async fn gets_tx_receipt_opt() {
+        init_tracing();
+        let url = "https://opt-mainnet.g.alchemy.com/v2/demo";
```
mattsse, line 1304: we want avoid using live rpc providers because this can degrade CI drastically.

for these tests we should use the raw json objects and test deserialization like this.

mattsse (changes requested):

mattsse (approved):

PR #877, opened 2024-06-12
feat(alloy): add `"full"` feature flag

### LINKED ISSUE #876: [Feature] Add `"full"` feature flag to `alloy` meta-crate
### Component

alloy

Should:

- Include all the core re-exports
- Features: 

```
full = [
    "consensus",
    "contract",
    "eips",
    "k256",
    "kzg",
    "network",
    "provider-http", # includes `providers`
    "provider-ws",   # includes `providers`
    "provider-ipc",  # includes `providers`
    "rpc-types",     # includes `rpc-types-eth`
    "signer-wallet", # includes `signers`
]
```

### Additional context

cc @gakonst

zerosnacks:
## Motivation

Closes #876 

## Solution

- Makes sure to enable `rpc-client` when enabling `rpc-client-ws` so that you do not also need to add the `rpc-client` flag if you if `rpc-client-ws`. This behavior is in line with the other feature flags.
- For easy of access make select `rpc-client` features (`WsConnect`) and (`IpcConnect`) directly available (if defined as flag) on the provider. Previously this required separately defining the `rpc-client` flag and importing it from the `rpc::client` namespace.

Related PR (see impact of proposed changes): https://github.com/alloy-rs/examples/pull/106

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -194,11 +194,14 @@ eip712 = [
 
 # full
 full = [
-    "providers",
-    "rpc-client",
-    "rpc-types",
-    "network",
-    "signer-wallet",
     "consensus",
+    "contract",
     "k256",
+    "network",
+    "provider-http", # includes `providers`
+    "provider-ws",   # includes `providers`
+    "provider-ipc",  # includes `providers`
+    "rpc-client",
+    "rpc-types",     # includes `rpc-types-eth`
+    "signer-wallet", # includes `signers`
 ]
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -72,6 +72,21 @@ std = [
     "alloy-consensus?/std",
 ]
 
+# full
+full = [
+    "consensus",
+    "contract",
+    "eips",
+    "k256",
+    "kzg",
+    "network",
+    "provider-http", # includes `providers`
+    "provider-ws",   # includes `providers`
+    "provider-ipc",  # includes `providers`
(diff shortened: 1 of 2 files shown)


crates/alloy/Cargo.toml
```diff
@@ -72,6 +72,21 @@ std = [
...
+    "eips",
+    "k256",
+    "kzg",
+    "network",
+    "provider-http", # includes `providers`
+    "provider-ws",   # includes `providers`
+    "provider-ipc",  # includes `providers`
+    "rpc-types",     # includes `rpc-types-eth`
+    "signer-wallet", # includes `signers`
+]
```
zerosnacks, line 88: In `examples` we did not need any specific flags from `core` as all commonly used functionality is already available by default

zerosnacks:
Moving to review, open to feedback / suggestions

gakonst (approved):
Tested it out and I like this as a set of minimal features. I wonder if we could trim KZG, but seems fine for now.

crates/alloy/Cargo.toml
```diff
@@ -72,6 +72,21 @@ std = [
     "alloy-consensus?/std",
 ]
 
+# full
+full = [
+    "consensus",
+    "contract",
+    "eips",
+    "k256",
+    "kzg",
```
mattsse, line 81: imo this makes sense to include in `full`

full will probably include more stuff than necessary but for convenience this is okay imo.

mattsse (approved):

Issue #575, opened 2024-04-18
[Refactor]: use "signer" for single credentials and "wallet" for credential stores


prestwich:
Current naming of signer types is somewhat confusing. Move to standardize naming by reviewing usage of the words "signer" and "wallet".

A signer is a single credential capable of signing hashes or txns or other data under a single address

A wallet is a credential store capable of storing multiple signers, and selecting a signer to use to sign hashes, or txns, or other data

              I want to rename `NetworkSigner` to `NetworkWallet`, and `LocalWallet` to `LocalSigner`, and always use `Wallet` to mean "a collection of credentials" and signer to mean "a single credential". but that is a more extensive refactor

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/569#discussion_r1570940200_

EmperorOrokuSaki:
I will give this a try

prestwich:
Cool. To be clear, this is just renaming existing types/functions/etc, should be no behavior changes

PR #883, opened 2024-06-12
refactor(signers): use `signer` for single credentials and `wallet` for credential stores 

zerosnacks:
## Motivation

Closes: https://github.com/alloy-rs/alloy/issues/575

Related downstream PR: https://github.com/alloy-rs/examples/issues/92

## Solution

Current naming of signer types is somewhat confusing. Move to standardize naming by reviewing usage of the words "signer" and "wallet":

- A signer is a single credential capable of signing hashes or txns or other data under a single address.
- A wallet is a credential store capable of storing multiple signers, and selecting a signer to use to sign hashes, or txns, or other data.

Changes:

- `NetworkSigner` -> `NetworkWallet`
- `LocalWallet` -> `LocalSigner`
- `signer-wallet` -> `signer-local` (note: updated crate name!)
- `LocalSigner<SigningKey>` previously exported as `LocalSigner` -> `PrivateKeySigner`

Expected change in syntax:

```diff
-    // Set up wallet from the first default Anvil account (Alice).
-    let wallet: LocalWallet = anvil.keys()[0].clone().into();

+    // Set up signer from the first default Anvil account (Alice).
+    let signer: PrivateKeySigner = anvil.keys()[0].clone().into();

     let provider = ProviderBuilder::new()
         .with_recommended_fillers()
-        .signer(EthereumSigner::from(wallet))
+        .wallet(EthereumSigner::from(signer))
         .on_http(rpc_url);
```

Major breaking / invasive changes, in the context of @prestwich's request

>yes all things named "Signer" or "Wallet" should conform to new naming convention. refactor should be as extensive as possible
>the signer- crates contain only signers. ~~however, we may want to introduce wallet- crates to take over GCP/AWS/Ledger/Trezor, which are inappropriately handled as single credentials~~

Renaming of `signer-*` to `wallet-*` crates is not handled in this PR per comment: https://github.com/alloy-rs/alloy/pull/883#issuecomment-2163792065

Included deprecation notices where I thought it made sense (and was possible)

Note: @yash-atreya  / @zerosnacks / @gakonst  requires changes in `alloy-rs/book` / `alloy-rs/examples` and examples in public relations

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


crates/signer-local/src/lib.rs
```diff
@@ -33,57 +33,57 @@ pub use yubihsm;
...
-/// A wallet instantiated with a locally stored private key
-pub type LocalWallet = Wallet<k256::ecdsa::SigningKey>;
+/// A signer instantiated with a locally stored private key.
+pub type FilledLocalSigner = LocalSigner<SigningKey>;
 
-/// A wallet instantiated with a YubiHSM
+/// A signer instantiated with a YubiHSM.
 #[cfg(feature = "yubihsm")]
-pub type YubiWallet = Wallet<yubihsm::ecdsa::Signer<k256::Secp256k1>>;
+pub type FilledYubiSigner = LocalSigner<yubihsm::ecdsa::Signer<k256::Secp256k1>>;
```
zerosnacks, line 41: Not a fan of the `Filled*` name, any suggestions @prestwich?

prestwich:
hold off on moving the `signer-*` crates for now. Right now they all still are signers, as they manage only a single credential. introducing `wallet-*` is significant future work

crates/signer-local/src/lib.rs
zerosnacks: Went with `PrivateKeySigner` (prev. `LocalSigner`) and `YubiSigner`

crates/signer-local/src/lib.rs
```diff
@@ -33,57 +33,58 @@ pub use yubihsm;
...
 /// # Ok::<_, Box<dyn std::error::Error>>(())
 /// ```
 #[derive(Clone)]
-pub struct Wallet<D> {
-    /// The wallet's private key.
-    pub(crate) signer: D,
-    /// The wallet's address.
+pub struct LocalSigner<C> {
+    /// The signers' credential.
+    pub(crate) credential: C,
```
zerosnacks, line 78: Referring to inner `private_key` as `credential` for flexibility, open to changing it

README.md
```diff
@@ -52,8 +52,8 @@ This repository contains the following crates:
   - [`alloy-signer-aws`] - [AWS KMS] signer implementation
   - [`alloy-signer-gcp`] - [GCP KMS] signer implementation
   - [`alloy-signer-ledger`] - [Ledger] signer implementation
+  - [`alloy-signer-local`] - Local signer (Keystore/Mnemonic/Yubihsm) signer implementation
```
DaniPopes, line 55:
```suggestion
  - [`alloy-signer-local`] - Local (keystore, mnemonic, YubiHSM) signer implementations
```

DaniPopes: maybe "private key" or similar too

crates/signer-local/src/lib.rs
```diff
@@ -33,57 +33,58 @@ pub use yubihsm;
...
+    /// The signers' address.
     pub(crate) address: Address,
-    /// The wallet's chain ID (for EIP-155).
+    /// The signers' chain ID (for EIP-155).
```
DaniPopes, line 81:
```suggestion
    /// The signer's chain ID (for EIP-155).
```

crates/signer-local/src/lib.rs
```diff
@@ -33,57 +33,58 @@ pub use yubihsm;
...
 /// ```
 #[derive(Clone)]
-pub struct Wallet<D> {
-    /// The wallet's private key.
-    pub(crate) signer: D,
-    /// The wallet's address.
+pub struct LocalSigner<C> {
+    /// The signers' credential.
+    pub(crate) credential: C,
+    /// The signers' address.
```
DaniPopes, line 79:
```suggestion
    /// The signer's address.
```

DaniPopes (approved):
lgtm

crates/signer/README.md
```diff
@@ -6,8 +6,8 @@ You can implement the [`Signer`][Signer] trait to extend functionality to other
 such as Hardware Security Modules, KMS etc. See [its documentation][Signer] for more.
 
 Signer implementation in Alloy:
```
DaniPopes, line 8:
```suggestion
Signer implementations in Alloy:
```

crates/signer-local/src/lib.rs
```diff
@@ -33,57 +33,58 @@ pub use yubihsm;
 #[cfg(feature = "mnemonic")]
 pub use coins_bip39;
 
-/// A wallet instantiated with a locally stored private key
-pub type LocalWallet = Wallet<k256::ecdsa::SigningKey>;
+/// A signer instantiated with a locally stored private key.
+pub type PrivateKeySigner = LocalSigner<k256::ecdsa::SigningKey>;
```
DaniPopes, line 37: needs hidden-deprecated type aliases

zerosnacks: good call, thanks

crates/provider/src/builder.rs
```diff
@@ -437,11 +437,12 @@ impl<L, F> ProviderBuilder<L, F, Ethereum> {
         let anvil_layer = crate::layers::AnvilLayer::from(f(Default::default()));
         let url = anvil_layer.endpoint_url();
 
-        let wallet = alloy_signer_wallet::Wallet::from(anvil_layer.instance().keys()[0].clone());
+        let signer =
+            alloy_signer_local::LocalSigner::from(anvil_layer.instance().keys()[0].clone());
```
prestwich, line 441: drive-by, while we're here, should we populate this with more than 1 key?

crates/signer-gcp/src/signer.rs
```diff
@@ -106,17 +106,17 @@ impl KeySpecifier {
 /// # }
 /// ```
 #[derive(Clone)]
-pub struct GcpSigner {
+pub struct GcpWallet {
```
prestwich, line 109: this is a signer, no? as it manages only a single key

crates/signer-trezor/src/signer.rs
```diff
@@ -10,22 +10,22 @@ use trezor_client::client::Trezor;
...
-/// A Trezor Ethereum signer.
+/// A Trezor Ethereum wallet.
 ///
 /// This is a simple wrapper around the [Trezor transport](Trezor).
 ///
-/// Note that this signer only supports asynchronous operations. Calling a non-asynchronous method
+/// Note that this wallet only supports asynchronous operations. Calling a non-asynchronous method
 /// will always return an error.
-pub struct TrezorSigner {
+pub struct TrezorWallet {
```
prestwich, line 19: also manages only 1 key

crates/signer-ledger/src/signer.rs
```diff
@@ -16,14 +16,14 @@ use alloy_dyn_abi::TypedData;
...
+/// A Ledger Ethereum wallet.
 ///
 /// This is a simple wrapper around the [Ledger transport](Ledger).
 ///
-/// Note that this signer only supports asynchronous operations. Calling a non-asynchronous method
+/// Note that this wallet only supports asynchronous operations. Calling a non-asynchronous method
 /// will always return an error.
 #[derive(Debug)]
-pub struct LedgerSigner {
+pub struct LedgerWallet {
```
prestwich, line 26: also manages only 1 key

prestwich (changes requested):
few things now called wallets that should not have been changed

crates/signer-gcp/src/signer.rs
zerosnacks: >A wallet is a credential store capable of storing multiple signers, and selecting a signer to use to sign hashes, or txns, or other data.

I was under the impression that it was more about if the device or service is capable of holding and managing multiple keys / a way to derive the next key

Reverted to use signer for these, also allows for removing the newly introduced deprecation notices

prestwich: Yeah, while GCP or the Ledger Device itself is a "wallet" according to our definition, the structs we have that manage access to those wallets function only as signers, i.e. each of these structs manage a single credential

this implies that we designed the structs wrong in the first place. the ledger struct acquiring the whole device for a single key is bad behavior. GCP instantianting a client for each key is bad behavior. We should be refactoring these from signers into wallets, but that's separate PR scope

crates/provider/src/builder.rs
```diff
@@ -437,11 +439,17 @@ impl<L, F> ProviderBuilder<L, F, Ethereum> {
         let anvil_layer = crate::layers::AnvilLayer::from(f(Default::default()));
         let url = anvil_layer.endpoint_url();
 
-        let wallet = alloy_signer_wallet::Wallet::from(anvil_layer.instance().keys()[0].clone());
+        let default_keys = anvil_layer.instance().keys().to_vec();
+        let (default_key, remaining_keys) = default_keys.split_first().expect("no keys available");
```
prestwich, line 443: prevent panic from having 0 keys when calling `on_anvil_with_wallet_and_config`

This can panic if the `f` function sets accounts to 0. This is not concerning (and is the pre-existing behavior) but we should probably fix it

followup work (not this PR)

prestwich: https://github.com/alloy-rs/alloy/issues/889

prestwich (approved):

prestwich:
tests are broken because they iterate over the hashmap in the `NetworkWallet`. just remove the use of iteration and use `contains` instead 👍

crates/provider/src/builder.rs
zerosnacks: Thought this way trivial to add but noticed some unexpected behavior when implementing this due to the way BTreeMap sorts the addresses, therefore if you get an iterator `signer_addresses` the first address is not the first that was inserted. This may have unexpected consequences when working with derivations / mnemonics where insertion order matters to users. The default address, being assigned upon first insertion stays the same so no issue there.

I think it is out of scope for this ticket, happy to implement it in another.

zerosnacks:
> tests are broken because they iterate over the hashmap in the `NetworkWallet`. just remove the use of iteration and use `contains` instead 👍

Will do

crates/provider/src/builder.rs
zerosnacks: Fixed for now as suggested here: https://github.com/alloy-rs/alloy/pull/883#issuecomment-2165759893

Still something we may want to evaluate the behavior of in a future ticket

prestwich (approved):

mattsse (approved):
lgtm

PR #899, opened 2024-06-14
feat: add getter methods for `FilterChanges`

tcoratger:


mattsse (approved):

crates/rpc-types-eth/src/filter.rs
```diff
@@ -894,6 +894,35 @@ impl From<Vec<Transaction>> for FilterChanges {
     }
 }
 
+impl<T> FilterChanges<T> {
+    /// Get the hashes if present.
+    pub fn hashes(&self) -> Option<&[B256]> {
```
prestwich, line 899: repo standard api is 

is_hashes -> bool
as_hashes -> Option<>

prestwich (changes requested):

crates/rpc-types-eth/src/filter.rs
tcoratger: fixed

prestwich (approved):

PR #900, opened 2024-06-14
feat(provider): support ethCall optional blockId serialization

### LINKED ISSUE #896: [Feature] Support ethCall optional blockId serialization
### Component

rpc

for eth_call/estimategas the blockId argument is option and might not be supported by clients:

https://github.com/foundry-rs/foundry/issues/8122

we should skip serializing this field if, blockid is None and overrides is none:

https://github.com/alloy-rs/alloy/blob/4157e2664fab08941a378198293c7ad376078b5e/crates/provider/src/provider/call.rs#L97-L97

https://github.com/alloy-rs/alloy/blob/4157e2664fab08941a378198293c7ad376078b5e/crates/provider/src/provider/call.rs#L17-L17

### Additional context

image

https://ethereum.github.io/execution-apis/api-documentation/

_No response_

zerosnacks:
## Motivation

Closes #896 

## Solution

- Implements optional serialization of block, correctly handling the `len`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/provider/call.rs
```diff
@@ -14,19 +14,25 @@ type RunningFut<'req, 'state, T, N, Resp, Output, Map> =
...
-        let len = if self.overrides.is_some() { 3 } else { 2 };
+        let len = 1 + self.block.is_some() as usize + self.overrides.is_some() as usize;
+
         let mut seq = serializer.serialize_seq(Some(len))?;
         seq.serialize_element(&self.data)?;
-        seq.serialize_element(&self.block)?;
+
+        if let Some(block) = self.block {
+            seq.serialize_element(&block)?;
+        }
```
mattsse, line 30: here we need to always serialize if overrides are set because overrides must be third argument

mattsse (changes requested):
could we also add a simple serde test ethcallparams?

crates/provider/src/provider/call.rs
```diff
@@ -254,3 +259,79 @@ where
         })
     }
 }
+
+#[cfg(test)]
+
```
DaniPopes, line 264:
```suggestion
```

crates/provider/Cargo.toml
```diff
@@ -67,6 +67,7 @@ alloy-transport-http = { workspace = true, features = ["reqwest"] }
 
 itertools.workspace = true
 reqwest.workspace = true
+serde_json = { workspace = true, features = ["preserve_order"] }
```
mattsse: please undo, we dont want to enable this because this can cause a ton of side-effects

mattsse (changes requested):

PR #902, opened 2024-06-14
feat: derive serde for header

mattsse:
we also have serde support for other consensus types.

this can be useful when we want to store these as json: https://github.com/foundry-rs/foundry/pull/8160

cc @samlaf


Evalir (approved):

prestwich (approved):

prestwich:
we also want to be embedding the consensus object in the rpc object

DaniPopes (approved):

PR #906, opened 2024-06-15
feat: integrate `EvmOverrides` to rpc types

tcoratger:
To replace https://github.com/paradigmxyz/reth/blob/c694433675c3fed05fbe72c50476f4a195acafc6/crates/rpc/rpc/src/eth/revm_utils.rs#L30-L41

### DIFF
--- a/crates/rpc-types-eth/src/state.rs
+++ b/crates/rpc-types-eth/src/state.rs
@@ -1,5 +1,6 @@
 //! bindings for state overrides in eth_call
 
+use crate::BlockOverrides;
 use alloy_primitives::{Address, Bytes, B256, U256, U64};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
@@ -28,6 +29,64 @@ pub struct AccountOverride {
     /// the call.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub state_diff: Option<HashMap<B256, B256>>,
+}
+
+/// Helper type that bundles various overrides for EVM Execution.
+///
+/// By `Default`, no overrides are included.
+#[derive(Debug, Clone, Default)]
+pub struct EvmOverrides {
+    /// Applies overrides to the state before execution.
+    pub state: Option<StateOverride>,
+    /// Applies overrides to the block before execution.
+    ///
+    /// This is a `Box` because less common and only available in debug trace endpoints.
+    pub block: Option<Box<BlockOverrides>>,
+}
+
+impl EvmOverrides {
+    /// Creates a new instance with the given overrides
+    pub const fn new(state: Option<StateOverride>, block: Option<Box<BlockOverrides>>) -> Self {
+        Self { state, block }
+    }
+
+    /// Creates a new instance with the given state overrides.
+    pub const fn state(state: Option<StateOverride>) -> Self {
+        Self { state, block: None }
+    }
+
+    /// Creates a new instance with the given block overrides.
+    pub const fn block(block: Option<Box<BlockOverrides>>) -> Self {
+        Self { state: None, block }
+    }
+
+    /// Returns `true` if the overrides contain state overrides.
+    pub const fn has_state(&self) -> bool {
+        self.state.is_some()
+    }
+
+    /// Returns `true` if the overrides contain block overrides.
+    pub const fn has_block(&self) -> bool {
+        self.block.is_some()
+    }
+
+    /// Adds state overrides to an existing instance.
+    pub fn with_state(mut self, state: StateOverride) -> Self {
+        self.state = Some(state);
+        self
+    }
+
+    /// Adds block overrides to an existing instance.
+    pub fn with_block(mut self, block: Box<BlockOverrides>) -> Self {
+        self.block = Some(block);
+        self
+    }
+}
+
+impl From<Option<StateOverride>> for EvmOverrides {
+    fn from(state: Option<StateOverride>) -> Self {
+        Self::state(state)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/state.rs
```diff
@@ -30,6 +31,64 @@ pub struct AccountOverride {
...
+}
+
+impl EvmOverrides {
+    /// Creates a new instance with the given overrides
+    pub const fn new(state: Option<StateOverride>, block: Option<Box<BlockOverrides>>) -> Self {
+        Self { state, block }
+    }
+
+    /// Creates a new instance with the given state overrides.
+    pub const fn state(state: Option<StateOverride>) -> Self {
```
prestwich, line 54: why even allow `None` as an input?

crates/rpc-types-eth/src/state.rs
```diff
@@ -30,6 +31,64 @@ pub struct AccountOverride {
...
+        Self { state, block: None }
+    }
+
+    /// Creates a new instance with the given block overrides.
+    pub const fn block(block: Option<Box<BlockOverrides>>) -> Self {
```
prestwich, line 59: same question

crates/rpc-types-eth/src/state.rs
```diff
@@ -30,6 +31,64 @@ pub struct AccountOverride {
...
+    }
+
+    /// Adds block overrides to an existing instance.
+    pub fn with_block(mut self, block: Box<BlockOverrides>) -> Self {
+        self.block = Some(block);
+        self
+    }
+}
+
+impl From<Option<StateOverride>> for EvmOverrides {
```
prestwich, line 86:
```
impl From<StateOverrides>
impl From<Box<BlockOverrides>>
impl From<Option<Box<BlockOverrides>>>
```

crates/rpc-types-eth/src/state.rs
```diff
@@ -30,6 +31,64 @@ pub struct AccountOverride {
...
+/// Helper type that bundles various overrides for EVM Execution.
+///
+/// By `Default`, no overrides are included.
+#[derive(Debug, Clone, Default)]
+pub struct EvmOverrides {
+    /// Applies overrides to the state before execution.
+    pub state: Option<StateOverride>,
+    /// Applies overrides to the block before execution.
+    ///
+    /// This is a `Box` because less common and only available in debug trace endpoints.
```
prestwich, line 43: if this type's only purpose is to be serialized, it could instead have borrows of each component?

mattsse: these are only really useful from the server POV and are intended to bundle overrides that are separate arguments

crates/rpc-types-eth/src/state.rs
mattsse: this is also mainly useful from server POV because some methods only support `StateOverrides` or `BlockOverrides` and usually as Option.

crates/rpc-types-eth/src/state.rs
mattsse: I'd actually like to remove this from impl entirely, the caller should call `EvmOverrides::state` instead

mattsse (changes requested):
pedantic nit re From impl

I'd like to have this type here as a helper even if not useful from the client POV, but on the server side this is very useful, hence this also does not require serde because it's not intended to be serialized

for example:

https://github.com/paradigmxyz/reth/blob/01b3757771496a6676a4e3c2b0ebde3978b70463/crates/rpc/rpc/src/eth/api/call.rs#L68-L74

crates/rpc-types-eth/src/state.rs
prestwich: so does this get deserialized as a unit? or does it get deser separately and then constructed?

crates/rpc-types-eth/src/state.rs
prestwich: My main point is that passing `None` to this function is always more clearly represented as `Default::default` which means this API is an artifact of a specific usage, not a good API on its own, and is not something we should make available to non-reth users

```
// `fn state(state:StateOverride) -> Self`
state_opt.map(EvmOverrides::state).unwrap_or_default()

// current api
EvmOverrides::state(state_opt)

// From<Option<_>> feels cleanest
state_opt.into()
```

prestwich:
```
/// Matches some other spots in reth
EvmOverrides::default().maybe_state(state_opt)
```

crates/rpc-types-eth/src/state.rs
mattsse: separately then constructed, so in the method implementation we can pass `overrides: EvmOverrides` instead of two Option arguments

tcoratger:
> pedantic nit re From impl
> 
> I'd like to have this type here as a helper even if not useful from the client POV, but on the server side this is very useful, hence this also does not require serde because it's not intended to be serialized
> 
> for example:
> 
> https://github.com/paradigmxyz/reth/blob/01b3757771496a6676a4e3c2b0ebde3978b70463/crates/rpc/rpc/src/eth/api/call.rs#L68-L74

@mattsse As a result of these discussions I've just removed the extra `From` implementation, leading the user having to specify `EvmOverrides::state(...)`. Let me know if this is ok to you

crates/rpc-types-eth/src/state.rs
prestwich: this smells like the API problem is minimizing diff for some existing-but-messy code. don't love it, but i won't block you merging it

mattsse (approved):

crates/rpc-types-eth/src/state.rs
mattsse: the issue here is that some endpoints support both override args, some only state, some only block, but a lot require the same setup, like eth_call, trace_call, call_many, estimate. 
this helper type makes it possible to reuse the same setup

PR #907, opened 2024-06-15
 chore: bump rust version 1.79 and bump crates

tcoratger:


DaniPopes:
There's no reason to do either of these things, and I prefer not to fully specify all dependency versions.

tcoratger:
> There's no reason to do either of these things, and I prefer not to fully specify all dependency versions.

@DaniPopes 
- Should we do the same thing on reth to not fully specify all the dependency version in situations like this?
   - https://github.com/paradigmxyz/reth/blob/925d40618890b1ec8bb5c2902ee306c7fbf7ee9e/Cargo.toml#L435
   - https://github.com/paradigmxyz/reth/blob/925d40618890b1ec8bb5c2902ee306c7fbf7ee9e/Cargo.toml#L400
   - https://github.com/paradigmxyz/reth/blob/925d40618890b1ec8bb5c2902ee306c7fbf7ee9e/Cargo.toml#L391
   - ...
 - Why not sync to the same version of Rust as reth?
 
 Just questions, I may be wrong, no problem at all with closing this :)

DaniPopes:
you don't want to fully specify the version to allow users to select the minor/patch version. If should only be increased if there's a reason to, e.g. a method was added in 1.2.3 while we're specifying 1.2 or 1. Same for MSRV

prestwich:
> Why not sync to the same version of Rust as reth?

alloy msrv policy is stated in the readme. We actually currently seem to be in violation of our policy on it

tcoratger:
> > Why not sync to the same version of Rust as reth?
> 
> alloy msrv policy is stated in the readme. We actually currently seem to be in violation of our policy on it

@prestwich yes on this I didn't read carefully the README my bad. 

But I mean about the crates, as stated by @DaniPopes it could be bad to write explicitly the full version of each so maybe we should apply that rule for reth too and replace for example `tokio = { version = "1.21", default-features = false }` by `tokio = { version = "1", default-features = false }` here

https://github.com/paradigmxyz/reth/blob/925d40618890b1ec8bb5c2902ee306c7fbf7ee9e/Cargo.toml#L435

wdyt about this?

DaniPopes:
Sure that's fine

PR #908, opened 2024-06-15
feat: Add trace_replay_transaction

Ebolon:
Add `trace_replay_transaction` method.

Example usage:
```rust
use alloy_primitives::{b256};
use alloy_provider::{ext::TraceApi, ProviderBuilder};
use alloy_rpc_types_trace::parity::TraceType;

#[tokio::main]
async fn main() {
    let url = std::env::var("ETH_MAINNET_HTTP").expect("$ETH_MAINNET_HTTP must be set.");

    let result = provider
        .trace_replay_transaction(
            b256!("45162e66a30e64c5a23c5cee2180874b72322d28ee05439d2dff8d1f6a100417"),
            &[TraceType::Trace, TraceType::StateDiff],
        )
        .await
        .unwrap();
    println!("{:#?}", result);
}
```

### DIFF
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -56,6 +56,13 @@ where
         &self,
         block: BlockNumberOrTag,
     ) -> TransportResult<Vec<LocalizedTransactionTrace>>;
+
+    /// Replays a transaction.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
these were indeed missing

lgtm, ty

PR #916, opened 2024-06-17
Convenience method for conversion from `BlockId` to `u64`

jmcph4:
## Motivation

Consider the case where we need to convert a `BlockId` into a `u64`. Currently, this is achievable by the following:

```rust
fn block_id_to_u64(block_id: BlockId) -> Option<u64> {
    if let BlockId::Number(number_or_tag) = block_id {
        number_or_tag.as_number()
    } else {
        None
    }
}
```

This works but given that the `BlockId::as_block_hash` method already exists, this seems like a logical inclusion into Alloy itself.

https://github.com/alloy-rs/alloy/blob/060de7871a76e99863917e62717c97423c37cadf/crates/eips/src/eip1898.rs#L275-L281

## PR Checklist

- [x] Added Tests
- [x] ~~Added Documentation~~
- [x] ~~Breaking changes~~

### DIFF
--- a/crates/eips/src/eip1898.rs
+++ b/crates/eips/src/eip1898.rs
@@ -280,6 +280,14 @@ impl BlockId {
         }
     }
 
+    /// Returns the block number if it is [BlockId::Number] and not a tag
+    pub const fn as_block_number(&self) -> Option<u64> {
+        match self {
+            Self::Number(x) => x.as_number(),
+            _ => None,
+        }
+    }
+
     /// Returns true if this is [BlockNumberOrTag::Latest]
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip1898.rs
```diff
@@ -280,6 +280,14 @@ impl BlockId {
         }
     }
 
+    /// Returns the block number if it is [BlockId::Number] and not a tag
```
onbjerg, line 283:
```suggestion
    /// Returns the block number if it is [`BlockId::Number`] and not a tag
```
nit

onbjerg (approved):
makes sense

PR #927, opened 2024-06-17
feat: add `is_` and `as_` utils for `FilterBlockOption`

tcoratger:


crates/rpc-types-eth/src/filter.rs
```diff
@@ -193,6 +193,47 @@ impl FilterBlockOption {
...
+    pub const fn set_to_block(&self, block: BlockNumberOrTag) -> Self {
+        let from_block = if let Self::Range { from_block, .. } = self { *from_block } else { None };
+
+        Self::Range { from_block, to_block: Some(block) }
+    }
+
+    /// Pins the block hash this filter should target.
+    #[must_use]
+    #[doc(alias = "set_block_hash")]
+    pub const fn set_hash(&self, hash: B256) -> Self {
```
mattsse, line 234: we should just name this set_block_hash

mattsse (changes requested):
pedantic naming nit

crates/rpc-types-eth/src/filter.rs
tcoratger: fixed

crates/rpc-types-eth/src/filter.rs
```diff
@@ -193,6 +193,46 @@ impl FilterBlockOption {
...
+    }
+
+    /// Returns true if this is a block hash filter.
+    pub const fn is_block_hash(&self) -> bool {
+        matches!(self, Self::AtBlockHash(_))
+    }
+
+    /// Sets the block number this range filter should start at.
+    #[must_use]
+    pub const fn set_from_block(&self, block: BlockNumberOrTag) -> Self {
```
DaniPopes, line 217: set_ is wrong naming because it takes by reference

mattsse: right, perhaps we can either drop the prefix or replace with `with_`

tcoratger: Initially I just wanted to move this implementation that was already done here to reduce to a single `impl FilterBlockOption` instead of two. I hope this is ok to change the API :)

crates/rpc-types-eth/src/filter.rs
```diff
@@ -193,6 +193,46 @@ impl FilterBlockOption {
...
+        let to_block = if let Self::Range { to_block, .. } = self { *to_block } else { None };
```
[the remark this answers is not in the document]
tcoratger: Fixed

mattsse (approved):

PR #928, opened 2024-06-17
feat: add EIP-7702 support

lattejed:
Sharing our work for adding EIP-7702 transaction support to reth / revm / alloy

Current state:

- 7702 transaction support has been added to reth and alloy
- EOA code loading has been added to revm but full 7702 execution has not been tested yet
- The teardown of 7702 transactions in revm has not been added yet
- Gas calculations have not been updated and are not correct
- Compact encoding of added types in reth is currently byte-for-byte
- Overall correctness needs to be checked
- Not all tests have been added or updated

Incomplete items and points of discussion are marked with `TODO(eip7702)` 

If you have any questions about any of this work let us know

The three related PRs:

https://github.com/paradigmxyz/reth/pull/8895
https://github.com/bluealloy/revm/pull/1541
https://github.com/alloy-rs/alloy/pull/928

### DIFF
--- a/crates/eips/src/eip7702.rs
+++ b/crates/eips/src/eip7702.rs
@@ -1,0 +1,158 @@
+//! [EIP-7702] types.
+//!
+//! [EIP-7702]: https://eips.ethereum.org/EIPS/eip-7702
+
+use alloy_primitives::{Address, ChainId, U256};
+use alloy_rlp::{Decodable, Encodable};
+use core::mem;
+
+/// A list of [`Authorization`] the current transaction will use
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(
+    any(test, feature = "arbitrary"),
+    derive(proptest_derive::Arbitrary, arbitrary::Arbitrary)
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct AuthorizationList(pub Vec<Authorization>);
+
+impl AuthorizationList {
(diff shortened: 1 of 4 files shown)


prestwich:
hey :) quick organization note, in the past we've done new crates for pre-acceptance eips (like `crates/eip7547`). given the high probability of acceptance, i wouldn't be opposed to housing this in `crates/eips`

cc @mattsse this is the first time we'll be implementing a new TX type that is not active on mainnet. How would reth like to handle HF enforcement?

E.g. should we have
- `PragueTransactionEnvelope` as its own type
- `impl TransactionEnvelope { fn requires_prague(&self) -> bool }`
- something else?

crates/eips/src/eip7702.rs
```diff
@@ -0,0 +1,158 @@
...
+        let consumed = started_len - buf.len();
+        if consumed != rlp_head.payload_length {
+            return Err(alloy_rlp::Error::ListLengthMismatch {
+                expected: rlp_head.payload_length,
+                got: consumed,
+            });
+        }
+        Ok(Self { chain_id, address, nonce, y_parity, r, s })
+    }
+}
```
Rjected, line 156: For encoding it might be easier to create another struct, where the nonce is in a vec, and derive encodable / decodable on that struct, with conversions to/from Authorization instead of doing a manual implementation.

Rjected (changes requested):
one initial comment on encoding impl

onbjerg (commented):
please also update the relevant changelogs

gakonst:
do we still need this?

PR #933, opened 2024-06-18
feat: add TryFrom for GethTrace for all inner variants

### LINKED ISSUE #929: [Feature] Add more convenience functions for GethTrace
### Component

rpc

this type is a catch all enum:

https://github.com/alloy-rs/alloy/blob/e44e7dc61bba892b02ba3f2aa1d7264312420deb/crates/rpc-types-trace/src/geth/mod.rs#L109-L109

we should make it easier to convert into the variants because it's reasonable to expect a certain variant depending on the call argument.

## TODO

* add `try_into_{variant}` functions, this should return a result, where the error is new error type that contains the original value: `Result<Variant, UnexpectedTraceErr>`

```rust
#[thiserror]
struct UnexpectedTraceErr(pub GethTrace)
```

cc @Ebolon 

### Additional context

_No response_

Ebolon:
Fixes #929 

For me it feels more rust like to use `{variant}::try_from(geth_trace)` instead of `geth_trace.try_into_{variant}()`. How do you feel about that @mattsse?

Example:
```rust
use alloy_primitives::b256;
use alloy_provider::{ext::DebugApi, ProviderBuilder};
use alloy_rpc_types_trace::geth::{
    GethDebugBuiltInTracerType, GethDebugTracerConfig, GethDebugTracerType, GethDebugTracingOptions, GethDefaultTracingOptions,
    PreStateConfig, PreStateFrame, TraceResult,
};
use std::convert::{TryFrom, TryInto};

#[tokio::main]
async fn main() {
    let url = std::env::var("ETH_MAINNET_HTTP").expect("$ETH_MAINNET_HTTP must be set.");

    let tracer_opts = GethDebugTracingOptions {
        tracer: Some(GethDebugTracerType::BuiltInTracer(GethDebugBuiltInTracerType::PreStateTracer)),
        config: GethDefaultTracingOptions::default().disable_storage().disable_stack().disable_memory().disable_return_data(),
        timeout: None,
        tracer_config: GethDebugTracerConfig::default(),
    }
    .with_prestate_config(PreStateConfig { diff_mode: Some(true) });

    let call_result =
        provider.debug_trace_block_by_hash(b256!("1b9f245b55515d17fd5b3467bfe595670c9599b3d54837cd32cfde5e2e667830"), tracer_opts).await;

    let trace_results = match call_result {
        Ok(trace_results) => trace_results,
        Err(e) => {
            println!("Error: {:?}", e);
            return;
        }
    };

    for trace_result in trace_results {
        match trace_result {
            TraceResult::Success { result, tx_hash } => {
                let pre_state_frame = PreStateFrame::try_from(result);
                assert!(pre_state_frame.is_ok());
            }
            TraceResult::Error { .. } => {}
        }
    }
}

```

### DIFF
--- a/crates/rpc-types-trace/src/geth/mod.rs
+++ b/crates/rpc-types-trace/src/geth/mod.rs
@@ -130,7 +130,7 @@ pub enum GethTrace {
 
 impl GethTrace {
     /// Try to convert the inner tracer to [DefaultFrame]
-    pub fn try_into_default(self) -> Result<DefaultFrame, UnexpectedTracerError> {
+    pub fn try_into_default_frame(self) -> Result<DefaultFrame, UnexpectedTracerError> {
         match self {
             Self::Default(inner) => Ok(inner),
             _ => Err(UnexpectedTracerError(self)),
@@ -138,7 +138,7 @@ impl GethTrace {
     }
 
     /// Try to convert the inner tracer to [CallFrame]
-    pub fn try_into_call(self) -> Result<CallFrame, UnexpectedTracerError> {
+    pub fn try_into_call_frame(self) -> Result<CallFrame, UnexpectedTracerError> {
         match self {
             Self::CallTracer(inner) => Ok(inner),
             _ => Err(UnexpectedTracerError(self)),
@@ -146,7 +146,7 @@ impl GethTrace {
     }
 
     /// Try to convert the inner tracer to [FourByteFrame]
-    pub fn try_into_four_byte(self) -> Result<FourByteFrame, UnexpectedTracerError> {
+    pub fn try_into_four_byte_frame(self) -> Result<FourByteFrame, UnexpectedTracerError> {
         match self {
             Self::FourByteTracer(inner) => Ok(inner),
             _ => Err(UnexpectedTracerError(self)),
@@ -154,7 +154,7 @@ impl GethTrace {
     }
 
     /// Try to convert the inner tracer to [PreStateFrame]
-    pub fn try_into_pre_state(self) -> Result<PreStateFrame, UnexpectedTracerError> {
+    pub fn try_into_pre_state_frame(self) -> Result<PreStateFrame, UnexpectedTracerError> {
         match self {
             Self::PreStateTracer(inner) => Ok(inner),
             _ => Err(UnexpectedTracerError(self)),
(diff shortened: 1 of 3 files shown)


onbjerg (changes requested):
please also update the corresponding changelog

mattsse:
> please also update the corresponding changelog

aren't those auto-generated?

mattsse (changes requested):
I'm okay with the tryfrom impls, but I personally would prefer gethtrace functions instead, so you can do something like:

```rust
let trace = provider.trace(..).await?.try_into_default()?;
```

without providing type hints, but we can do both, tryfrom and the functions

onbjerg:
> > please also update the corresponding changelog
> 
> aren't those auto-generated?

if they are it should be documented, because i am not aware

Ebolon:
@mattsse please have a look again.

PS: If you wonder why the clippy fixes: I used the command from the `CONTRIBUTING.md` in differs from the CI job.

crates/rpc-types-trace/src/geth/mod.rs
```diff
@@ -123,6 +128,64 @@ pub enum GethTrace {
     JS(serde_json::Value),
 }
 
+impl GethTrace {
+    /// Try to convert the inner tracer to [DefaultFrame]
+    pub fn try_into_default(self) -> Result<DefaultFrame, UnexpectedTracerError> {
+        match self {
+            Self::Default(inner) => Ok(inner),
+            _ => Err(UnexpectedTracerError(self)),
+        }
```
mattsse, line 137: these are now redundant duplicates and can just reuse the tryfrom impl `self.try_into()`

mattsse (changes requested):

crates/rpc-types-trace/src/geth/mod.rs
Ebolon: I removed them fully. See next comment for reasoning.

Ebolon:
I gave the topic a fresh look and came to the conclusion that the TryFrom implementation is redundant at all. From api perspective you initial idea has a better flow and and is easier to use. I adapted the naming of the variants in the function names. My feeling is that it is much better to have the full type name in there. It is easier to read without IDE type hinting and does not maybe lead to a mixup with e.g. Default(). 

@mattsse feel free to have a look again

mattsse (approved):

PR #938, opened 2024-06-18
docs: update get_balance docs

### LINKED ISSUE #937: [Bug] Some RPC methods don't expose BlockId
### Component

provider, pubsub

v0.1.1

### Operating System

macOS (Apple Silicon)

### Describe the bug

Some rpc methods are documented as also accepting a block height parameter, but they currently don't expose it.

Example: https://github.com/alloy-rs/alloy/blob/041c7c47b33907523cbe4023db03ae7bb57bf7bd/crates/provider/src/provider/trait.rs#L250-L254

This is the case for: get_account, get_balance, get_proof (as far as I can tell, I didn't check all other methods)

If you have a solution in mind I'm happy to open a PR for it

mattsse:
closes #937

outdated docs re blockid

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -248,7 +248,9 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         RpcWithBlock::new(self.weak_client(), "eth_getAccount", address)
     }
 
-    /// Gets the balance of the account at the specified tag, which defaults to latest.
+    /// Gets the balance of the account.
+    ///
+    /// Defaults to the latest block. See also [RpcWithBlock::block_id].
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -248,7 +248,9 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         RpcWithBlock::new(self.weak_client(), "eth_getAccount", address)
     }
 
-    /// Gets the balance of the account at the specified tag, which defaults to latest.
+    /// Gets the balance of the account.
+    ///
+    /// Defaults to the latest block. See also [RpcWithBlock::block_id].
```
onbjerg, line 253:
```suggestion
    /// Defaults to the latest block. See also [`RpcWithBlock::block_id`].
```

PR #939, opened 2024-06-18
add range test in `FilterBlockOption`

tcoratger:
## Summary

Currently there are no checks in the `with_from_block` and `with_to_block` methods to ensure the validity of the given block numbers and thus avoid false ranges.

This pull request adds unit tests for the `FilterBlockOption` struct, specifically focusing on the `with_from_block` and `with_to_block` methods. These tests ensure that the methods correctly handle various scenarios, including valid and invalid block ranges.

### `FilterBlockOption` Methods

- **`with_from_block`**: Added a check to ensure that the `from_block` value is not greater than the `to_block` value. If this condition is violated, the method will panic with an appropriate error message.
- **`with_to_block`**: Added a check to ensure that the `to_block` value is not less than the `from_block` value. If this condition is violated, the method will panic with an appropriate error message.

### Unit Tests

Unit tests are added to verify the behavior of the `with_from_block` and `with_to_block` methods.

### DIFF
--- a/crates/rpc-types-eth/src/filter.rs
+++ b/crates/rpc-types-eth/src/filter.rs
@@ -153,6 +153,15 @@ impl From<U256> for Topic {
     }
 }
 
+/// Represents errors that can occur when setting block filters in `FilterBlockOption`.
+#[derive(Debug, PartialEq, Eq)]
+pub enum FilterBlockError {
+    /// Error indicating that the `from_block` is greater than the `to_block`.
+    FromBlockGreaterThanToBlock,
+    /// Error indicating that the `to_block` is less than the `from_block`.
+    ToBlockLessThanFromBlock,
+}
+
 /// Represents the target range of blocks for the filter
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum FilterBlockOption {
@@ -213,8 +222,7 @@ impl FilterBlockOption {
     }
 
     /// Sets the block number this range filter should start at.
-    #[must_use]
-    pub fn with_from_block(&self, block: BlockNumberOrTag) -> Self {
+    pub fn with_from_block(&self, block: BlockNumberOrTag) -> Result<Self, FilterBlockError> {
         // Check if from_block is greater than to_block
         if self
             .get_to_block()
@@ -222,15 +230,14 @@ impl FilterBlockOption {
             .zip(block.as_number())
             .map_or(false, |(to, from)| from > to)
         {
-            panic!("from_block cannot be greater than to_block");
+            return Err(FilterBlockError::FromBlockGreaterThanToBlock);
         }
 
-        Self::Range { from_block: Some(block), to_block: self.get_to_block().copied() }
+        Ok(Self::Range { from_block: Some(block), to_block: self.get_to_block().copied() })
     }
 
     /// Sets the block number this range filter should end at.
-    #[must_use]
-    pub fn with_to_block(&self, block: BlockNumberOrTag) -> Self {
+    pub fn with_to_block(&self, block: BlockNumberOrTag) -> Result<Self, FilterBlockError> {
         // Check if to_block is less than from_block
         if self
             .get_from_block()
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/filter.rs
```diff
@@ -215,12 +215,32 @@ impl FilterBlockOption {
...
     pub fn with_from_block(&self, block: BlockNumberOrTag) -> Self {
+        // Check if from_block is greater than to_block
+        if self
+            .get_to_block()
+            .and_then(|to| to.as_number())
+            .zip(block.as_number())
+            .map_or(false, |(to, from)| from > to)
+        {
+            panic!("from_block cannot be greater than to_block");
+        }
```
mattsse, line 226: hmm, we should not panic here.

worst thing that can happen when sending an invalid range is an rpc error as response

mattsse (changes requested):
I'm not convinced that panicking is the right call here.

perhaps we can solve this with a separate function `fn ensure_valid_block_range(self) -> Result` instead?
this way it's up to the caller to include this check or not

crates/rpc-types-eth/src/filter.rs
tcoratger: Yes true, didn't think about this properly

tcoratger:
> I'm not convinced that panicking is the right call here.
> 
> perhaps we can solve this with a separate function `fn ensure_valid_block_range(self) -> Result` instead? this way it's up to the caller to include this check or not

@mattsse I suggest something else, tell me what you think.

I renamed the methods that existed before with an `_unchecked` suffix so the user knows explicitly that nothing will be checked for them.

I created two new methods without the suffix for cases where the user wants there to be checking.

I know that this breaks the previous API but the user will notice it because the return type is not the same.

crates/rpc-types-eth/src/filter.rs
```diff
@@ -213,34 +222,44 @@ impl FilterBlockOption {
     }
 
     /// Sets the block number this range filter should start at.
-    #[must_use]
-    pub fn with_from_block(&self, block: BlockNumberOrTag) -> Self {
+    pub fn with_from_block(&self, block: BlockNumberOrTag) -> Result<Self, FilterBlockError> {
```
mattsse, line 225: this is very inconvenient, we should not do this.

I'm fine with an additional ensure_ function that does this check, but this should not be part of the builder functions because not useful

mattsse (changes requested):

crates/rpc-types-eth/src/filter.rs
tcoratger: Just fixed that, let me know :)

I tried, apparently you didn't like it haha

In fact, it was while testing our RPC that I realized that it seemed much too easy to produce erroneous block ranges and then spend several minutes before understanding what was wrong.

For example in our case we filter with the MongoDB Rust client and therefore obviously if I filter the blocks from 15 to 2 it doesn't work but I had to do a lot of prints before understanding that I was doing that :)

So I thought that an unchecked method (for the more confident) and a checked method (for the more cautious) could be a good idea. But let's do this with an `ensure_` method that at least allows us to have the option of adding risk insurance.

crates/rpc-types-eth/src/filter.rs
```diff
@@ -153,6 +153,15 @@ impl From<U256> for Topic {
     }
 }
 
+/// Represents errors that can occur when setting block filters in `FilterBlockOption`.
+#[derive(Debug, PartialEq, Eq)]
+pub enum FilterBlockError {
```
mattsse, line 158: this should be an error type

crates/rpc-types-eth/src/filter.rs
```diff
@@ -212,6 +221,20 @@ impl FilterBlockOption {
         matches!(self, Self::AtBlockHash(_))
     }
 
+    /// Ensure block range validity
+    pub fn ensure_block_range(&self) -> Result<(), FilterBlockError> {
```
mattsse, line 225:
```suggestion
    pub fn ensure_valid_block_range(&self) -> Result<(), FilterBlockError> {
```
```suggestion
    pub fn ensure_block_range(&self) -> Result<(), FilterBlockError> {
```

mattsse (changes requested):
sorry forgot about this one,

one smol thing missing

crates/rpc-types-eth/src/filter.rs
tcoratger: Fixed

crates/rpc-types-eth/src/filter.rs
tcoratger: Fixed

mattsse (approved):

PR #940, opened 2024-06-18
chore: relax version in workspace dependencies

DaniPopes:
https://github.com/crate-ci/cargo-release/blob/f0b5af40b903aedd103992366d0875fe2b9f2f99/docs/faq.md?plain=1#L133-L137

> ## How do I do a release when there is dependency cycle in my workspace?
> 
> If this is for dev-dependencies, just declare your dev-dependency with only a path, no version, and it should work out.
> 
> If you have other cycles, open an issue, we'd love to hear about your use case and see how we can help!

This should be equivalent since we pin to latest version anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,36 +31,36 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [workspace.dependencies]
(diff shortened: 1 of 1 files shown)


mattsse (approved):

DaniPopes:
Hmm actually not specifying `version` means wildcard `*`

Issue #926, opened 2024-06-17
[Feature] Add missing RPC namespaces and missing RPC methods


zerosnacks:
### Component

rpc

### Describe the feature you would like

Achieve full compatibility with Reth's JSON-RPC methods and namespaces

Link: https://ethereum.org/en/developers/docs/apis/json-rpc/#json-rpc-methods

Missing namespaces:

- [x] [`rpc-types-web3`](https://reth.rs/jsonrpc/web3.html): https://github.com/alloy-rs/alloy/pull/996
- [x] [`rpc-types-net`](https://reth.rs/jsonrpc/net.html)
- [x] [`rpc-types-rpc`](https://reth.rs/jsonrpc/rpc.html)
- [x] [`rpc-types-mev`](https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-types/src/mev.rs): https://github.com/paradigmxyz/reth/issues/9027)

Missing methods:

For [Trace](https://reth.rs/jsonrpc/trace.html):

- [x] trace_rawTransaction: #925
- [x] trace_replayBlockTransactions: #925
- [x] trace_filter: #946
- [x] trace_get: https://github.com/alloy-rs/alloy/pull/987

For [Admin](https://reth.rs/jsonrpc/admin.html):

- [x] admin_peerEvents_subscribe: https://github.com/alloy-rs/alloy/pull/991
- [x] ~~admin_peerEvents_unsubscribe~~

For [Debug](https://reth.rs/jsonrpc/debug.html):

- [x] debug_getRawHeader
- [x] debug_getRawBlock
- [x] debug_getRawTransaction
- [x] debug_getRawReceipts
- [x] debug_getBadBlocks
- [x] debug_traceChain
- [x] debug_traceBlock

DaniPopes:
Yeah, sorry about that, I updated the remaining methods with their PR. Thanks to everyone!

PR #946, opened 2024-06-19
feat: add trace_filter method

DoTheBestToGetTheBest:
this should mark trace_filter as done here https://github.com/alloy-rs/alloy/issues/926

### DIFF
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -56,6 +56,12 @@ where
         trace_type: &[TraceType],
     ) -> TransportResult<TraceResults>;
 
+    /// Traces matching given filter.
+    async fn trace_filter(
+        &self,
+        tracer: TraceFilter,
+    ) -> TransportResult<Vec<LocalizedTransactionTrace>>;
+
     /// Trace all transactions in the given block.
     ///
     /// # Note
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
ty.

this should be moved to the traceApi trait instead and use &

DoTheBestToGetTheBest:
> ty.
> 
> this should be moved to the traceApi trait instead and use &

ty so much for your fast review, changes has been made right now

crates/provider/src/ext/trace.rs
```diff
@@ -56,6 +57,12 @@ where
         trace_type: &[TraceType],
     ) -> TransportResult<TraceResults>;
 
+    /// Traces matching given filter.
+    async fn trace_filter(
+        &self,
+        tracer: TraceFilter,
```
mattsse, line 63:
```suggestion
        tracer: &TraceFilter,
```

crates/provider/src/ext/trace.rs
```diff
@@ -112,6 +119,13 @@ where
         self.client().request("trace_transaction", (hash,)).await
     }
 
+    async fn trace_filter(
+        &self,
+        tracer: TraceFilter,
```
mattsse, line 124:
```suggestion
        tracer: &TraceFilter,
```

mattsse (approved):

PR #949, opened 2024-06-20
chore(eips): make `sha2` optional, add `kzg-sidecar` feature

DaniPopes:


mattsse (commented):
I believe we use this without c-kzg feature in some places

mattsse (approved):
that makes sense

PR #950, opened 2024-06-20
feat: add eip-7702 helpers

onbjerg:
## Motivation

EIP-7702 is up for inclusion in Prague.

## Solution

Adds helpers and some types specified in [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702).

The transaction type itself will be added in another PR as it is not relevant right now (for example, the RPC spec has not been updated yet).

This also adds a `k256` feature to `alloy_eips` since authority recovery requires signature recovery which is only possible with `k256` enabled in `alloy_primitives`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/Cargo.toml
+++ b/crates/eips/Cargo.toml
@@ -70,6 +70,7 @@ serde = [
 kzg = ["kzg-sidecar", "sha2", "dep:derive_more", "dep:c-kzg", "dep:once_cell"]
 kzg-sidecar = ["sha2"]
 sha2 = ["dep:sha2"]
+k256 = ["alloy-primitives/k256"]
 ssz = [
     "std",
     "dep:ethereum_ssz",
--- a/crates/eips/src/eip7702/auth_list.rs
+++ b/crates/eips/src/eip7702/auth_list.rs
@@ -64,7 +64,7 @@ impl Authorization {
     }
 
     /// Convert to a signed authorization by adding a signature.
-    pub fn into_signed<S>(self, signature: S) -> SignedAuthorization<S> {
+    pub const fn into_signed<S>(self, signature: S) -> SignedAuthorization<S> {
         SignedAuthorization { inner: self, signature }
     }
 }
--- a/crates/eips/src/eip7702/auth_list.rs
+++ b/crates/eips/src/eip7702/auth_list.rs
@@ -6,7 +6,7 @@ use alloy_primitives::{keccak256, Address, ChainId, B256};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodable, RlpEncodable};
 
 /// An unsigned EIP-7702 authorization.
-#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable, Eq, PartialEq)]
 pub struct Authorization {
     /// The chain ID of the authorization.
     pub chain_id: ChainId,
@@ -110,7 +110,7 @@ impl<S> Deref for SignedAuthorization<S> {
 /// nonce was specified (i.e. `None`). If there is 1 item, this is the same as `Some`.
 ///
 /// The wrapper type is used for RLP encoding and decoding.
-#[derive(Default, Debug, Copy, Clone)]
+#[derive(Default, Debug, Copy, Clone, Eq, PartialEq)]
 pub struct OptionalNonce(Option<u64>);
 
 impl OptionalNonce {
@@ -161,5 +161,46 @@ impl Deref for OptionalNonce {
 
     fn deref(&self) -> &Self::Target {
         &self.0
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn test_encode_decode_roundtrip(auth: Authorization) {
+        let mut buf = Vec::new();
+        auth.encode(&mut buf);
+        let decoded = Authorization::decode(&mut buf.as_ref()).unwrap();
+        assert_eq!(buf.len(), auth.length());
+        assert_eq!(decoded, auth);
+    }
+
+    #[test]
+    fn test_encode_decode_auth() {
+        // fully filled
+        test_encode_decode_roundtrip(Authorization {
+            chain_id: 1u64,
+            address: Address::left_padding_from(&[6]),
+            nonce: Some(1u64).into(),
+        });
+
+        // no nonce
+        test_encode_decode_roundtrip(Authorization {
+            chain_id: 1u64,
+            address: Address::left_padding_from(&[6]),
+            nonce: None.into(),
+        });
+    }
+
+    #[test]
+    fn opt_nonce_too_many_elements() {
+        let mut buf = Vec::new();
+        vec![1u64, 2u64].encode(&mut buf);
+
+        assert_eq!(
+            OptionalNonce::decode(&mut buf.as_ref()),
+            Err(alloy_rlp::Error::UnexpectedLength)
+        )
     }
 }
--- a/crates/eips/src/eip7702/auth_list.rs
+++ b/crates/eips/src/eip7702/auth_list.rs
@@ -2,18 +2,17 @@ use core::ops::Deref;
 
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
+use alloy_primitives::{keccak256, Address, ChainId, B256};
 #[cfg(feature = "k256")]
-use alloy_primitives::{keccak256, SignatureError, B256};
-use alloy_primitives::{Address, ChainId, Signature};
-use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpEncodable};
+use alloy_primitives::{Signature, SignatureError};
+use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodable, RlpEncodable};
 
-/// An EIP-7702 authorization.
-#[derive(Debug, Clone, RlpEncodable)]
+/// An unsigned EIP-7702 authorization.
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
 pub struct Authorization {
     chain_id: ChainId,
     address: Address,
     nonce: OptionalNonce,
-    signature: Signature,
 }
 
 impl Authorization {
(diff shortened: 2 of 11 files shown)


crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// An EIP-7702 authorization list.
+pub type AuthorizationList = Vec<Authorization>;
+
+/// An EIP-7702 authorization.
+#[derive(Debug, Clone, RlpEncodable)]
+pub struct Authorization {
+    chain_id: ChainId,
+    address: Address,
+    nonce: OptionalNonce,
```
mattsse, line 18: these fields are the raw authorization, right?

so would it make sense to have this as a standalone type? because I guess this authorization is then signed?

similar to Tx and the Signed type

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,119 @@
...
+    /// item.
+    ///
+    /// The authority prehash is `keccak(MAGIC || rlp([chain_id, [nonce], address]))`
+    #[inline]
+    #[cfg(feature = "k256")]
+    fn authority_prehash(&self) -> B256 {
+        #[derive(RlpEncodable)]
+        struct Auth {
+            chain_id: ChainId,
+            nonce: OptionalNonce,
```
mattsse, line 62: is this the correct order?

this is a bit weird, I would expect that the optional stuff is the last thing

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,119 @@
...
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+use alloy_primitives::{keccak256, Address, ChainId, Signature, SignatureError, B256};
+use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpEncodable};
+
+use super::constants::MAGIC;
+
+/// An EIP-7702 authorization list.
+pub type AuthorizationList = Vec<Authorization>;
```
mattsse, line 11: I'd like to remove this aliases, imo aliases for vec types are always bad

mattsse (changes requested):

crates/eips/src/eip7702/auth_list.rs
onbjerg: yes, it is correct.

>At the start of executing the transaction, for each [chain_id, address, [nonce], y_parity, r, s] tuple:
>
>authority = ecrecover(keccak(MAGIC || rlp([chain_id, [nonce], address])), y_parity, r, s]

crates/eips/src/eip7702/auth_list.rs
onbjerg: makes sense

crates/eips/src/eip7702/auth_list.rs
onbjerg: ok

crates/eips/src/eip7702/auth_list.rs
onbjerg: a bit annoying, can't reuse `Signed<T>` because it's in `alloy-consensus`, will take a gander though

crates/eips/src/eip7702/auth_list.rs
mattsse: jfc

crates/eips/src/eip7702/auth_list.rs
onbjerg: I split it into a signed/unsigned type, and I also made it generic over the signature in line with the transaction type. The only hurdle here is that we cannot default `Sig` to `Signature`. wdyt?

Rjected (changes requested):
one comment on validation

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,140 @@
...
+#[cfg(feature = "k256")]
+use alloy_primitives::{Signature, SignatureError};
+use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodable, RlpEncodable};
+
+/// An unsigned EIP-7702 authorization.
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+pub struct Authorization {
+    chain_id: ChainId,
+    address: Address,
+    nonce: OptionalNonce,
```
mattsse, line 15: can we make these pub, otherwise there's no way to instantiate them

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,140 @@
+use core::ops::Deref;
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+use alloy_primitives::{keccak256, Address, ChainId, B256};
+#[cfg(feature = "k256")]
+use alloy_primitives::{Signature, SignatureError};
```
mattsse, line 7: can we use the full path instead of putting imports behind features?

crates/eips/src/eip7702/auth_list.rs
DaniPopes: I'd prefer a constructor

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// The signature hash is `keccak(MAGIC || rlp([chain_id, [nonce], address]))`
+    #[inline]
+    pub fn signature_hash(&self) -> B256 {
+        use super::constants::MAGIC;
+
+        #[derive(RlpEncodable)]
+        struct Auth {
+            chain_id: ChainId,
+            nonce: OptionalNonce,
```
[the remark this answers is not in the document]
onbjerg: yes because the order of the fields are different:)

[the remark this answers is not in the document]
onbjerg: https://github.com/ethereum/EIPs/pull/8679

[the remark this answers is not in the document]
mattsse: I see, I really hope this is just an oversight in the eip -.-

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,158 @@
...
+                nonce.encode(out);
+            }
+            None => Header { list: true, payload_length: 0 }.encode(out),
+        }
+    }
+}
+
+impl Decodable for OptionalNonce {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let list: Vec<u64> = Vec::decode(buf)?;
```
mattsse, line 143: this should be okay, but could be done either manually or perhaps even with the arrayvec feature I assume, for now this is reasonable.

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+}
+
+/// An internal wrapper around an `Option<u64>` for optional nonces.
+///
+/// In EIP-7702 the nonce is encoded as a list of either 0 or 1 items, where 0 items means that no
+/// nonce was specified (i.e. `None`). If there is 1 item, this is the same as `Some`.
+///
+/// The wrapper type is used for RLP encoding and decoding.
+#[derive(Debug, Copy, Clone)]
```
mattsse, line 113: add Default ?

mattsse (approved):

crates/eips/src/eip7702/auth_list.rs
onbjerg: did manual impl tal

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,165 @@
...
+impl Decodable for OptionalNonce {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let mut bytes = Header::decode_bytes(buf, true)?;
+        if bytes.is_empty() {
+            return Ok(Self(None));
+        }
+
+        let payload_view = &mut bytes;
+        let nonce = u64::decode(payload_view)?;
+        if !payload_view.is_empty() {
```
Rjected, line 150: wouldn't this error if there are for example more things in left to be decoded?

Rjected (approved):
lgtm

crates/eips/src/eip7702/auth_list.rs
mattsse: I belive this is correct.

maybe we can a few simple rlp roundtrip tests for Authorization

mattsse (changes requested):
can we has some sanity tests? :)

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -0,0 +1,206 @@
...
+
+    /// Convert to a signed authorization by adding a signature.
+    pub const fn into_signed<S>(self, signature: S) -> SignedAuthorization<S> {
+        SignedAuthorization { inner: self, signature }
+    }
+}
+
+/// A signed EIP-7702 authorization.
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+pub struct SignedAuthorization<S> {
```
rakita, line 75: Can we add  `RecoveredAuthorization`? Where we will cache the recovered account as Option<Address> (Option as it can fail).
Recovery can be expensive and can be done before revm gets called. We are okay to do that as it is covered by gas.

Idea that I have is to have a enum as in Revm input:
```
AuthorisationList {
   Recovered(Vec<RecoveredAuthorization>),
   SignedAuthorization(Vec<SignedAuthorization>),
}
```

crates/eips/Cargo.toml
```diff
@@ -70,6 +70,7 @@ serde = [
 kzg = ["kzg-sidecar", "sha2", "dep:derive_more", "dep:c-kzg", "dep:once_cell"]
 kzg-sidecar = ["sha2"]
 sha2 = ["dep:sha2"]
+k256 = ["alloy-primitives/k256"]
```
rakita, line 73: Would be good to have `seck256k1` lib/feature flag as it is faster. But looking at it as a low priority.

mattsse (approved):

crates/eips/src/eip7702/auth_list.rs
onbjerg: added

rakita (approved):
lgtm

DaniPopes (approved):

Issue #770, opened 2024-05-22
[Feature] Add a `CachedReqwestProvider` to cache RPC requests using a `ReqwestProvider`


puma314:
### Component

provider, pubsub

### Describe the feature you would like

For use-cases like `SP1-Reth` or `Kona`, we often want to execute a (historical) block, but we don't have the entire state in memory and we execute this block with a `ProviderDb` that fetches accounts, storage, etc. using an RPC. Fetching from the network is slow and often takes minutes for all of the accesses required for an entire block.

Often we re-run these blocks to debug things or tune performance, etc. and each time the feedback loop on iteration is very slow because it requires waiting for all the network requests each time. It would be nice to add a very simple caching layer on top of `ReqwestProvider` that can cache the results of RPC calls to a file (or some other easy to set up format) and then first check the cache before sending a network request. 

This would speed up iteration time for use-cases like `Kona` and `SP1-Reth` tremendously.

An interface like this might make sense:
```
let provider = ReqwestProvider::new_http(rpc_url).cache("my_file.txt")
```

In our case, we are usually querying old blocks (not near the tip of the chain), so re-org awareness is not important for our use-case. We just want a really simple caching layer.

gakonst:
Could this be a tower layer?

Seeing https://docs.rs/tower/latest/tower/ready_cache/cache/struct.ReadyCache.html - cc @mattsse does this work?

prestwich:
We won't add caching at the `Transport` layer via `tower` because caching (unlike rate limiting or retrying) needs to be aware of the RPC semantics and potentially the provider heartbeat task, so that it can invalidate caches on new blocks and reorgs. This means we need it to be a provider `alloy_provider::Layer` producing `CachingProvider<P, T, N>`, rather than a `tower::Layer` producing `CachingTransport<T>`. 

This is blocked by #736 (which is pretty straightforward to resolve)

Is the use case here making a high volume of requests against specific deep historical states? It sounds like you actually don't want to cache to a file. You want an in-memory cache that is persisted to a file when your program stops? I'm in general not in favor of caching to/from a file directly, as responses get invalidated so regularly, fs access degrades perf, and the target user for alloy doesn't have an archive node and doesn't make queries against the deep state. Would it be enough to have the cache internals be (de)serializable and a way to instantiate the cache with data in it?

gakonst:
[...]

PR #954, opened 2024-06-21
feat(provider): LRUCache Layer

yash-atreya:
## Motivation

Ref: #770 

## Solution

- Adds a provider caching layer that dumps cache to fs. 
- Responses are saved as key-value pairs where `key` is the hash of the request and its params. 

Overrides the following RPC methods

- [x] `get_block_by_number`
- [x] `get_block_by_hash`
- [x] `get_block_receipts`
- [x] `get_proof`
- [x] `get_storage_at`
- [x] `get_code_at`
- [x] `get_transaction_count`
- [x] `get_logs`
- [x] `get_transaction_by_hash`
- [x] `get_raw_transaction_by_hash`
- [x] `get_transaction_receipt`

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,150 @@
...
+use std::{marker::PhantomData, num::NonZeroUsize, path::PathBuf};
+use tokio::sync::Mutex;
+
+// TODO: Populate load cache from file on initialization.
+// TODO: Add method to dump cache to file.
+/// A provider layer that caches RPC responses and serves them on subsequent requests.
+#[derive(Debug, Clone)]
+pub struct CacheLayer {
+    path: PathBuf,
+    max_items: usize,
```
mattsse, line 17: we can put these types into a config struct, so we can easily extend this, perhaps this should even store the actual cache

mattsse (changes requested):

crates/provider/src/layers/cache.rs
yash-atreya: +1 on moving it to a config struct. But don't think it should store the actual cache since that would requires us to:

```rust
struct CacheLayer {
cache: RwLock<Lru>,
}

struct CacheProvider {
layer: CacheLayer
}
```

This is an anti-pattern as the user initializes the layer like so `.layer(CacheLayer::new(path, max_items)` while initializing the `Provider` using `ProviderBuilder`. And in turn, the `CacheProvider` is initialized.

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,349 @@
...
+        let val = cache.get(key).cloned();
+        Ok(val)
+    }
+
+    /// Saves the cache to a file specified by the path.
+    /// If the files does not exist, it creates one.
+    /// If the file exists, it overwrites it.
+    pub async fn save_cache(&self, path: PathBuf) -> TransportResult<()> {
+        let cache = self.cache.read().await;
+        let file = std::fs::File::create(path).map_err(TransportErrorKind::custom)?;
```
mattsse, line 118: flip

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,349 @@
...
+    use super::*;
+
+    #[tokio::test]
+    async fn test_cache_provider() {
+        let cache = CacheLayer::new(100);
+        let anvil = Anvil::new().block_time_f64(0.3).spawn();
+        let provider = ProviderBuilder::default().layer(cache).on_http(anvil.endpoint_url());
+
+        let path = PathBuf::from_str("./rpc-cache.txt").unwrap();
+        provider.load_cache(path.clone()).await.unwrap();
```
mattsse, line 306: this shouldn't be necessary, path should be part of the config instead

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,349 @@
...
+        hydrate: bool,
+    ) -> TransportResult<Option<Block>> {
+        let hash =
+            RequestType::GetBlockByNumber((number, hydrate)).params_hash(self.inner.client())?;
+
+        if let Some(block) = self.get(&hash).await? {
+            let block = serde_json::from_str(&block).map_err(TransportErrorKind::custom)?;
+            println!("Cache hit");
+            return Ok(Some(block));
+        }
```
mattsse, line 172: this has to be repeated for every endpoint that we cache.

this should be simpler, either with a helper function or a macro

mattsse (changes requested):

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,214 @@
...
+use alloy_json_rpc::{RequestPacket, Response, ResponsePacket, ResponsePayload};
+use alloy_primitives::B256;
+use lru::LruCache;
+use serde::{Deserialize, Serialize};
+use serde_json::value::RawValue;
+use std::{
+    io::BufReader,
+    num::NonZeroUsize,
+    path::PathBuf,
+    sync::{Arc, RwLock},
```
mattsse, line 11: this should use parking_lot rwlock

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,214 @@
...
+            .map(|(key, value)| FsCacheEntry { key: *key, value: value.clone() })
+            .collect::<Vec<_>>();
+        serde_json::to_writer(file, &entries).map_err(TransportErrorKind::custom)?;
+        Ok(())
+    }
+
+    /// Loads the cache from a file specified by the path.
+    /// If the file does not exist, it returns without error.
+    pub fn load_cache(&self) -> TransportResult<()> {
+        println!("Loading cache...");
```
mattsse, line 127: replace this with a trace

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,214 @@
...
+                                }
+                                Err(e) => Err(e),
+                            }
+                        })
+                    }
+                    Err(e) => Box::pin(async move { Err(e) }),
+                }
+            }
+            RequestPacket::Batch(_reqs) => {
+                todo!()
```
mattsse, line 202: we should just forward these as well

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,214 @@
...
+                let resp = this.get(&params_hash);
+                match resp {
+                    Ok(Some(resp)) => {
+                        println!("Cache hit!");
+                        let raw = RawValue::from_string(resp).unwrap();
+                        let payload: ResponsePayload<Box<RawValue>, Box<RawValue>> =
+                            ResponsePayload::Success(raw);
+                        let response = Response { id: ser_req.id().clone(), payload };
+
+                        Box::pin(async move { Ok(ResponsePacket::Single(response)) })
```
mattsse, line 178: we can extract this to a new function that returns the rawvalue from the cache, if this is none, we send the request, if some we return

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,214 @@
...
+        let loaded = service.load_cache();
+
+        match loaded {
+            Ok(_) => {
+                tracing::info!("Loaded cache");
+            }
+            Err(e) => {
+                tracing::info!("Error loading cache: {:?}", e);
+            }
+        }
```
mattsse, line 89: I think you can use inspect and/or inspect_err here

mattsse (changes requested):
some suggestions

crates/transport/Cargo.toml
```diff
@@ -31,6 +32,8 @@ tower.workspace = true
 url.workspace = true
 tracing.workspace = true
 tokio = { workspace = true, features = ["rt", "time"] }
+lru = "0.12"
+parking_lot = "0.12.3"
```
mattsse, line 36: can we move this to workspace, if not already?

crates/provider/src/provider/trait.rs
```diff
@@ -1022,6 +1024,25 @@ mod tests {
         assert_eq!(0, num);
     }
 
+    #[tokio::test]
+    async fn test_transport_cache_layer() {
+        init_tracing();
+        let anvil = Anvil::new().block_time_f64(0.3).spawn();
+        let cache_layer = CacheLayer::new(10, "./transport-rpc-cache.txt".into());
+        let client = ClientBuilder::default().layer(cache_layer).http(anvil.endpoint_url());
```
mattsse, line 1032: when will the service be created?
is this on a per-request level? or only once and the service is reused across requests?

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub max_items: usize,
+    /// Path of the cache file.
+    pub path: PathBuf,
+}
+
+impl<S> Layer<S> for CacheLayer {
+    type Service = CachingService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        CachingService::new(inner, self.config.clone())
```
mattsse, line 55: this will create a new cache per service

I assume we'd like to reuse the cache here? so we'd need to lift the cache to the layer?

mattsse (commented):
some questions

crates/transport/src/layers/cache.rs
yash-atreya: Yeah moving it.

crates/provider/src/provider/trait.rs
yash-atreya: The service is created on a per-request level.

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,230 @@
...
+    type Service = CachingService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        CachingService::new(inner, self.clone())
+    }
+}
+
+impl Drop for CacheLayer {
+    fn drop(&mut self) {
+        let _ = self.save_cache();
```
mattsse, line 126: does this save on every request?

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,230 @@
...
+                        let resp = this.handle_cache_hit(ser_req.id().to_owned(), raw);
+                        Box::pin(async move { Ok(resp) })
+                    }
+                    Ok(None) => {
+                        Box::pin(async move {
+                            match inner.call(req).await {
+                                Ok(resp) => {
+                                    // Store success response into cache.
+                                    if let Some(res) = resp.single_response() {
+                                        let ser = res.payload.as_success().unwrap().to_string();
```
mattsse, line 200: this unwrap seems dangerous

crates/transport/src/layers/cache.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), TransportError>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: RequestPacket) -> Self::Future {
+        let mut inner = self.inner.clone();
+        let this = self.clone();
+        match req.clone() {
+            RequestPacket::Single(ser_req) => {
+                let params_hash = ser_req.params_hash();
```
mattsse, line 188: hmm this logic has a critical flaw, because block tags like latest will result in the same hash and therefor return an outdated response.

not sure how to mitigate this

yash-atreya:
Decided to make this a provider-layer, this lets us achieve two very important features that wouldn't work as a transport layer: 

1. Let users access `save_cache` and `load_cache` on demand via the provider interface. In the case of a ransport layer this would only work when the layer is dropped and initialized respectively. 
2. Requests that use `BlockId` tags like `latest`, `pending` etc. should not use cache. This becomes trivial to implement as a provider layer. 

To accomplish this we need to merge #788

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,436 @@
...
+
+        provider.save_cache(path).unwrap();
+    }
+
+    #[tokio::test]
+    async fn test_get_proof() {
+        let cache = CacheLayer::new(100);
+        let anvil = Anvil::new().block_time_f64(0.3).spawn();
+        let provider = ProviderBuilder::default()
+            // .with_recommended_fillers() - TODO: Issue here. Layer doesn't work with fillers. Fix.
```
yash-atreya, line 390: Enabling fillers would wrap the `CacheProvider` or any other layer provider in a `FillProvider`. Making methods on `CacheProvider` inaccessible. i.e `provider.load_cache(path)` won't work but `provider.inner.load_cache(path)` works.

This is not ideal.

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+        let cache = Arc::new(RwLock::new(LruCache::<B256, String>::new(
+            NonZeroUsize::new(max_items).unwrap(),
+        )));
+        Self { inner, cache, _pd: PhantomData }
+    }
+
+    /// Puts a value into the cache, and returns the old value if it existed.
+    pub fn put(&self, key: B256, value: String) -> TransportResult<Option<String>> {
+        let mut cache = self.cache.write();
+        Ok(cache.put(key, value))
```
mattsse, line 84: this can be a one liner I believe

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+
+        ProviderCall::BoxedFuture(Box::pin(async move {
+            let client = client?;
+
+            let res = client.request_noparams(req.method()).await?;
+
+            let json_str = serde_json::to_string(&res).map_err(TransportErrorKind::custom)?;
+            let hash = req.params_hash()?;
+            let mut cache = cache.write();
+            let _ = cache.put(hash, json_str);
```
mattsse, line 239: this should be a function

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+    T: Transport + Clone,
+{
+    #[inline(always)]
+    fn root(&self) -> &RootProvider<T> {
+        self.inner.root()
+    }
+
+    /// Gets the accounts in the remote node. This is usually empty unless you're using a local
+    /// node.
+    fn get_accounts(&self) -> ProviderCall<T, NoParams, Vec<Address>> {
```
mattsse, line 209: supporting this call is pretty pointless because this isn't even used really

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+            let json_str = serde_json::to_string(&res).map_err(TransportErrorKind::custom)?;
+            let hash = req.params_hash()?;
+            let mut cache = cache.write();
+            let _ = cache.put(hash, json_str);
+            Ok(res)
+        }))
+    }
+
+    /// Gets the chain ID.
+    fn get_chain_id(&self) -> ProviderCall<T, NoParams, U64, u64> {
```
mattsse, line 245: this is not useful imo

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    async fn get_block_by_number(
+        &self,
+        number: BlockNumberOrTag,
+        hydrate: bool,
+    ) -> TransportResult<Option<Block>> {
+        let hash = RequestType::new("eth_getBlockByNumber", (number, hydrate));
+
+        cache_get_or_fetch!(self, hash, self.inner.get_block_by_number(number, hydrate))
```
mattsse, line 290: this is problematic because of reorgs, we'd need to track chain height somehow, but out of scope for this pr imo

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+        if let Some(ref data) = result {
+            let json_str = serde_json::to_string(data).map_err(TransportErrorKind::custom)?;
+            let _ = $self.put(hash, json_str)?;
+        }
+
+        Ok(result)
+    }};
+}
+
+macro_rules! rpc_prov_call {
```
mattsse, line 148: not obvious what these do, please add docs

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_get_accounts() {
+        let cache = CacheLayer::new(100);
+        let anvil = Anvil::new().block_time_f64(0.3).spawn();
+        let provider = ProviderBuilder::default().layer(cache).on_http(anvil.endpoint_url());
+
+        let path = PathBuf::from_str("./rpc-cache-accounts.txt").unwrap();
```
mattsse, line 421:
```suggestion
        let path = PathBuf::from("./rpc-cache-accounts.txt");
```

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+        provider.load_cache(path.clone()).unwrap();
+
+        let accounts = provider.get_accounts().await.unwrap();
+        let accounts2 = provider.get_accounts().await.unwrap();
+        assert_eq!(accounts, accounts2);
+
+        provider.save_cache(path).unwrap();
```
mattsse, line 428: won't this write a new file and pollute the disk?

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+    type Provider = CacheProvider<P, T>;
+
+    fn layer(&self, inner: P) -> Self::Provider {
+        CacheProvider::new(inner, self.max_items())
+    }
+}
+
+/// A provider that caches responses to RPC requests.
+#[derive(Debug, Clone)]
+pub struct CacheProvider<P, T> {
```
mattsse, line 59: this needs a few more docs,

not obvious what is being cached

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// Inner provider.
+    inner: P,
+    /// In-memory LRU cache, mapping requests to responses.
+    cache: Arc<RwLock<LruCache<B256, String>>>,
```
mattsse, line 63: I'd be nicer to have this as a separate type, so you can clone and share this

mattsse (changes requested):

crates/provider/src/layers/cache.rs
yash-atreya: Yeah changing this to use tmp_dir

mattsse (changes requested):
almost there

few more suggestions

crates/provider/src/layers/cache.rs
```diff
@@ -23,22 +23,29 @@ use std::{io::BufReader, marker::PhantomData, num::NonZeroUsize, path::PathBuf,
 /// file system by calling `save_cache`.
 #[derive(Debug, Clone)]
 pub struct CacheLayer {
+    /// Configuration for the cache layer.
     config: CacheConfig,
```
mattsse: this isn't used anymore?

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,762 @@
...
+        Self { inner }
+    }
+
+    /// Puts a value into the cache, and returns the old value if it existed.
+    pub fn put(&self, key: B256, value: String) -> TransportResult<bool> {
+        Ok(self.inner.write().insert(key, value))
+    }
+
+    /// Gets a value from the cache, if it exists.
+    pub fn get(&self, key: &B256) -> TransportResult<Option<String>> {
```
mattsse, line 487: this does not need to return result

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,762 @@
...
+        // Need to acquire a write guard to change the order of keys in LRU cache.
+        let val = self.inner.write().get(key).cloned();
```
mattsse, line 489: this only needs a read lock

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,762 @@
...
+        Ok(val)
+    }
+
+    /// Get deserialized value from the cache.
+    pub fn get_deserialized<T>(&self, key: &B256) -> TransportResult<Option<T>>
+    where
+        T: for<'de> Deserialize<'de>,
+    {
+        if let Some(cached) = self.get(key)? {
```
mattsse, line 498: if get returns None then this can use let Some else return notation

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,762 @@
...
+        } else {
+            Ok(None)
+        }
+    }
+
+    /// Saves the cache to a file specified by the path.
+    /// If the files does not exist, it creates one.
+    /// If the file exists, it overwrites it.
+    pub fn save_cache(&self, path: PathBuf) -> TransportResult<()> {
+        let cache = self.inner.read();
```
mattsse, line 510: can we clone the content first?
then we wouldn't need to keep the lock over io

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,762 @@
...
+/// Shareable cache.
+#[derive(Debug, Clone)]
+pub struct SharedCache {
+    inner: Arc<RwLock<LruMap<B256, String>>>,
+}
+
+impl SharedCache {
+    /// Instantiate a new shared cache.
+    pub fn new(max_items: u32) -> Self {
+        let inner = Arc::new(RwLock::new(LruMap::<B256, String>::new(ByLength::new(max_items))));
```
mattsse, line 477: is this type hint here reuqired?

mattsse (changes requested):

crates/provider/src/layers/cache.rs
yash-atreya: `.get` of LruMap is &mut self and requires a write lock as it promotes the key that was accessed if found

yash-atreya:
@mattsse addressed the suggestions

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,753 @@
...
+    }
+
+    fn get_block_receipts(
+        &self,
+        block: BlockId,
+    ) -> ProviderCall<T, (BlockId,), Option<Vec<N::ReceiptResponse>>> {
+        let req = RequestType::new("eth_getBlockReceipts", (block,));
+
+        let redirect =
+            !matches!(block, BlockId::Hash(_) | BlockId::Number(BlockNumberOrTag::Number(_)));
```
mattsse, line 209: the number case is also slightly problematic, but we can worry about this separately

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,753 @@
...
+            let req = RequestType::new("eth_getTransactionCount", address).with_block_id(block_id);
+
+            cache_rpc_call_with_block!(cache, client, req)
+        })
+    }
+
+    async fn get_logs(&self, filter: &Filter) -> TransportResult<Vec<Log>> {
+        let req = RequestType::new("eth_getLogs", filter.clone());
+
+        let params_hash = req.params_hash().ok();
```
mattsse, line 291: this is also a bit nuanced because the blockrange variants allows the server to chose the range for example if upper bound is missing

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,753 @@
...
+
+    /// Puts a value into the cache, and returns the old value if it existed.
+    pub fn put(&self, key: B256, value: String) -> TransportResult<bool> {
+        Ok(self.inner.write().insert(key, value))
+    }
+
+    /// Gets a value from the cache, if it exists.
+    pub fn get(&self, key: &B256) -> Option<String> {
+        // Need to acquire a write guard to change the order of keys in LRU cache.
+        self.inner.write().get(key).cloned()
```
mattsse, line 483: this can be read

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,753 @@
...
+        let result = serde_json::from_str(&cached).map_err(TransportErrorKind::custom)?;
+        Ok(Some(result))
+    }
+
+    /// Saves the cache to a file specified by the path.
+    /// If the files does not exist, it creates one.
+    /// If the file exists, it overwrites it.
+    pub fn save_cache(&self, path: PathBuf) -> TransportResult<()> {
+        let cloned = self.inner.clone();
+        let cache = cloned.read();
```
mattsse, line 501: this is just read lock with extra steps,
what meant here is that should just clone the guarded value if we clone the values later anyway

crates/provider/src/layers/cache.rs
```diff
@@ -0,0 +1,753 @@
...
+
+        let result = $fetch_fn.await?;
+        if let Some(ref data) = result {
+            let json_str = serde_json::to_string(data).map_err(TransportErrorKind::custom)?;
+            let _ = $self.cache.put(hash, json_str)?;
+        }
+
+        Ok(result)
+    }};
+}
```
mattsse, line 109: I'm a bit sceptical that we even need these macros if those could also just be functions, but this also works, so lets stick with that

mattsse (changes requested):
last suggestions

mattsse (approved):

PR #955, opened 2024-06-21
fix: deserialization of null storage keys in AccessListItem

Wodann:
## Motivation

We recently adopted the `AccessList` and `AccessListItem` types for EDR/Hardhat. This fixes an issue when deserializing an `AccessListItem` where the `null` value for storage keys was not accepted.

## Solution

Added a custom serde deserialization handler.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -31,7 +31,7 @@ pub struct AccessListItem {
         )
     )]
     // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
-    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::null_as_default"))]
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::null_as_default"))]
     pub storage_keys: Vec<B256>,
 }
 
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -30,6 +30,7 @@ pub struct AccessListItem {
             strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
         )
     )]
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::or_default"))]
     pub storage_keys: Vec<B256>,
 }
 
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -31,7 +31,7 @@ pub struct AccessListItem {
         )
     )]
     // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
-    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::or_default"))]
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::null_as_default"))]
     pub storage_keys: Vec<B256>,
 }
 
--- a/crates/eips/src/eip2930.rs
+++ b/crates/eips/src/eip2930.rs
@@ -30,6 +30,7 @@ pub struct AccessListItem {
             strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
         )
     )]
+    // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
     #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::or_default"))]
     pub storage_keys: Vec<B256>,
 }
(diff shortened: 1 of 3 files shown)


crates/serde/src/optional.rs
```diff
@@ -0,0 +1,21 @@
+//! Serde functions for encoding optional values.
+//!
+//! This is defined as a "hex encoded unsigned integer", with a special case of 0 being `0x0`.
+//!
+//! A regex for this format is: `^0x([1-9a-f]+[0-9a-f]*|0)$`.
+//!
+//! This is only valid for human-readable [`serde`] implementations.
+//! For non-human-readable implementations, the format is unspecified.
+//! Currently, it uses a fixed-width big-endian byte-array.
```
mattsse, line 9: unrelated?

crates/serde/src/optional.rs
```diff
@@ -0,0 +1,21 @@
...
+use serde::{Deserialize, Deserializer};
+
+/// For use with serde's `deserialize_with` on a sequence that must be
+/// deserialized as a single but optional (i.e. possibly `null`) value.
+pub fn or_default<'de, T, D>(deserializer: D) -> Result<T, D::Error>
```
mattsse, line 14: this only interprets "null" as default, right?

so perhaps we rename this to `null_as_default` or smth?

mattsse (changes requested):
supportive, the spec does not properly specify this but I think it's reasonable to interpret null as [].

crates/serde/src/optional.rs
Wodann: Copy+paste remnant. Resolved in https://github.com/alloy-rs/alloy/pull/955/commits/f33ce645b4a08876ea65393e419a9f1b417e639e

crates/serde/src/optional.rs
Wodann: Addressed in https://github.com/alloy-rs/alloy/pull/955/commits/f33ce645b4a08876ea65393e419a9f1b417e639e

crates/eips/src/eip2930.rs
```diff
@@ -30,6 +30,8 @@ pub struct AccessListItem {
             strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
         )
     )]
+    // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::or_default"))]
```
mattsse, line 34:
```suggestion
    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::optional::null_as_default"))]
```

crates/eips/src/eip2930.rs
```diff
@@ -30,6 +30,8 @@ pub struct AccessListItem {
             strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
         )
     )]
+    // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
```
mattsse, line 33:
```suggestion
```

mattsse (changes requested):

crates/serde/src/lib.rs
```diff
@@ -18,6 +18,8 @@ use serde::Serializer;
 mod bool;
 pub use self::bool::*;
 
+pub mod optional;
```
DaniPopes, line 21: I would prefer private module and re-export the function

DaniPopes (changes requested):

crates/serde/src/lib.rs
Wodann: Addressed in https://github.com/alloy-rs/alloy/pull/955/commits/fdd21940e2046775d223b23ffa02cc67f855379d

mattsse (approved):
ty

DaniPopes (approved):

Wodann:
Thank you for your reviews!

What timeline do alloy releases normally follow?

mattsse:
likely when we have #950 in

jtguibas:
This PR unfortunately makes it so all types inheriting AccessListItem cannot be serialized/deserialized with serializers like bincode.

I think a good solution could be to instead make the type Option<Vec<B256>> for storage keys, if we really want to support null deserialization.

mattsse:
hey @Wodann please remind me, why should a null value be supported? I believe a storagekey should always be present

Wodann:
> hey @Wodann please remind me, why should a null value be supported? I believe a storagekey field should always be required:
> 
> https://github.com/ethereum/go-ethereum/blob/37590b2c5579c36d846c788c70861685b0ea240e/core/types/tx_access_list.go#L34-L35

Historically, Hardhat allows specification of the null value in storage keys to signal the same thing as an empty array of storage keys.

It could very well be that the spec has changed since. If this change is causing issues, we're happy to perform a breaking change in Hardhat's JSON-RPC spec.

Wodann:
We decided to remove support for `null` lists to match geth and alloy.

PR #959, opened 2024-06-23
feat(network): Add `input` method to `TransactionResponse`

moricho:
## Motivation
It would be nice to have a method to get input data from `TransactionResponse`.
During my development, there was a situation in which I wanted to know what method was being called and with what arguments from `TransactionResponse`.

## Solution
Adds `input` method to `TransactionResponse` and implements it to `TransactionResponse` type for each `Network`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/ethereum/mod.rs
+++ b/crates/network/src/ethereum/mod.rs
@@ -1,4 +1,5 @@
 use crate::{Network, ReceiptResponse, TransactionResponse};
+use alloy_primitives::Bytes;
 
 mod builder;
 
@@ -61,5 +62,9 @@ impl TransactionResponse for alloy_rpc_types_eth::Transaction {
 
     fn gas(&self) -> u128 {
         self.gas
+    }
+
+    fn input(&self) -> Bytes {
+        self.input.clone()
     }
 }
--- a/crates/network/src/ethereum/mod.rs
+++ b/crates/network/src/ethereum/mod.rs
@@ -64,7 +64,7 @@ impl TransactionResponse for alloy_rpc_types_eth::Transaction {
         self.gas
     }
 
-    fn input(&self) -> Bytes {
-        self.input.clone()
+    fn input(&self) -> &Bytes {
+        &self.input
     }
 }
--- a/crates/network/src/lib.rs
+++ b/crates/network/src/lib.rs
@@ -74,6 +74,7 @@ pub trait TransactionResponse {
     fn gas(&self) -> u128;
 
     /// Input data
+    #[doc(alias = "calldata")]
     fn input(&self) -> Bytes;
 }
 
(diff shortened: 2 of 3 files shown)


crates/network/src/lib.rs
```diff
@@ -72,6 +72,9 @@ pub trait TransactionResponse {
 
     /// Gas limit
     fn gas(&self) -> u128;
+
+    /// Input data
+    fn input(&self) -> Bytes;
```
mattsse, line 77:
```suggestion
    #[doc(alias = "calldata")]
    fn input(&self) -> Bytes;
```

mattsse, line 77: perhaps we should change this to &[u8] to match:

https://github.com/alloy-rs/alloy/blob/5eadbf18c468d6b77aa0f3210b80245aa431f3f4/crates/consensus/src/transaction/mod.rs#L58-L58

but I could see how Bytes would be more useful in-case you want ownership.

wdyt @DaniPopes

mattsse (changes requested):
thanks, this is indeed useful

crates/network/src/lib.rs
moricho: Fixed 👉  https://github.com/alloy-rs/alloy/pull/959/commits/1a714705b820f0e04a227c8fff766d8d1bddfa92

moricho: I realized that `input` type of `alloy_rpc_types_eth::Transaction` (`Bytes`) is different from `alloy_consensus::Transaction` (`&[u8]`). 
https://github.com/alloy-rs/alloy/blob/b79257c4de39b2486adc78ecb12cc72fdcb5253e/crates/rpc-types-
eth/src/transaction/mod.rs#L76

I set `Bytes` for the `input` type to match the `alloy_rpc_types_eth::Transaction` type in this PR since `TransactionResponse` is defined as  `alloy_rpc_types_eth::Transaction` for ethereum network:
https://github.com/alloy-rs/alloy/blob/main/crates/network/src/ethereum/mod.rs#L25

Should we fix that as well to match the type?

crates/network/src/ethereum/mod.rs
```diff
@@ -62,4 +63,8 @@ impl TransactionResponse for alloy_rpc_types_eth::Transaction {
     fn gas(&self) -> u128 {
         self.gas
     }
+
+    fn input(&self) -> Bytes {
```
mattsse, line 67: wdyt @DaniPopes Bytes or &[u8] ?

mattsse (approved):

crates/network/src/ethereum/mod.rs
DaniPopes: &Bytes would be the best choice here

mattsse: I guess we can reasonably assume that every tx always has an input, so that makes sense

PR #960, opened 2024-06-23
feat(contract): implement Filter's builder methods on Event

### LINKED ISSUE #824: [Feature] Implement `Filter`'s builder methods on `alloy_contract::Event`
### Component

contract

It would be nice to have method like `event_filter_with_address::<Event>(address: impl Into<ValueOrArray<Address>>)`. For example, you want to subscribe to all DEXs like uniswap.

```rust
/// Event filters.
#[automatically_derived]
impl<
    T: alloy_contract::private::Transport + ::core::clone::Clone,
    P: alloy_contract::private::Provider<T, N>,
    N: alloy_contract::private::Network,
> RouterInstance<T, P, N> {
    /// Creates a new event filter using this contract instance's provider and address.
    ///
    /// Note that the type can be any event, not just those defined in this contract.
    /// Prefer using the other methods for building type-safe event filters.
    pub fn event_filter<E: alloy_sol_types::SolEvent>(
        &self,
    ) -> alloy_contract::Event<T, &P, E, N> {
        alloy_contract::Event::new_sol(&self.provider, &self.address)
    }
    ///Creates a new event filter for the [`CreateProgram`] event.
    pub fn CreateProgram_filter(
        &self,
    ) -> alloy_contract::Event<T, &P, CreateProgram, N> {
        self.event_filter::<CreateProgram>()
    }
}
 ```
 
 https://github.com/alloy-rs/alloy/blob/c6c91cc197c79c456433c79d60f81c8c5aa13001/crates/contract/src/event.rs#L32-L45

### Additional context

_No response_

alexfertel:
## Motivation

Resolves #824 

## Solution

This PR adds builder methods to `alloy_contract::Event` that mirror `Filter`'s builder methods, forwarding calls such that you can customize the filter from the event.

Tbh, I don't like copy-pasting all of this, but I think the DX improvement is worth it.

Adds a single test, since the inner methods are tested in Filter's tests.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -84,21 +84,18 @@ impl<T: Transport + Clone, P: Provider<T, N>, E: SolEvent, N: Network> Event<T, 
     /// Sets the inner filter object
     ///
     /// See [`Filter::select`].
-    #[must_use]
     pub fn select(mut self, filter: impl Into<FilterBlockOption>) -> Self {
         self.filter.block_option = filter.into();
         self
     }
 
     /// Sets the from block number
-    #[must_use]
     pub fn from_block<B: Into<BlockNumberOrTag>>(mut self, block: B) -> Self {
         self.filter.block_option = self.filter.block_option.with_from_block(block.into());
         self
     }
 
     /// Sets the to block number
-    #[must_use]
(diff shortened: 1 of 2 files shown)


mattsse (approved):
seems appropriate.

don't mind duplicating this

pending @DaniPopes

PR #964, opened 2024-06-24
chore(eips): Add serde to Authorization types

rakita:
Small PR that adds serde to new types.

And adds `Eq, PartialEq` for `SignedAuthorization`

### DIFF
--- a/crates/eips/src/eip7702/auth_list.rs
+++ b/crates/eips/src/eip7702/auth_list.rs
@@ -8,6 +8,7 @@ use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodable, RlpEncodable
 /// An unsigned EIP-7702 authorization.
 #[derive(Debug, Clone, RlpEncodable, RlpDecodable, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct Authorization {
     /// The chain ID of the authorization.
     pub chain_id: ChainId,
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip7702/auth_list.rs
```diff
@@ -71,7 +72,8 @@ impl Authorization {
 }
 
 /// A signed EIP-7702 authorization.
-#[derive(Debug, Clone, RlpEncodable, RlpDecodable)]
+#[derive(Debug, Clone, RlpEncodable, RlpDecodable, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SignedAuthorization<S> {
     inner: Authorization,
```
mattsse, line 78: this should probably be serde flatten?

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -142,6 +145,7 @@ impl Deref for RecoveredAuthorization {
 ///
 /// The wrapper type is used for RLP encoding and decoding.
 #[derive(Default, Debug, Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct OptionalNonce(Option<u64>);
```
mattsse, line 149: do we want to serialize this as list?

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -7,6 +7,7 @@ use alloy_rlp::{BufMut, Decodable, Encodable, Header, RlpDecodable, RlpEncodable
 
 /// An unsigned EIP-7702 authorization.
 #[derive(Debug, Clone, RlpEncodable, RlpDecodable, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Authorization {
```
mattsse, line 11: rename camelCase?

mattsse (changes requested):
few suggestions

crates/eips/src/eip7702/auth_list.rs
rakita: Probably, have added it

crates/eips/src/eip7702/auth_list.rs
rakita: I am not sure, this would depend on jsonrpc and how it is defined there. Are you okay with leaving it as it is, until more info is known?

mattsse (approved):

PR #966, opened 2024-06-24
chore: remove proptest in all crates and Arbitrary derives

Rjected:
## Motivation
This was only required because reth required these types to implement `proptest::Arbitrary`. This is no longer the case, and `arbitrary::Arbitrary` is already usable.

## Solution

Remove ~~almost~~ all proptest derives, remove unused proptest dependencies.

~~There is a single `proptest!` left, which is for other fields, would like to know whether that's necessary or not.~~

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -140,8 +140,6 @@ serde_with = "3.3.0"
 ## misc-testing
 arbitrary = "1.3"
 assert_matches = "1.5"
-proptest = { version = "1.4", default-features = false, features = ["alloc"] }
-proptest-derive = { version = "0.4", default-features = false }
 serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.10"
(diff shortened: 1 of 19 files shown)


DaniPopes (approved):
which derives are missing?

Rjected:
> which derives are missing?

just this:
https://github.com/alloy-rs/alloy/blob/c2dd7cee159a31000d189c2cf851527fd56e3815/crates/serde/src/other/mod.rs#L175

used for this:
https://github.com/alloy-rs/alloy/blob/c2dd7cee159a31000d189c2cf851527fd56e3815/crates/serde/src/other/arbitrary_.rs#L91

DaniPopes:
Ah yeah the custom impl can be removed, it's only there because proptest derive doesn't support recursion

mattsse (approved):

PR #970, opened 2024-06-25
feat: move mev.rs from reth to rpc-types-mev

0xqd:
Ref https://github.com/paradigmxyz/reth/issues/9027


crates/rpc-types-mev/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+exclude.workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[dependencies]
+alloy-eips = { workspace = true, features = ["serde"] }
+alloy-primitives.workspace = true
+alloy-rpc-types.workspace = true
```
mattsse, line 21: this should directly depend on alloy-rpc-types-eth

mattsse (changes requested):
ty, some early comments

we also want a feature in rpc-types for this

crates/rpc-types/Cargo.toml
```diff
@@ -25,6 +25,7 @@ alloy-rpc-types-anvil = { workspace = true, optional = true }
 alloy-rpc-types-beacon = { workspace = true, optional = true }
 alloy-rpc-types-engine = { workspace = true, optional = true }
 alloy-rpc-types-eth = { workspace = true, optional = true }
+alloy-rpc-types-mev = { workspace = true, optional = true }
```
mattsse, line 28: needs a feature and reexport in lib.rs

mattsse (changes requested):
I'd like to not have a ton of modules just for single types

I'm fine with separating eth and mev modules

0xqd:
yeah, it's reasonable, I tried to split it by endpoint but it seems to be a bit much at the end.

mattsse (approved):
lgtm

PR #971, opened 2024-06-25
DO NOT MERGE: showing diffs

johntaiko:


mattsse (commented):
is this intended to be integrated in reth's API?

if so then this should be a standalone taiko type because payloadattributes can already be configured in reth

johntaiko:
> is this intended to be integrated in reth's API?
> 
> if so then this should be a standalone taiko type because payloadattributes can already be configured in reth

So sorry, I used wrong based branch

PR #976, opened 2024-06-25
chore(provider): simplify nonce filler

DaniPopes:
Removes dependency on tokio mutex and make the map smaller.

### DIFF
--- a/crates/provider/src/fillers/nonce.rs
+++ b/crates/provider/src/fillers/nonce.rs
@@ -7,8 +7,8 @@ use alloy_network::{Network, TransactionBuilder};
 use alloy_primitives::Address;
 use alloy_transport::{Transport, TransportResult};
 use dashmap::DashMap;
+use futures::lock::Mutex;
 use std::sync::Arc;
-use tokio::sync::Mutex;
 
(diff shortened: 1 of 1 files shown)


crates/provider/src/fillers/nonce.rs
```diff
@@ -92,33 +94,57 @@ impl NonceFiller {
...
+        let nonce = Arc::clone(
+            self.nonces.entry(from).or_insert_with(|| Arc::new(AtomicU64::new(NONE))).value(),
+        );
+
+        let prev_nonce = nonce.fetch_add(1, Ordering::SeqCst);
+        let current_nonce;
+        if prev_nonce == NONE {
+            // Initialize the nonce if we haven't seen this account before.
+            current_nonce = provider.get_transaction_count(from).await?;
+            nonce.store(current_nonce, Ordering::SeqCst);
```
mattsse, line 111: this is now reachable by two concurrent calls that can return the same value.

I guess we need a mutex at least for the initial fetch?

mattsse (changes requested):

mattsse (approved):

PR #977, opened 2024-06-26
fix(signer-trezor): fix zero gas price when sending legacy tx with trezor

TuDo1403:
## Motivation
This PR introduces a modification to the way gas prices are handled in the creation of legacy transactions for trezor wallet. The change ensures that the trezor-signer safely handles cases where the gas price might not be explicitly provided in the transaction details.

## Solution
Previously, the gas price was statically set to zero. This modification allows the trezor-signer to fetch the current gas price from the transaction and defaults to zero if none is specified.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
ty

huyhuynh3103 (approved):

PR #978, opened 2024-06-26
feat: add missing eth bundle args

mattsse:
ref https://github.com/paradigmxyz/reth/issues/9112

add missing (undocumented) fields


DaniPopes (approved):
classic

PR #981, opened 2024-06-26
Allow to convert CallBuilderTo TransactionRequest

leruaa:
## Motivation

Sometimes we don't want to `send()` a transaction created with `CallBuilder`, but instead convert it to `TransactionRequest`, in order to later include it in a bundle.

`CallBuilder::as_ref()` could be used, but it would require to clone the resulting `&TransactionRequest` to call `FillProvider::fill()`.

## Solution

Adding `into_transaction_request()` to `CallBuilder`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

Cc @igorline

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -131,6 +131,13 @@ pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
     transport: PhantomData<T>,
 }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this is reasonable

PR #986, opened 2024-06-28
feat: add missing debug_* rpc methods

alexfertel:
## Motivation

Adds the missing `debug_*` methods in https://github.com/alloy-rs/alloy/issues/926

Note that even though the issue says that we want to achieve full compatibility with reth, some of these methods are not implemented, like [`debug_getBadBlocks`](https://github.com/paradigmxyz/reth/blob/5416adf97b8e333c9a04ed9f3a18cdffd1a4f5f2/crates/rpc/rpc/src/debug.rs#L713).

Also note that the tests are dummies because:

- Proper tests should probably live in the node implementations. 
- These methods are not supported by anvil.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/debug.rs
+++ b/crates/provider/src/ext/debug.rs
@@ -1,7 +1,7 @@
 //! This module extends the Ethereum JSON-RPC provider with the Debug namespace's RPC methods.
 use crate::Provider;
 use alloy_network::Network;
-use alloy_primitives::{Bytes, TxHash, B256};
+use alloy_primitives::{hex, Bytes, TxHash, B256};
 use alloy_rpc_types_eth::{Block, BlockNumberOrTag, TransactionRequest};
 use alloy_rpc_types_trace::geth::{
     BlockTraceResult, GethDebugTracingCallOptions, GethDebugTracingOptions, GethTrace, TraceResult,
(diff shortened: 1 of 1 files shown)


crates/provider/src/ext/debug.rs
```diff
@@ -99,6 +136,42 @@ where
...
+    ) -> TransportResult<Vec<BlockTraceResult>> {
+        self.client().request("debug_traceChain", (start_exclusive, end_inclusive)).await
+    }
+
+    async fn debug_trace_block(
+        &self,
+        rlp_block: &[u8],
+        trace_options: GethDebugTracingOptions,
+    ) -> TransportResult<Vec<TraceResult>> {
+        self.client().request("debug_traceBlock", (rlp_block, trace_options)).await
```
DaniPopes, line 172: you likely want to hex encode here

mattsse: +0x prefix

alexfertel: Ah, you're totally right. Not having tests bites back.

mattsse (approved):
there are a few more debug_ functions that are very geth specific, these were initially added for geth debugging purposes

PR #987, opened 2024-06-28
feat: Add trace_get

SozinM:
## Motivation
Adds the missing trace_get method in https://github.com/alloy-rs/alloy/issues/926

## Problems
Currently trace_get returns null if queries with multiple indices (params example: "params":["0x346b152d421c095a2eab383d9c5d0309bb28c31e8c5b84fdc6a4f38e2194558d", ["0x0", "0x1"]]) https://github.com/paradigmxyz/reth/pull/3852

Because of this, i choose to use param `index` with type u64, to remove footgun.

## Solution
Usage example. Replace RETH_ENDPOINT with endpoint
```
use alloy_primitives::{b256};
use alloy_provider::{ext::TraceApi, ProviderBuilder};

#[tokio::main]
async fn main() {
    let url = String::from("RETH_ENDPOINT");


    let result = provider
        .trace_get(
            b256!("346b152d421c095a2eab383d9c5d0309bb28c31e8c5b84fdc6a4f38e2194558d"),
            1,
        )
        .await
        .unwrap();
    println!("{:#?}", result);
}
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -52,10 +52,10 @@ where
     ) -> TransportResult<Vec<LocalizedTransactionTrace>>;
 
     /// Traces of the transaction on the given positions
-    /// 
+    ///
     /// # Note
-    /// 
-    /// This function accepts single index and build list with it under the hood because 
+    ///
+    /// This function accepts single index and build list with it under the hood because
     /// trace_get method accepts list of indices but limits this list to len == 1.
     async fn trace_get(
         &self,
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -50,6 +50,13 @@ where
         hash: TxHash,
     ) -> TransportResult<Vec<LocalizedTransactionTrace>>;
 
+    /// Traces of the transaction on the given positions
+    async fn trace_get(
+        &self,
+        hash: TxHash,
+        index: u64,
+    ) -> TransportResult<LocalizedTransactionTrace>;
+
     /// Trace the given raw transaction.
     async fn trace_raw_transaction(
         &self,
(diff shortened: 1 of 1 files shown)


SozinM:
Open to discussion about whether to use indices: Vec<u64> or index: u64 param.
The first variant is compatible with documentation (which is misleading).
The second variant helps with correct understanding but is confusing as contradict docs.

crates/provider/src/ext/trace.rs
```diff
@@ -119,6 +126,15 @@ where
         self.client().request("trace_transaction", (hash,)).await
     }
 
+    async fn trace_get(
+        &self,
+        hash: TxHash,
+        index: u64,
+    ) -> TransportResult<LocalizedTransactionTrace> {
+        // We are using vec![indices] because API accepts a list, but in fact works only if list.len == 1
+        self.client().request("trace_get", (hash, vec![index])).await
```
DaniPopes, line 135: you can use tuples for lists in JSON since they're equivalent to avoid allocating on the heap, see above in `trace_transaction`

mattsse, line 135: this should also use the Index type of rpc-types-eth, because this should ideally be formatted as hex

https://docs.alchemy.com/reference/trace-get

crates/provider/src/ext/trace.rs
```diff
@@ -119,6 +126,15 @@ where
         self.client().request("trace_transaction", (hash,)).await
     }
 
+    async fn trace_get(
+        &self,
+        hash: TxHash,
+        index: u64,
```
mattsse, line 132: since this is an index, I wonder if we should use usize here because I assume most of the time this value is derived from doing some vec.len() +- ops

mattsse (changes requested):

crates/provider/src/ext/trace.rs
```diff
@@ -129,10 +135,10 @@ where
     async fn trace_get(
         &self,
         hash: TxHash,
-        index: u64,
+        index: usize,
     ) -> TransportResult<LocalizedTransactionTrace> {
         // We are using vec![indices] because API accepts a list, but in fact works only if list.len == 1
```
mattsse, line 140:
```suggestion
        // We are using `[index]` because API accepts a list, but only supports a single index
```

mattsse (approved):
lgtm,
doc nit

crates/provider/src/ext/trace.rs
SozinM: That's a good point, put usize in there.
I think it would be less convenient to put Index as type because then all calls to trace_get would require a .into() call from the user. It probably would be less convenient.

SozinM:
@mattsse could you approve workflows, please?

DaniPopes (approved):

PR #989, opened 2024-06-29
feat: add net rpc namespace

alexfertel:
## Motivation

Adds the `net` namespace to the json-rpc methods supported. Work towards completing #926.

Decided to convert from `PeerCount` to `u64` since it's more convenient for the end-user, but lmk if this is not expected.
There was no need for an `rpc-types-net` crate, so decided not to add one.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm, ty

PR #991, opened 2024-06-29
feat: add missing admin_* methods

alexfertel:
## Motivation

Adds missing `admin_*` RPC methods. Work towards completing #926. 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/admin.rs
+++ b/crates/provider/src/ext/admin.rs
@@ -41,7 +41,7 @@ pub trait AdminApi<N, T>: Send + Sync {
     #[cfg(feature = "pubsub")]
     async fn subscribe_peer_events(
         &self,
-    ) -> TransportResult<Subscription<alloy_rpc_types_admin::PeerEvent>>;
+    ) -> TransportResult<alloy_pubsub::Subscription<alloy_rpc_types_admin::PeerEvent>>;
 }
 
 #[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
--- a/crates/provider/src/ext/admin.rs
+++ b/crates/provider/src/ext/admin.rs
@@ -33,9 +33,6 @@ pub trait AdminApi<N, T>: Send + Sync {
     async fn node_info(&self) -> TransportResult<NodeInfo>;
 
     /// Subscribe to events received by peers over the network.
-    ///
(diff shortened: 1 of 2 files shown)


alexfertel:
Interestingly, the `wasm-wasi` check also fails on `main`.

mattsse (changes requested):
ty, this is a subscription for

https://github.com/ethereum/go-ethereum/blob/473ee8fc07a3f89cf3978e164a6fad218de9a6b5/p2p/peer.go#L92-L103

and it looks like we're missing this type in the admin crate

DaniPopes:
> 
Interestingly, the wasm-wasi check also fails on main. Unsure why ext::anvil::tests::test_anvil_set_interval_mining fails on windows. Damn, why is this PR breaking this stuff...

Not your fault, fixed on main

mattsse:
@alexfertel 

see also 

https://github.com/alloy-rs/alloy/blob/1807965627d29a3e6d9cab7774da1d92be971df5/crates/rpc-client/src/call.rs#L202-L208

"eth_subscribe" is detected as subscription by default, for others we'd need to set it after calling ".request("admin_peeEvents").set_is_subscription"

alexfertel:
@mattsse Does this make more sense?

I'm worried that `PeerEvent` may not serialize/deserialize correctly, is there a place where I can find a few json examples of the responses so that I can test this?

crates/provider/src/ext/admin.rs
```diff
@@ -31,6 +31,17 @@ pub trait AdminApi<N, T>: Send + Sync {
     /// Returns general information about the node as well as information about the running p2p
     /// protocols (e.g. `eth`, `snap`).
     async fn node_info(&self) -> TransportResult<NodeInfo>;
+
+    /// Subscribe to events received by peers over the network.
+    ///
+    /// Like other subscription methods, this returns the ID of the subscription, which is then used
+    /// in all events subsequently.
+    ///
+    /// To unsubscribe from peer events, call `unsubscribe_peer_events`.
```
DaniPopes, line 40: doc is outdated

DaniPopes (approved):
Thanks!

PR #994, opened 2024-06-29
Add rpc namespace

SozinM:
## Motivation
Adds RPC namespace for https://github.com/alloy-rs/alloy/issues/926

## Considerations
I took the type from reth rpc-types, so it could be migrated to alloy crate later.

The name of the crate came up a little bit silly `rpc-types-rpc`, so maybe better to rename it to rpc-types-namespaces, or something like that

## Solution
Usage example. Replace RETH_ENDPOINT with endpoint
```
use alloy_provider::{ext::RpcApi, ProviderBuilder};


#[tokio::main]
async fn main() {
    let url = String::from("RETH_ENDPOINT");



    let result = provider
        .rpc_modules()
        .await
        .unwrap();
    println!("{:#?}", result);
}
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


SozinM:
@mattsse hi!
Feel free to review it when you have spare time :)

SozinM:
@DaniPopes, maybe you want to take a look until it's gone stale?

crates/rpc-types-rpc/src/rpc.rs
mattsse, line 1: I think this crate is a bit too much,

this only consists of 1 type which we can just move to the rpc-types crate, it's only one type.

mattsse (changes requested):
sorry about the delay.

let's just move this type to the rpc-types crate, not worth having a standalone crate just for this.

crates/rpc-types-rpc/src/rpc.rs
SozinM: Changed!
One comment is that I didn't hide mod rpc behind the feature flag, because it will cause serde crate to be unused. If the feature flag is needed I could dig deeper into it.

mattsse (approved):
lgtm

PR #997, opened 2024-06-30
fix: Fix watching already mined transactions

### LINKED ISSUE #941: [Bug] Execution halting on .watch() for pending transactions.
### Component

provider, pubsub

alloy v0.1.1 (https://github.com/alloy-rs/alloy#041c7c47)

### Operating System

Windows

### Describe the bug

Execution is hanging on .watch() for pending transactions. The transactions go through, which can be seen on chain, however all execution halts on the watch before an transaction hash is returned.

This replicated using quick node and alchemy and is also both on websockets and http. Only tested so far on arbitrum. There does seem to be differences in behavior between operating systems where Ubuntu seems to hang on the second transaction but has no problem with the first and windows getting stuck on the first. The transactions did go through however when using anvil without any issues. Tested around with a few different alloy versions and the issue still persisted.

```
use alloy::{
    network::EthereumWallet,
    primitives::{address, U256},
    providers::ProviderBuilder,
    signers::local::PrivateKeySigner,
    sol,
};

sol!(
    #[sol(rpc)]
    IERC20,
    "src/abi.json"
);

#[tokio::main]
    let http_url = std::env::var("RPC_HTTP")?;
    let private_key: PrivateKeySigner = std::env::var("PRIVATE_KEY")?.parse()?;

    // For running on anvil. Anvil seems to have no issue with hanging on any of the watch functions.
    // let http_url = "http://localhost:8545";
    // let wallet_address = address!("f39Fd6e51aad88F6F4ce6aB8827279cffFb92266");
    // let private_key: PrivateKeySigner = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80".parse()?;

    let usdc_address = address!("af88d065e77c8cC2239327C5EDb3A432268e5831");
    let weth_address = address!("82aF49447D8a07e3bd95BD0d56f35241523fBab1");

    let wallet = EthereumWallet::from(private_key);
        .with_recommended_fillers()
        .wallet(wallet)
        .on_http(http_url.parse()?);

    let iweth = IERC20::new(weth_address, &provider);
    let iusdc = IERC20::new(usdc_address, &provider);

    // This should just be the uniswap router for arbitrum, pick what ever address to approve you like
    let approval_address = address!("4752ba5dbc23f44d87826276bf6fd6b1c372ad24");
    println!("trying first tx");
    let tx_hash = iweth
        .approve(approval_address, U256::MAX)
        .send()
        .await?
        // On windows it seems to hang execution here
        .watch()
        .await?;
    println!("{tx_hash:?}");
    println!("trying second tx");
    let tx_hash_2 = iusdc
        .approve(approval_address, U256::MAX)
        .send()
        .await?
        // On ubuntu execution seems to hang here
        .watch()
        .await?;
    println!("{tx_hash_2:?}");
    Ok(())
}
```

popzxc:
Potentially fixes #941

Changes provider and `Heartbeat` in a way that accounts for recently mined transactions.

## Motivation

I've met a problem where `.watch()` would get stuck even though a transaction is mined on the node.
When I checked the code, I noticed that we only start looking for it in blocks that come _after_ we subscribe for transaction.

## Solution

- Check if the transaction is already mined in the Provider trait.
  - ⚠️ There is an edge case where the transaction is mined but also has custom number of confirmations. I didn't find a way to extract the block number from an abstract transaction receipt, so just left a TODO there. I hope that it's an edge case that can be tackled separately, though -- the solution here would cover all the "simple" cases.
- Add a lookbehind buffer into `Heartbeat`, so that we can check if the transaction was confirmed recently.
- Make sure that the heartbeat is spawned before we submit the first transaction.

Added test was failing before the fix, now it passes.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/heart.rs
```diff
@@ -467,6 +496,26 @@ impl<S> Heartbeat<S> {
...
+            self.past_blocks.pop_front();
+        }
+        if let Some((last_height, _)) = self.past_blocks.back().as_ref() {
+            // Check that the chain is continuous.
+            if *last_height + 1 != *block_height {
+                // We cannot do anything about it, but we may at least log it.
+                warn!(%block_height, last_height, "non-continuous chain");
+            }
+        }
+        self.past_blocks.push_back((*block_height, block.transactions.hashes().copied().collect()));
```
mattsse, line 516: this doesn't handle reorgs, right?
here we'd append blocks with an existing height?
and we could find a tx confirmed in a reorged block  in L473

so I think this should replace the block with same number if it exists

mattsse (changes requested):
thanks, I believe this should work.

left some comments/qs

crates/provider/src/heart.rs
```diff
@@ -467,6 +496,26 @@ impl<S> Heartbeat<S> {
...
+        error!(?self.past_blocks, "lookbehind");
```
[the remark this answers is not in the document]
popzxc: Prototyping artifact, fixed.

crates/provider/src/provider/trait.rs
```diff
@@ -951,12 +955,26 @@ impl<T: Transport + Clone, N: Network> Provider<T, N> for RootProvider<T, N> {
         &self,
         config: PendingTransactionConfig,
     ) -> TransportResult<PendingTransaction> {
+        if self.get_transaction_receipt(*config.tx_hash()).await?.is_some() {
+            // The transaction is already confirmed.
+            if config.required_confirmations() <= 1 {
+                return Ok(PendingTransaction::ready(*config.tx_hash()));
+            }
+            // TODO: There exists a corner case, if we subscribe to an already mined transaction and
+            // require a custom number of confirmations. The `TransactionReceipt` trait
```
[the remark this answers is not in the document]
popzxc: Implemented via #1003, thanks.

crates/provider/src/heart.rs
popzxc: @mattsse That's right!

However, turns out the issue with reorgs is much deeper.
Once reorg occurs, we must also drop all the existing subscriptions received at block >= `*block_height`.
This part is easy, but it doesn't end here.
`ChainStreamPoller` doesn't seem to account for reorgs at all, e.g. [here](https://github.com/alloy-rs/alloy/blob/main/crates/provider/src/chain.rs#L72) it always expects the chain to be continuous.
And to fix that we probably need to compare block tip against `next_yield` [here](https://github.com/alloy-rs/alloy/blob/main/crates/provider/src/chain.rs#L56), and if `block_number < next_yield`, we should do some kind of search for the last non-diverging block (so probably we need to poll not only for block number, but rather for block header to use root hashes for comparison).

There is also an issue with the poller stream sporadically ending on transient network error, but that's a separate problem.

I've pushed a WIP commit with reorg handling in the `Heartbeat` and a (very flaky due to issues above) test.
PTAL, and let me know how you want to proceed.

If it works, I suggest the following:
- I will polish the commit, so that heartbeat supports reorgs, but will remove the test for now.
- We will merge this PR as-is, and I will create an issue for reorg handling in `ChainStreamPoller`.
- I will create a separate pull request for said issue, and will restore the test for reorg detection. (I cannot fully commit that I will do it due to limited capacity; but given that I already did some research and the problem seems interesting, I'll try my best to deliver it this week).

mattsse: thanks,
suggestions sgtm!

popzxc: I've removed the parts that are not relevant for this PR and created [an issue](https://github.com/alloy-rs/alloy/issues/1005) for the remaining work.
PTAL.

crates/provider/src/heart.rs
```diff
@@ -448,38 +463,119 @@ impl<S> Heartbeat<S> {
         }
     }
 
+    /// Reap transactions overridden by the reorg.
+    fn reap_reorg(&mut self, new_height: u64) {
```
mattsse, line 467: what does this do?
needs some additional docs, not obvious what this does?

looks like this drops the channel? which would mean on reorg, the listeners is unable to obtain the receipt because the watcher dropped?

crates/provider/src/heart.rs
```diff
@@ -394,6 +405,9 @@ pub(crate) struct Heartbeat<S> {
     /// The stream of incoming blocks to watch.
     stream: futures::stream::Fuse<S>,
 
+    /// Lookbehind blocks in form of mapping block number -> vector of transaction hashes.
+    past_blocks: VecDeque<(u64, Vec<B256>)>,
```
mattsse, line 409: this should probably be a hashmap of hashes?

otherwise with >200txs per block we'll spend some time searching those vecs

crates/provider/src/heart.rs
```diff
@@ -448,38 +463,119 @@ impl<S> Heartbeat<S> {
...
+            self.past_blocks.pop_front();
+        }
+        if let Some((last_height, _)) = self.past_blocks.back().as_ref() {
+            // Check that the chain is continuous.
+            if *last_height + 1 != *block_height {
+                // Remove all the transactions that were reset by the reorg.
+                warn!(%block_height, last_height, "reorg detected");
+                self.reap_reorg(*block_height);
+                // Remove past blocks that are now invalid.
+                self.past_blocks.retain(|(h, _)| h < block_height);
```
mattsse, line 551: maybe block_height <= last_height is more accurate here?

mattsse (changes requested):
one more q re reap_reorg

crates/provider/src/heart.rs
popzxc: > what does this do?

It erases all the subscriptions for the blocks that were affected by the reorg. Basically, it works the same as `reap_timeouts` above.

> looks like this drops the channel?

It does, similarly to how it does on timeout. If the block with your tx was reorganized, you are not guaranteed to receive a receipt for it, so IMHO it makes sense to drop it.

A dedicated error type would be nice (even if just for timeout), but it's a change in public interfaces, and not sure if it belongs to this PR (I can submit it in a separate PR though).

crates/provider/src/heart.rs
popzxc: You mean `HashMap<u64, HashSet<B256>>`? 
I agree about `HashSet`, but for the wrapper IMHO we have a nice `pop_back` / `push_front` flow here, so probably I'd suggest making it `VecDeque<(u64, HashSet<B256>)>`. WDYT?

crates/provider/src/heart.rs
popzxc: No, `block_height` is the first block of the "new" chain after reorg, so transactions that belong to this block should also be removed. We will push the block with `block_height` right below.

crates/provider/src/heart.rs
popzxc: Changed to `VecDeque<HashSet<(u64, HashSet<B256>)>`.

crates/provider/src/heart.rs
popzxc: Added more information to the method doc-comment.

mattsse: > so IMHO it makes sense to drop it.

I don't think this makes sense, because there's a high chance this tx will also be included in the new chain.

crates/provider/src/heart.rs
```diff
@@ -448,38 +463,121 @@ impl<S> Heartbeat<S> {
         }
     }
 
+    /// Reap transactions overridden by the reorg.
+    /// Accepts new chain height as an argument, and drops any subscriptions
+    /// that were received in blocks affected by the reorg (e.g. >= new_height).
```
mattsse, line 468: I think dropping reorged transactions isn't ideal, they become unconfirmed, because they can still be included in the new chain.

we could treat them as unconfirmed again after their block has been reorged.

mattsse (changes requested):

crates/provider/src/heart.rs
popzxc: I see, so would putting it back to `unconfirmed` make sense then?

crates/provider/src/heart.rs
popzxc: Sure, will do.

popzxc: Changed. Now transactions are returned to `self.unconfirmed`.

mattsse (approved):
this should behave as expected now.

but ngl, handling all of this is more complex than it should be -.-

PR #998, opened 2024-06-30
fix: correct trace_get method parameter

DoTheBestToGetTheBest:
the index should be an array of Index according to alchemy docs ; ref: https://docs.alchemy.com/reference/trace-get

### DIFF
--- a/crates/provider/src/ext/trace.rs
+++ b/crates/provider/src/ext/trace.rs
@@ -60,7 +60,7 @@ where
     async fn trace_get(
         &self,
         hash: TxHash,
-        index: usize,
+        index: Vec<Index>,
     ) -> TransportResult<LocalizedTransactionTrace>;
 
     /// Trace the given raw transaction.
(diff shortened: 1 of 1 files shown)


DaniPopes:
https://github.com/paradigmxyz/reth/blob/7401f46790f46c87cb1040f2292e38b88e14a07e/crates/rpc/rpc/src/trace.rs#L217-L233 cc @mattsse this should be correct

DoTheBestToGetTheBest:
> https://github.com/paradigmxyz/reth/blob/7401f46790f46c87cb1040f2292e38b88e14a07e/crates/rpc/rpc/src/trace.rs#L217-L233 cc @mattsse this should be correct, but we may want to keep the API the same and just serialize the one index as an array?

we don't want to use vector here?

mattsse (commented):
this isn't useful because this endpoint only supports a single index even if it accepts an array

PR #1009, opened 2024-07-03
feat(provider) : introduction to eth_sendRawTransactionConditional  RPC endpoint type

DoTheBestToGetTheBest:
reference to the implementation : https://notes.ethereum.org/@yoav/SkaX2lS9j#

this method allows for the conditional processing of transactions based on predefined network states


mattsse (changes requested):
I'd like to limit this pr to the rpc type first, need to think about how to integrate this in the provider API

DoTheBestToGetTheBest:
> I'd like to limit this pr to the rpc type first, need to think about how to integrate this in the provider API

thanks for your review, does it look better for you ?

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -767,6 +766,34 @@ impl From<TxEnvelope> for TransactionRequest {
 #[non_exhaustive]
 pub struct TransactionInputError;
 
+/// Options for conditional raw transaction submissions.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct ConditionalTxOptions {
```
mattsse, line 771: these should be moved to a separate module

mattsse, line 771: I believe all of these are camelCase?

mattsse, line 771: can we references the "spec" here?

mattsse (changes requested):

crates/rpc-types-eth/src/transaction/request.rs
DoTheBestToGetTheBest: > these should be moved to a separate module

yeah you're totally right here boss, can you please suggest a module or i should create new file for this ?

DoTheBestToGetTheBest: > can we references the "spec" here?

sure, thanks for the remind!

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -767,6 +766,34 @@ impl From<TxEnvelope> for TransactionRequest {
...
+    pub known_accounts: HashMap<Address, KnownAccountState>,
+}
+/// Represents the expected state of an account for a transaction to be conditionally accepted.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(untagged)]
+pub enum KnownAccountState {
+    /// Expected storage root hash of the account.
+    StorageRoot(B256),
+    /// Explicit storage slots and their expected values.
+    Slots(HashMap<U256, B256>),
```
DoTheBestToGetTheBest, line 795: i'm also not sure if this should be   Slots(HashMap<B256, B256>), or  Slots(HashMap<U256, B256>),

crates/rpc-types-eth/src/transaction/request.rs
DoTheBestToGetTheBest: > I believe all of these are camelCase?

changes has been done, ty!

Vid201:
Hey @DoTheBestToGetTheBest , this is the implementation in `ethers-rs` : https://github.com/gakonst/ethers-rs/pull/2595

Didn't get merged, but it's working since it's used in production in Silius (account abstraction bundler)

DoTheBestToGetTheBest:
> Hey @DoTheBestToGetTheBest , this is the implementation in `ethers-rs` : [gakonst/ethers-rs#2595](https://github.com/gakonst/ethers-rs/pull/2595)
> 
> Didn't get merged, but it's working since it's used in production in Silius (account abstraction bundler)

ty, currently in this pr we just want to add the type

DoTheBestToGetTheBest:
@mattsse beffore i start to work on eth_simulateV1 , please take a look on this so we can done this

mattsse (approved):
doesn't hurt

PR #1014, opened 2024-07-04
chore: convert rcp-types-eth block Header to consensus Header

teddav:
There are currently multiple Block Header types (@mattsse told me it's because of legacy naming).  
This PR adds a way to convert one into the other. I needed that in order to then RLP encode the Header. I added tests at the bottom to make sure the RLP encoding is correct and correctly recomputes the block hash

### DIFF
--- a/crates/rpc-types-eth/src/block.rs
+++ b/crates/rpc-types-eth/src/block.rs
@@ -1,6 +1,6 @@
 //! Block RPC types.
 
-use crate::{Transaction, Withdrawal};
+use crate::{ConversionError, Transaction, Withdrawal};
 use alloy_primitives::{Address, BlockHash, Bloom, Bytes, B256, B64, U256, U64};
 use alloy_serde::OtherFields;
 use serde::{ser::Error, Deserialize, Serialize, Serializer};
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/block.rs
```diff
@@ -149,6 +149,60 @@ impl Header {
     }
 }
 
+impl TryFrom<Header> for alloy_consensus::Header {
+    type Error = String;
```
mattsse, line 153: This should be `ConversionError`

crates/rpc-types-eth/src/block.rs
```diff
@@ -149,6 +149,60 @@ impl Header {
...
+            state_root,
+            transactions_root,
+            receipts_root,
+            logs_bloom,
+            difficulty,
+            number,
+            gas_limit,
+            gas_used,
+            timestamp,
+            total_difficulty: _total_difficulty,
```
mattsse, line 169: can we group this and hash and separate with a `// not included in header` comment?

mattsse (changes requested):
nits

crates/rpc-types-eth/src/block.rs
teddav: done. i just added a "::Custom" error, since the ConversionError seems to be made for Transactions, i didnt want to add other variants to it

mattsse (approved):

PR #1016, opened 2024-07-04
Export eip1559 related codes in `alloy-provider` to `alloy-eips`

moricho:
## Motivation
`utils.rs` in alloy-provider had some EIP1559-related code, such as gas fee estimator helpers and constants (e.g., min priority fee). Now we have the `alloy-eips` crate, so it would be helpful and easier to maintain if all these EIP1559-related codes were gathered in one place.

However, `EIP1559_FEE_ESTIMATION_REWARD_PERCENTILE` and `EIP1559_FEE_ESTIMATION_PAST_BLOCKS` remain in alloy-provider as they are specific to alloy-provider.

## Solution

Move some EIP1559 related constants to `alloy-eips/eip1559/constants.rs` and the gas fee estimation codes to the `alloy-eips/eip1559/estimator.rs`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
hmm, the estimation part doesn't really fit into the eips crate as this is more of a provider-level use case and an opinionated impl.

is there a particular reason to move this?

moricho:
@mattsse Since it had the prefix EIP1559, I simply thought it should be moved to the eips crate 🙏  However, it seems that it is only used by the provider for now, so there is no need to transfer it. So this can be closed!

PR #1018, opened 2024-07-04
feat: use kzg-rs for kzg point evaluation

### LINKED ISSUE #1017: [Feature] Use `kzg-rs` for KZG point evaluation
### Component

consensus, eips, genesis

Currently, alloy uses c-kzg for KZG point evaluation. It would be useful to have a [no_std] endpoint for this, which [kzg-rs](https://github.com/0xWOLAND/kzg-rs) can be used for.

For instance, it would allow to use alloy `consensus` and `eips` in a [SP1](https://github.com/succinctlabs/sp1) zkVM context.

I'm starting to work on a PR.

### Additional context

_No response_

leruaa:
## Motivation

Closes #1017

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,3 +144,8 @@ proptest-derive = { version = "0.4", default-features = false }
 serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.10"
+
+# TODO: Remove this once https://github.com/succinctlabs/kzg-rs/pull/3 is merged
+[patch.crates-io]
+kzg-rs = { git = "https://github.com/leruaa/kzg-rs", branch = "verify_blobs" }
+bls12_381 = { git = "https://github.com/leruaa/bls12_381", branch = "msm_variable_base" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,6 +80,7 @@ ethereum_ssz = "0.5"
 
 # crypto
 c-kzg = { version = "1.0", default-features = false }
+kzg-rs = { version = "0.1", default-features = false }
 elliptic-curve = { version = "0.13", default-features = false }
 k256 = { version = "0.13", default-features = false, features = ["ecdsa"] }
 sha2 = { version = "0.10", default-features = false }
@@ -125,6 +126,7 @@ thiserror = "1.0"
 thiserror-no-std = "2.0.2"
 url = "2.5"
 derive_more = "0.99.17"
+cfg-if = "1.0.0"
 
 ## serde
 serde = { version = "1.0", default-features = false, features = [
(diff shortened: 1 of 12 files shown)


crates/eips/src/eip4844/sidecar.rs
```diff
@@ -228,12 +228,21 @@ where
...
-    KZGError(c_kzg::Error),
+    
+    cfg_if::cfg_if! {
+        if #[cfg(feature = "c-kzg")] {
+            /// An error returned by [`c_kzg`].
+            KZGError(c_kzg::Error),
+        } else if #[cfg(feature = "kzg-rs")] {
+            /// An error returned by [`kzg-rs`].
+            KZGError(kzg_rs::KzgError),
+        }
```
mattsse, line 243: these type defs should then be isolated in a separate module and conditionally exported, so that we don't have to do this multiple times

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -96,7 +96,7 @@ impl BlobTransactionSidecar {
     /// Returns [BlobTransactionValidationError::InvalidProof] if any blob KZG proof in the response
     /// fails to verify, or if the versioned hashes in the transaction do not match the actual
     /// commitment versioned hashes.
-    #[cfg(feature = "kzg")]
+    #[cfg(feature = "c-kzg")]
     pub fn validate(
```
mattsse, line 100: is there a kzg-rs equivalent to this?

mattsse (changes requested):
I'd like to keep c-kzg as "kzg" for now so we don't break anything just yet

crates/eips/src/eip4844/sidecar.rs
leruaa: No, but I'm on it.

Cargo.toml
```diff
@@ -142,3 +144,8 @@ proptest-derive = { version = "0.4", default-features = false }
 serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.10"
+
+# TODO: Remove this once https://github.com/succinctlabs/kzg-rs/pull/3 is merged
+[patch.crates-io]
+kzg-rs = { git = "https://github.com/leruaa/kzg-rs", branch = "verify_blobs" }
+bls12_381 = { git = "https://github.com/leruaa/bls12_381", branch = "msm_variable_base" }
```
leruaa, line 151: These patches may be removed once https://github.com/succinctlabs/kzg-rs/pull/3 is merged.

PR #1019, opened 2024-07-05
feat: add helper to set both input and data fields

mattsse:
https://github.com/foundry-rs/foundry/issues/7764

some node impls still only accept the legacy data field...

this makes it easy to set both


DaniPopes:
Yikes!

PR #1021, opened 2024-07-07
feat: add From<Index> for some numeric primitive type

DoTheBestToGetTheBest:
This is useful for some Trace


mattsse (commented):
hmm, don't think this is very useful because this type is supposed to be used as usize

PR #1025, opened 2024-07-09
chore: add payloadbodies v2 to capabilities set

mattsse:


DaniPopes:
Does this constant belong in alloy?

mattsse:
not really, 

I will move this to reth separately, but imo, it's useful to a a set of all engine API endpoints

DaniPopes (approved):
Please move to reth

PR #1028, opened 2024-07-09
fix: export stream types part of public api

mattsse:
these types are part of public api but can't be imported.


DaniPopes:
https://github.com/alloy-rs/alloy/pull/1029

PR #1034, opened 2024-07-10
fix(eip7702): Add correct rlp decode/encode

rakita:
eip7702 AuthorizationList rlp format is not correct.

The format that gets encoded is: `0xf85ed801940000000000000000000000000000000000000006c101f8431ba048b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353a0efffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804`
That decoded is:
`[["0x01","0x0000000000000000000000000000000000000006",["0x01"]],["0x1b","0x48b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353","0xefffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804"]]`
https://toolkit.abdk.consulting/ethereum#rlp

There are two inner lists that should not be there.

Removed Generic over `SignedAuthorization` as it seems it is not needed. We default to Signature (can revert back if this is not the case)

### DIFF
--- a/crates/consensus/src/transaction/eip2930.rs
+++ b/crates/consensus/src/transaction/eip2930.rs
@@ -206,7 +206,7 @@ impl TxEip2930 {
         let tx = Self::decode_fields(buf)?;
         let signature = Signature::decode_rlp_vrs(buf)?;
 
-        let signed = tx.into_signed(signature);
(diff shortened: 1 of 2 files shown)


mattsse (approved):
nice

PR #1036, opened 2024-07-11
feat: add arbitrary to auth

### LINKED ISSUE #1031: [Feature] Implement Arbitrary for SignedAuthorization
### Component

consensus, eips, genesis

Currently in reth we implement `Arbitrary` manually, to remove this we need to implement Arbitrary for the eip7702 types, specifically `SignedAuthorization`

### Additional context

_No response_

mattsse:
closes #1031

note we don't have an easy way to generate random valid k256 sigs, so this just uses a hardcoded one

### DIFF
--- a/crates/eips/src/eip7702/auth_list.rs
+++ b/crates/eips/src/eip7702/auth_list.rs
@@ -12,6 +12,10 @@ use alloy_rlp::{
 #[derive(Debug, Clone, RlpEncodable, RlpDecodable, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(
+    any(test, feature = "arbitrary"),
+    derive(proptest_derive::Arbitrary, arbitrary::Arbitrary)
+)]
 pub struct Authorization {
     /// The chain ID of the authorization.
     pub chain_id: ChainId,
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip7702/auth_list.rs
```diff
@@ -160,6 +164,32 @@ impl Deref for SignedAuthorization {
...
 
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a> arbitrary::Arbitrary<'a> for SignedAuthorization {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        use alloy_primitives::Parity;
+
+        let inner = u.arbitrary::<Authorization>()?;
+        let parity = u.arbitrary::<Parity>()?;
+
+        // TODO: find an easy way to generate random signatures
```
DaniPopes, line 175: why not run it in a loop with random values until it works?

mattsse: tried this first, couldn't figure out what random values to use, random u256 just ran forever

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -160,6 +164,32 @@ impl Deref for SignedAuthorization {
...
+        let signature = Signature::from_scalars_and_parity(
+            <B256 as core::str::FromStr>::from_str(
+                "0xc569c92f176a3be1a6352dd5005bfc751dcb32f57623dd2a23693e64bf4447b0",
+            )
+            .unwrap(),
+            <B256 as core::str::FromStr>::from_str(
+                "0x1a891b566d369e79b7a66eecab1e008831e22daa15f91a0a0cf4f9f28f47ee05",
```
DaniPopes, line 182: b256! macro

DaniPopes (approved):

PR #1038, opened 2024-07-11
fix(admin): id in NodeInfo is string instead of B256

jsvisa:
## Motivation

Return the same response type as geth's, ref https://github.com/ethereum/go-ethereum/blob/v1.14.0/p2p/server.go#L1078-L1079

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/geth.rs
+++ b/crates/node-bindings/src/geth.rs
@@ -2,7 +2,7 @@
 
 use crate::unused_port;
 use alloy_genesis::{CliqueConfig, Genesis};
-use alloy_primitives::{hex, Address, B256};
+use alloy_primitives::Address;
 use k256::ecdsa::SigningKey;
 use std::{
     borrow::Cow,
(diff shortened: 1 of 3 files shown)


crates/node-bindings/src/geth.rs
```diff
@@ -121,7 +121,7 @@ impl GethInstance {
     /// Blocks until geth adds the specified peer, using 20s as the timeout.
     ///
     /// Requires the stderr to be present in the `GethInstance`.
-    pub fn wait_to_add_peer(&mut self, id: B256) -> Result<(), GethInstanceError> {
+    pub fn wait_to_add_peer(&mut self, id: String) -> Result<(), GethInstanceError> {
```
DaniPopes, line 124: use &str

crates/node-bindings/src/geth.rs
```diff
@@ -132,8 +132,8 @@ impl GethInstance {
             err_reader.read_line(&mut line).map_err(GethInstanceError::ReadLineError)?;
 
             // geth ids are truncated
-            let truncated_id = hex::encode(&id.0[..8]);
-            if line.contains("Adding p2p peer") && line.contains(&truncated_id) {
+            let truncated_id = &id[0..8];
```
DaniPopes, line 135: this panics if too short

DaniPopes (changes requested):

mattsse (commented):
nice find, I believe I couldn't figure out what the correct value here is

crates/node-bindings/src/geth.rs
jsvisa: thanks for the tip, then I change it to get the first 16bytes instead.

mattsse (approved):
lgtm
ty

@DaniPopes if we release alloy with this fix alloy as a patch, this will break 1.0.1 reth build for some folks

PR #1039, opened 2024-07-11
feat(genesis): rm EIP150Hash

jsvisa:
## Motivation

As EIP150Hash was removed in geth, so I think we can also remove them, ref https://github.com/ethereum/go-ethereum/blob/v1.14.0/params/config.go#L326

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes (approved):
Removed over a year ago in https://github.com/ethereum/go-ethereum/pull/27087

mattsse (approved):

PR #1041, opened 2024-07-11
feat: generate valid signed auth signatures

Rjected:
## Motivation

This makes it so the auth list signature is not hard coded, although this gates the arbitrary implementation behind the k256 feature

## Solution

Just generate a hash, based on the auth, and sign it

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/Cargo.toml
+++ b/crates/eips/Cargo.toml
@@ -41,6 +41,9 @@ arbitrary = { workspace = true, features = ["derive"], optional = true }
 proptest = { workspace = true, optional = true }
 proptest-derive = { workspace = true, optional = true }
 
+# for signed authorization list arbitrary
+k256 = { workspace = true, optional = true }
+
 [dev-dependencies]
 alloy-primitives = { workspace = true, features = [
     "rand",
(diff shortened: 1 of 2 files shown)


crates/eips/Cargo.toml
```diff
@@ -41,6 +41,9 @@ arbitrary = { workspace = true, features = ["derive"], optional = true }
 proptest = { workspace = true, optional = true }
 proptest-derive = { workspace = true, optional = true }
 
+# for signed authorization list arbitrary
+k256 = { workspace = true, optional = true }
```
mattsse, line 45: this is fine because if if the feature is enabled on alloy-primitives this dep is already part of the dep tree

crates/eips/src/eip7702/auth_list.rs
```diff
@@ -174,21 +174,21 @@ impl Deref for SignedAuthorization {
     }
 }
 
-#[cfg(any(test, feature = "arbitrary"))]
+#[cfg(all(any(test, feature = "arbitrary"), feature = "k256"))]
```
mattsse, line 177: we could also make this work without the k256 feature but not useful imo

mattsse (approved):

PR #1043, opened 2024-07-12
feat(otterscan): add ots slim block and serialze OperationType to int

jsvisa:
## Motivation

I'm running reth along with otterscan, and find some more rpc mismatches:

1. `ots_getInternalOperations` returns `type` as int, but the default serde json serialize will return as String instead, use `serde_repr` to serialize it;
2. the response of `ots_getBlockDetails` without transactions, https://github.com/otterscan/otterscan/blob/3adcc62c4eedc56f812fab5ca4b85e3c5fd4b595/docs/custom-jsonrpc.md#L76 https://github.com/ledgerwatch/erigon/blob/6097c05cc914801132e7897427956f196efbbba1/turbo/jsonrpc/otterscan_api.go#L259-L261
3. add a new function to set `issuance` and `total_fees` for `BlockDetails`

After this prototypes changed, I think reth can almost run along with otterscan(without the transaction before/after listing and ots2 API)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/otterscan.rs
+++ b/crates/rpc-types-trace/src/otterscan.rs
@@ -5,10 +5,10 @@
 
 use alloy_primitives::{Address, Bloom, Bytes, TxHash, B256, U256};
 use alloy_rpc_types_eth::{Block, Header, Rich, Transaction, TransactionReceipt, Withdrawal};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 
 /// Operation type enum for `InternalOperation` struct
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Deserialize)]
 pub enum OperationType {
     /// Operation Transfer
     OpTransfer = 0,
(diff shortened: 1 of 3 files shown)


crates/rpc-types-trace/src/otterscan.rs
```diff
@@ -3,12 +3,14 @@
...
-use alloy_rpc_types_eth::{Block, Rich, Transaction, TransactionReceipt};
+use alloy_primitives::{Address, Bloom, Bytes, TxHash, B256, U256};
+use alloy_rpc_types_eth::{Block, Header, Rich, Transaction, TransactionReceipt, Withdrawal};
 use serde::{Deserialize, Serialize};
+use serde_repr::{Deserialize_repr, Serialize_repr};
 
 /// Operation type enum for `InternalOperation` struct
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize_repr, Deserialize_repr)]
+#[repr(u8)]
```
mattsse, line 13: I'd like to manually implement serde for this by doing match -> u8 -> string and String -> parse u8
so that we don't need the serde_repr dep

mattsse (changes requested):
awesome, tysm for all of these
one nit

crates/rpc-types-trace/src/otterscan.rs
jsvisa: sounds more reasonable, no need to introduce a new crate for only small things

crates/rpc-types-trace/src/otterscan.rs
```diff
@@ -20,6 +23,37 @@ pub enum OperationType {
     OpCreate2 = 3,
 }
 
+// Implement Serialize for OperationType
+impl Serialize for OperationType {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        serializer.serialize_u8(*self as u8)
```
mattsse, line 32: so this is actually a number and not a string?

mattsse (approved):

crates/rpc-types-trace/src/otterscan.rs
jsvisa: Yes, I think it is:

image

PR #1046, opened 2024-07-13
feat: add 7702 tx type

mattsse:
adds consensus 7702 tx type

this is the same as 1559, just with an additional field.


onbjerg (approved):
lgtm

onbjerg:
we should add this to the tx enum later on in a separate pr

PR #1047, opened 2024-07-13
chore: make auth mandatory in recovered auth

mattsse:
optional authority if recovery failed doesn't make sense.

this should always be Address


Evalir (approved):
agreed, this shouldn't be an option

rakita:
Main point of Authority is it can be invalid, that was why it is made as a option

PR #1049, opened 2024-07-13
feat(core): update core version

0xfourzerofour:
## Motivation

Update version of core libs to access the new features in latest release from alloy

## Solution

Update version


mattsse (approved):
patches shouldn't matter but doesn't hurt

Issue #1032, opened 2024-07-10
[Feature] Add 7702 authorization list to rpc tx and receipt types


mattsse:
### Component

rpc

### Describe the feature you would like

ref https://github.com/ethereum/execution-apis/issues/561

## TODO

* add `authorizationList: Option<Vec<Authorization>>` as an additional field to rpc Transaction and tx receipt types

https://github.com/alloy-rs/alloy/blob/83343b172585fe4e040fb104b4d1421f58cbf9a2/crates/eips/src/eip7702/auth_list.rs#L8-L19

PanGan21:
Hey @mattsse I did some effort for this in #1051 . Please let me know if something is missing! 🙂

onbjerg:
Closing this as closed by #1051, thanks for the PR:)

PR #1051, opened 2024-07-14
feat: add authorization list to rpc transaction and tx receipt types

PanGan21:
## Motivation

Adds 7702 authorization list to rpc transaction and transaction receipt types.
Related: #1032 

## PR Checklist

- [X] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -103,7 +106,10 @@ pub struct Transaction {
     )]
     #[doc(alias = "tx_type")]
     pub transaction_type: Option<u8>,
-
+    /// The authorization list is a list of tuples that store the address to code which the signer
+    /// desires to execute in the context of their EOA.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
mattsse, line 111:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

crates/rpc-types-eth/src/transaction/receipt.rs
```diff
@@ -56,6 +57,10 @@ pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
     /// EIP98 makes this optional field, if it's missing then skip serializing it
     #[serde(skip_serializing_if = "Option::is_none", rename = "root")]
     pub state_root: Option<B256>,
+    /// The authorization list is a list of tuples that store the address to code which the signer
+    /// desires to execute in the context of their EOA.
+    #[serde(skip_serializing_if = "Option::is_none")]
```
mattsse, line 62:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -103,7 +106,10 @@ pub struct Transaction {
...
+    pub authorization_list: Option<Vec<Authorization>>,
```
mattsse, line 112: I believe we actually want the SignedAuthoriazion here because we also need the signature

mattsse (changes requested):
ty!

DaniPopes (approved):

crates/rpc-types-eth/src/transaction/mod.rs
PanGan21: Updated

PanGan21: Seems fixed!

PR #1052, opened 2024-07-15
fix(provider): Do not overflow LRU cache capacity in ChainStreamPoller

popzxc:
## Motivation

I started to work on handling reorgs in `ChainStreamPoller`, but noticed a few adjacent problems that probably should be fixed first.
Here, I noticed that we don't check whether we've reached the LRU cache capacity. As a result, if a lot of blocks are created in a short time (e.g. via `anvil_mine`), some blocks may be skipped.

This PR fixes that, and also adds a few basic tests for `ChainStreamPoller`.

## Solution

Only add new items to the cache if we don't reach capacity.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/chain.rs
+++ b/crates/provider/src/chain.rs
@@ -31,10 +31,16 @@ impl<T: Transport + Clone, N: Network> ChainStreamPoller<T, N> {
     }
 
     pub(crate) fn new(client: WeakClient<T>) -> Self {
+        Self::with_next_yield(client, NO_BLOCK_NUMBER)
+    }
+
+    /// Can be used to force the poller to start at a specific block number.
+    /// Mostly useful for tests.
+    fn with_next_yield(client: WeakClient<T>, next_yield: BlockNumber) -> Self {
         Self {
             client: client.clone(),
             poll_task: PollerBuilder::new(client, "eth_blockNumber", ()),
-            next_yield: NO_BLOCK_NUMBER,
+            next_yield,
             known_blocks: LruCache::new(BLOCK_CACHE_SIZE),
             _phantom: PhantomData,
         }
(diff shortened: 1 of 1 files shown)


crates/provider/src/chain.rs
```diff
@@ -31,10 +31,16 @@ impl<T: Transport + Clone, N: Network> ChainStreamPoller<T, N> {
...
+
+    /// Can be used to force the poller to start at a specific block number.
+    /// Mostly useful for tests.
+    fn with_next_yield(client: WeakClient<T>, next_yield: BlockNumber) -> Self {
         Self {
             client: client.clone(),
             poll_task: PollerBuilder::new(client, "eth_blockNumber", ()),
-            next_yield: NO_BLOCK_NUMBER,
+            next_yield,
             known_blocks: LruCache::new(BLOCK_CACHE_SIZE),
```
popzxc, line 44: Also, I'm not quite sure why an LRU cache is used here. Wouldn't a simple `VecDeque` suffice?

crates/provider/src/chain.rs
```diff
@@ -106,8 +112,77 @@ impl<T: Transport + Clone, N: Network> ChainStreamPoller<T, N> {
                     }
                 };
                 self.known_blocks.put(number, block);
+                if self.known_blocks.len() == BLOCK_CACHE_SIZE.get() {
```
popzxc, line 115: The implication here is that we will have to wait for the next value from `poller_task` while we technically know already that there are more blocks we can fetch. However, implementing it may be somewhat complicated and feels like a premature optimization.

crates/provider/src/chain.rs
```diff
@@ -106,8 +112,77 @@ impl<T: Transport + Clone, N: Network> ChainStreamPoller<T, N> {
...
+                    debug!(number, "cache full");
+                    break;
+                }
             }
         }
         }
     }
 }
+
+#[cfg(all(test, feature = "anvil-api"))] // Tests rely heavily on ability to mine blocks on demand.
```
popzxc, line 126: Writing tests without `anvil_mine` would be more complex, so I hope it's fine given that tests are run with `--all-features` in CI.

crates/provider/src/chain.rs
mattsse: hmm, imo like this loop's logic is flawed because this allows iterating next_yield..tip before we start yielding blocks again

I consider this a hotfix, but ideally we can solve this my improving this stream impl.

crates/provider/src/chain.rs
mattsse: yeah, imo vecdeque would be more appropriate here actually

mattsse (approved):
this stream impl is a bit weird because this does things sequentially that should be done sequentially?

I'm fine with this hotfix for now, but we should rethink this and get rid of the lru, seems weird

Issue #889, opened 2024-06-13
prevent panic from having 0 keys when calling `on_anvil_with_wallet_and_config`


prestwich:
This can panic if the `f` function sets accounts to 0. This is not concerning (and is the pre-existing behavior) but we should probably fix it

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/883#discussion_r1638268737_

PR #1055, opened 2024-07-15
fix(provider): Prevent panic from having 0 keys when calling `on_anvil_with_wallet_and_config`

moricho:
## Motivation
Closes https://github.com/alloy-rs/alloy/issues/889.

`on_anvil_with_wallet_and_config` can panic if the available key in an anvil instance is 0.

## Solution

Adds the new error `NoKeysAvailable` to `AnvilError` in `alloy-node-bindings` and uses it as the error of `on_anvil_with_wallet_and_config`

```rs
pub enum AnvilError {
  ...
  /// No keys available in anvil instance.
  #[error("no keys available in anvil instance")]
  NoKeysAvailable,
}
```

```rs
pub fn on_anvil_with_wallet_and_config(
        self,
        f: impl FnOnce(alloy_node_bindings::Anvil) -> alloy_node_bindings::Anvil,
    ) -> Result<
        ...,
        AnvilError,
    >
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/builder.rs
+++ b/crates/provider/src/builder.rs
@@ -376,12 +376,10 @@ impl<L, F> ProviderBuilder<L, F, Ethereum> {
     /// use in tests.
     pub fn on_anvil_with_wallet(
         self,
-    ) -> AnvilProviderResult<
-        <JoinedEthereumWalletFiller<F> as ProviderLayer<
-            L::Provider,
-            alloy_transport_http::Http<reqwest::Client>,
-        >>::Provider,
-    >
+    ) -> <JoinedEthereumWalletFiller<F> as ProviderLayer<
+        L::Provider,
+        alloy_transport_http::Http<reqwest::Client>,
+    >>::Provider
     where
         F: TxFiller<Ethereum>
             + ProviderLayer<L::Provider, alloy_transport_http::Http<reqwest::Client>, Ethereum>,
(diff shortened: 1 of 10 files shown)


crates/provider/src/builder.rs
```diff
@@ -420,10 +422,13 @@ impl<L, F> ProviderBuilder<L, F, Ethereum> {
     pub fn on_anvil_with_wallet_and_config(
         self,
         f: impl FnOnce(alloy_node_bindings::Anvil) -> alloy_node_bindings::Anvil,
-    ) -> <JoinFill<F, WalletFiller<alloy_network::EthereumWallet>> as ProviderLayer<
-        L::Provider,
-        alloy_transport_http::Http<reqwest::Client>,
-    >>::Provider
+    ) -> Result<
```
mattsse, line 425: I'd like to keep this function and not return a result, but we can add `fn try_on_anvil...` for this and on_anvil calls this and unwraps

mattsse (changes requested):

crates/provider/src/builder.rs
```diff
@@ -347,6 +346,11 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
     }
 }
 
+type JoinedEthereumWalletFiller<F> = JoinFill<F, WalletFiller<alloy_network::EthereumWallet>>;
+
+#[cfg(any(test, feature = "anvil-node"))]
+type AnvilProviderResult<T> = Result<T, alloy_node_bindings::anvil::AnvilError>;
```
moricho, line 352: Added these type aliases to avoid clippy error

moricho: @mattsse By the way, should we create a type alias for `Http<reqwest::Client>`? It appears so many times in this repository and it can cause clippy warnings `very complex type used. Consider factoring parts into type definitions` when a type is complex, as in this case.

crates/provider/src/builder.rs
moricho: fixed: https://github.com/alloy-rs/alloy/pull/1055/commits/bd30ae136f9ecc4f61354c5a2a17fca832b4ee6f

mattsse (approved):

PR #1056, opened 2024-07-15
chore(eip7702): remove `OptionalNonce`

### LINKED ISSUE #1048: [Feature] 7702 auth nonce is no longer optional
### Component

consensus, eips, genesis

converted from list with optional item to regular value

https://github.com/ethereum/EIPs/pull/8677

## TODO
* remove OptionalNonce type

https://github.com/alloy-rs/alloy/blob/8abc6071cbae6e41579c4ee21610d20848ab253c/crates/eips/src/eip7702/auth_list.rs#L22-L23

### Additional context

_No response_

onbjerg:
## Motivation

The nonce is no longer optional for devnet2.

## Solution

Remove `OptionalNonce` and make the nonce a `u64`.

This should **not be merged** before devnet2.

Closes #1048

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip7702/auth_list.rs
```diff
@@ -20,7 +20,7 @@ pub struct Authorization {
     /// The address of the authorization.
     pub address: Address,
     /// The nonce for the authorization.
-    pub nonce: OptionalNonce,
+    pub nonce: u64,
```
mattsse, line 23: we should serde serialize this with alloy_serde::quantity

mattsse (approved):

crates/eips/src/eip7702/auth_list.rs
onbjerg: 10ce80a

onbjerg:
I think we should get this merged seeing as we are gunning for the current EIP7702 spec. Then, we'll merge #1059 as well.

PR #1058, opened 2024-07-15
fix: require storageKeys value broken bincode serialization from #955

jtguibas:
Fixes bincode serialization issues from #955.

## Motivation

Bincode is a common crate used for efficient serialization/deserialization in the Rust ecosystem.

## Solution

Removes deserialize_with = "alloy_serde::null_as_default".

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip2930.rs
```diff
@@ -19,8 +19,6 @@ pub struct AccessListItem {
     /// Account addresses that would be loaded at the start of execution
     pub address: Address,
     /// Keys of storage that would be loaded at the start of execution
-    // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
-    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::null_as_default"))]
```
mattsse, line 23: this will fail the existing access_list_null_storage_keys test, we should just use serde(default) for now and clarify if we even need to support the null case

mattsse (approved):

PR #1059, opened 2024-07-16
feat: Add 7702 tx enum

moricho:
## Motivation
The 7702 tx type has been added at https://github.com/alloy-rs/alloy/pull/1046, so it would be better to support 7702 tx enum in the consensus layer as well.

TODO
- [ ] Add an encoding/decoding test for 7702 tx
- [ ] Update `rpc-types-eth`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
thanks for this!

this type is will be supported by devnet1 soon, but I think we can actually include it already, though we should wait until devnet is up

moricho:
thanks, it’s fine to leave this until devnet is ready :)

onbjerg (approved):
lgtm, I will rebase it and resolve the conflicts:) thanks for your contribution

mattsse (approved):
lgtm

PR #1061, opened 2024-07-16
refactor: Refactor ChainStreamPoller

### LINKED ISSUE #1053: [Feature] Improve Chainpoller stream impl
### Component

rpc

ref https://github.com/alloy-rs/alloy/pull/1052

this stream loop is a bit weird, because this listens for new heads and then fetches the missing range in full, using a cache to store to tmp store the full block:

https://github.com/alloy-rs/alloy/blob/c461d6885784e45482e8b6962c4a098d0b493d57/crates/provider/src/chain.rs#L49-L52

fetching the blocks can also be done sequentially and we should yield the next block immediately, like a `FuturesOrdered` or similar

cc @popzxc 

### Additional context

_No response_

popzxc:
Fixes #1053

This PR turned out a bit larger than I wanted it to be, but I don't think it makes much sense to split it into several ones.
Here's what changed:

- `TransportErrorKind::recoverable` marked as deprecated, because there already is a more advanced `is_retry_err` method. The deprecation message suggests using it, or `RetryBackoffLayer` for HTTP.
- Retries were removed from `PollerBuilder` and `ChainStreamPoller`. The logic there was pretty basic and it wouldn't solve most problems that can realistically occur. IMHO, recovering belongs to the transport domain; otherwise, there will be one recovery implementation per client user, and a good solution is already present in the project - `RetryBackoffLayer`. I guess something like this is needed for WebSocket as well.
- `ChainPollerStream` was split into two parts. One handles initialization (works like a tiny builder), and one actually stores the stream state. It allowed us to get rid of `NO_BLOCK_NUMBER` used as a placeholder.
- Dedicated error type was added to simplify implementation.
- Finally, batches are now requested via JSON RPC batch request instead of sequential single requests.


crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+            self.chain_tip = new_number;
+        }
+        Ok(())
+    }
+
+    /// Tries to fetch the range of blocks that can surely be yielded via single batch request.
+    async fn fetch_blocks(&self) -> Result<Vec<Block>, ChainStreamError<T>> {
+        // We don't want to request too many batches by accident.
+        // TODO: Should be a property on transport or client.
+        const MAX_BATCH_SIZE: usize = 10;
```
popzxc, line 108: In most realistic cases, batches will be smaller than that, and if not, we remember `chain_tip` so we will send a new request on the next iteration without having to wait for the block number stream.

Still, probably it makes sense to make max batch size configurable, but it's out of scope for this PR.

popzxc:
@mattsse could you please take a look?

mattsse:
ah sorry, thanks for the ping, will check in a bit

crates/provider/src/chain/mod.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    fn init_tracing() {
+        let _ = tracing_subscriber::fmt::try_init();
+    }
+
+    async fn with_timeout<T: Future>(fut: T) -> T::Output {
+        tokio::select! {
+            _ = tokio::time::sleep(Duration::from_secs(1)) => panic!("Operation timed out"),
+            out = fut => out,
+        }
```
mattsse, line 66: there's also 

https://docs.rs/tokio/latest/tokio/time/fn.timeout.html

crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+    _phantom: PhantomData<N>,
+}
+
+impl<T: Transport + Clone, N: Network> PollerStream<T, N> {
+    pub(super) fn stream(
+        client: WeakClient<T>,
+        next_yield: Option<u64>,
+    ) -> impl Stream<Item = Block> + 'static {
+        stream! {
+            let mut self_ = match Self::new(client, next_yield).await {
```
mattsse, line 32:
```suggestion
            let mut this = match Self::new(client, next_yield).await {
```

crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+use tokio_stream::wrappers::BroadcastStream;
+
+use super::ChainStreamError;
+
+/// Stream state.
+#[derive(Debug)]
+pub(super) struct PollerStream<T, N = Ethereum> {
+    client: WeakClient<T>,
+    /// Next block that should be yielded.
+    next_yield: BlockNumber,
```
mattsse, line 18: can we name this block or something, I find this a bit hard to follow otherwise

crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+                let blocks = match self_.new_blocks().await {
+                    Ok(blocks) => blocks,
+                    Err(err) => {
+                        warn!(%err, "irrecoverable error, stopping the stream");
+                        break;
+                    }
+                };
+                for block in blocks {
+                    yield block;
+                    self_.next_yield += 1;
```
mattsse, line 49: hmm, this doesn't check block numbers?
but we apparently also didn't check this in the original impl

crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+        let futures: FuturesOrdered<Waiter<Option<Block>>> = range
+            .map(|number| {
+                batch_request
+                    .add_call::<_, Option<Block>>(
+                        "eth_getBlockByNumber",
+                        &(U64::from(number), false),
+                    )
+                    .expect("Cannot serialize call params")
+            })
+            .collect();
```
mattsse, line 128: I'm actually not sure how batch request works under the hood, but I believe there's no guarantee that responses are in the same order as requested

mattsse (changes requested):
I think I can follow along, but have some questions and would like to see some additional docs.

crates/provider/src/chain/stream.rs
popzxc: Yup. I think we can say that the invariant here is that `PollerStream::new_blocks` is responsible for making sure that the state of the object is unambiguous, and whatever it produces is a sequential list of blocks that _must_ be yielded.
Otherwise, handling situations like reorgs (which are not implemented yet -- will be done later) will be spread among many different places and it will be hard to manage.
I've added a comment for that.

crates/provider/src/chain/stream.rs
popzxc: Each request is represented by an individual future, so what we collect here is futures for responses we've created, so I think it's safe to assume that the order will be the same.
I've added a comment for that.

crates/provider/src/chain/stream.rs
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+        Ok(blocks)
+    }
+
+    /// Performs a single step of the polling loop.
+    /// Either returns a list of blocks to yield, or an irrecoverable error.
+    async fn new_blocks(&mut self) -> Result<Vec<Block>, ChainStreamError<T>> {
+        self.update_chain_tip().await?;
+        self.fetch_blocks().await
```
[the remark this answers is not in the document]
popzxc: Yes, that's deliberate. I think a "fully parallel" solution would be significantly more complex, and not sure if that's justified here -- in most realistic cases the blocks will be consumed one by one right after they're fetched.
This solution is more efficient than the previous one, though, since we don't have to wait for the underlying poller if we know that we're lagging behind.

popzxc:
GM @mattsse
Sorry for pinging again; do you think anything else should be done in the scope of this PR?

popzxc:
@mattsse this PR got a bit stale :)
If this functionality is not something alloy will benefit from, I can close the PR (though I've promised to ship some currently lacking tests that depend on this functionality), but I would prefer not to leave it hanging.

popzxc:
@DaniPopes @gakonst Sorry for pinging, but I'm unsure how to proceed with this PR.

popzxc:
I'll close this PR as stale. If you think it's worth merging, I'm happy to re-open and resolve merge conflicts.

Issue #325, opened 2024-03-15
[Feature] Add methods to retrieve full_blocks and full_blocks_with_txs in `watch` and `subscribe` methods.


yash-atreya:
### Component

provider, pubsub

### Motivation

Currently, the `watch_blocks` methods only returns block hashes and `subscribe_blocks` returns complete blocks but not with transaction bodies.

### Solution

Add the following methods:

1. `watch_full_blocks` which returns complete block bodies.
2. `watch_full_blocks_with_txs`, which returns the block body with the tx objects in that block.
3. `subscribe_blocks_with_txs` which returns the block body with the tx objects in that block.

07Vaishnavi-Singh:
heyy @mattsse @jenpaff would like to work on this issue

PR #1065, opened 2024-07-16
feat(rpc-client): WIP - post poll execution in PollerBuilder

yash-atreya:
## Motivation

Ref https://github.com/alloy-rs/alloy/issues/325  

## Solution

WIP 

Adds a provision to call another RPC method after a success response is received from a polled request. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


yash-atreya:
Closing this for now. Will pick up once again after rethinking the stream approach such that the `type Item` of the stream is the return type of the post-poll RPC request.

PR #1069, opened 2024-07-18
fix: deserialized with bincode

CeciliaZ030:
Encounter error when deserializing TxEip2930 which contains AccessList: https://github.com/paradigmxyz/reth/blob/c3347f323c0a6d7784772624d91edebf4a76d36d/crates/primitives/src/transaction/eip2930.rs#L54 In my version the deserialization was overridden with:
```
pub struct AccessListItem {
    /// Account addresses that would be loaded at the start of execution
    pub address: Address,
    /// Keys of storage that would be loaded at the start of execution
    #[cfg_attr(
        any(test, feature = "arbitrary"),
        proptest(
            strategy = "proptest::collection::vec(proptest::arbitrary::any::<B256>(), 0..=20)"
        )
    )]
    // In JSON, we have to accept `null` for storage key, which is interpreted as an empty array.
    #[cfg_attr(feature = "serde", serde(deserialize_with = "alloy_serde::null_as_default"))]
    pub storage_keys: Vec<B256>,
}
```
and it returns Error when deserializing with `bincode` because null and None from Option is interpreted differently.
```
called `Result::unwrap()` on an `Err` value: InvalidTagEncoding(2)
```
The solution turn erroneous results into None

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/serde/src/optional.rs
+++ b/crates/serde/src/optional.rs
@@ -8,6 +8,6 @@ where
     T: Deserialize<'de> + Default,
     D: Deserializer<'de>,
 {
-    let s: Option<T> = Deserialize::deserialize(deserializer)?;
+    let s: Option<T> = Deserialize::deserialize(deserializer).ok();
     Ok(s.unwrap_or_default())
 }


mattsse (commented):
this is no longer used on storage keys

https://github.com/alloy-rs/alloy/blob/423a4c53c1bffdfb2acca2daee60215194f7711d/crates/eips/src/eip2930.rs#L20-L21

but this change is now just serde(default)

PR #1079, opened 2024-07-19
Added stages to the sync info rpc type

loocapro:
## Motivation
Ref: https://github.com/paradigmxyz/reth/issues/7838

CC: @mattsse 

## Solution
Adding a new optional field to sync info to show stages sync detail on eth_syncing rpc method.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/syncing.rs
+++ b/crates/rpc-types-eth/src/syncing.rs
@@ -1,6 +1,6 @@
 use alloy_primitives::{B512, U256, U64};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use std::collections::{BTreeMap, HashMap};
+use std::collections::BTreeMap;
 
 /// Syncing info
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/syncing.rs
```diff
@@ -16,6 +16,9 @@ pub struct SyncInfo {
     pub warp_chunks_amount: Option<U256>,
     /// Warp sync snapshot chunks processed.
     pub warp_chunks_processed: Option<U256>,
+    /// The details of the sync stages as an hashmap
+    /// where the key is the name of the stage and the value is the block number.
+    pub stages: Option<HashMap<String, U256>>,
```
mattsse, line 21: let's do a vec here because we expect that these are ordered
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub stages: Option<Vec<String, U64>>,
```

we can also add a helper type for this, so that this is compat with erigon see also:

https://github.com/ledgerwatch/erigon/blob/aefb97b07d1c4fd32a66097a24eddd8f6ccacae0/turbo/jsonrpc/eth_system.go#L55-L60

mattsse (changes requested):

mattsse (changes requested):

gakonst:
will also need to wire it here! https://github.com/paradigmxyz/reth/blob/9e685163deb57b2a2e1661195c093071d1e85647/crates/rpc/rpc/src/eth/helpers/spec.rs#L55-L62 thx for taking this on

mattsse (changes requested):
last nits eheh

mattsse (approved):

Issue #1074, opened 2024-07-18
[Bug] all `*_many` methods don't work on Reth


0xlosh:
### Component

rpc

### What version of Alloy are you on?

0.1/0.2

### Operating System

Linux

### Describe the bug

Following methods do not work with a reth node
```    
    async fn call_many: not implemented
    
    async fn debug_trace_call_many(
        &self,
        txs: Vec<TransactionRequest>,
        block: BlockNumberOrTag,
        trace_options: GethDebugTracingCallOptions,
    )
    
    async fn trace_call_many<'a>(
        &self,
        request: TraceCallList<'a, N>,
    )
```
reason being that reth accepts different types
```
    async fn call_many(
        &self,
        bundle: Bundle,
        state_context: Option<StateContext>,
        state_override: Option<StateOverride>,
    )
    
     pub async fn debug_trace_call_many(
        &self,
        bundles: Vec<Bundle>,
        state_context: Option<StateContext>,
        opts: Option<GethDebugTracingCallOptions>,
    ) 
    
     pub async fn trace_call_many(
        &self,
        calls: Vec<(TransactionRequest, HashSet<TraceType>)>,
        block_id: Option<BlockId>,
    )
``` 

instead of `Vec<TransactionRequest>`, reth expects a `Vec<Bundle>` (debug_trace_call_many)
instead of `BlockNumberOrTag`, reth expects a `StateContext` (debug_trace_call_many)
instead of `Vec<TraceType>`, reth expects a `HashSet<TraceType>` (trace_call_many)

mattsse:
@zerosnacks mind looking into this, we likely have wrong arg order here

zerosnacks:
Spotted a discrepancy, specifically in `debug_trace_call_many` where the `bundles` parameter doesn't skip over `block_override`, incompatible with `Vec<TransactionRequest>`.

https://github.com/alloy-rs/alloy/blob/067cc464bd1357e745653709db051707b949cc6b/crates/rpc-types-eth/src/call.rs#L11-L12

https://github.com/alloy-rs/alloy/blob/067cc464bd1357e745653709db051707b949cc6b/crates/provider/src/ext/debug.rs#L125

[eth_callMany](https://www.chainnodes.org/docs/ethereum/eth_callMany) does not seem to be implemented in Alloy so that will have to be added in the `eth_` namespace

[...]

PR #1085, opened 2024-07-20
feat: Add call_many 

DoTheBestToGetTheBest:
as mentionned here https://github.com/alloy-rs/alloy/issues/1074 call_many is not implemented

### DIFF
--- a/crates/consensus/src/request.rs
+++ b/crates/consensus/src/request.rs
@@ -41,6 +41,12 @@ impl From<WithdrawalRequest> for Request {
     }
 }
 
+impl From<ConsolidationRequest> for Request {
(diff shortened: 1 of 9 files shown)


0xlosh:
`call_many` should either take a `Vec<TransactionRequest` or a `Bundle`, idk which one the alloy devs want to use, reth wants a `Bundle`,

DoTheBestToGetTheBest:
> `call_many` should either take a `Vec<TransactionRequest>` or a `Bundle`, idk which one the alloy devs want to use, reth wants a `Bundle`

you're right, thank you!

zerosnacks:
Thanks! Merging into `zerosnacks/fix-all-many-methods-rpc` to be included

PR #1090, opened 2024-07-23
feat(rpc/trace): filter matches with trace

jsvisa:
## Motivation

Filter the trace with trace {action,result}'s address, not by tx's {from, to}_address, inspired by https://github.com/openethereum/openethereum/blob/6c2d392d867b058ff867c4373e40850ca3f96969/crates/ethcore/src/trace/types/filter.rs#L95-L131 and https://github.com/ledgerwatch/erigon/blob/72ab70be8b28964a8446e52b7a9975958c6726bf/turbo/jsonrpc/trace_filtering.go#L238-L289

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/filter.rs
+++ b/crates/rpc-types-trace/src/filter.rs
@@ -1,4 +1,8 @@
 //! `trace_filter` types and support
+use crate::parity::{
+    Action, CallAction, CreateAction, CreateOutput, RewardAction, SelfdestructAction, TraceOutput,
+    TransactionTrace,
+};
 use alloy_primitives::Address;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/filter.rs
```diff
@@ -93,38 +97,79 @@ pub enum TraceFilterMode {
...
-                    self.from_addresses.contains(&from)
-                        || to.map_or(false, |to_addr| self.to_addresses.contains(&to_addr))
-                }
-                TraceFilterMode::Intersection => {
-                    self.from_addresses.contains(&from)
-                        && to.map_or(false, |to_addr| self.to_addresses.contains(&to_addr))
-                }
-            },
+    /// Returns `true` if the given `TransactionTrace` matches this filter.
+    pub fn matches(&self, trace: &TransactionTrace) -> bool {
```
mattsse, line 140: can we please document this a bit more

crates/rpc-types-trace/src/filter.rs
```diff
@@ -93,38 +97,79 @@ pub enum TraceFilterMode {
     Intersection,
 }
 
+/// Address filter.
+/// This is a set of addresses to match against.
+/// An empty set matches all addresses.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct AddressFilter(HashSet<Address>);
```
mattsse, line 104:
```suggestion
pub struct AddressFilter(pub HashSet<Address>);
```

crates/rpc-types-trace/src/filter.rs
```diff
@@ -93,38 +97,79 @@ pub enum TraceFilterMode {
...
+    }
+
+    /// Returns `true` if the filter is empty.
+    pub fn matches_all(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
 /// Helper type for matching `from` and `to` addresses. Empty sets match all addresses.
 #[derive(Clone, Debug, PartialEq, Eq)]
```
mattsse, line 131: this needs clear docs about waht this is supposed to match

crates/rpc-types-trace/src/filter.rs
```diff
@@ -93,38 +97,79 @@ pub enum TraceFilterMode {
...
+}
+
+impl From<Vec<Address>> for AddressFilter {
+    fn from(addrs: Vec<Address>) -> Self {
+        Self::from_iter(addrs)
+    }
+}
+
+impl AddressFilter {
+    /// Returns `true` if the given address is in the filter.
```
mattsse, line 119: this is not entirely correct, this also matches if the filter is empty

mattsse (changes requested):
all of these things are not well document, so I'd like to see a few more docs here

crates/rpc-types-trace/src/filter.rs
jsvisa: sure, let me add some

mattsse (approved):
ty!

PR #1098, opened 2024-07-24
feat: add helper for decoding custom errors

### LINKED ISSUE #787: [Feature] Get abi error data from an error response on call / send transaction
### Component

contract

Right now, assuming I have a contract Foo with a function bar, if it has an error, I have to do a lot of parsing / handling to get the abi error. I would like to have a more convenient function to handle this.

Example of how I'm doing it right now:

```rust
sol!(
    #[allow(missing_docs)]
    #[sol(rpc)]
    Foo,
    "abis/Foo.json"
);

let foo_contract = Foo::new(FOO_ADDRESS, &provider);

let tx_receipt = match foo_contract
        .bar()
        .from(signer)
        .send()
        .await
    {
        Ok(result) => result.get_receipt().await?,
        Err(error) => match error {
            Error::TransportError(TransportError::ErrorResp(err)) => {
                let data = err.data.unwrap_or_default();
                let data = data.get().trim_matches('"');
                let data = Bytes::from_str(data)?;
                let decoded_error = Foo::FooErrors::abi_decode(&data, true)?;

            }
            _ => {
                return Err(eyre::eyre!(
                    "Received TransportError with data: {:?}",
                    eyre::eyre!(error)
                ))
            }
        },
    };
```

### Additional context

_No response_

klkvr:
## Motivation

Closes #787 

## Solution

Adds helper method allowing decoding of custom errors from `ErrorPayload`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-rpc/src/response/error.rs
+++ b/crates/json-rpc/src/response/error.rs
@@ -1,8 +1,10 @@
+use alloy_primitives::Bytes;
+use alloy_sol_types::SolInterface;
 use serde::{
     de::{DeserializeOwned, MapAccess, Visitor},
     Deserialize, Deserializer, Serialize,
 };
-use serde_json::value::RawValue;
+use serde_json::{value::RawValue, Value};
 use std::{borrow::Borrow, fmt, marker::PhantomData};
 
 /// A JSONRPC-2.0 error object.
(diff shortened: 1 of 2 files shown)


crates/json-rpc/src/response/error.rs
```diff
@@ -67,6 +69,18 @@ impl<E> ErrorPayload<E> {
...
 }
 
+/// Recursively traverses the value, looking for hex data that it can extract.
+///
+/// Inspired by ethers-js logic:
+/// <https://github.com/ethers-io/ethers.js/blob/9f990c57f0486728902d4b8e049536f2bb3487ee/packages/providers/src.ts/json-rpc-provider.ts#L25-L53>
+fn spelunk_revert(value: &Value) -> Option<Bytes> {
+    match value {
+        Value::String(s) => s.parse().ok(),
+        Value::Object(o) => o.values().flat_map(spelunk_revert).next(),
```
DaniPopes, line 79: `values().find(|| )`

DaniPopes (approved):

mattsse (approved):
nice!

I think we could add an example for this @zerosnacks @yash-atreya

PR #1101, opened 2024-07-25
refactor: add network-primitives

klkvr:
## Motivation

`alloy-rpc-types-eth` are supposed to be concrete, thus having generics on types defined in it is not helpful and makes it hard to define implementations based on traits from other crates.

## Solution

- Move `ReceiptResponse` and `TransactionResponse` from `alloy-network` to new crate `alloy-network-primitives`, allowing their usage without depending on `alloy-network`
- Move `BlockTransactions<T>` into `network-primitives` and provide implementations for `T: TransactionResponse` instead of concrete `rpc_types_eth::Transaction`

This sets us up for block response abstraction, making it possible to add getters for `BlockTransactions` generic over `N::TransactionResponse` ref https://github.com/alloy-rs/alloy/issues/267

Note: This changed `Iterator::Item` for `BlockTransactionHashes` from `&B256` to `B256` as this is what `TransactionResponse::tx_hash` returns. Is it fine or should I update the trait instead? It seems that it's OK to expect `TransactionResponse` to own a hash.

This also removes `BlockTransactionHashesMut`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


crates/rpc-types-eth/src/block.rs
mattsse, line 1: can we make sure we reexport everything from network-primitives so this change doesn't break existing imports? e.g. BlockTransactions

mattsse (approved):
lgtm, one nit

as dicussed we need this split so we can avoid cyclic dependencies for network/consensus/rpc

> This also removes BlockTransactionHashesMut

this is reasonable I believe, I don't think we want to support this

crates/rpc-types-eth/src/block.rs
```diff
@@ -14,7 +15,7 @@ pub use alloy_eips::{
 /// Block representation
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Block<T = Transaction> {
+pub struct Block {
```
mattsse, line 18: hmm, can we keep this generic? otherwise we can't reuse that block type?

crates/rpc-types-eth/src/block.rs
```diff
@@ -204,324 +205,6 @@ impl TryFrom<Header> for alloy_consensus::Header {
...
-    }
-}
-
-impl<'a> std::iter::FusedIterator for BlockTransactionHashes<'a, Transaction> {}
-
-/// An Iterator over the transaction hashes of a block.
-///
-/// See [`BlockTransactions::hashes_mut`].
-#[derive(Debug)]
-pub struct BlockTransactionHashesMut<'a, T = Transaction>(BlockTransactionHashesInnerMut<'a, T>);
```
mattsse, line 425: on second thought, we should be able to keep this type?

klkvr: if we'd add `fn tx_hash_mut() -> &mut TxHash` ot `TransactionResponse`, then yes

crates/rpc-types-eth/src/block.rs
klkvr: right, this could be useful in some cases, though I think for most of the rollups entire `Block` would be different as there'll be no withdrawals, so we'd likely want to just add more reusable generic components for custom blocks (like `BlockTransactions<T>`)

added generic back as it doesn't hurt anyway

mattsse (approved):
ah now I get why we can't reuse the BlocktransactionsMut

lgtm

crates/rpc-types-eth/src/block.rs
mattsse: yep I also think this will be the case but this can be useful for rollups that only have 1 additional tx type for example

PR #1102, opened 2024-07-25
fix: make Parity TraceResults output optional

m1stoyanov:
That will resolve the issue when the RPC server returns "output": null, as Nethermind does.

## Motivation

That will resolve the issue when the RPC server returns "output": null, as Nethermind does.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/parity.rs
+++ b/crates/rpc-types-trace/src/parity.rs
@@ -29,7 +29,7 @@ pub enum TraceType {
 #[serde(rename_all = "camelCase")]
 pub struct TraceResults {
     /// Output of the trace
-    pub output: Bytes,
+    pub output: Option<Bytes>,
     /// Enabled if [TraceType::StateDiff] is provided
     pub state_diff: Option<StateDiff>,
     /// Enabled if [TraceType::Trace] is provided, otherwise an empty vec
--- a/crates/rpc-types-trace/src/parity.rs
+++ b/crates/rpc-types-trace/src/parity.rs
@@ -29,7 +29,8 @@ pub enum TraceType {
 #[serde(rename_all = "camelCase")]
 pub struct TraceResults {
     /// Output of the trace
-    pub output: Option<Bytes>,
+    #[serde(deserialize_with = "alloy_serde::null_as_default")]
+    pub output: Bytes,
     /// Enabled if [TraceType::StateDiff] is provided
     pub state_diff: Option<StateDiff>,
     /// Enabled if [TraceType::Trace] is provided, otherwise an empty vec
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/parity.rs
```diff
@@ -29,7 +29,7 @@ pub enum TraceType {
 #[serde(rename_all = "camelCase")]
 pub struct TraceResults {
     /// Output of the trace
-    pub output: Bytes,
```
mattsse, line 32: can we do serde(default) instead and ideally a test with a nethermind response?

mattsse (changes requested):

crates/rpc-types-trace/src/parity.rs
m1stoyanov: With serde(default), deserialization fails again with the same error:

> invalid type: null

We can use something like #[serde(deserialize_with = "deserialize_null_default")], but wouldn't it be better to make it optional, exactly like the vm_trace?

mattsse: hmm, I guess you're right, curious why this can be null in the first place?

mattsse: if you know why this can be null, please add it as comment

m1stoyanov: In my opinion, this is a bug in versions prior to Nethermind/v1.26.0. Unfortunately, many EVMs use those versions.

mattsse: I see, then it would make more sense to use serde(default) imo

m1stoyanov: I just confirmed that the bug is still not resolved in Nethermind version 1.27.0.
Using #[serde(default)] makes no difference.
#[serde(default)] will work only with workaround like here:
[https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d9ffd2fb157f8316a2abc85e8c08dcd1](url)

So my suggestion is just making it optional.

DaniPopes: We can use null_as_default https://github.com/alloy-rs/alloy/blob/4ba532373eb085a5426d2de8c71c0b7273e9b62c/crates/serde/src/optional.rs#L6

m1stoyanov: null_as_default is ok imo
I will rewrite pr accordingly

mattsse: please use serde(default)

null_as_default has a slightly different purpose

crates/rpc-types-trace/src/parity.rs
```diff
@@ -821,4 +822,31 @@ mod tests {
...
+            "from": "0x0000000000000000000000000000000000000000000000000000000000042f66",
+            "to": "0x0000000000000000000000000000000000000000000000000000000000042f67"
+          }
+        }
+      }
+    }
+  },
+  "trace": [],
+  "vmTrace": null,
+  "transactionHash": "0xe56a5e7455c45b1842b35dbcab9d024b21870ee59820525091e183b573b4f9eb"
```
mattsse, line 846: is this a real transaction?
wasn't able to find it

crates/rpc-types-trace/src/parity.rs
```diff
@@ -821,4 +822,31 @@ mod tests {
         let serialized = serde_json::to_string_pretty(&trace).unwrap();
         similar_asserts::assert_eq!(serialized, reference_data);
     }
+    #[test]
+    fn test_nethermind_trace_result_null_output_value() {
+        let reference_data = r#"{
+  "output": null,
```
mattsse, line 828: is there a nethermind issue for this, I want to understand when this can be null to rule out something we may have missed

crates/rpc-types-trace/src/parity.rs
m1stoyanov: https://gnosisscan.io/tx/0xe56a5e7455c45b1842b35dbcab9d024b21870ee59820525091e183b573b4f9eb

I reduced it slightly

crates/rpc-types-trace/src/parity.rs
m1stoyanov: To be honest, I didn't dig very deeply, but it seems that vm_trace can also be null.

mattsse (approved):
looking at the og docs
https://openethereum.github.io/JSONRPC-trace-module#trace_transaction
it looks like output should always be there

so treating null as empty makes sense

DaniPopes (approved):

PR #1108, opened 2024-07-27
feat(alloy-provider): add abstraction for `NonceFiller` behavior

### LINKED ISSUE #1073: [Bug] NonceFiller doesn't work after some contracts are deployed.
### Component

provider, pubsub

0.2.0

### Operating System

macOS (Apple Silicon)

### Describe the bug

NonceFiller doesn't work after contracts deployment. Use Anvil with tests. To tell the truth for testing purposes I would suggest not to cache nonce, it is really very inconvenient and use two fillers or initialize this one with the parameter. 

`TransportError(ErrorResp(ErrorPayload { code: -32003, message: "nonce too low", data: None })`

StackOverflowExcept1on:
### Motivation

Resolves #1073

### Solution

I added `trait NonceManager` and two implementations: `SimpleNonceManager` and `CachedNonceManager` to allow the user to choose the behavior of nonces filling. In the recommended fillers, I also changed the implementation from `CachedNonceManager` to `SimpleNonceManager` to make the recommended fillers more resilient to chain reorganizations. The issue was described in more detail in https://github.com/alloy-rs/alloy/issues/1073#issuecomment-2251181724.

### PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


zerosnacks:
Good point on the re-org resistance. An alternative solution could be to have a way of having a user be able to indicate that the nonce needs to be refetched (assuming they have a system that is re-org aware and can simply call this function).
I would prefer that over having two distinct implementations.

StackOverflowExcept1on:
@zerosnacks

> An alternative solution could be to have a way of having a user be able to indicate that the nonce needs to be refetched.

Even if we indicate whether to fetch nonce from the rpc when sending the rpc request, it will still be inconvenient since you won't be able to fully customize the behavior as you wish.

zerosnacks:
> Even if we indicate whether to fetch nonce from the rpc when sending the rpc request, it will still be inconvenient since you won't be able to fully customize the behavior as you wish.

Could you expand on that? What kind of behavior would you be customizing that is not covered by my proposed approach?

StackOverflowExcept1on:
@zerosnacks this will help implement any behavior, but on the caller side you will need to pass a boolean value or use some wrapper, which seems inconvenient on the caller side. We will also need wrappers for providers, alloy::contract, etc.

StackOverflowExcept1on:
Also consider that you can't just get a NonceFiller instance and indicate it to fetch rpc, since it might be wrapped with JoinFill. So you'll have to invent some mechanism to interact with the NonceFiller via interior mutability.

zerosnacks:
cc @onbjerg / @prestwich curious to hear your thoughts on this proposal / approach

onbjerg:
It makes sense to have other types of nonce management strategies. You can already do this using a custom filler, but this simplifies implementing a new nonce strategy somewhat, so I think this is ok.

StackOverflowExcept1on:
Hi, can someone take a look at the PR?

crates/provider/src/fillers/nonce.rs
```diff
@@ -6,15 +6,89 @@ use crate::{
...
+}
+
+/// This [`NonceManager`] implementation will fetch the transaction count for any new account it
+/// sees.
+///
+/// Unlike [`CachedNonceManager`], this implementation does not store the transaction count locally,
+/// which results in more frequent calls to the provider, but it is more resilient to chain
+/// reorganizations.
+#[derive(Clone, Debug, Default)]
+pub struct SimpleNonceManager;
```
mattsse, line 33:
```suggestion
#[non_exhaustive]
pub struct SimpleNonceManager;
```

crates/provider/src/fillers/nonce.rs
```diff
@@ -6,15 +6,89 @@ use crate::{
...
+use async_trait::async_trait;
 use dashmap::DashMap;
 use futures::lock::Mutex;
 use std::sync::Arc;
 
-/// A [`TxFiller`] that fills nonces on transactions.
+/// A trait that determines the behavior of filling nonces.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait NonceManager: Clone + Send + Sync + std::fmt::Debug + Default {
```
mattsse, line 17: hmm, do we need default here?

this could be a problem for nonce managers that require external context

crates/provider/src/builder.rs
```diff
@@ -144,8 +144,11 @@ impl<L, N> ProviderBuilder<L, Identity, N> {
     /// Add nonce management to the stack being built.
     ///
     /// See [`NonceFiller`]
-    pub fn with_nonce_management(self) -> ProviderBuilder<L, JoinFill<Identity, NonceFiller>, N> {
-        self.filler(NonceFiller::default())
+    pub fn with_nonce_management<M: NonceManager>(
+        self,
+        nonce_manager: M,
+    ) -> ProviderBuilder<L, JoinFill<Identity, NonceFiller<M>>, N> {
```
mattsse, line 150: I'm fine with this but I'd like a new function that calls this function the previous nonce manager (now called CachedNonce I believe)

crates/provider/src/fillers/nonce.rs
```diff
@@ -40,11 +114,18 @@ use std::sync::Arc;
 /// # }
 /// ```
 #[derive(Clone, Debug, Default)]
-pub struct NonceFiller {
-    nonces: DashMap<Address, Arc<Mutex<u64>>>,
+pub struct NonceFiller<M: NonceManager> {
```
mattsse, line 117: could we make the CachedNonceMg the default type here? so that we can rid of type hints

mattsse (changes requested):
sorry for the delay,

supportive, but have some nits and suggestions

crates/provider/src/fillers/nonce.rs
StackOverflowExcept1on: yes, it is really not used

crates/provider/src/fillers/nonce.rs
StackOverflowExcept1on: ok, I'll try but why CachedNonceManager and not SimpleNonceManager? see PR description about reorgs

mattsse: I don't have a preference, also fine with SimpleNonceManager as the default, good point

crates/provider/src/builder.rs
StackOverflowExcept1on: ok, so you want something like `with_{simple, cached}_nonce(_management) `methods?

crates/provider/src/fillers/nonce.rs
StackOverflowExcept1on: so you want separate methods for simple and cached managers, and the user can implement their own type and pass it to this function?
```
error[E0423]: expected value, found struct `SimpleNonceManager`
  --> crates/provider/src/fillers/nonce.rs:110:28
   |
10 |     .with_nonce_management(SimpleNonceManager)
   |                            ^^^^^^^^^^^^^^^^^^ constructor is not visible here due to private fields
```

crates/provider/src/builder.rs
mattsse: yeah, I think def for the builtin cached one

mattsse (approved):
lgtm

PR #1120, opened 2024-08-02
chore(rpc): Remove redundant wrapper around `AnyNetwork::TransactionResponse` type

emhane:
## Motivation

`Transaction` type already has other fields

## Solution

Remove wrapper type `WtihOtherFields`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/any/mod.rs
+++ b/crates/network/src/any/mod.rs
@@ -69,7 +69,7 @@ impl Network for AnyNetwork {
 
     type TransactionRequest = WithOtherFields<TransactionRequest>;
 
-    type TransactionResponse = WithOtherFields<Transaction>;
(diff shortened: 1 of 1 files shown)


mattsse (dismissed):

mattsse:
I believe the idea here was to use that and then remove the other fields from transaction, which I believe we still want to do, so we should actually keep it that for now

emhane:
> I believe the idea here was to use that and then remove the other fields from transaction, which I believe we still want to do, so we should actually keep it that for now

make sense. imo we should build the op type in the same way, saves updating in two places by not having duplicate fields.

Issue #1122, opened 2024-08-03
[Feature] Add authorization list to TransactionRequest


mattsse:
### Component

rpc

## TODO

* add an additional optional field `authorization_list: Option<SignedAuthorization>` to TransactionRequest for 7702 transactions.

onbjerg:
@mattsse this should be an `Option<Vec<SignedAuthorization>>`:)

PR #1125, opened 2024-08-04
feat: add authorization list to TransactionRequest

DoTheBestToGetTheBest:
closes https://github.com/alloy-rs/alloy/issues/1122


onbjerg (changes requested):
this is not correct, an authorization list is a `Vec<SignedAuthorization>`.

DoTheBestToGetTheBest:
> this is not correct, an authorization list is a `Vec<SignedAuthorization>`, so this should be `Option<Vec<SignedAuthorization>>`.

make sens, ty for the review.

mattsse (approved):

PR #1129, opened 2024-08-06
fix(rpc): show data in when cast send result in custom error

cuiweixie:
## Motivation

```
➜  foundry_test git:(master) ✗ cast send 0x5FbDB2315678afecb367f032d93F642f64180aa3  "customError()" --private-key $PK
Error:
server returned an error response: error code 3: execution reverted:
```
cast send result in execution revert with out reason(data) is hard to debug.  

old version of cast send result would be like:  
```
(code: 3, message: execution reverted: CustomError3333, data: Some(String("0x8d6ea8be0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000f437573746f6d4572726f72333333330000000000000000000000000000000000"))) 
```

it's good to make cast send print data like "0x8d6ea8be...".

## Solution

add data to print in fmt

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-rpc/src/response/error.rs
+++ b/crates/json-rpc/src/response/error.rs
@@ -88,10 +88,7 @@ impl<ErrData: fmt::Display> fmt::Display for ErrorPayload<ErrData> {
             "error code {}: {}{}",
             self.code,
             self.message,
-            self.data
-                .as_ref()
-                .map(|data| format!(", data: {}",data.to_string()))
-                .unwrap_or_default()
+            self.data.as_ref().map(|data| format!(", data: {}", data)).unwrap_or_default()
         )
     }
 }
--- a/crates/json-rpc/src/response/error.rs
+++ b/crates/json-rpc/src/response/error.rs
@@ -81,9 +81,12 @@ fn spelunk_revert(value: &Value) -> Option<Bytes> {
     }
 }
 
-impl<ErrData> fmt::Display for ErrorPayload<ErrData> {
+impl<ErrData: Default + fmt::Display+ Clone > fmt::Display for ErrorPayload<ErrData> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "error code {}: {}", self.code, self.message)
+        write!(f, "error code {}: {}, data: {}", self.code, self.message,
+               <Option<ErrData> as Clone>::clone(&self.data).
+                   unwrap_or_else(|| ErrData::default()).to_string().trim_matches('"')
+        )
     }
 }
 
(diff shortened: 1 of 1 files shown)


cuiweixie:
see https://github.com/foundry-rs/foundry/issues/8606 
see https://github.com/foundry-rs/foundry/issues/8603 
see https://github.com/foundry-rs/foundry/pull/8607

crates/json-rpc/src/response/error.rs
```diff
@@ -81,9 +81,12 @@ fn spelunk_revert(value: &Value) -> Option<Bytes> {
...
 }
 
-impl<ErrData> fmt::Display for ErrorPayload<ErrData> {
+impl<ErrData: Default + fmt::Display+ Clone > fmt::Display for ErrorPayload<ErrData> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "error code {}: {}", self.code, self.message)
+        write!(f, "error code {}: {}, data: {}", self.code, self.message,
+               <Option<ErrData> as Clone>::clone(&self.data).
+                   unwrap_or_else(|| ErrData::default()).to_string().trim_matches('"')
+        )
```
mattsse, line 89: logging data makes sense and restricting ErrData to display makes sense I believe.

but this can be solved without out the default and the clone impl by mapping data, converting to string, if none this then should be an empty string

mattsse (changes requested):

crates/json-rpc/src/response/error.rs
cuiweixie: change to:
```
impl<ErrData: fmt::Display> fmt::Display for ErrorPayload<ErrData> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(
            f,
            "error code {}: {}, data: {}",
            self.code,
            self.message,
            self.data
                .as_ref()
                .map(|data| data.to_string())
                .unwrap_or_default()
        )
    }
}
```

mattsse (approved):

cuiweixie:
[cargo_test_workspace.log](https://github.com/user-attachments/files/16511804/cargo_test_workspace.log)
```
cargo test --workspace
```
run successfully in local machine, please rerun the workflows.

mattsse (approved):

mattsse:
needs cargo +nightly fmt and clippy

cuiweixie:
> needs cargo +nightly fmt and clippy

done.

mattsse (approved):

PR #1131, opened 2024-08-06
chore(dep): feature gate jwt in engine types

mattsse:
ref https://github.com/paradigmxyz/reth/pull/10141


DaniPopes (approved):
you can ignore engine and beacon types in the wasm workflow

mattsse:
I just forgot to also make rand optional...

PR #1140, opened 2024-08-12
rpc-types-eth: rm `PeerCount`

tcoratger:


crates/rpc-types-eth/src/syncing.rs
```diff
@@ -57,6 +57,18 @@ pub enum PeerCount {
...
+    fn from(number: u32) -> Self {
+        Self::Number(number)
+    }
+}
+
+impl From<U64> for PeerCount {
+    fn from(hex: U64) -> Self {
+        Self::Hex(hex)
+    }
+}
```
mattsse, line 70: we actually don't have any use for this type and this is legacy and should actually be removed.

so let's remove this type instead and I'll open an issue on reth to phase this out

mattsse (changes requested):

crates/rpc-types-eth/src/syncing.rs
tcoratger: Done

mattsse (approved):

DaniPopes (approved):

PR #1145, opened 2024-08-14
fix: return more user-friendly error on tx timeout

klkvr:
## Motivation

Currently on `TxWatcher` timeout we just drop it, resulting in `TransportError::BackendGone` error which is not really user-friendly.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm

DaniPopes (approved):

PR #1150, opened 2024-08-16
feat(geth/trace): add field log.position

jsvisa:
## Motivation

Add a field `log.position` in the response of debug_traceXXX RPC with `withLog:true` config(which was added in https://github.com/ethereum/go-ethereum/pull/28389)

ref https://github.com/paradigmxyz/reth/issues/10258

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -56,6 +56,9 @@ pub struct CallLogFrame {
     /// The data of the log.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub data: Option<Bytes>,
+    /// The position of the log relative to subcalls within the same trace.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
DaniPopes, line 60: pls use alloy_serde::quantity::opt with u64 instead of ruint

DaniPopes (changes requested):

crates/rpc-types-trace/src/geth/call.rs
jsvisa: fixed, ptal

onbjerg (approved):

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -56,6 +56,9 @@ pub struct CallLogFrame {
     /// The data of the log.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub data: Option<Bytes>,
+    /// The position of the log relative to subcalls within the same trace.
+    #[serde(default, with = "alloy_serde::quantity::opt")]
+    pub position: Option<u64>,
```
onbjerg, line 61: should we skip serializing this if `None`?

DaniPopes: we should because null is generally invalid for this type of stuff

jsvisa: Thanks for it, so we should also add skip_if_none

DaniPopes (approved):

PR #1153, opened 2024-08-18
consensus: upstream cfg in test for clippy

tcoratger:
Serde cfg can be upstreamed so that the import `super::*` only appends for the appropriate cfg.


onbjerg:
don't think this makes too much sense since we might add other tests

PR #1154, opened 2024-08-19
fix: remove `OtherFields` from Transaction and Block

klkvr:
## Motivation

Those currently consume all extra data, forcing `AnyNetwork` users to do `value.inner.other` instead of just `value.other`.

Should be merged after https://github.com/alloy-rs/alloy/pull/1106 which makes block responses use `WithOtherFields<Block>` while with this PR returned blocks are just `Block` which will miss extra fields

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


mattsse (approved):
good call, we can remove these now

I believe this only has one sideffect that we need to tune reth's rpc responses a bit

PR #1159, opened 2024-08-20
refactor(`ProviderCall`): don't rely on `Caller` in `RpcWithBlock`

klkvr:
## Motivation

Proposing different approach for integrating `ProviderCall` into `RpcWithBlock`

## Solution

- Add `map_params` method to `RpcCall`, allowing to change request params type. Useful for conversion of `Params` into `ParamsWithBlock`
- Add `WithBlockInner` type, containing two variants: https://github.com/alloy-rs/alloy/blob/1d49e4bd41c25e42c2e82fb24e438efdde301329/crates/provider/src/provider/with_block.rs#L40-L55
  The idea is to allow people providing arbitrary `Fn(BlockId) -> ProviderCall` methods
- Change `RpcWithBlock::Future` to `ProviderCall`. When awaited, `RpcWithBlock` will convert `WithBlockInner` into `ProviderCall` by either changing rpc call params, or invoking function.
- Also changed `ProviderCall::Ready` inner value to `Result` to allow error propagation from synchronous sources.

Altogether this allows API like this in user-defined `Provider` implementations:
```rust
fn get_code_at(&self, address: Address) -> RpcWithBlock<T, Address, Bytes> {
    let db = self.db_client();
    (move |block_id| {
        ProviderCall::Ready(Some(db.get_code(address, block_id)))
    }).into()
}
```

Motivation for separate `RpcCall` variant is that it's still getting special treatment from both `ProviderCall` and `Provider` trait itself. It is the only `ProviderCall` variant holding `Params`, and some of `Provider` functions rely on being able to use non-`static` params making it impossible to box type producing such `RpcCall`.

Another option is to ensure that `RpcWithBlock` holds `Params`, changing `ProviderCallProducer` to also accept it, however this would require all user-defined implementations of provider to re-assemble the params as expected by RPC which might be inconvenient, especially if method implementation does not interact with RPC at all. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -78,6 +78,11 @@ impl<Params> Request<Params> {
     pub fn set_subscription_status(&mut self, sub: bool) {
         self.meta.set_subscription_status(sub);
     }
+
+    /// Change type of the request parameters.
+    pub fn map_params<NewParams>(self, map: impl Fn(Params) -> NewParams) -> Request<NewParams> {
+        Request { meta: self.meta, params: map(self.params) }
+    }
 }
 
 /// A [`Request`] that has been partially serialized. The request parameters
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -80,7 +80,10 @@ impl<Params> Request<Params> {
     }
 
     /// Change type of the request parameters.
-    pub fn map_params<NewParams>(self, map: impl FnOnce(Params) -> NewParams) -> Request<NewParams> {
+    pub fn map_params<NewParams>(
+        self,
+        map: impl FnOnce(Params) -> NewParams,
+    ) -> Request<NewParams> {
         Request { meta: self.meta, params: map(self.params) }
     }
 }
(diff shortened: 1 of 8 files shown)


mattsse (commented):
cool, 
I like this a lot more and this is basically what we're looking for

crates/json-rpc/src/request.rs
```diff
@@ -78,6 +78,11 @@ impl<Params> Request<Params> {
     pub fn set_subscription_status(&mut self, sub: bool) {
         self.meta.set_subscription_status(sub);
     }
+
+    /// Change type of the request parameters.
+    pub fn map_params<NewParams>(self, map: impl Fn(Params) -> NewParams) -> Request<NewParams> {
```
DaniPopes, line 83:
```suggestion
    pub fn map_params<NewParams>(self, map: impl FnOnce(Params) -> NewParams) -> Request<NewParams> {
```

DaniPopes (approved):

mattsse (approved):

Issue #1158, opened 2024-08-20
[Feature] Support arbitrary subscription ids


mattsse:
### Component

provider, pubsub

### Describe the feature you would like

this currently only supports u256 ids:

https://github.com/alloy-rs/alloy/blob/4a270df340d21359283b3d7fb2c1204baf9f4084/crates/pubsub/src/managers/in_flight.rs#L58-L70

and causes issues where servers don't return u256 as subscription ids #1155

## TODO

* introduce a new type that can handle both u256 and string:

```
pub enum SubscriptionId {
	/// Numeric id
	Num(U267),
	/// String id
	Str(String),
}
```

and use this as a replacement for the `server_id: U256` where used:

https://github.com/alloy-rs/alloy/blob/4a270df340d21359283b3d7fb2c1204baf9f4084/crates/pubsub/src/service.rs#L170-L170

DaniPopes:
https://github.com/alloy-rs/alloy/blob/4a270df340d21359283b3d7fb2c1204baf9f4084/crates/json-rpc/src/common.rs#L26-L33

HowHsu:
why do we use an Enum rather than directly use String as Id

PR #1163, opened 2024-08-20
fix: allow arbitrary strings in subscription ids

klkvr:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/1158

This adds `SubId` enum similar to `json_rpc::Id` but without `None` variant and with `U256` in place of `Number variant`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
clean

PR #1171, opened 2024-08-21
Add trait methods for constructing `alloy_rpc_types_eth::Transaction` to `alloy_consensus::Transaction`

emhane:
## Motivation

## Solution

Adds trait methods to `alloy_consensus::Transaction`, and implements for respective types following implementation of https://github.com/paradigmxyz/reth/blob/302ce7f482d5c3189457fe43d8155676d01766ff/crates/transaction-pool/src/traits.rs#L766-L876

Part 2: https://github.com/emhane/alloy/pull/1

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


emhane:
closing in favour of duplicate https://github.com/alloy-rs/alloy/pull/1172

PR #1172, opened 2024-08-21
Add trait methods for constructing `alloy_rpc_types_eth::Transaction` to `alloy_consensus::Transaction`

emhane:
## Motivation

## Solution

Adds trait methods to `alloy_consensus::Transaction`, and implements for respective types following implementation of https://github.com/paradigmxyz/reth/blob/302ce7f482d5c3189457fe43d8155676d01766ff/crates/transaction-pool/src/traits.rs#L766-L876

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/consensus/src/transaction/mod.rs
```diff
@@ -59,6 +103,17 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
...
+    /// Returns the transaction type
+    fn ty(&self) -> u8;
+
+    /// Returns the EIP2930 `access_list` for the particular transaction type. Returns `None` for
+    /// older transaction types.
+    fn access_list(&self) -> Option<&AccessList>;
+
+    /// Blob versioned hashes for eip4844 transaction. For previous transaction types this is
+    /// `None`.
+    fn blob_versioned_hashes(&self) -> Option<Vec<&B256>>;
```
mattsse, line 116: blob versioned hashes is a part of the tx and should not require an allocation
```suggestion
    fn blob_versioned_hashes(&self) -> Option<&[B256]>;
```

mattsse (changes requested):
all of these we definitely want, one nit re blob versioned hashes

crates/consensus/src/transaction/mod.rs
```diff
@@ -51,6 +52,49 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
...
+    fn max_priority_fee_per_gas(&self) -> Option<u128>;
+
+    /// Return the max priority fee per gas if the transaction is an EIP-1559 transaction, and
+    /// otherwise return the gas price.
+    ///
+    /// # Warning
+    ///
+    /// This is different than the `max_priority_fee_per_gas` method, which returns `None` for
+    /// non-EIP-1559 transactions.
+    fn priority_fee_or_price(&self) -> u128;
```
klkvr, line 74: I think this can have a default impl like `self.max_priority_fee_per_gas().unwrap_or(self.max_fee_per_gas())`

though I'm wondering what's a usecase for this method?

crates/consensus/src/transaction/eip7702.rs
```diff
@@ -267,7 +267,8 @@ impl TxEip7702 {
         mem::size_of::<U256>() + // value
         self.access_list.size() + // access_list
         self.input.len() + // input
-        self.authorization_list.capacity() * mem::size_of::<SignedAuthorization>() // authorization_list
+        self.authorization_list.capacity() * mem::size_of::<SignedAuthorization>()
+        // authorization_list
```
klkvr, line 271:
```suggestion
        self.authorization_list.capacity() * mem::size_of::<SignedAuthorization>() // authorization_list
```

crates/consensus/src/transaction/mod.rs
klkvr: ah got it, its basically how execution layer treats legacy txses wrt priority fee

mattsse (approved):

PR #1173, opened 2024-08-21
Add trait methods to `alloy_network_primitives::TransactionResponse`

emhane:
## Motivation

## Solution

Adds trait methods to `alloy_network_primitives::TransactionResponse` for filling a `alloy_network_primitives::Transaction` from `alloy_consensus::Signed<impl alloy_consensus::Transaction>`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/network-primitives/src/traits.rs
```diff
@@ -44,6 +45,22 @@ pub trait TransactionResponse {
...
+
+    /// Assemble from signed transaction.
+    fn fill(
+        signed_tx: Signed<impl alloy_consensus::Transaction>,
+        signer: Address,
+        block_hash: Option<B256>,
+        block_number: Option<u64>,
+        base_fee: Option<u64>,
+        transaction_index: Option<usize>,
+    ) -> Self;
```
mattsse, line 63: unsure we can do this easily.

so I'd like to hold off on this for now

crates/network-primitives/src/traits.rs
```diff
@@ -44,6 +45,22 @@ pub trait TransactionResponse {
     /// Input data
     #[doc(alias = "calldata")]
     fn input(&self) -> &Bytes;
+
+    /// Returns the gas price formatted for the RPC response.
+    fn gas_price(tx: &impl alloy_consensus::Transaction, base_fee: Option<u64>) -> u128;
+
+    /// Returns the max fee per gas.
+    fn max_fee_per_gas(tx: impl alloy_consensus::Transaction) -> Option<u128>;
```
mattsse, line 53: these are a bit weird, unsure why they need to be part of this trait

mattsse (changes requested):
let's do the additional `trait Transaction` first and think about the TransactionResponse fn(&impl Transaction) functions separately

emhane:
> let's do the additional `trait Transaction` first and think about the TransactionResponse fn(&impl Transaction) functions separately

that's exactly how it is done. see base of prs. first pr: https://github.com/alloy-rs/alloy/pull/1172

mattsse:
sorry I missed that pr

crates/network-primitives/src/traits.rs
emhane: so computation for l1 tx types can be reused by l2, much better than l2s doing copy paste

crates/network-primitives/src/traits.rs
emhane: wdym? this is needed to make progress in reth

emhane:
re-request review @mattsse

crates/consensus/src/transaction/mod.rs
mattsse, line 1: can we extract this to a separate pr?

mattsse (changes requested):

emhane:
extracted yet another bit https://github.com/alloy-rs/alloy/pull/1198

crates/network-primitives/src/traits.rs
```diff
@@ -107,6 +124,25 @@ impl<T: TransactionResponse> TransactionResponse for WithOtherFields<T> {
...
         self.inner.input()
     }
+
+    fn gas_price(tx: &impl alloy_consensus::Transaction, base_fee: Option<u64>) -> u128 {
+        T::gas_price(tx, base_fee)
+    }
+
+    fn max_fee_per_gas(tx: impl alloy_consensus::Transaction) -> Option<u128> {
+        T::max_fee_per_gas(tx)
+    }
```
mattsse, line 134: these don't make much sense because the caller can just call them on tx directly

crates/network-primitives/src/traits.rs
```diff
@@ -107,6 +124,25 @@ impl<T: TransactionResponse> TransactionResponse for WithOtherFields<T> {
...
+
+    fn fill(
+        signed_tx: Signed<impl alloy_consensus::Transaction>,
+        signer: Address,
+        block_hash: Option<B256>,
+        block_number: Option<u64>,
+        base_fee: Option<u64>,
+        transaction_index: Option<usize>,
+    ) -> Self {
```
mattsse, line 143: this function is problematic because this now effectively freezes what additional fields can be included in a tx response.

mattsse (commented):
we can't really solve this on the type trait, this is a server side problem because only there this type is created

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -322,7 +322,66 @@ impl TransactionResponse for Transaction {
...
+            }
+        }
+    }
+
+    fn max_fee_per_gas(tx: impl alloy_consensus::Transaction) -> Option<u128> {
+        match TxType::try_from(tx.ty()).expect("should decode") {
+            TxType::Legacy | TxType::Eip2930 => None,
+            TxType::Eip1559 | TxType::Eip4844 | TxType::Eip7702 => Some(tx.max_fee_per_gas()),
+        }
+    }
```
emhane, line 346: this implementation varies from the impl on `alloy_consensus::Transaction`. this code is ported from reth, without modifying the logic @mattsse

crates/network-primitives/src/traits.rs
emhane: https://github.com/alloy-rs/alloy/pull/1173#discussion_r1732734183

crates/network-primitives/src/traits.rs
emhane: using `TransactionInfo` isn't possible due to circular dep, should I move `TransactionInfo` into this crate?

crates/rpc-types-eth/src/transaction/mod.rs
emhane: https://github.com/paradigmxyz/reth/blob/e6dc947a3728637aa3b0f78c893e8f24dc468357/crates/rpc/rpc-types-compat/src/transaction/mod.rs#L41-L60

crates/network-primitives/src/traits.rs
mattsse: I think this would be reasonable actually

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -319,7 +331,67 @@ impl TransactionResponse for Transaction {
...
+
+    fn fill(
+        signed_tx: Signed<impl alloy_consensus::Transaction>,
+        signer: Address,
+        tx_info: TransactionInfo,
+    ) -> Self {
```
mattsse, line 361: these functions won't work, because they constrain type creation, we can not lock this in in the trait function

mattsse (commented):
converting types from consensus to rpc can't really be solved like this because this constrains it to `alloy_consensus::Transaction`

crates/rpc-types-eth/src/transaction/mod.rs
emhane: how come?

mattsse:
closing this, because we can't support the fill feature like this, because this conversion is type specific.

and the additional gas functions doesn't really make sense like this because this info should be encapsulated in the transaction itself, especially if we proceed with

https://github.com/alloy-rs/alloy/pull/1169

PR #1179, opened 2024-08-22
feat: make block struct generic over header type

### LINKED ISSUE #1177: [Feature] Make `alloy_rpc_types_eth::Block` generic over header
### Component

rpc

Make `Block` generic over header type, for using `Block<alloy_network::Network::HeaderResponse>` in reth RPC.

https://github.com/alloy-rs/alloy/blob/3952a72346ff208c12fc0556a3547727aa9c7864/crates/rpc-types-eth/src/block.rs#L14-L37

### Additional context

_No response_

mvares:
This PR allows Header to assume a type based on generic.

Closes #1177

### DIFF
--- a/crates/rpc-types-eth/src/block.rs
+++ b/crates/rpc-types-eth/src/block.rs
@@ -14,10 +14,10 @@ pub use alloy_eips::{
 /// Block representation
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Block<T = Transaction> {
+pub struct Block<T = Transaction, H = Header> {
     /// Header of the block.
     #[serde(flatten)]
-    pub header: Header,
+    pub header: H,
     /// Uncles' hashes.
     #[serde(default)]
     pub uncles: Vec<B256>,
--- a/crates/rpc-types-eth/src/block.rs
+++ b/crates/rpc-types-eth/src/block.rs
@@ -14,7 +14,7 @@ pub use alloy_eips::{
 /// Block representation
 #[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct Block<T = Transaction, H = Header> {
+pub struct Block<H = Header, T = Transaction> {
     /// Header of the block.
     #[serde(flatten)]
     pub header: H,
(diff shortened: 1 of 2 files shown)


mvares:
hey @emhane, do you know why the CI is failing?

emhane (changes requested):
need to make sure that all the implementations for `Block` will also work for the non-default type. may advice is to add default generics last, after you got everything compiling.

emhane:
sorry wasn't clear enough:

remove the default for the generic.
compile.
fix bugs.
add default generic.

that's how I usually do.

`Block::into_full_block` is only impl for default header generic now

mvares:
> sorry wasn't clear enough:
> 
> remove the default for the generic. compile. fix bugs. add default generic.
> 
> that's how I usually do.
> 
> `Block::into_full_block` is only impl for default header generic now

Hmm. Looks a good start.

mvares:
@emhane, maybe `BlockResponse` should have a Header generic. What do you think?

emhane:
> @emhane, maybe `BlockResponse` should have a Header generic. What do you think?

block response does have generic header + transaction in form of associated types https://github.com/alloy-rs/alloy/blob/dbd9d7712da879ab1736a89caa37c2135fa9fafa/crates/network-primitives/src/traits.rs#L71-L74

crates/rpc-types-eth/src/block.rs
```diff
@@ -36,7 +36,7 @@ pub struct Block<T = Transaction> {
     pub withdrawals: Option<Vec<Withdrawal>>,
 }
 
-impl Block {
+impl Block<Header> {
```
emhane, line 39:
```suggestion
impl Block<H> {
```
nitpick

crates/rpc-types-eth/src/block.rs
```diff
@@ -334,7 +334,7 @@ pub struct BlockOverrides {
     pub block_hash: Option<BTreeMap<u64, B256>>,
 }
 
-impl<T> BlockResponse for Block<T> {
+impl<Header, T> BlockResponse for Block<Header, T> {
```
emhane, line 337:
```suggestion
impl<H, T> BlockResponse for Block<H, T> {
```
nitpick

emhane (approved):

klkvr:
@emhane wdyt about switching generics order? it would turn this update into a non-breaking change and also seems more convenient as imo users are most likely to change the tx generic

PR #1185, opened 2024-08-23
cjore: remove RichBlock and RichHeader types

### LINKED ISSUE #1184: [Feature] Deprecate RichBlock and RichHeader types
### Component

rpc

we no longer need those because we can use `WithOtherFields`

### Additional context

_No response_

mvares:
Closes #1184


mattsse (changes requested):
let's also remove this type entirely:

https://github.com/alloy-rs/alloy/blob/main/crates/rpc-types-eth/src/block.rs#L252-L252

mattsse (approved):

PR #1187, opened 2024-08-24
fix: add missing op fields

mattsse:
ref https://github.com/paradigmxyz/reth/issues/10462


DaniPopes (commented):
shouldnt this be in alloy optimism?

mattsse:
yeah want to move this now

PR #1196, opened 2024-08-26
fix: serde for `depositReceiptVersion`

klkvr:
Should we deprecate optimism types in favor of op-alloy's for the next release?


emhane (approved):
lgtm

mattsse:
@klkvr yeah we should move all of these to op-alloy

klkvr:
> @klkvr yeah we should move all of these to op-alloy now

yep, I've already added them to https://github.com/alloy-rs/op-alloy/pull/45

should we put them under `#[deprecated]` here or just remove on this/next release?

PR #1200, opened 2024-08-26
pub decoder for CallBuilder

Corn3333:
## Motivation
I'm trying to migrate Multicall from Ethers to Alloy. But the field decoder is private, which makes Multicall impossible!

I'm using the code from this pull request, which was ignored!
https://github.com/alloy-rs/alloy/pull/442

## Solution
Make the decoder field public 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -128,7 +128,7 @@ pub struct CallBuilder<T, P, D, N: Network = Ethereum> {
     /// The provider.
     // NOTE: This is public due to usage in `sol!`, please avoid changing it.
     pub provider: P,
-    decoder: D,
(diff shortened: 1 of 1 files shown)


DaniPopes:
Decoder and APIs around it are not stable, and should not be used outside, how are you using it?

Corn3333:
> Decoder and APIs around it are not stable, and should not be used outside, how are you using it?

Please check out this PR: https://github.com/alloy-rs/alloy/pull/442

Decoder is used there, to decode the result of a multicall request

yash-atreya:
This is not what we want currently. Thank you for your contribution.

PR #1202, opened 2024-08-27
feat(transport-http): Http with header auth

yash-atreya:
## Motivation

There is no easy way to send authenticated HTTP requests. Users would have to write a custom transport client implementation. 

## Solution

- Adds a new field in the `transport-http::Http<T>` called `auth: Option<Authorization>`.

- Made accommodations for `reqwest` and `hyper` implementations to incorporate and set the `header` authorization if `auth` is set.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
hmm, not sure we really need this for the `Http<reqwest::Client>` because this is a static auth header, and the user could install that as a default header on the client directly. so we should recommend doing that.

we could however use some examples for this

mattsse:
closing this in favor of #1227 which will allow dynamic headers and other features, static headers can be installed manually already on reqwest client

PR #1207, opened 2024-08-28
chore(features): enable `consensus` and `network` along with `providers`

klkvr:
## Motivation

Both of those are already included as `alloy-provider` dependencies, so this just enables useful re-exports for users doing just `features = ["providers"]`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense

PR #1209, opened 2024-08-28
feat(arbitrary): add suport for `f64`

### LINKED ISSUE #1208: [Feature] Add support for `f64` in `alloy-serde` arbitrary module
### Component

serde

Add de-/serialization of `f64` from/to `String` in module _crates/serde/src/other/arbitrary\_.rs_

https://github.com/alloy-rs/op-alloy/blob/f9514d5914abc6d2f4979db59d6081a60e8adfc7/crates/rpc-types/src/receipt.rs#L64-L86

### Additional context

_No response_

mvares:
Closes #1208

### DIFF
--- a/crates/serde/src/other/arbitrary_.rs
+++ b/crates/serde/src/other/arbitrary_.rs
@@ -47,11 +47,7 @@ impl ArbitraryValue {
     fn from_json_value(value: serde_json::Value) -> Option<Self> {
         match value {
             serde_json::Value::String(s) => {
-                if let Ok(f) = s.parse::<f64>() {
-                    Some(Self::Float(f))
-                } else {
-                    Some(Self::String(s))
-                }
+                s.parse::<f64>().map_or(Some(Self::String(s)), |f| Some(Self::Float(f)))
             }
             _ => todo!(),
         }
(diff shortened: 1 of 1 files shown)


mvares:
can you check out? 

cc @emhane 

PS: I added the `from_json_value` method due to deserialization.

crates/serde/src/other/arbitrary_.rs
```diff
@@ -31,6 +32,7 @@ impl ArbitraryValue {
         match self {
             Self::Null => serde_json::Value::Null,
             Self::Bool(b) => serde_json::Value::Bool(b),
+            Self::Float(n) => serde_json::Value::String(n.to_string()),
```
mattsse, line 35: we should be able to convert this into Value::Number via `Number::from_f64`

mattsse (changes requested):

crates/serde/src/other/arbitrary_.rs
mvares: WDYT?
`Self::Float(n) => serde_json::Number::from_f64(n).into()`

mattsse: that should work, we can treat the None case as 0.0 that should be okay given this is only used for testing

emhane (commented):
taking @mattsse 's review as a confirmation of the issue. i added the 'needs triage' label to the issue since I wasn't sure if ethereum stopped using f64 and hence if it made sense to impl this. OP uses `f64` for the `l1FeeScalar`, but that value is depreciated since ecotone.

mattsse:
I see, then I think this isn't that useful

PR #1210, opened 2024-08-28
fix: value of TxEip1559.ty

bbjubjub2494:
## Motivation

`TxEip1559::ty()` returns a value indicating EIP-2930 transactions without EIP-1559 features. This is not correct.

## Solution

Change the value

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
good catch! ty

PR #1213, opened 2024-08-29
fix: Anvil builder default port

MistApproach:
## Motivation

Anvil builder `new()` method docs states the default port to be 8545. In practice, however, current code sets random port number.

## Solution

Set default port inside `try_spawn()` to 8545, unless otherwise specified.
Unit test added.

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (changes requested):
we actually want it to be assigned to a random port, otherwise the diff instances in tests will conflict entirely which where all of these failures come from

i'd update the docs instead

MistApproach:
Updated doc-comment for `Anvil::new()` instead.

onbjerg (approved):

Issue #1119, opened 2024-08-02
[Feature] Any plans to support multicall like viem?


wozhishilaonanhai:
### Component

contract

### Describe the feature you would like

Are there any plans to support a multicall feature similar to viem? This would greatly improve the efficiency of making multiple calls, especially in scenarios where batch data retrieval is needed. Looking forward to your response, thank you!

PR #1217, opened 2024-08-29
feat: multicall

nhtyy:
## Motivation

https://github.com/alloy-rs/alloy/issues/1119
https://github.com/alloy-rs/alloy/issues/328

## Solution

Introduce a new type for each Multicall Type (`aggregate`, `try_aggregate`. `aggreagte3`) that holds an `R: AsRef<MultiCall>`

When a users creates a multicall for the first time we put it in an arc internally

when a users wants to make a multicall we give them a wrapper with an &Arc<_> to start, users can mutate this wrapper that holds the ref to add calls and set other options. 

You can execute the call from here, or convert it into an "Owned" version of the call (thru the From trait) which clones the arc and you can name the batch of calls without lifetime

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [?] Breaking changes: I think this breaks any downstream implementors of `TransactionRequest`


nhtyy:
Missing tests, If this change is supported I will finish them

nhtyy:
Hi Guys sorry for the changes after Ive already opened it...

I wanted to also share while this branch requires the users have mutable access to the type, ive made another that adds a type for each multicall type

this means a user effectively gets a "builder" from the multicall instance that it can mutate rather than needing the instance itself to be mutable 

Link is below, I think the alias naming could use a little work but its all there otherwise

https://github.com/nhtyy/alloy/pull/2

nhtyy:
To expand on the motivation a user might have calldata that needs to change overtime, or it doesnt. In the latter case a user would probably want to name this type and reuse the calldata

We also dont want to tie every multicall instance to a specific return type. Instead we return "call builders" for each multicall type

To solve these two problems similar to Danis approach we return "call builders" for each multicall type.

As mentioned in the description of this we internally create a multicall contract instance from a provider, and put it inside an Arc, this is the thing that returns the builders, if the builder contains an `&Arc<_>` we will consume the whole type (and the internal call buffer) on `call`

Otherwise if we want to reuse the calldata and name this builder we can use something like `DynAggreagate<T, P, N>` or `pub type SolAggreagate<T, P, C, N>`
and `call` on these types will clone the internal call buffer.

nhtyy:
Also I wasnt able to reproduce the test failures! Im on OSX

nhtyy:
So I think ive seen the error of my ways with the Arc and removed it, since the bindings are really cheap to store theres no use in trying to re use them.. You can still name a call builder for reuse without lifetime assuming you owned the provider you created the instance with.

However Im wondering if this doesn't belong in `alloy-contract`, I think It would be nice if I could call `.aggregate` directly on a `Provider`

nhtyy:
Close in favor of eth_callMany

PR #1218, opened 2024-08-30
eip2930: add payload methods for `TxEip2930`

tcoratger:
Related https://github.com/paradigmxyz/reth/pull/10623


mattsse (approved):

mattsse:
looks like we need to bump msrv

tcoratger:
> looks like we need to bump msrv

@mattsse done here https://github.com/alloy-rs/alloy/pull/1219

crates/consensus/src/transaction/eip2930.rs
```diff
@@ -223,6 +223,19 @@ impl TxEip2930 {
...
+
+    /// Output the length of the RLP signed transaction encoding, _without_ a RLP string header.
+    pub fn payload_len_with_signature_without_header(&self, signature: &Signature) -> usize {
+        let payload_length = self.fields_len() + signature.rlp_vrs_len();
+        // 'transaction type byte length' + 'header length' + 'payload length'
+        1 + length_of_length(payload_length) + payload_length
+    }
+
+    /// Output the length of the RLP signed transaction encoding. This encodes with a RLP header.
+    pub fn payload_len_with_signature(&self, signature: &Signature) -> usize {
```
mattsse, line 235: should all of those accept `S: EncodableSignature` instead?

mattsse (commented):
ptal @klkvr

crates/consensus/src/transaction/eip2930.rs
tcoratger: Same doesn't hurt to make the methods more generic imo, just modified

klkvr:
hmm, not sure we need those? we already have `encoded_len_with_signature` and `encode_with_signature` which accept `with_header: bool`

tcoratger:
Oh that's true thanks

Issue #1161, opened 2024-08-20
[Feature] Add JWT layer for transport


mattsse:
### Component

provider, pubsub

### Describe the feature you would like

when interacting with jwt authenticated endpoints (engine API) we need to refresh claims from time to time (60s).

this should be supported as some form of transport layer that can added to http.

PR #1227, opened 2024-09-02
feat(transport-http): layer client

yash-atreya:
## Motivation

Currently, there is no way to modify the absolute low-level request such as `http::Request`. 

This is due to the limitation that `transport` layers work over `alloy_json_rpc::RequestPacket` and not `http::Request`. 

## Solution

Making the `hyper` transport HTTP client generic over an underlying service. 

By default, it serves as a vanilla hyper client like before (use `HyperTransport::new()`), with the added benefit of adding tower layers on top of the `hyper::Client` and using that in `HyperTransport::with_service(service)`. 

This new set of layers will work over `http::Request` which is used by `hyper` allowing us to modify request properties such as headers. 

A `HyperTransport` with layers can be initialized like so:
```rust
let hyper_client = Client::builder(TokioExecutor::new()).build_http::<Full<HyperBytes>>();

let service = tower::ServiceBuilder::new().layer(SetRequestHeaderLayer::if_not_present(header::USER_AGENT, HeaderValue::from_static("alloy app"))).service(hyper_client);

let layer_transport = alloy_transport_http::HyperTransport::with_service(service);

let rpc_client = alloy_rpc_client::RpcClient::new(layer_transport, true);
```

This will help implement a JWT auth layer #1161 that refreshes auth tokens. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


yash-atreya:
The `LoggingLayer` is just for testing and demo, I'll be removing it

crates/transport-http/src/layers/logging.rs
mattsse, line 1: do we even need these?
we should try to reuse the tower-http things

mattsse (commented):
cool, this should work.

I'd like to see some more advanced tests/exampls here that make use of some tower-http components

for example auth

see also https://github.com/tower-rs/tower-http/blob/3f98dc19bc6c70aa6e4912b63538d89443f03b19/examples/axum-key-value-store/src/main.rs

crates/transport-http/src/layers/logging.rs
yash-atreya: Yeah to be removed https://github.com/alloy-rs/alloy/pull/1227#issuecomment-2324595402

yash-atreya:
What we currently have working is the following:

- A new HTTP transport client based on `hyper` that lets users create and layer services on top of each other that modify the underlying `http::Request` or `http::Response` types. 
- This is enabled through the use of `tower::ServiceBuilder` 
- One can now write a layer with a service that works over `http::Request`. Examples include #1161 that modifies the auth header,  and/or append a `X-Flashbots-Signature` https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication
- tower-http layers that do not change the response type or error type work as well, this is due to the restrictive trait bounds we've used. 
- All of this is very useful but not ideal, we want to work with all tower-http layers

TODO

- tower-http layers like `TimeoutLayer` or `TraceLayer` don't work because they modify the response type to something like `http::Response<TimeoutBody<Incoming>>` which is incompatible with expected response type i.e `type HyperResponse = http::Response<Incoming>` type. Thinking of ways to make this more flexible. 

cc @mattsse @klkvr

crates/transport-http/Cargo.toml
```diff
@@ -35,7 +35,7 @@ hyper = { workspace = true, default-features = false, optional = true }
 hyper-util = { workspace = true, features = ["full"], optional = true }
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "hyper"]
```
mattsse, line 38: can you compare how many additional deps/build time this will now use as default?

crates/transport-http/src/hyper_layer_transport.rs
```diff
@@ -0,0 +1,144 @@
...
+where
+    S: Service<HyperRequest<B>, Response = HyperResponse> + Clone + Send + Sync + 'static,
+    S::Future: Send,
+    S::Error: std::error::Error + Send + Sync + 'static,
+    B: From<Bytes> + Buf + Send + 'static + Clone,
+{
+    /// Create a new [HyperLayerTransport] with the given URL and service.
+    pub const fn new(url: Url, service: S) -> Self {
+        Self { url, service, _pd: PhantomData }
+    }
```
mattsse, line 44: this should not enforce trait bounds here

crates/transport-http/src/hyper_layer_transport.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Make a request to the server using the given service.
+    pub fn request(&mut self, req: RequestPacket) -> TransportFut<'static> {
+        let this = self.clone();
+        let span = debug_span!("HyperLayerTransport", url = %this.url);
+        Box::pin(
+            async move {
+                debug!(count = req.len(), "sending request packet to server");
+                let ser = req.serialize().map_err(TransportError::ser_err)?;
+                // convert the Box<RawValue> into a hyper request<B>
+                let body = Full::from(Bytes::from(<Box<[u8]>>::from(<Box<str>>::from(ser))));
```
mattsse, line 55: this is just ser.get().as_bytes().into()?

mattsse (changes requested):
nits

overall this looks good

crates/transport-http/src/hyper_layer_transport.rs
```diff
@@ -0,0 +1,144 @@
...
+pub type HyperRequest<B> = Request<Full<B>>;
+
+/// Alias for [`Response<Incoming>`]
+pub type HyperResponse = Response<Incoming>;
+
+/// Alias for pinned box future that results in [`HyperResponse`]
+pub type HyperResponseFut<T = HyperResponse, E = Error> =
+    Pin<Box<dyn Future<Output = Result<T, E>> + Send + 'static>>;
+
+impl<S, B> HyperLayerTransport<S, B>
```
klkvr, line 34: this now almost completely duplicates `HyperTransport` impl.

given that hyper's client implements `tower::Service` on it's own, I am wondering if we could unify this in `HyperTransport` by making underlying hyper client generic?

klkvr:
> * tower-http layers like `TimeoutLayer` or `TraceLayer` don't work because they modify the response type to something like `http::Response<TimeoutBody<Incoming>>` which is incompatible with expected response type i.e `type HyperResponse = http::Response<Incoming>` type. Thinking of ways to make this more flexible.

I think we could just use `Body` trait to restrict response body? https://docs.rs/http-body/1.0.0/http_body/trait.Body.html

something like `Body<Data = Bytes>`

crates/transport-http/Cargo.toml
yash-atreya: Adds about ~20s to build time for `transport-http` crate along with 21 deps.

crates/transport-http/src/hyper_transport.rs
```diff
@@ -1,63 +1,78 @@
...
 pub type HyperClient = hyper_util::client::legacy::Client<
     hyper_util::client::legacy::connect::HttpConnector,
     http_body_util::Full<::hyper::body::Bytes>,
 >;
 
-/// An [`Http`] transport using [`hyper`].
-pub type HyperTransport = Http<HyperClient>;
+/// A [hyper] client that can be used with tower layers.
+#[derive(Clone, Debug)]
+pub struct HyperTransport<B = Full<Bytes>, S = HyperClient> {
```
klkvr, line 26: I am wondering if removing `Http` here is fine? I think it's possible to keep it, though unsure how useful is it in general

crates/transport-http/src/lib.rs
```diff
@@ -37,19 +36,28 @@ use url::Url;
...
+        Self { url, transport: Some(transport) }
     }
 
     /// Get a reference to the URL.
     pub const fn url(&self) -> &Url {
         &self.url
     }
+
+    /// Get a reference to the client.
+    pub const fn transport(&self) -> &Option<T> {
```
klkvr, line 58: why do we need to add this?

crates/transport-http/src/lib.rs
```diff
@@ -37,19 +36,28 @@ use url::Url;
...
 }
 
 impl<T> HttpConnect<T> {
     /// Create a new [`HttpConnect`] with the given URL.
     pub const fn new(url: Url) -> Self {
-        Self { url, _pd: PhantomData }
+        Self { url, transport: None }
+    }
+
+    pub const fn with_transport(transport: T, url: Url) -> Self {
```
klkvr, line 48: why do we need to add this?

crates/transport-http/src/hyper_transport.rs
```diff
@@ -1,63 +1,78 @@
...
-    B: From<Bytes> + Buf + Send + 'static,
+    S: Service<Request<B>, Response = HyperResponse> + Clone + Send + Sync + 'static,
+    S::Future: Send,
+    S::Error: std::error::Error + Send + Sync + 'static,
+    B: From<Vec<u8>> + Send + 'static + Clone,
 {
-    /// Make a request.
-    fn request_hyper(&self, req: RequestPacket) -> TransportFut<'static> {
+    /// Make a request to the server using the given service.
+    pub fn request(&mut self, req: RequestPacket) -> TransportFut<'static> {
```
klkvr, line 66: I don't think this should be `pub`

crates/transport-http/src/hyper_transport.rs
```diff
@@ -105,43 +122,63 @@ where
     }
 }
 
-impl<C, B> Service<RequestPacket> for &Http<Client<C, Full<B>>>
+impl<B, S> Http<HyperTransport<B, S>>
```
klkvr, line 125: hmm, why do we use `Http` here? it means we now have url both in `Http` and `HyperTransport`

crates/transport-http/src/lib.rs
klkvr: ah I see, that's to have `TransportConnect` impl for any instance of `HyperTransport`

I am not sure we should have that. I believe that point of `TransportConnect` is in knowing how to instantiate transport. If we require user to provide it with a ready transport instance, then user can just directly configure this instance instead of relying on `TransportConnect`

klkvr (changes requested):

yash-atreya:
> I am not sure we should have that. I believe that point of TransportConnect is in knowing how to instantiate transport. If we require user to provide it with a ready transport instance, then user can just directly configure this instance instead of relying on TransportConnect

> we can only have TransportConnect implemented for default hyper transport I think

This makes sense!
Removed the `transport` from `HttpConnect`, and impl `TransportConnect` for default HyperTransport

crates/transport-http/src/hyper_transport.rs
yash-atreya: Resolved, `HyperTransport` now infers the url from `Http<T>`

crates/transport-http/src/hyper_transport.rs
```diff
@@ -1,74 +1,96 @@
...
 
-impl<C, B> Http<Client<C, Full<B>>>
+impl<B, S> HyperTransport<B, S> {
+    /// Create a new [HyperTransport] with the given URL and service.
+    pub const fn with_service(service: S) -> Self {
+        Self { service, _pd: PhantomData }
+    }
+}
+
+impl<B, S> HyperTransport<B, S>
```
klkvr, line 63: can we keep this impl block identical to the one we had before? previous implementation was for `Http<...>`, so we didn't have to pass `url` arg, there's no reason to change that

crates/transport-http/src/hyper_transport.rs
klkvr: I'd prefer us to keep `HyperTransport = Http<..>` type alias here, because right now `HyperTransport` on its own is not a transport

We basically only need the new type as a wrapper around hyper client with phatomdata, so we can rename it into `HyperClient`

klkvr (changes requested):
nits

crates/transport-http/src/hyper_transport.rs
```diff
@@ -105,43 +126,39 @@ where
     }
 }
 
-impl<C, B> Service<RequestPacket> for &Http<Client<C, Full<B>>>
```
klkvr, line 108: why did we remove impl for `&Http`?

yash-atreya: Added back

klkvr (approved):
lgtm

PR #1229, opened 2024-09-03
feat(network-primitives): expose more fields via block response traits

yash-atreya:
## Motivation

`HeaderResponse` exposed very limited fields. `BlockResponse` doesn't expose `other` fields in case of `WithOtherFields<Block>`. 

## Solution

- Expose `miner, gas_limit, mix_hash, difficulty` in `HeaderResponse`. Exposing these fields helps to initialize a `revm::BlockEnv` like so: 

```rust
BlockEnv {
    number: U256::from(block.header().number()),
    timestamp: U256::from(block.header().timestamp()),
    coinbase: block.header().miner(),
    difficulty: block.header().difficulty(),
    prevrandao: block.header().mix_hash(),
    basefee: U256::from(block.header().base_fee_per_gas().unwrap_or_default()),
    gas_limit: U256::from(block.header().gas()),
    ..Default::default()
}
```

- Add `other_fields(&self) -> Option<OtherFields>` to `BlockResponse` which returns `None` by default unless overridden. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -68,16 +68,28 @@ pub trait HeaderResponse {
     /// Blob fee for the next block (if EIP-4844 is supported)
     fn next_block_blob_fee(&self) -> Option<u128>;
 
-    /// Miner/Coinbase of the block
-    fn miner(&self) -> Address;
+    /// Coinbase/Miner of the block
+    fn coinbase(&self) -> Address;
 
     /// Gas limit of the block
-    fn gas(&self) -> u128;
+    fn gas_limit(&self) -> u128;
 
     /// Mix hash of the block
+    ///
+    /// Before the merge this proves, combined with the nonce, that a sufficient amount of
+    /// computation has been carried out on this block: the Proof-of-Work (PoW).
+    ///
+    /// After the merge this is `prevRandao`: Randomness value for the generated payload.
+    ///
+    /// This is an Option because it is not always set by non-ethereum networks.
+    ///
+    /// See also <https://eips.ethereum.org/EIPS/eip-4399>
+    /// And <https://github.com/ethereum/execution-apis/issues/328>
     fn mix_hash(&self) -> Option<B256>;
 
     /// Difficulty of the block
+    ///
+    /// Unused after the Paris (AKA the merge) upgrade, and replaced by `prevrandao`.
     fn difficulty(&self) -> U256;
 }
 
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/traits.rs
```diff
@@ -67,6 +67,18 @@ pub trait HeaderResponse {
...
+    fn miner(&self) -> Address;
+
+    /// Gas limit of the block
+    fn gas(&self) -> u128;
+
+    /// Mix hash of the block
+    fn mix_hash(&self) -> Option<B256>;
+
+    /// Difficulty of the block
+    fn difficulty(&self) -> U256;
```
mattsse, line 81: maybe we don't need this, I assume this could eventually be removed

crates/network-primitives/src/traits.rs
```diff
@@ -67,6 +67,18 @@ pub trait HeaderResponse {
 
     /// Blob fee for the next block (if EIP-4844 is supported)
     fn next_block_blob_fee(&self) -> Option<u128>;
+
+    /// Miner/Coinbase of the block
+    fn miner(&self) -> Address;
```
mattsse, line 72: I'd like to use coinbase here

crates/network-primitives/src/traits.rs
```diff
@@ -67,6 +67,18 @@ pub trait HeaderResponse {
...
+
+    /// Gas limit of the block
+    fn gas(&self) -> u128;
+
+    /// Mix hash of the block
+    fn mix_hash(&self) -> Option<B256>;
```
mattsse, line 78: this needs docs, that is is now `prevrandao` the  randomness value for the block

crates/network-primitives/src/traits.rs
```diff
@@ -67,6 +67,18 @@ pub trait HeaderResponse {
 
     /// Blob fee for the next block (if EIP-4844 is supported)
     fn next_block_blob_fee(&self) -> Option<u128>;
+
+    /// Miner/Coinbase of the block
+    fn miner(&self) -> Address;
+
+    /// Gas limit of the block
+    fn gas(&self) -> u128;
```
mattsse, line 75: I'd like to name this gas_limit

mattsse (changes requested):
a few nits

crates/network-primitives/src/traits.rs
yash-atreya: OP stack based L2s use `block.difficulty` for generating random values.

yash-atreya: But I guess `block.prevrandao` serves the same purpose on L2s

mattsse (approved):
lgtm, pending @klkvr

klkvr (approved):

PR #1233, opened 2024-09-04
feat(rpc-types-engine): no-std support

refcell:
### Description

> [!NOTE]
>
> Blocked by https://github.com/alloy-rs/alloy/pull/1252

Adds `no_std` for the `alloy-rpc-types-engine` crate.


crates/rpc-types-engine/src/lib.rs
```diff
@@ -5,6 +5,10 @@
 )]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![cfg_attr(not(feature = "jwt"), no_std)]
```
mattsse, line 8: this should use std feature to disable no_std

crates/rpc-types-engine/src/lib.rs
```diff
@@ -5,6 +5,10 @@
...
+
+#[cfg(not(feature = "jwt"))]
```
mattsse, line 10: same

crates/rpc-types-engine/Cargo.toml
```diff
@@ -43,7 +43,7 @@ rand = { workspace = true, optional = true }
 
 [features]
 default = ["jwt"]
```
mattsse, line 45: this needs a std feature

mattsse (changes requested):
let's also add this crate to

https://github.com/alloy-rs/op-alloy/blob/bd669da1ea0239411b317d81e30a0477aabb0808/scripts/check_no_std.sh#L4-L5

so we check in ci

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1,7 +1,7 @@
 //! Payload types.
 
-#[cfg(not(feature = "jwt"))]
-use alloc::vec::Vec;
+#[cfg(not(feature = "std"))]
+use alloc::{string::{ToString, String}, vec::Vec};
```
mattsse: let's use extern crate alloc so we dont need to feature gate these

crates/rpc-types-engine/src/payload.rs
```diff
@@ -771,72 +771,75 @@ impl<'de> Deserialize<'de> for ExecutionPayload {
...
     InvalidVersionedHashes,
     /// Encountered decoding error.
-    #[error(transparent)]
-    Decode(#[from] alloy_rlp::Error),
+    #[display("{_0}")]
+    Decode(alloy_rlp::Error),
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for PayloadError {}
```
mattsse, line 844: this needs to impl fn source and return the rlp error

crates/rpc-types-engine/src/payload.rs
```diff
@@ -771,72 +771,75 @@ impl<'de> Deserialize<'de> for ExecutionPayload {
...
     },
     /// Expected blob versioned hashes do not match the given transactions.
-    #[error("expected blob versioned hashes do not match the given transactions")]
+    #[display("expected blob versioned hashes do not match the given transactions")]
     InvalidVersionedHashes,
     /// Encountered decoding error.
-    #[error(transparent)]
-    Decode(#[from] alloy_rlp::Error),
+    #[display("{_0}")]
+    Decode(alloy_rlp::Error),
```
mattsse, line 840: this is now missing the from impl

crates/rpc-types-engine/src/forkchoice.rs
```diff
@@ -34,22 +34,25 @@ pub struct ForkchoiceState {
 ///
 /// These are considered hard RPC errors and are _not_ returned as [PayloadStatus] or
 /// [PayloadStatusEnum::Invalid].
-#[derive(Clone, Copy, Debug, PartialEq, Eq, thiserror::Error)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, derive_more::Display)]
```
mattsse, line 37: once we bump msrv to 1.81 we can also use derive_more error and remove the feature entirely

https://docs.rs/derive_more/latest/derive_more/derive.Error.html

mattsse (changes requested):
nits

refcell:
Replaced with https://github.com/alloy-rs/alloy/pull/1268

PR #1234, opened 2024-09-04
chore: replace once_cell with std::sync

estensen:
## Motivation

Remove external crate

### DIFF
--- a/crates/eips/src/eip4844/env_settings.rs
+++ b/crates/eips/src/eip4844/env_settings.rs
@@ -48,15 +48,14 @@ impl EnvKzgSettings {
                 };
                 #[cfg(feature = "std")]
                 {
-                    use once_cell as _;
                     use std::sync::OnceLock;
                     static DEFAULT: OnceLock<KzgSettings> = OnceLock::new();
                     DEFAULT.get_or_init(load)
                 }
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip4844/env_settings.rs
```diff
@@ -48,15 +48,14 @@ impl EnvKzgSettings {
...
                     use std::sync::OnceLock;
                     static DEFAULT: OnceLock<KzgSettings> = OnceLock::new();
                     DEFAULT.get_or_init(load)
                 }
                 #[cfg(not(feature = "std"))]
                 {
-                    use once_cell::race::OnceBox;
-                    static DEFAULT: OnceBox<KzgSettings> = OnceBox::new();
+                    use std::sync::OnceLock;
+                    static DEFAULT: OnceLock<KzgSettings> = OnceLock::new();
```
mattsse, line 58: unsure what the reason for this was, but these feature branches are now identical and should be removed if we can make this change, but could have some no-std implications which made this necessary in the first place, because don't think we can use this in in no-std?

estensen: Sorry, missed the conditional attribute. Probably not any use then since the crate has to be pulled down for no-std. Closing.

PR #1247, opened 2024-09-06
consensus: add `Receipts` struct

tcoratger:
Related https://github.com/paradigmxyz/reth/issues/7238


crates/consensus/src/receipt/receipts.rs
```diff
@@ -102,6 +103,45 @@ impl<T> From<ReceiptWithBloom<T>> for Receipt<T> {
...
 }
 
+/// A collection of receipts organized as a two-dimensional vector.
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, From, derive_more::Deref, DerefMut, IntoIterator,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Receipts {
+    /// A two-dimensional vector of optional `Receipt` instances.
+    pub receipt_vec: Vec<Vec<Option<Receipt>>>,
```
mattsse, line 113: I don't want to do this, this is massive debt in reth because of bad pruning decisions

crates/consensus/src/receipt/receipts.rs
```diff
@@ -102,6 +103,45 @@ impl<T> From<ReceiptWithBloom<T>> for Receipt<T> {
     }
 }
 
+/// A collection of receipts organized as a two-dimensional vector.
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, From, derive_more::Deref, DerefMut, IntoIterator,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Receipts {
```
mattsse, line 111: I can see this being helpful, but this should be generic over the internal type

mattsse (commented):
the reth receipts type is very bad because the option was due to contract level pruning.

I'd prefer if we could remove the Option entirely, I assume we can do this if we change how we prune during pipeline execution, or use two types one non optional and one optional for pruned receipts

crates/consensus/src/receipt/receipts.rs
tcoratger: Oh yeah absolutely agree, maybe this is a good time to review the reth design here when integrating this primitive, just changed the definition here

crates/consensus/src/receipt/receipts.rs
tcoratger: Indeed better, just fixed

tcoratger:
@mattsse Nice bump here, can we merge or any blocker?

crates/consensus/src/receipt/receipts.rs
```diff
@@ -100,6 +101,45 @@ impl<T> From<ReceiptWithBloom<T>> for Receipt<T> {
...
 }
 
+/// Receipt containing result of transaction execution.
+#[derive(
+    Clone, Debug, PartialEq, Eq, Default, From, derive_more::Deref, DerefMut, IntoIterator,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Receipts<T> {
+    /// A two-dimensional vector of [`Receipt`] instances.
+    pub receipt_vec: Vec<Vec<Receipt<T>>>,
```
mattsse, line 111: actually, let's do:

```suggestion
    pub receipt_vec: Vec<Vec<T>>,
```

this way this is also reusable by multiple things

mattsse (changes requested):
sorry, forgot about this, one nit

tcoratger:
> sorry, forgot about this, one nit

@mattsse done

mattsse (approved):

PR #1248, opened 2024-09-06
feat: no_std network primitives

refcell:
### Description

> [!NOTE]
>
> Stacked ontop of https://github.com/alloy-rs/alloy/pull/1250

Makes `alloy-network-primitives` support `no_std`.

This is a pre-requisite to supporting `no_std` in `alloy-rpc-types-eth`.

### DIFF
--- a/crates/network-primitives/src/block.rs
+++ b/crates/network-primitives/src/block.rs
@@ -1,6 +1,9 @@
 use alloy_primitives::B256;
 use serde::{Deserialize, Serialize};
 
+use alloc::{vec, vec::Vec};
+use core::slice;
+
 use crate::TransactionResponse;
 
 /// Block Transactions depending on the boolean attribute of `eth_getBlockBy*`,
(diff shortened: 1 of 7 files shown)


crates/network-primitives/src/block.rs
```diff
@@ -1,6 +1,16 @@
...
 
+#[cfg(not(feature = "std"))]
+use alloc::{vec, vec::Vec};
+#[cfg(feature = "std")]
+use std::{vec, vec::Vec};
+
+#[cfg(not(feature = "std"))]
+use alloc::slice;
+#[cfg(feature = "std")]
+use std::slice;
```
DaniPopes, line 12: This is available in core, please use core for this without any cfgs

crates/network-primitives/src/block.rs
```diff
@@ -1,6 +1,16 @@
 use alloy_primitives::B256;
 use serde::{Deserialize, Serialize};
 
+#[cfg(not(feature = "std"))]
+use alloc::{vec, vec::Vec};
+#[cfg(feature = "std")]
+use std::{vec, vec::Vec};
```
DaniPopes, line 7: should just use `extern crate alloc;` and no cfgs

DaniPopes (changes requested):

crates/network-primitives/src/block.rs
mattsse: yeah we should do that in more places

DaniPopes: (see alloy-core)

DaniPopes (approved):

PR #1251, opened 2024-09-06
rpc: add getter trait methods to `ReceiptResponse`

### LINKED ISSUE #1245: [Feature] Add getter trait methods to `ReceiptResponse`
### Component

rpc

Add getters for all fields of `alloy_rpc_types_eth::TransactionReceipt` as trait methods of `ReceiptResponse`. This is needed to replace concrete types with network specific receipt type in reth.

https://github.com/alloy-rs/alloy/blob/37b0c294a2baa5e2ead36259f9b5c7d9a000ad7c/crates/rpc-types-eth/src/transaction/receipt.rs#L9-L65

https://github.com/alloy-rs/alloy/blob/37b0c294a2baa5e2ead36259f9b5c7d9a000ad7c/crates/network-primitives/src/traits.rs#L6-L26

### Additional context

_No response_

tcoratger:
Should close #1245


crates/network-primitives/src/traits.rs
```diff
@@ -23,6 +27,36 @@ pub trait ReceiptResponse {
...
+    fn blob_gas_price(&self) -> Option<u128>;
+
+    /// Address of the sender.
+    fn from(&self) -> Address;
+
+    /// Address of the receiver.
+    fn to(&self) -> Option<Address>;
+
+    /// Post-transaction state root.
+    fn state_root(&self) -> Option<B256>;
```
mattsse, line 56: we can remove this, this isn't filled anymore afaik

crates/network-primitives/src/traits.rs
```diff
@@ -23,6 +27,36 @@ pub trait ReceiptResponse {
...
+
+    /// Authorization list.
+    fn authorization_list(&self) -> Option<Vec<SignedAuthorization>>;
```
mattsse, line 59: this should be &[] otherwise this requires a clone

mattsse (changes requested):
nits

all of these should be fine 

we also need to impl this for 

https://github.com/alloy-rs/alloy/blob/122c4b187af3c621bdd10f829d44538300c0dae0/crates/network/src/any/mod.rs#L72-L72

I believe

crates/network-primitives/src/traits.rs
tcoratger: fixed

crates/network-primitives/src/traits.rs
tcoratger: fixed

mattsse (approved):
all of these are good

PR #1252, opened 2024-09-06
feat: no_std eth rpc types

refcell:
### Description

Adds support for `no_std` in `alloy-rpc-types-eth`.


DaniPopes:
Can you update the workflow and no_std script from `alloy-rs/core`?

refcell:
> Can you update the workflow and no_std script from `alloy-rs/core`? Would like them to be the same

Yes definitely, mind if I do this in another PR to unblock this?

crates/rpc-types-eth/src/block.rs
```diff
@@ -247,13 +249,14 @@ impl HeaderResponse for Header {
 }
 
 /// Error that can occur when converting other types to blocks
-#[derive(Clone, Copy, Debug, thiserror::Error)]
+#[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "std", derive(thiserror::Error))]
```
mattsse, line 253: the problem with this is that this now removes the display impl for no-std

perhaps we should use derive_more display and manually impl error for all errors @DaniPopes ?

crates/rpc-types-eth/src/state.rs
```diff
@@ -1,9 +1,10 @@
 //! bindings for state overrides in eth_call
 
 use crate::BlockOverrides;
+use alloc::boxed::Box;
 use alloy_primitives::{Address, Bytes, B256, U256};
+use hashbrown::HashMap;
```
mattsse, line 6: we could either use std hashmap or hashbrown because this change will break in a few places

mattsse (changes requested):

mattsse (approved):
pending @DaniPopes

.github/workflows/no_std.yml
DaniPopes, line 1: Can you add the job from alloy-rs/core in ci.yml?

crates/rpc-types-eth/src/state.rs
DaniPopes: this is a breaking change

DaniPopes (approved):

mattsse (approved):

PR #1253, opened 2024-09-07
chore: add aliases for Num Hash

mattsse:
this type is useful when dealing with any combinations of number+hash and introduces aliases that remove Block form the name so that this can be used when the number or the hash don't originate from a block

### DIFF
--- a/crates/eips/src/eip1898.rs
+++ b/crates/eips/src/eip1898.rs
@@ -612,6 +612,9 @@ pub struct BlockNumHash {
 /// Block number and hash of the forked block.
 pub type ForkBlock = BlockNumHash;
 
+/// Either a hash _or_ a number
+pub type NumHash = BlockNumHash;
+
 impl BlockNumHash {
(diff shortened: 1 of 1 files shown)


yash-atreya (approved):

DaniPopes (approved):

crates/eips/src/eip1898.rs
```diff
@@ -612,6 +612,9 @@ pub struct BlockNumHash {
 /// Block number and hash of the forked block.
 pub type ForkBlock = BlockNumHash;
 
+/// Either a hash _or_ a number
+pub type NumHash = BlockNumHash;
```
DaniPopes, line 616: maybe should do the opposite, rename the structs and have the aliases be for Block*

mattsse: yeah, but didn't want to rename, although this should be a non-breaking operation

DaniPopes: I would prefer we did

PR #1257, opened 2024-09-09
chore(provider): elided lifetime

jsvisa:
## Motivation

`cargo check release` with warnings as below:

```
warning: elided lifetime has a name
   --> crates/provider/src/provider/trait.rs:659:52
    |
657 |         &self,
    |         - lifetime `'life0` declared here
658 |         tx: N::TransactionRequest,
659 |     ) -> TransportResult<PendingTransactionBuilder<'_, T, N>> {
    |                                                    ^^ this elided lifetime gets resolved as `'life0`
    |
    = note: `#[warn(elided_named_lifetimes)]` on by default


...
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -4,9 +4,6 @@ use alloy_serde::WithOtherFields;
 
 use crate::BlockTransactions;
 
-#[cfg(not(feature = "std"))]
-use alloc::vec::Vec;
-
 /// Receipt JSON-RPC response.
 pub trait ReceiptResponse {
     /// Address of the created contract, or `None` if the transaction was not a deployment.
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -627,7 +627,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     async fn send_raw_transaction(
         &self,
         encoded_tx: &[u8],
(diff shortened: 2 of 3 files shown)


crates/network-primitives/src/traits.rs
```diff
@@ -4,9 +4,6 @@ use alloy_serde::WithOtherFields;
 
 use crate::BlockTransactions;
 
-#[cfg(not(feature = "std"))]
-use alloc::vec::Vec;
-
```
emhane, line 9: sure this should be removed?

onbjerg: it's not used in the file

onbjerg (approved):

crates/provider/src/provider/trait.rs
```diff
@@ -627,7 +627,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
     async fn send_raw_transaction(
         &self,
         encoded_tx: &[u8],
-    ) -> TransportResult<PendingTransactionBuilder<'_, T, N>> {
+    ) -> TransportResult<PendingTransactionBuilder<'life0, T, N>> {
```
onbjerg, line 630: can we just call it 'a

onbjerg (changes requested):

mattsse (commented):
this is an issue with async_trait that is resolved with a new update

PR #1258, opened 2024-09-09
feat(rpc-types-trace): always serialize result if no error

jsvisa:
## Motivation

I'm checking with reth's trace results with erigon's, seems erigon will always return `result:null` if no error occurred:

image

And found the OG OpenEthereum's serialized logic as below:

https://github.com/openethereum/openethereum/blob/main/crates/rpc/src/v1/types/trace.rs#L532

```rust
  match self.result {
        Res::Call(ref call) => struc.serialize_field("result", call)?,
        Res::Create(ref create) => struc.serialize_field("result", create)?,
        Res::FailedCall(ref error) => struc.serialize_field("error", &error.to_string())?,
        Res::FailedCreate(ref error) => struc.serialize_field("error", &error.to_string())?,
        Res::None => struc.serialize_field("result", &None as &Option<u8>)?,
    }
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/parity.rs
+++ b/crates/rpc-types-trace/src/parity.rs
@@ -497,12 +497,11 @@ impl Serialize for LocalizedTransactionTrace {
 
         if let Some(error) = error {
             s.serialize_field("error", error)?;
-        } else {
-            match result {
-                Some(TraceOutput::Call(call)) => s.serialize_field("result", call)?,
-                Some(TraceOutput::Create(create)) => s.serialize_field("result", create)?,
-                None => s.serialize_field("result", &None as &Option<u8>)?,
-            }
+        }
+        match result {
+            Some(TraceOutput::Call(call)) => s.serialize_field("result", call)?,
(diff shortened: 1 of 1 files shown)


mattsse (commented):
is this causing issues?

do you know when this can be null?

jsvisa:
> is this causing issues?

Not really, I'm comparing the trace results between reth and erigon, and then found this issue(see the diff image)

> do you know when this can be null?

suicide's results were null

after thinking twice, I think we can keep the same format as erigon's, `result` always present, `error` omitted if none https://github.com/erigontech/erigon/blob/13b4b7768485736e54ff5ca3270ebeec5c023ba8/turbo/jsonrpc/trace_types.go#L62-L63

```go
type ParityTrace struct {
...
	Error               string       `json:"error,omitempty"`
	Result              interface{}  `json:"result"`
...
}
```

@mattsse WDYT

mattsse:
I see, then this makes sense

mattsse (approved):

PR #1259, opened 2024-09-09
refactor: separate transaction builders for tx types

klkvr:
## Motivation

Not all netwoks support 7702 and 4844 transactions, however, current network abstraction requires `TransactionRequest` to contain setters for both of those tx types. This results in custom network implementations either having no-op setters/getters on builders, or introducing additional fallible steps during transaction conversion pipeline to reject attempts to construct unsupported transactions.

## Solution

Extracts part of `TransactionBuilder` methods into `TransactionBuilder4844` and `TransactionBuilder7702`. Extracts blob fee filling from `GasFiller` into `BlobGasFiller` which enforces `N::TxRequest: TransactionBuilder4844` bound.

Given that filling logic now depends on transaction request properties, I've extracted recomended fillers into `RecommendedFillers` trait which can be implemented on custom network implementations, adding support for filling custom transaction types. Right now it is implemented on `Ethereum` and `AnyNetwork`.

This makes it a bit harder to use transaction builder due to need to import 2 additional traits, so I'm wondering if we should consider adding some of the methods directly to it?

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -155,7 +155,7 @@ impl<N: Network> TxFiller<N> for GasFiller {
         P: Provider<T, N>,
         T: Transport + Clone,
     {
-        if tx.gas_price().is_some() {
+        if tx.gas_price().is_some() || tx.access_list().is_some() {
             self.prepare_legacy(provider, tx).await
         } else {
             match self.prepare_1559(provider, tx).await {
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -155,7 +155,7 @@ impl<N: Network> TxFiller<N> for GasFiller {
         P: Provider<T, N>,
         T: Transport + Clone,
     {
-        if tx.gas_price().is_some() || tx.access_list().is_some() {
+        if tx.gas_price().is_some() {
             self.prepare_legacy(provider, tx).await
         } else {
             match self.prepare_1559(provider, tx).await {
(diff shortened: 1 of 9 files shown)


emhane (approved):
type safety ftw

crates/provider/src/fillers/gas.rs
```diff
@@ -210,8 +157,6 @@ impl<N: Network> TxFiller<N> for GasFiller {
     {
         if tx.gas_price().is_some() || tx.access_list().is_some() {
```
klkvr, line 158: Our `GasFiller` automatically treats transaction as legacy if it has an `accessList` set which is not really correct. We also have a test enforcing this, wondering if this should be changed?

mattsse: I believe this is a bug

klkvr: updated in 5697944

crates/provider/src/fillers/mod.rs
```diff
@@ -308,3 +308,42 @@ where
         self.inner.send_transaction_internal(tx).await
     }
 }
+
+/// A trait which may be used to configure default fillers for [Network] implementations.
+pub trait RecommendedFillers {
+    /// Recommended fillers for this network.
+    type RecomendedFillters: TxFiller;
```
mattsse, line 315: I think this makes a ton of sense, so we can network specific defaults

mattsse (approved):
> This makes it a bit harder to use transaction builder due to need to import 2 additional traits, so I'm wondering if we should consider adding some of the methods directly to it?

this should be okay

klkvr:
should we merge this now or hold off until next breaking release?

PR #1261, opened 2024-09-09
fix(rpc-types-trace): use rpc-types Log in OtsReceipt

yash-atreya:
## Motivation

Ref https://github.com/foundry-rs/foundry/issues/8830#issuecomment-2338108826

`OtsReceipt` uses `alloy_primitives::Log`. 

## Solution

Use `rpc_types_eth::Log` instead of `alloy_primitives::Log` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
this is indeed correct

https://otterscan.github.io/execution-apis/api-documentation/

image

PR #1262, opened 2024-09-09
chore: require destination for 7702

mattsse:
> The fields chain_id, nonce, max_priority_fee_per_gas, max_fee_per_gas, gas_limit, destination, value, data, and access_list of the outer transaction follow the same semantics as [EIP-4844](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md). Note, this means a null destination is not valid.

https://github.com/alloy-rs/eips/

to field is required for 7702


crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -314,6 +314,11 @@ impl TransactionRequest {
     fn build_7702(self) -> Result<TxEip7702, &'static str> {
         let checked_to = self.to.ok_or("Missing 'to' field for Eip7702 transaction.")?;
 
+        let to_address = match checked_to {
```
DaniPopes, line 317: `.to().ok_or()`

crates/consensus/src/transaction/eip7702.rs
```diff
@@ -263,7 +261,7 @@ impl TxEip7702 {
         mem::size_of::<u64>() + // gas_limit
         mem::size_of::<u128>() + // max_fee_per_gas
         mem::size_of::<u128>() + // max_priority_fee_per_gas
-        self.to.size() + // to
+            mem::size_of::<Address>() + // to
```
DaniPopes, line 264:
```suggestion
        mem::size_of::<Address>() + // to
```

crates/consensus/src/header.rs
```diff
@@ -641,7 +641,6 @@ impl<'a> arbitrary::Arbitrary<'a> for Header {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
```
DaniPopes, line 644: move cfg serde up?

DaniPopes (changes requested):

DaniPopes (approved):

PR #1275, opened 2024-09-11
Add trait methods `cumulative_gas_used` and `state_root` to `ReceiptResponse`

emhane:
## Motivation

Use generic `ReceiptResponse` type in place of `alloy_rpc_types_eth::TransactionReceipt` in reth RPC

## Solution

Adds trait methods `ReceiptResponse::cumulative_gas_used` and `ReceiptResponse::state_root`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -51,6 +51,14 @@ pub trait ReceiptResponse {
 
     /// EIP-7702 Authorization list.
     fn authorization_list(&self) -> Option<&[SignedAuthorization]>;
+
+    /// Returns the cumulative gas used at this receipt.
+    fn cumulative_gas_used(&self) -> u128;
+
+    /// The post-transaction state root (pre Byzantium)
+    ///
+    /// EIP98 makes this field optional.
+    fn state_root(&self) -> Option<&B256>;
 }
 
 /// Transaction JSON-RPC response.
@@ -220,6 +228,14 @@ impl<T: ReceiptResponse> ReceiptResponse for WithOtherFields<T> {
 
     fn authorization_list(&self) -> Option<&[SignedAuthorization]> {
         self.inner.authorization_list()
+    }
+
+    fn cumulative_gas_used(&self) -> u128 {
+        self.inner.cumulative_gas_used()
+    }
+
+    fn state_root(&self) -> Option<&B256> {
+        self.inner.state_root()
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/traits.rs
```diff
@@ -51,6 +51,14 @@ pub trait ReceiptResponse {
...
     /// EIP-7702 Authorization list.
     fn authorization_list(&self) -> Option<&[SignedAuthorization]>;
+
+    /// Returns the cumulative gas used at this receipt.
+    fn cumulative_gas_used(&self) -> u128;
+
+    /// The post-transaction state root (pre Byzantium)
+    ///
+    /// EIP98 makes this field optional.
+    fn state_root(&self) -> Option<&B256>;
```
mattsse, line 61: we can ingore this because no longer useful

crates/network-primitives/src/traits.rs
```diff
@@ -51,6 +51,14 @@ pub trait ReceiptResponse {
 
     /// EIP-7702 Authorization list.
     fn authorization_list(&self) -> Option<&[SignedAuthorization]>;
+
+    /// Returns the cumulative gas used at this receipt.
+    fn cumulative_gas_used(&self) -> u128;
```
mattsse, line 56: this makes sense

mattsse (changes requested):

crates/network-primitives/src/traits.rs
emhane: when converting from `ReceiptResponse` to `TransactionReceipt<T>`, would it not be a bug to assume the field to be `None` for historic blocks?

mattsse: > when converting from ReceiptResponse to TransactionReceipt<T>
this isn't the purpose of this trait, this trait just serves the purpose of providing expected values

but I guess it doesn't really hurt to add this function, even if this is useless.

mattsse, line 61:
```suggestion
    fn state_root(&self) -> Option<B256>;
```

mattsse (changes requested):
nit

crates/network-primitives/src/traits.rs
mattsse: this is at odds with the other trait functions. I'd like to unify this and return a copy

PR #1276, opened 2024-09-11
feat(alloy-rpc-types-eth): Optional serde

refcell:
### Description

Downstream crates would like to use types from `alloy-rpc-types-eth` without the need for `serde` support. This PR places serialization for core eth types behind the `serde` feature flag that is enabled-by-default.

### DIFF
--- a/crates/contract/Cargo.toml
+++ b/crates/contract/Cargo.toml
@@ -22,7 +22,7 @@ workspace = true
 alloy-network.workspace = true
 alloy-network-primitives.workspace = true
 alloy-provider.workspace = true
-alloy-rpc-types-eth.workspace = true
+alloy-rpc-types-eth = { workspace = true, features = ["serde"] }
 alloy-transport.workspace = true
 
 alloy-dyn-abi = { workspace = true, features = ["std"] }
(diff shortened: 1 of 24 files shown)


mattsse (approved):
this seems fine, since all tests basically use serde we could derive serde for tests as well, but this would now require changing all attributes, so I'm fine with this.

would appreciate one additional sanity review @klkvr @DaniPopes

klkvr (approved):
lgtm! wondering if need for this means we are missing something in consensus types?

had a feeling that rpc-types were supposed to always implement serde as they should only be used to communicate with RPC, so curious to learn more about usecases needing this change

mattsse:
actually, @refcell is this a size related issue?

because this will still pull in serde via alloy-serde I believe

refcell:
> actually, @refcell is this a size related issue?
> 
> because this will still pull in serde via alloy-serde I believe

Good point, it was to decrease size and generally make available lighter types. Wonder if I can place alloy-serde behind the feature flag as well

mattsse:
should be possible to make it optional

mattsse:
actually, let me take a look at the engine type crate

mattsse:
ref https://github.com/alloy-rs/alloy/pull/1280

PR #1076, opened 2024-07-19
feat: improve Reth bindings and fix `debug_traceCallMany` and `trace_callMany` methods

zerosnacks:
## Motivation

Closes: https://github.com/alloy-rs/alloy/issues/1074 and improves on Reth bindings added in https://github.com/alloy-rs/alloy/pull/1092

### Fixes

- [x] Fix `debug_traceCallMany`
- [x] Fix `trace_callMany`

### Other

- `keys()` / `address()` method for Reth binding - used in tracing but useful in general so you can perform transactions
- adds `random_instance()` to reduce flakiness of concurrent spawned Reth nodes
- Adds `run_with_tempdir_*` methods for proper handling and clean up when using tempdirs + updating tests
- Unification Anvil errors with other nodes

Does not include the addition of `eth_callMany`, to be added in https://github.com/alloy-rs/alloy/issues/1274 and requires more consideration. Ties into #328 / #1119.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

PR #1279, opened 2024-09-12
feat: improve node bindings

zerosnacks:
## Motivation

Improves on existing Reth bindings:

Preferably blocked until https://github.com/alloy-rs/alloy/pull/1278 is merged so that the tests can be updated to use the new utilities.

## Solution

- Now picks random instance for Reth by default to reduce flakiness in tests (busy ports)
- Adds `run_with_tempdir_*` methods for proper handling and clean up when using tempdirs + updating tests
- Unifies Anvil errors with other node errors and generalizes

I've decided not to include the hardcoded `keys` and `addresses` as proposed in https://github.com/alloy-rs/alloy/pull/1076 as it is easy to derive if necessary.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/reth.rs
+++ b/crates/node-bindings/src/nodes/reth.rs
@@ -2,6 +2,7 @@
 
 use crate::{utils::extract_endpoint, NodeError, NODE_STARTUP_TIMEOUT};
 use alloy_genesis::Genesis;
+use rand::Rng;
 use std::{
     fs::create_dir,
     io::{BufRead, BufReader},
@@ -23,6 +24,7 @@ const RETH: &str = "reth";
 #[derive(Debug)]
 pub struct RethInstance {
     pid: Child,
+    instance: u16,
     http_port: u16,
     ws_port: u16,
     auth_port: Option<u16>,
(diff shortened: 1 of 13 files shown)


crates/node-bindings/src/nodes/reth.rs
```diff
@@ -142,12 +149,14 @@ pub struct Reth {
...
-    pub const fn new() -> Self {
+    /// The instance number is set to a random number between 1 and 200 by default to reduce the
+    /// odds of port conflicts. This can be changed with [`Reth::instance`]. Set to 0 to use the
+    /// default ports. 200 is the maximum number of instances that can be run set by Reth.
+    pub fn new() -> Self {
         Self {
             dev: false,
             block_time: None,
-            instance: 0,
+            instance: rand::thread_rng().gen_range(1..200),
```
mattsse, line 159: hmm, there's currently no way to set the port, right?
the `Geth` type has his and allows port 0 so that the os can choose a random port and it should get parsed from cli output.
could we do the same here?

mattsse (approved):
one suggestion
but this can be solved as followup

crates/node-bindings/src/nodes/reth.rs
zerosnacks: you can set the instance to roughly control the port mapping but it indeed would be better to have that fully configurable as well

added the ability to customize the ports

zerosnacks: you can now define `.http_port(0)` (same for other ports) to use a random port picked by `reth`

zerosnacks:
Removing `blocked` label, will update https://github.com/alloy-rs/alloy/pull/1278 tests

mattsse (approved):

PR #1283, opened 2024-09-12
feat(engine): Optional Serde

refcell:
### Description

Replaces #1282 

Since `alloy-rpc-types-engine` doesn't depend on `alloy-rpc-types-eth`, we can split this out from `serde` feature support in `alloy-rpc-types-eth`.

This PR makes `serde` optional on the `alloy-rpc-types-engine` crate. Some JWT functionality now sits behind the `serde` feature flag since it uses `serde` internally for encoding/decoding. `serde` is also enabled-by-default.


mattsse (approved):
lgtm

PR #1284, opened 2024-09-12
fix(types-eth): Optional Alloy Serde

refcell:
### Description

This makes the `alloy-serde` dep optional, but hides types behind the `serde` feature flag. Open to suggestions for alternative ways to make these types accessible in a no-serde environment.


DaniPopes (approved):

mattsse (approved):
this seems fine

PR #1293, opened 2024-09-15
chore(rpc): rename witness fields

rkrasiuk:
## Description

Rename witness fields and amend docs to clarify what they represent.

### DIFF
--- a/crates/rpc-types-debug/src/debug.rs
+++ b/crates/rpc-types-debug/src/debug.rs
@@ -9,9 +9,11 @@ use std::collections::HashMap;
 pub struct ExecutionWitness {
     /// Map of all hashed trie nodes to their preimages that were required during the execution of
     /// the block, including during state root recomputation.
-    pub witness: HashMap<B256, Bytes>,
-    /// Map of all hashed account addresses and storage slots to their preimages (unhashed account
-    /// addresses and storage slots, respectively) that were required during the execution of the
-    /// block. during the execution of the block.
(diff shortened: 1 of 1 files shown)


crates/rpc-types-debug/src/debug.rs
```diff
@@ -9,9 +9,11 @@ use std::collections::HashMap;
...
     /// Map of all hashed trie nodes to their preimages that were required during the execution of
     /// the block, including during state root recomputation.
-    pub witness: HashMap<B256, Bytes>,
-    /// Map of all hashed account addresses and storage slots to their preimages (unhashed account
-    /// addresses and storage slots, respectively) that were required during the execution of the
-    /// block. during the execution of the block.
-    pub state_preimages: Option<HashMap<B256, Bytes>>,
+    /// keccak(rlp(node)) => rlp(node)
+    pub state: HashMap<B256, Bytes>,
+    /// Map of all hashed account and storage keys (addresses and slots) to their preimages
```
0x00101010, line 14: Can we add code here as well?

0x00101010: Similar to geth's implementation: https://github.com/ethereum/go-ethereum/pull/30069/files#diff-f0800971a600b0acfe84ea41db6150cd6f45b5376a6600d85613d4414fcd5c66R74

rkrasiuk: we can add those separately once we resolve some questions regarding compatibility

mattsse (approved):

PR #1299, opened 2024-09-17
feat(`node-bindings`): support appending extra args

yash-atreya:
## Motivation

Currently, users cannot append any extra args to pass to `reth` or `geth`. We can only launch their respective instances with settings we have available in the builder api. 

## Solution

Adds `args: Vec<OsString>` to both `Reth` and `Geth` that get appended to the `cmd` before spawning the instance.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/geth.rs
+++ b/crates/node-bindings/src/nodes/geth.rs
@@ -204,6 +204,7 @@ pub struct Geth {
     genesis: Option<Genesis>,
     mode: NodeMode,
     clique_private_key: Option<SigningKey>,
+    args: Vec<String>,
 }
 
 impl Geth {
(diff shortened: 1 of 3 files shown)


crates/node-bindings/src/nodes/geth.rs
```diff
@@ -367,6 +368,28 @@ impl Geth {
...
+        self
+    }
+
+    /// Adds multiple arguments to pass to `geth`.
+    ///
+    /// Pass any args that is not supported by the builder.
+    pub fn args<I, S>(mut self, args: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: Into<String>,
```
DaniPopes, line 385: should use OsString, see signatures on std command

DaniPopes (approved):

mattsse (approved):

PR #1302, opened 2024-09-17
consensus: add `BlockHeader` getter trait

### LINKED ISSUE #1301: [Feature] Add getters of `alloy_consensus::Header` to a new trait `BlockHeader`
### Component

consensus, eips, genesis

Add new trait, called something like `BlockHeader`. Add a getter for all fields of `alloy_consensus::Header` as trait methods to this new trait. Set the new trait as trait bound for `alloy_network::Network`.

### Additional context

_No response_

tcoratger:
Should close #1301

### DIFF
--- a/crates/consensus/src/header.rs
+++ b/crates/consensus/src/header.rs
@@ -638,6 +638,158 @@ impl<'a> arbitrary::Arbitrary<'a> for Header {
     }
 }
 
+/// Trait for extracting specific Ethereum block data from a header
+pub trait BlockHeader {
+    /// Retrieves the parent hash of the block
+    fn parent_hash(&self) -> B256;
+
+    /// Retrieves the ommers hash of the block
+    fn ommers_hash(&self) -> B256;
+
+    /// Retrieves the beneficiary (miner) of the block
+    fn beneficiary(&self) -> Address;
+
+    /// Retrieves the state root hash of the block
+    fn state_root(&self) -> B256;
+
+    /// Retrieves the transactions root hash of the block
+    fn transactions_root(&self) -> B256;
+
+    /// Retrieves the receipts root hash of the block
+    fn receipts_root(&self) -> B256;
+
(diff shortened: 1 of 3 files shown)


crates/consensus/src/header.rs
```diff
@@ -638,6 +638,158 @@ impl<'a> arbitrary::Arbitrary<'a> for Header {
     }
 }
 
+/// Trait for extracting specific Ethereum block data from a header
+pub trait BlockHeader {
```
mattsse, line 642: from reth pov there's an argument for limiting these functions to the absolute necessary fields, or introduce sub traits, for example for evm you only need certain fields.
but since this doesn't clash with rpc, I think all of those are fine, and we could still do sub traits

mattsse (approved):
ah now I get it @emhane 

pending @klkvr and @emhane

crates/consensus/src/header.rs
tcoratger: Let me know if you think we should split in subtraits now or if we should do that in a follow up if needed

mattsse: I think we can do in a followup if needed, realistically we only need this trait for one type really, the chain's header type and for op we can even reuse the L1 type

emhane (approved):

tcoratger:
@emhane @mattsse Now that https://github.com/paradigmxyz/reth/pull/10691 is merged, in order to completely remove `BlockHeader` trait from reth and have it in alloy, do we want to merge this? Or reduce the implementation strictly to what has be done in reth (less getter)?

emhane:
also for rpc header we want `HeaderResponse: BlockHeader`, as for rpc transaction
https://github.com/alloy-rs/alloy/pull/1328#issuecomment-2368075961

tcoratger:
@emhane FYI just opened a follow up issue here on reth https://github.com/paradigmxyz/reth/issues/11127

PR #1305, opened 2024-09-18
fix: enforce correct parity encoding for typed transactions

klkvr:
## Motivation

Enforces that we only accept 0 or 1 as parity for typed transactions, and only 27, 28 or 35.. for legacy transactions

ref https://github.com/paradigmxyz/reth/pull/10758

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,7 +1,7 @@
 use crate::{EncodableSignature, SignableTransaction, Signed, Transaction, TxType};
 use alloc::vec::Vec;
 use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization};
-use alloy_primitives::{keccak256, Bytes, ChainId, Signature, TxKind, B256, U256};
+use alloy_primitives::{keccak256, Bytes, ChainId, Parity, Signature, TxKind, B256, U256};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use core::mem;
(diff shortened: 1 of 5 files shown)


klkvr:
Wondering if we should consider introducing parity generic on signature? this might end up hacky, but likely can be done in a nice way with some helper traits

klkvr:
eg we could do trait SignableTransaction { type Signature }

```rust
struct Signed<Unsigned: SignableTransaction> {
    tx: Unsigned,
    signature: Unsigned::Signature,
}

impl SignableTransaction for TxLegacy {
    type Signature = primitives::Signature<LegacyParity>;
    ...
}

impl SignableTransaction for TxLegacy {
    // default generic would be a simple bool
    type Signature = primitives::Signature;
    ...
}
```

mattsse (approved):
we def need to enforce this on decoding, but I think we have a problem an encoding where it would be possible to use an incorrect signature, right?

so making the sig tx specific makes sense, if we can make the ergonomics right, because dealing with these things manually is a pita

klkvr:
> we def need to enforce this on decoding, but I think we have a problem an encoding where it would be possible to use an incorrect signature, right?

yeah, we enforce correct parity in `into_signed` right now:
https://github.com/alloy-rs/alloy/blob/e57eb86f5b2332833d927719f3b5abb08d3e3397/crates/consensus/src/transaction/eip1559.rs#L339-L342

PR #1314, opened 2024-09-19
feat(transport-http): JWT auth layer

### LINKED ISSUE #1161: [Feature] Add JWT layer for transport

yash-atreya:
## Motivation

Post #1227,  we can close #1161.

## Solution

Adds a JWT layer based on hyper transport client that refreshes the bearer tokens in requests.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/transport-http/Cargo.toml
```diff
@@ -51,6 +55,10 @@ hyper = [
     "dep:serde_json",
     "dep:tower",
     "dep:tracing",
+    "dep:alloy-rpc-types-engine",
+    "alloy-rpc-types-engine/jwt",
+    "alloy-rpc-types-engine/serde",
+    "dep:jsonwebtoken",
```
mattsse, line 61: this will now enable all of these by default, do we want this?

I think we can feature gate this with engine-jwt or similar

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,125 @@
...
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<B>) -> Self::Future {
+        let headers = req.headers();
+
+        match get_bearer_token(headers).and_then(|token| self.secret.validate(&token).ok()) {
```
mattsse, line 85: this validation is slightly problematic, because this can be valid on client side but invalid on server if the client time is just below the iat, like 1ms

I think could track the most recent claims separately, and use some grace period, like iat - 5s or similar.

mattsse (changes requested):

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Create a new [`AuthService`] with the given inner service.
+    pub const fn new(inner: S, secret: JwtSecret, latency_buffer: u64) -> Self {
+        Self { inner, secret, latency_buffer }
+    }
+
+    /// Validate the token in the request headers.
+    ///
+    /// Returns `true` if the token is valid and `iat` is beyond the grace buffer.
+    pub fn validate(&self, headers: &HeaderMap) -> bool {
+        get_bearer_token(headers).map_or(false, |token| {
```
mattsse, line 71: lets use if let else return false synstax here to improve readability

or a helper function that returns Option so we can make use of `?` operator

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,158 @@
...
+                let validation = Validation::new(jsonwebtoken::Algorithm::HS256);
+                let decoding_key = DecodingKey::from_secret(self.secret.as_bytes());
+                decode::<Claims>(token.as_str(), &decoding_key, &validation).ok().and_then(|data| {
+                    let curr_secs = get_current_timestamp();
+                    if data.claims.iat.abs_diff(curr_secs) * 1000 <= self.latency_buffer {
+                        None
+                    } else {
+                        Some(())
+                    }
+                })
```
mattsse, line 82: all of this is a bit costly and kinda redundant because all we do here is validating our own claims, if we track the most recent claim instead we can make this easier

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,158 @@
...
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: Request<B>) -> Self::Future {
+        let headers = req.headers();
+
+        match self.validate(headers) {
+            true => Box::pin(self.inner.call(req)),
+            false => {
```
mattsse, line 120: I prefer no matches on bool

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+fn get_bearer_token(headers: &HeaderMap) -> Option<String> {
+    let header = headers.get(AUTHORIZATION)?;
+    let auth: &str = header.to_str().ok()?;
+    let prefix = "Bearer ";
+    let index = auth.find(prefix)?;
+    let token: &str = &auth[index + prefix.len()..];
+    Some(token.into())
+}
```
mattsse, line 148: this is a bit pointless if we track the most recent claims

mattsse (changes requested):

crates/transport-http/src/layers/auth.rs
```diff
@@ -48,41 +46,34 @@ impl<S> Layer<S> for AuthLayer {
...
-                    if data.claims.iat.abs_diff(curr_secs) * 1000 <= self.latency_buffer {
-                        None
-                    } else {
-                        Some(())
-                    }
-                })
-            });
-            is_valid.is_some()
-        })
+    pub fn validate(&self) -> bool {
```
mattsse, line 68:
```suggestion
   fn validate(&self) -> bool {
```

crates/transport-http/src/layers/auth.rs
```diff
@@ -48,41 +46,34 @@ impl<S> Layer<S> for AuthLayer {
...
 impl<S> AuthService<S> {
     /// Create a new [`AuthService`] with the given inner service.
     pub const fn new(inner: S, secret: JwtSecret, latency_buffer: u64) -> Self {
-        Self { inner, secret, latency_buffer }
+        Self { inner, secret, latency_buffer, most_recent_claim: None }
     }
 
     /// Validate the token in the request headers.
     ///
     /// Returns `true` if the token is valid and `iat` is beyond the grace buffer.
```
mattsse, line 67:
```suggestion
    /// Returns `true` if the token is still valid and `iat` is beyond the grace buffer.
```

crates/transport-http/src/layers/auth.rs
```diff
@@ -113,40 +104,32 @@ where
...
-    let auth: &str = header.to_str().ok()?;
-    let prefix = "Bearer ";
-    let index = auth.find(prefix)?;
-    let token: &str = &auth[index + prefix.len()..];
-    Some(token.into())
-}
-
 fn create_token_from_secret(secret: &JwtSecret) -> Result<String, jsonwebtoken::errors::Error> {
     let token = secret.encode(&Claims {
         iat: (SystemTime::now().duration_since(UNIX_EPOCH).unwrap() + Duration::from_secs(60))
```
mattsse, line 135: let's convert this to a function of auth that then also updates the most recent claims

crates/transport-http/src/layers/auth.rs
```diff
@@ -113,40 +104,32 @@ where
...
-                            .insert(AUTHORIZATION, HeaderValue::from_str(&token).unwrap());
-                        Box::pin(self.inner.call(req))
-                    }
-                    Err(e) => {
-                        let e = TransportErrorKind::custom(e);
-
-                        Box::pin(async move { Err(e) })
-                    }
+        if self.validate() {
+            Box::pin(self.inner.call(req))
```
mattsse, line 108: this does not set the auth header

crates/transport-http/src/layers/auth.rs
```diff
@@ -113,40 +104,32 @@ where
...
+        } else {
+            // Generate a fresh token from the secret and insert it into the request.
+            match create_token_from_secret(&self.secret) {
+                Ok(token) => {
+                    // Decode the token to get the claims.
+                    let validation = Validation::new(jsonwebtoken::Algorithm::HS256);
+                    let decoding_key = DecodingKey::from_secret(self.secret.as_bytes());
+                    let token_data = decode::<Claims>(&token, &decoding_key, &validation);
+                    self.most_recent_claim = token_data.ok().map(|data| data.claims);
```
mattsse, line 117: this is redundant, we already created a new claims, there's no need to parse this again.

mattsse (changes requested):

crates/transport-http/src/layers/auth.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Returns `true` if the token is still valid and `iat` is beyond the grace buffer.
+    fn validate(&self) -> bool {
+        if let Some(claim) = self.most_recent_claim.as_ref() {
+            let curr_secs = get_current_timestamp();
+            if claim.iat.abs_diff(curr_secs) * 1000 <= self.latency_buffer {
+                return false;
+            }
+        }
+
+        true
```
mattsse, line 73: I believe this is incorrect and should return false, otherwise self.most_recent_claim.as_ref().unwrap() panics

mattsse (changes requested):
nit, can we add a simple provider test, just to make sure this can also be used as regular transport layer

mattsse (approved):
last pedantic nit

PR #1317, opened 2024-09-20
rpc: add `ExtendedTransaction` type

tcoratger:
In reth, we have quite a few places in the code where we are using `WithOtherFields<Transaction>` and this is quite cumbersome to manipulate and a little bit long to write/read.

https://github.com/paradigmxyz/reth/blob/05f862af41e74319ba00a1254d3a74192d304cce/crates/rpc/rpc/src/txpool.rs#L34

As a consequence an `ExtendedTransaction` could be useful for easier manipulations.


emhane:
this type is on it's way out anyway https://github.com/paradigmxyz/reth/pull/9774

Issue #1290, opened 2024-09-14
[Feature] Introduce Block type in alloy-consensus


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

currently we don't have a block type in the consensus crate

ref 

https://github.com/paradigmxyz/reth/blob/97cc9dd8ecaaf7ebfa0b401ed080da28dc791376/crates/primitives/src/block.rs#L28-L48

this type should be structured slightly differently, so that there's a body container struct that wraps transactions, requests, withdrawals

which is this type

https://github.com/paradigmxyz/reth/blob/97cc9dd8ecaaf7ebfa0b401ed080da28dc791376/crates/primitives/src/block.rs#L557-L566

PR #1319, opened 2024-09-20
feat(consensus): Generic Block Type

refcell:
### Description

Introduces a `Block` type for the `alloy-consensus` crate. See #1290

### DIFF
--- a/crates/consensus/src/block.rs
+++ b/crates/consensus/src/block.rs
@@ -1,0 +1,23 @@
+//! Genesic Block Type
+
+use crate::Header;
+use alloy_eips::eip4895::Withdrawal;
+use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
+
+/// Ethereum full block.
+///
+/// Withdrawals can be optionally included at the end of the RLP encoded message.
+///
+/// Taken from [reth-primitives](https://github.com/paradigmxyz/reth)
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct Block<T: Encodable + Decodable> {
+    /// Block header.
+    pub header: Header,
+    /// Transactions in this block.
+    pub body: Vec<T>,
+    /// Ommers/uncles header.
+    pub ommers: Vec<Header>,
+    /// Block withdrawals.
+    pub withdrawals: Option<Vec<Withdrawal>>,
+}
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -12,6 +12,9 @@ extern crate alloc;
 mod account;
 pub use account::Account;
 
+mod block;
+pub use block::Block;
+
 pub mod constants;
 
 mod encodable_signature;


crates/consensus/src/block.rs
```diff
@@ -0,0 +1,35 @@
...
+    pub header: Header,
+    /// Block body.
+    #[rlp(flatten)]
+    pub body: BlockBody<T>,
+}
+
+/// A block body.
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct BlockBody<T: Encodable + Decodable + Encodable2718 + Decodable2718> {
```
mattsse, line 27: let's not require anything here

crates/consensus/src/block.rs
```diff
@@ -0,0 +1,35 @@
...
+use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
+
+/// Ethereum full block.
+///
+/// Withdrawals can be optionally included at the end of the RLP encoded message.
+///
+/// Taken from [reth-primitives](https://github.com/paradigmxyz/reth)
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct Block<T: Encodable + Decodable + Encodable2718 + Decodable2718> {
```
mattsse, line 16: same

crates/consensus/src/block.rs
```diff
@@ -0,0 +1,35 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct BlockBody<T: Encodable + Decodable + Encodable2718 + Decodable2718> {
+    /// Transactions in this block.
+    pub transactions: Vec<T>,
+    /// Ommers/uncles header.
+    pub ommers: Vec<Header>,
+    /// Block withdrawals.
+    pub withdrawals: Option<Vec<Withdrawal>>,
+    // TODO: add request with rlp encoding support
```
mattsse, line 34: we should move the request's type here:

https://github.com/paradigmxyz/reth/issues/11082

mattsse (changes requested):

crates/consensus/src/request.rs
```diff
@@ -4,7 +4,42 @@ use alloy_eips::{
...
+
+impl Decodable for Requests {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        Ok(<Vec<Bytes> as Decodable>::decode(buf)?
+            .into_iter()
+            .map(|bytes| Request::decode_7685(&mut bytes.as_ref()))
+            .collect::<Result<Vec<_>, alloy_eips::eip7685::Eip7685Error>>()
+            .map(Self)?)
+    }
+}
```
emhane, line 42: pls add tests for these impls!

emhane (commented):
wondering if the en-/decoding of requests is a bit out of scope?

crates/consensus/src/request.rs
refcell: There already was a Requests object upstream :)

refcell: I've removed this now

crates/consensus/src/block.rs
refcell: I think we need Encodable + Decodable for the `RlpEncodable, RlpDecodable` auto-derive above.

mattsse: I see

emhane (approved):

crates/consensus/src/block.rs
```diff
@@ -0,0 +1,38 @@
...
+use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
+
+/// Ethereum full block.
+///
+/// Withdrawals can be optionally included at the end of the RLP encoded message.
+///
+/// Taken from [reth-primitives](https://github.com/paradigmxyz/reth)
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct Block<T: Encodable + Decodable + Encodable2718 + Decodable2718> {
```
mattsse, line 18: this should only use `Encodable + Decodable`

and it is expected that this is equivalent to 

https://github.com/alloy-rs/alloy/blob/21bc07ebbcb327486fc260cb1a8fb9d23f4bae53/crates/eips/src/eip2718.rs#L219-L219

see:

https://github.com/alloy-rs/alloy/blob/21bc07ebbcb327486fc260cb1a8fb9d23f4bae53/crates/consensus/src/transaction/envelope.rs#L329-L332

because this type is the p2p type

crates/consensus/src/block.rs
```diff
@@ -0,0 +1,38 @@
...
+    eip2718::{Decodable2718, Encodable2718},
+    eip4895::Withdrawal,
+};
+use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
+
+/// Ethereum full block.
+///
+/// Withdrawals can be optionally included at the end of the RLP encoded message.
+///
+/// Taken from [reth-primitives](https://github.com/paradigmxyz/reth)
```
mattsse, line 15: can we add this as context:

https://github.com/paradigmxyz/reth/blob/3fa1e28e12cf7b05a365e93ade19a6de9e2aafd3/crates/primitives/src/block.rs#L567-L569

and reference 

https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-encoding-and-validity

mattsse (changes requested):
suggestions

refcell:
Think I've added what you're asking for but LMK if I'm missing something :)

crates/consensus/src/block.rs
klkvr: hmm, I don't think it's needed? those derives will add bounds for respective impl blocks

crates/consensus/src/block.rs
```diff
@@ -0,0 +1,39 @@
...
+///
+/// See p2p block encoding reference: <https://github.com/ethereum/devp2p/blob/master/caps/eth.md#block-encoding-and-validity>
+#[derive(Debug, Clone, PartialEq, Eq, Default, RlpEncodable, RlpDecodable)]
+#[rlp(trailing)]
+pub struct Block<T: Encodable + Decodable> {
+    /// Block header.
+    pub header: Header,
+    /// Block body.
+    #[rlp(flatten)]
```
klkvr, line 21: I don't think `#[rlp(flatten)]` works? this should probably be manual impl instead

mattsse (approved):

PR #1322, opened 2024-09-20
feat(rpc-types-beacon): `BuilderBlockValidationRequestV4`

ryanschneider:
## Motivation

As suggested by @mattsse [here](https://github.com/alloy-rs/alloy/pull/1310#issuecomment-2360727290) this is a simple type alias for `BuilderBlockValidationRequestV4`.

I still think we should consider refactoring these structs to use an explicit versioned `ExecutionPayloadV*` instead of the enum, but I'm not sure how we feel about breaking compat on the V1 and V2 structs.

We could also move them from `::relay` to `::validation` to mirror the layout in reth: https://github.com/paradigmxyz/reth/blob/3daec1d9b97c2d654c7f84e8521ac552dcaf4928/crates/rpc/rpc-api/src/validation.rs

## Solution

Anyways the type alias works for Pectra in the short-term, but happy to make the other changes if people are ok with that.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-beacon/src/relay.rs
+++ b/crates/rpc-types-beacon/src/relay.rs
@@ -225,6 +225,9 @@ pub struct BuilderBlockValidationRequestV3 {
     pub parent_beacon_block_root: B256,
 }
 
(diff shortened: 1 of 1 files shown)


mattsse:
> We could also move them from ::relay to ::validation to mirror the layout in reth:

that makes sense

mattsse (approved):

PR #1326, opened 2024-09-21
feat(consensus): move requests struct definition from reth

PanGan21:
It moves the `Requests` definition from reth to where `Request` is defined.

## Motivation
Related to: https://github.com/paradigmxyz/reth/issues/11082


mattsse (approved):
lgtm

fyi @refcell 

once there's a new release we can upstream this to reth

emhane (approved):

Issue #1297, opened 2024-09-16
[Feature] Add getters to `alloy_network_primitives::TransactionResponse`


emhane:
### Component

rpc

### Describe the feature you would like

Add all getters for all fields of `alloy_rpc_types_eth::Transaction` as trait methods to `alloy_network_primitives::TransactionResponse`

emhane:
also this one @garwahl?

garwahl:
Yep can take this one too, thanks

PR #1328, opened 2024-09-22
feat: Add getters into TransactionResponse and update implementations 

garwahl:
Closes https://github.com/alloy-rs/alloy/issues/1297

Adds getters into the `TransactionResponse` trait and updates implementations accordingly.

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -1,5 +1,6 @@
 use alloy_eips::eip7702::SignedAuthorization;
-use alloy_primitives::{Address, BlockHash, Bytes, TxHash, B256, U256};
+use alloy_eips::eip2930::AccessList;
+use alloy_primitives::{Address, Signature, ChainId, BlockHash, Bytes, TxHash, B256, U256};
 use alloy_serde::WithOtherFields;
 
 use crate::BlockTransactions;
@@ -67,6 +68,18 @@ pub trait TransactionResponse {
     #[doc(alias = "transaction_hash")]
     fn tx_hash(&self) -> TxHash;
 
+    /// Nonce
+    fn nonce(&self) -> u64;
+
+    /// Block hash
+    fn block_hash(&self) -> Option<BlockHash>;
+
+    /// Block number
+    fn block_number(&self) -> Option<u64>;
+
+    /// Transaction Index
+    fn transaction_index(&self) -> Option<u64>;
+
     /// Sender of the transaction
     fn from(&self) -> Address;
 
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/traits.rs
```diff
@@ -76,12 +88,42 @@ pub trait TransactionResponse {
...
+
+    /// Configured max fee per blob gas for eip-4844 transactions
+    fn max_fee_per_blob_gas(&self) -> Option<u128>;
+
     /// Input data
     #[doc(alias = "calldata")]
     fn input(&self) -> &Bytes;
+
+    /// Transaction signature
+    fn signature(&self) -> Option<Signature>;
```
garwahl, line 111: @emhane `Signature` comes from `alloy-rpc-types-eth` which can't be imported `network-primitives/Cargo.toml` due to a cyclic dependency since `rpc-types-eth/Cargo.toml` already pulls in network-primitives.

Do you have a suggestion here?

emhane: probably best to add signature as associated type to the trait

emhane (approved):

garwahl:
@emhane I've updated this to use an associated type, feel free to merge if LGTY or happy to wait for a +1 from another reviewer

crates/network-primitives/src/traits.rs
```diff
@@ -63,10 +64,24 @@ pub trait ReceiptResponse {
 
 /// Transaction JSON-RPC response.
 pub trait TransactionResponse {
```
mattsse, line 66: Is this now just `Transaction` + additional functions to get the block context?

mattsse (commented):
wdyt @klkvr

crates/network-primitives/src/traits.rs
garwahl: Yep, although I'm not entirely across the context. It looks similar to how `ReceiptResponse` is a projection of the fields on a `TransactionReceipt`. Happy to be guided by owners on this

klkvr (approved):
lgtm, this is now very similar to `Transaction` trait, we can dedup those in a follow-up

mattsse (approved):

mattsse:
another thing we can do is restrict the signature to `EncodeableSignature` perhaps

garwahl:
Thanks @klkvr @mattsse @emhane! Can one of you kick off a merge if LGTY, happy to address in a follow up. 

> another thing we can do is restrict the signature to EncodeableSignature perhaps
@mattsse `alloy_rpc_types_eth::transaction::Signature` doesn't implement `EncodeableSignature`. If this is something we want perhaps we can do it in a follow up + enforcement?

mattsse:
imo, that would make sense

emhane:
> lgtm, this is now very similar to `Transaction` trait, we can dedup those in a follow-up

probably will be nicest if we have the consensus `Transaction` trait as super trait of the rpc `TransactionResponse` trait, because the way the transaction response is built is: `Transaction` type + signature + singer address + block context. hence, makes sense if `TransactionResponse` is an extension of core `Transaction`.

klkvr:
> probably will be nicest if we have the consensus `Transaction` trait as super trait of the rpc `TransactionResponse` trait

agree, was thinking about this as well

PR #1330, opened 2024-09-23
make factory and paymaster fields optional in `PackedUserOperation`

dastanbeksamatov:
## Motivation

Factory and paymaster related fields don't need to be required, as it forces to pass some default value, i.e `null address` for `Address`, 0 for gas limits, etc. which leads to invalid user ops.

## Solution

make factory and paymaster related fields optional

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/erc4337.rs
+++ b/crates/rpc-types-eth/src/erc4337.rs
@@ -85,10 +85,12 @@ pub struct PackedUserOperation {
     pub nonce: U256,
     /// Deployer contract address: Required exclusively for deploying new accounts that don't yet
     /// exist on the blockchain.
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
all of these should probably be ignored if missing?

see for ref

https://github.com/alloy-rs/alloy/blob/21bc07ebbcb327486fc260cb1a8fb9d23f4bae53/crates/rpc-types-eth/src/pubsub.rs#L48-L48

makes sense, q about optional fields

mattsse (approved):

mattsse (approved):

PR #1332, opened 2024-09-23
`ExecutionPayloadV1`: make gas limit `u128`

tcoratger:
Related https://github.com/paradigmxyz/reth/pull/11130


mattsse (commented):
this goes against engine api spec, cant do this

PR #1256, opened 2024-09-07
feat: Add iterator to BlobTransactionSidecar

DoTheBestToGetTheBest:
closes https://github.com/alloy-rs/alloy/issues/1255

PR #1334, opened 2024-09-23
feat: Blob Tx Sidecar Iterator

### LINKED ISSUE #1255: [Feature] Add iterator to BlobTransactionSidecar
### Component

consensus, eips, genesis

this type is a container type https://github.com/alloy-rs/alloy/blob/122c4b187af3c621bdd10f829d44538300c0dae0/crates/eips/src/eip4844/sidecar.rs#L22-L33

all arrays are supposed to have the same length.

an iterator type would be helpful here that yields

```rust
struct BlobTransactionSidecarItem {
  index, blob, kzg_commitment, kzg_proof
}
```

which is basically a subset of 

https://github.com/alloy-rs/alloy/blob/122c4b187af3c621bdd10f829d44538300c0dae0/crates/rpc-types-beacon/src/sidecar.rs#L65-L75

## TODO
* introduce type
* add an iterator for BlobTransactionSidecar
* add verify function for BlobTransactionSidecarItem, see also: https://github.com/anton-rs/kona/blob/4ba28121ac0665f268876d6bbf15f1b4dfcbc011/crates/primitives/src/sidecar.rs#L66

### Additional context

_No response_

refcell:
### Description

> [!NOTE]
>
> Replaces #1256 

Implements `IntoIter` for the blob tx sidecar.

Closes #1255

### DIFF
--- a/crates/eips/src/eip4844/sidecar.rs
+++ b/crates/eips/src/eip4844/sidecar.rs
@@ -3,6 +3,7 @@
 use crate::eip4844::{
     kzg_to_versioned_hash, Blob, Bytes48, BYTES_PER_BLOB, BYTES_PER_COMMITMENT, BYTES_PER_PROOF,
 };
+use alloc::boxed::Box;
 use alloy_primitives::{bytes::BufMut, B256};
 use alloy_rlp::{Decodable, Encodable};
 
@@ -11,6 +12,10 @@ use crate::eip4844::MAX_BLOBS_PER_BLOCK;
 
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
+
+/// The versioned hash version for KZG.
+#[cfg(feature = "kzg")]
+pub(crate) const VERSIONED_HASH_VERSION_KZG: u8 = 0x01;
 
 /// This represents a set of blobs, and its corresponding commitments and proofs.
 ///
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+            .into_iter()
+            .zip(self.commitments)
+            .zip(self.proofs)
+            .map(|((blob, commitment), proof)| BlobTransactionSidecarItem {
+                index: 0,
+                blob,
+                kzg_commitment: commitment,
+                kzg_proof: proof,
+            })
+            .collect::<Vec<_>>()
```
mattsse, line 54: this collect is a bit pointless and could be avoided with a custom iterator type that is generic over I: Iterator (blob,commitment, proof) and uses `Enumerate<I>`

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+
+/// A single blob sidecar.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[repr(C)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct BlobTransactionSidecarItem {
+    /// The index.
+    pub index: usize,
+    /// The index.
+    pub blob: Blob,
```
mattsse, line 67: serde is problematic here because this will overflow in debug mode, 

we can move 

https://github.com/alloy-rs/alloy/blob/21bc07ebbcb327486fc260cb1a8fb9d23f4bae53/crates/rpc-types-beacon/src/sidecar.rs#L105-L108

into this crate

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+    /// The KZG commitment.
+    #[cfg_attr(feature = "serde", serde(rename = "kzg_commitment"))]
+    pub kzg_commitment: Bytes48,
+    /// The KZG proof.
+    #[cfg_attr(feature = "serde", serde(rename = "kzg_proof"))]
```
mattsse, line 72: these renames are redundant

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+            Ok(true)
+        } else {
+            Err(BlobTransactionValidationError::InvalidProof)
+        }
+    }
+
+    /// Verify the blob sidecar against its [crate::BlockNumHash].
+    pub fn verify_blob(
+        &self,
+        hash: &crate::BlockNumHash,
```
mattsse, line 114: this should use `NumHash` because this is not the block number, right?

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+                have: blob_hash_part,
+                expected: hash.hash,
+            });
+        }
+
+        let computed_hash = self.to_kzg_versioned_hash();
+        if computed_hash != hash.hash {
+            return Err(BlobTransactionValidationError::WrongVersionedHash {
+                have: computed_hash.into(),
+                expected: alloy_primitives::FixedBytes(*hash.hash),
```
mattsse, line 128: not needed

mattsse (changes requested):

crates/eips/src/eip4844/sidecar.rs
```diff
@@ -32,6 +36,107 @@ pub struct BlobTransactionSidecar {
...
+    type Item = BlobTransactionSidecarItem;
+    type IntoIter = alloc::vec::IntoIter<BlobTransactionSidecarItem>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.blobs
+            .into_iter()
+            .zip(self.commitments)
+            .zip(self.proofs)
+            .map(|((blob, commitment), proof)| BlobTransactionSidecarItem {
+                index: 0,
```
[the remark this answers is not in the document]
refcell: Oops ty

crates/eips/src/eip4844/sidecar.rs
mattsse: can open a good first for this

mattsse (approved):

PR #1335, opened 2024-09-23
dep: add full feature to `derive_more`

tcoratger:
As per the `derive_more` documentation and specifically https://github.com/JelteF/derive_more?tab=readme-ov-file#installation, it seems that in last versions of Rust, to compile properly, we need to specify features to avoid errors of this type:

```shell
error[E0432]: unresolved imports `derive_more::Deref`, `derive_more::DerefMut`, `derive_more::From`, `derive_more::IntoIterator`
   --> crates/consensus/src/request.rs:10:19
    |
10  | use derive_more::{Deref, DerefMut, From, IntoIterator};
    |                   ^^^^^  ^^^^^^^^  ^^^^  ^^^^^^^^^^^^ no `IntoIterator` in the root
    |                   |      |         |
    |                   |      |         no `From` in the root
    |                   |      no `DerefMut` in the root
    |                   no `Deref` in the root
```

Since we are using quite a few elements currently, we are choosing the full configuration for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ semver = "1.0"
 thiserror = "1.0"
 thiserror-no-std = "2.0.2"
 url = "2.5"
-derive_more = { version = "1.0.0", default-features = false, features = ["full"] }
+derive_more = { version = "1.0.0", default-features = false }
 http = "1.1.0"
 
 ## serde
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ semver = "1.0"
 thiserror = "1.0"
 thiserror-no-std = "2.0.2"
 url = "2.5"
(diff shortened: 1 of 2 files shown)


mattsse (commented):
how to reproduce the compiler error,
can we instead enable the required features in the consensus crate?

tcoratger:
> how to reproduce the compiler error, can we instead enable the required features in the consensus crate?

Maybe my version of rust or something, anyway seems to be gone now but I've added the specific features we need for the consensus crate to be consistent with the other usages on the other crates.

DaniPopes (approved):
Probably due to outdated deps locally, eg alloy-primitives might've enabled these already more recently, but you have an older version of it that doesn't do that; anyway doesn't hurt

PR #1343, opened 2024-09-24
fix: use vec for flat call frame

mattsse:
this should be a vec of course because it's a flattened callframe..

ref https://github.com/paradigmxyz/reth/pull/11114


0xkrane:
yeah i wasn't sure why it wasn't a vec, thanks!

mattsse:
we can merge the pr without the vec until we bump alloy

PR #1348, opened 2024-09-24
fix(serde): encode optional types as Some

shekhirin:
## Motivation

There are binary formats such as bincode that care whether an `Option` is serialized using `serialize_some` or just `serialize`, because they prepend the enum variant of `Option` as a starting byte.

The tests were passing because JSON doesn't care about this.

## Solution

Serialize `quantity::opt` as `Some` when the value is `Some`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/serde/src/quantity.rs
+++ b/crates/serde/src/quantity.rs
@@ -34,7 +34,7 @@ where
 /// See [`quantity`](self) for more information.
 pub mod opt {
     use super::private::ConvertRuint;
-    use serde::{Deserialize, Deserializer, Serializer};
+    use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
     /// Serializes an optional primitive number as a "quantity" hex string.
     pub fn serialize<T, S>(value: &Option<T>, serializer: S) -> Result<S::Ok, S::Error>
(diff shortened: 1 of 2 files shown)


crates/serde/src/quantity.rs
```diff
@@ -43,7 +43,7 @@ pub mod opt {
         S: Serializer,
     {
         match value {
-            Some(value) => super::serialize(value, serializer),
+            Some(value) => Some(value.into_ruint()).serialize(serializer),
```
mattsse, line 46: this should be

```suggestion
            Some(value) => serializer.serialize_some(value.into_ruint()),
```

see Serialize impl on option which does another match

mattsse (changes requested):
nit

crates/serde/src/quantity.rs
shekhirin: right

shekhirin:
It made me think, if that's a bug, do we want to have it on other modules too such as https://github.com/alloy-rs/alloy/blob/cf9a98fd7ed41f1e86af1d58d3eba9bc1c9799b0/crates/serde/src/num.rs#L23-L33

mattsse:
nice catch we should actually use the serialize some here, this doesn't matter for json but I guess it does for bincode

mattsse (approved):

mattsse (approved):

PR #1350, opened 2024-09-24
feat(serde): remove deprecated `num` module

DaniPopes:
Deprecated over 3 months ago in https://github.com/alloy-rs/alloy/pull/855.

Unfortunately the `#[deprecated]` attribute warnings don't show if used inside of `#[serde(...)]` so this may cause more breakage than it should.


crates/serde/src/quantity.rs
```diff
@@ -89,6 +89,48 @@ pub mod vec {
     }
 }
 
+/// serde functions for handling `Vec<Vec<u128>>` via [U128](alloy_primitives::U128)
+pub mod u128_vec_vec_opt {
```
DaniPopes, line 93: Just moved

mattsse (approved):
reasonable

PR #1359, opened 2024-09-25
feat: impl From<Eip2718Error> for alloy_rlp::Error

klkvr:
## Motivation

`Eip2718Error` is just a more specific decoding error and it is useful to have an easy way to convert it to alloy-rlp error

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -45,6 +45,15 @@ impl From<alloy_rlp::Error> for Eip2718Error {
     }
 }
 
+impl From<Eip2718Error> for alloy_rlp::Error {
+    fn from(err: Eip2718Error) -> Self {
+        match err {
+            Eip2718Error::RlpError(err) => err,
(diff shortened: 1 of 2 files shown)


crates/eips/src/eip2718.rs
```diff
@@ -45,6 +45,15 @@ impl From<alloy_rlp::Error> for Eip2718Error {
...
+    fn from(err: Eip2718Error) -> Self {
+        match err {
+            Eip2718Error::RlpError(err) => err,
+            Eip2718Error::UnexpectedType(_) => alloy_rlp::Error::Custom("Unexpected type flag"),
+        }
+    }
+}
+
 #[cfg(feature = "std")]
 impl std::error::Error for Eip2718Error {}
```
mattsse, line 58: unrelated: this lacks `fn source` which should return the wrapped rlperr

mind fixing this in a separate pr quickly?

mattsse (approved):
makes sense, CI failures unrelated

PR #1368, opened 2024-09-25
feat: replacing the map types  (#1365)

dhruvmalik007:
- replacing the HashMap and HashSet from the std::collections to the alloy-primitives along with setting up the cargo file.

- setting the dependencies in the cargo files

- [ ]Bug fixes and new features should include tests: TBD.

## Motivation

to close the PR #1365 . 

## Solution

migrating the types as defined in the given issue.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/lib.rs
+++ b/crates/rpc-types-eth/src/lib.rs
@@ -15,7 +15,7 @@ extern crate alloc;
 pub mod collections {
     cfg_if::cfg_if! {
         if #[cfg(feature = "std")] {
-            pub use std::collections::{hash_set, HashMap, HashSet};
+            pub use alloy_primitives::map::{hash_set, HashMap, HashSet};
             use hashbrown as _;
         } else {
             pub use hashbrown::{hash_set, HashMap, HashSet};
--- a/crates/rpc-types-eth/src/lib.rs
+++ b/crates/rpc-types-eth/src/lib.rs
@@ -12,16 +12,7 @@
 extern crate alloc;
 
 /// Standardized collections across `std` and `no_std` environments.
-pub mod collections {
-    cfg_if::cfg_if! {
-        if #[cfg(feature = "std")] {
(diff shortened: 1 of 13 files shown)


crates/rpc-types-eth/src/lib.rs
```diff
@@ -15,7 +15,7 @@ extern crate alloc;
 pub mod collections {
```
mattsse, line 15: I believe this is now obsolete?

crates/rpc-types/Cargo.toml
```diff
@@ -30,6 +30,7 @@ alloy-rpc-types-mev = { workspace = true, optional = true }
 alloy-rpc-types-trace = { workspace = true, optional = true }
 alloy-rpc-types-txpool = { workspace = true, optional = true }
 serde = { workspace = true, features = ["derive", "std"] }
+alloy-primitives = { version = "0.8.5", features = ["map"] }
```
mattsse, line 33: this should use workspace

crates/rpc-types/src/rpc.rs
```diff
@@ -28,13 +27,16 @@ mod tests {
     #[test]
     fn test_parse_module_versions_roundtrip() {
         let s = r#"{"txpool":"1.0","trace":"1.0","eth":"1.0","web3":"1.0","net":"1.0"}"#;
-        let module_map = HashMap::from([
+        let mut module_map: HashMap<String, String> = HashMap::default(); // Change to FxHashMap
+
+         module_map.extend([
```
mattsse, line 32: this should use `from_iter`

mattsse (changes requested):
needs closer look by @DaniPopes

crates/rpc-types-eth/src/lib.rs
dhruvmalik007: yes Its removed as redundant.

PR #1369, opened 2024-09-25
feat: errors for responses

prestwich:
- Adds shortcut methods to instantiate response errors
- Adds `serialize_payload` to the response objects to allow easy converting to the default `Response`

**This is a breaking change, as it changes the ErrorPayload message to a Cow**. However, it doesn't break anything in the library/

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
all of these seem fine, the Cow makes sense, though only really for server POV

PR #1370, opened 2024-09-25
crate: alloy-json-rpc-router

prestwich:
## Motivation

See how hard it would be to make a better json rpc router than jsonrpsee

turns out not that hard I think 🤔 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


prestwich:
now based on #1402 to fix CI

mattsse:
closing since this is now https://github.com/init4tech/ajj

PR #1383, opened 2024-09-26
chore: flatten eip-7685 requests into a single opaque list

onbjerg:
## Motivation

The three different request fields have been replaced by a single field.

Requests over engine API will now be in an opaquely encoded format as defined by EIP-7685, and they will be in a single list.

See https://github.com/ethereum/execution-apis/pull/591

## Solution

Requests are now a list of opaque `Bytes`. Additionally, the changes to the engine API also removes the need for `ExecutionPayloadV4`, so I've removed that (and all variants of it).

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/lib.rs
+++ b/crates/rpc-types-engine/src/lib.rs
@@ -59,6 +59,4 @@ pub const CAPABILITIES: &[&str] = &[
     "engine_newPayloadV4",
     "engine_getPayloadBodiesByHashV1",
     "engine_getPayloadBodiesByRangeV1",
-    "engine_getPayloadBodiesByHashV2",
-    "engine_getPayloadBodiesByRangeV2",
 ];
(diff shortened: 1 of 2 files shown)


onbjerg:
After some thinking we likely do not gain much by retaining the type here instead of using blobs, since validation is mostly a CL responsibility now, and system contracts adhering to EIP-7685 return the requests pre-serialized. For user-facing RPC it will matter more, but at that point in time it's ok to pay for the ser/de cost, while we'd likely want to avoid it during execution.

onbjerg:
Looks like I need to update `ExecutionPayloadBodyV2` as well?

mattsse (approved):
lgtm

DaniPopes (approved):

onbjerg:
Sorry I meant `ExecutionPayloadBodyV4`

onbjerg:
>why?

https://github.com/paradigmxyz/reth/blob/8d32fd788bfa8c2e041bd1b93ecca37093a09c5a/crates/rpc/rpc-types-compat/src/engine/payload.rs#L384-L428

mattsse (approved):

Issue #1344, opened 2024-09-24
[Feature] Unify Transaction and TransactionResponse trait


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

```
trait TransactionResponse: Transaction {
  // fns for block context
}
```

ref

https://github.com/alloy-rs/alloy/pull/1328#issuecomment-2368079765

cc @garwahl

garwahl:
Yep I'll take this

PR #1387, opened 2024-09-26
Add supertrait alloy_consensus::Transaction to RPC TransactionResponse

emhane:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/1344

## Solution

- Unifies getters for mutually exclusive values `max_fee_per_gas` and `gas_price` in trait `alloy_consensus::Transaction`. Removes getter for deprecated name `gas_price` from trait `alloy_consensus::Transaction`.
- Adds `alloy_consensus::Transaction` as super trait of `TransactionResponse`
- Removes redundant methods from `TransactionResponse`

Follow up would be to manually implement de-/serialisation for `alloy_rpc_types_eth::Transaction`, and remove trait methods of `TransactionResponse`: `to`, `gas_price`, `max_fee_per_gas` and `transaction_type`, which are just matching the values returned by `alloy_consensus::Transaction`, with the right label w.r.t. to the json object standard for the transaction type.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/mod.rs
+++ b/crates/consensus/src/transaction/mod.rs
@@ -4,6 +4,7 @@ use crate::Signed;
 use alloc::vec::Vec;
 use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization};
 use alloy_primitives::{keccak256, ChainId, TxKind, B256, U256};
+use alloy_serde::WithOtherFields;
 use core::any;
 
 mod eip1559;
@@ -47,9 +48,6 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
     /// Get `gas_limit`.
     fn gas_limit(&self) -> u64;
 
-    /// Get `gas_price`.
-    fn gas_price(&self) -> Option<u128>;
-
     /// Returns the EIP-1559 the maximum fee per gas the caller is willing to pay.
     ///
     /// For legacy transactions this is `gas_price`.
@@ -57,10 +55,9 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
     /// This is also commonly referred to as the "Gas Fee Cap".
     fn max_fee_per_gas(&self) -> u128;
 
-    /// Returns the EIP-1559 Priority fee the caller is paying to the block author.
-    ///
-    /// This will return `None` for non-EIP1559 transactions
-    fn max_priority_fee_per_gas(&self) -> Option<u128>;
+    /// Return the max priority fee per gas if the transaction is an EIP-1559 transaction, and
+    /// otherwise return the gas price.
+    fn max_priority_fee_per_gas(&self) -> u128;
 
     /// Max fee per blob gas for EIP-4844 transaction.
     ///
@@ -68,15 +65,6 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
     ///
     /// This is also commonly referred to as the "Blob Gas Fee Cap".
     fn max_fee_per_blob_gas(&self) -> Option<u128>;
-
-    /// Return the max priority fee per gas if the transaction is an EIP-1559 transaction, and
-    /// otherwise return the gas price.
-    ///
-    /// # Warning
-    ///
-    /// This is different than the `max_priority_fee_per_gas` method, which returns `None` for
-    /// non-EIP-1559 transactions.
-    fn priority_fee_or_price(&self) -> u128;
 
     /// Returns the effective tip for this transaction.
     ///
@@ -96,8 +84,11 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
         let fee = max_fee_per_gas - base_fee;
 
         // Compare the fee with max_priority_fee_per_gas (or gas price for non-EIP1559 transactions)
-        self.max_priority_fee_per_gas()
-            .map_or(Some(fee), |priority_fee| Some(fee.min(priority_fee)))
+        if self.ty() < 2 {
+            return Some(fee);
+        }
+
+        Some(fee.min(self.max_priority_fee_per_gas()))
     }
 
     /// Get `to`.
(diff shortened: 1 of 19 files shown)


crates/consensus/src/transaction/mod.rs
```diff
@@ -47,20 +47,16 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
...
     /// This is also commonly referred to as the "Gas Fee Cap".
     fn max_fee_per_gas(&self) -> u128;
 
-    /// Returns the EIP-1559 Priority fee the caller is paying to the block author.
-    ///
-    /// This will return `None` for non-EIP1559 transactions
-    fn max_priority_fee_per_gas(&self) -> Option<u128>;
+    /// Return the max priority fee per gas if the transaction is an EIP-1559 transaction, and
+    /// otherwise return the gas price.
+    fn max_priority_fee_per_gas(&self) -> u128;
```
mattsse, line 59: why is this no longer an option?

mattsse (changes requested):
this changes the fee functions, which is a bit confusing, seems like we want all of those optional because these are tx specific

crates/consensus/src/transaction/mod.rs
```diff
@@ -47,20 +47,16 @@ pub trait Transaction: any::Any + Send + Sync + 'static {
     /// Get `gas_limit`.
     fn gas_limit(&self) -> u64;
 
-    /// Get `gas_price`.
-    fn gas_price(&self) -> Option<u128>;
```
klkvr, line 51: what's our motivation for removing this? I can see how it's redundant from execution perspective, though it is an actual separate property of transaction which is important from RPC PoV

I think it's pretty much a question on how we expect this trait to be consumed and whether separate getters might be useful

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -338,87 +338,89 @@ impl TryFrom<Transaction> for TxEnvelope {
...
 
-    fn access_list(&self) -> Option<AccessList> {
-        self.access_list.clone()
+    fn from(&self) -> Address {
+        self.from
     }
 
-    fn transaction_type(&self) -> Option<u8> {
-        self.transaction_type
+    fn to(&self) -> Option<Address> {
```
klkvr, line 418: I think this can be removed? given that we already have `to` on consensus `Transaction`

crates/consensus/src/transaction/mod.rs
emhane: removed redundancy within the consensus `Transaction` trait. `gas_price` has been deprecated in favour of `max fee per gas` and `max priority fee per gas`. checking if this returns an option is a bad way to check if this is a pre-eip1559 tx, and instead trait method `Transaction::ty` should be used.

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -338,87 +338,89 @@ impl TryFrom<Transaction> for TxEnvelope {
...
-    fn block_number(&self) -> Option<u64> {
-        self.block_number
+    fn max_fee_per_gas(&self) -> u128 {
+        self.max_fee_per_gas.unwrap_or(self.gas_price.unwrap_or_default())
     }
 
-    fn transaction_index(&self) -> Option<u64> {
-        self.transaction_index
+    fn max_priority_fee_per_gas(&self) -> u128 {
+        self.max_fee_per_gas.unwrap_or(self.gas_price.unwrap_or_default())
```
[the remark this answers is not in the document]
emhane: will get cleaned up by manual impl of serialisation for the the `alloy_rpc_types_eth::Transaction` type. then we don't need to have one field each for the essentially mutually exclusive fields `max fee per gas` and `gas price`, we just need the field `max fee per gas`, and the manual serialisation impl will just check `if self.ty() < 2` then use `gasPrice` as label, otherwise use `maxBaseFee`

crates/consensus/src/transaction/mod.rs
emhane: the gas price field is never used on tx types since max fee per gas was introduced

crates/rpc-types-eth/src/transaction/mod.rs
emhane: consensus `Transaction` returns `TxKind`. requiring every caller of `to` in rpc context to do this conversion to `Option<Address>` isn't good.

crates/consensus/src/transaction/mod.rs
mattsse: this should be an option because not every tx type has this

mattsse (changes requested):
I'd like to keep the functions as they were before, because this issue was intended to unify the traits. we can still discuss changes separately, but this is now becoming a blocker and I'd like to make this change now and discuss nuances later.

crates/rpc-types-eth/src/transaction/mod.rs
klkvr: I think this is fine? the conversion is as simple as calling `kind.to()`

and we can probably change `to` type to `TxKind` on rpc types as it's done in consensus serde https://github.com/alloy-rs/alloy/blob/599e57751fd986a4b3fb64935e80cc512b87a018/crates/consensus/src/transaction/legacy.rs#L48-L49

emhane: following matze's directive, prio of this pr should be not to change the API. pls open an issue.

crates/network-primitives/src/traits.rs
```diff
@@ -85,48 +83,38 @@ pub trait TransactionResponse {
...
-
-    /// Gas limit
-    fn gas(&self) -> u64;
-
-    /// Max BaseFeePerGas the user is willing to pay
-    fn max_fee_per_gas(&self) -> Option<u128>;
-
-    /// The miner's tip
-    fn max_priority_fee_per_gas(&self) -> Option<u128>;
+    /// Recipient of the transaction. Returns `None`if transaction is a contract creation.
```
klkvr, line 86:
```suggestion
    /// Recipient of the transaction. Returns `None` if transaction is a contract creation.
```

klkvr (approved):

mattsse (approved):

PR #1389, opened 2024-09-27
fix: ensure `max_fee_per_blob_gas` field handles `Some(0)` gracefully

zerosnacks:
## Motivation

Closes: https://github.com/alloy-rs/alloy/issues/1371

## Solution

Related Foundry PR: https://github.com/foundry-rs/foundry/pull/8963 that depends on this implementation

If `max_fee_per_blob_gas` is `Some(0)` or `None` we set it to a valid value `next_block_blob_fee()`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -279,7 +279,7 @@ mod tests {
 
         let receipt = tx.get_receipt().await.unwrap();
 
-        assert_eq!(receipt.effective_gas_price, 10_000_000_01);
+        assert_eq!(receipt.effective_gas_price, 1_000_000_001);
         assert_eq!(receipt.gas_used, 21000);
     }
 
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -247,17 +247,22 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{ProviderBuilder, WalletProvider};
+    use crate::ProviderBuilder;
+    use alloy_consensus::{SidecarBuilder, SimpleCoder};
+    use alloy_eips::eip4844::DATA_GAS_PER_BLOB;
     use alloy_primitives::{address, U256};
     use alloy_rpc_types_eth::TransactionRequest;
+
+    fn init_tracing() {
+        let _ = tracing_subscriber::fmt::try_init();
+    }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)


crates/provider/src/fillers/gas.rs
```diff
@@ -196,8 +197,11 @@ where
     type Fillable = u128;
 
     fn status(&self, tx: &<N as Network>::TransactionRequest) -> FillerControlFlow {
-        // nothing to fill if non-eip4844 tx or max_fee_per_blob_gas is already set
-        if tx.blob_sidecar().is_none() || tx.max_fee_per_blob_gas().is_some() {
+        // Nothing to fill if non-eip4844 tx or `max_fee_per_blob_gas` is already set to a valid
+        // value.
+        if tx.blob_sidecar().is_none()
+            || tx.max_fee_per_blob_gas().is_some_and(|gas| gas >= BLOB_TX_MIN_BLOB_GASPRICE)
```
zerosnacks, line 203: If the field is set to `Some(0)` (which happens because of serialization) we force setting it to `.next_block_blob_fee()` as 0 is always invalid as it is less than `BLOB_TX_MIN_BLOB_GASPRICE` (1).

crates/provider/src/fillers/gas.rs
```diff
@@ -288,6 +301,65 @@ mod tests {
...
+
+        let provider = ProviderBuilder::new().with_recommended_fillers().on_anvil_with_wallet();
+
+        let sidecar: SidecarBuilder<SimpleCoder> = SidecarBuilder::from_slice(b"Hello World");
+        let sidecar = sidecar.build().unwrap();
+
+        let tx = TransactionRequest {
+            to: Some(address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into()),
+            sidecar: Some(sidecar),
+            ..Default::default()
```
zerosnacks, line 319: Tests both cases, implicit `Some(0) / None` and explicit `Some(0)`

mattsse (approved):
lgtm, this is effectively a safe guard to prevent user error, right?

zerosnacks:
> lgtm, this is effectively a safe guard to prevent user error, right?

It is but it mostly enables you do not define the field and rely on the filler. What happens is that the transaction is serialized as 0 as the field is required. This is then interpreted as `Some(0)` rather than `None` that would previously then skip the filler.

Issue #1349, opened 2024-09-24
[Feature] Bincode support for `alloy-consensus` and `alloy-primitive` types


shekhirin:
### Component

consensus, eips, genesis, serde

## Problem

[Bincode](https://github.com/bincode-org/bincode/issues/326) does not support `#[serde(skip*)]` attributes. Additionally, there's a similar issue with [rmp-serde](https://github.com/3Hren/msgpack-rust/issues/86).

It means that we cannot use a lot of Alloy types for serialization/deserialization using these formats.

## Solution

Introduce helper serde modules for those structs in `alloy-consensus` that have `#[serde(skip*)]` attributes on fields and remove those attributes. Modules should be similar to https://github.com/alloy-rs/alloy/blob/5c7470c113df23a625456413bdf8c87e0b7438bc/crates/rpc-types-beacon/src/payload.rs#L275-L298

Then, these modules should be used in structs that have problematic types as fields. For example, for the `Block` type we will need to use a `#[serde_as]` on the `Block` struct and `#[serde_as(as = "HeaderBincode")]` on the `header` field https://github.com/alloy-rs/alloy/blob/cf9a98fd7ed41f1e86af1d58d3eba9bc1c9799b0/crates/consensus/src/block.rs#L15-L18

Both of the attributes should be [gated](https://docs.rs/serde_with/latest/x86_64-pc-windows-msvc/serde_with/guide/serde_as/index.html#gating-serde_as-on-features) by the `serde-bincode` feature.

The types that need special handling are:
- `Header`
- `TxEip2930`
- `TxEip1559`
- `TxLegacy`

As a test, we should be able to serialize and deserialize the `Block` struct using `bincode`.

PR #1397, opened 2024-09-28
feat(consensus): bincode compatibility for header and transaction types

shekhirin:
Closes https://github.com/alloy-rs/alloy/issues/1349

Introduces the `serde-bincode-compat` feature that provides bincode-compatible implementations of serde serialization/deserialization to use with `serde_as` macro.

Example usage https://github.com/alloy-rs/alloy/blob/af593d2e725047aeccf79328f978faac8805a76c/crates/consensus/src/header.rs#L835-L846


crates/consensus/Cargo.toml
```diff
@@ -64,3 +67,4 @@ serde = [
     "dep:alloy-serde",
     "alloy-eips/serde",
 ]
+serde-bincode-compat = ["serde_with"]
```
shekhirin, line 70: open to renaming this feature to something more generic than bincode, because it also unlocks messagepack via https://github.com/3Hren/msgpack-rust

crates/consensus/src/header.rs
```diff
@@ -820,3 +820,168 @@ mod tests {
...
+        ommers_hash: B256,
+        beneficiary: Address,
+        state_root: B256,
+        transactions_root: B256,
+        receipts_root: B256,
+        #[serde(default)]
+        withdrawals_root: Option<B256>,
+        logs_bloom: Bloom,
+        difficulty: U256,
+        #[serde(with = "alloy_serde::quantity")]
```
shekhirin, line 861: hmm actually think these can be removed as well, since we're not in the RPC land anymore

mattsse (commented):
why is this required for all the tx types?

shekhirin:
> why is this required for all the tx types?

they have a `skip_serializing_if` attribute, and we use tx types in the block body in Reth

mattsse (approved):

PR #1406, opened 2024-10-01
ci: do not limit feature powerset check by depth

shekhirin:
## Motivation

Help us to catch errors like https://github.com/alloy-rs/alloy/commit/9eb1409dfd4d4af5f87aa3cef73f2bdb7a99208b

## Solution

Check full feature powerset without limiting by depth for `alloy-consensus`, `alloy-eips` and `alloy-genesis` crates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,7 +138,7 @@ jobs:
         with:
           cache-on-failure: true
       - name: cargo hack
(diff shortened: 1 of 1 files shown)


DaniPopes:
That's not even gonna compute because it goes OOM lol

DaniPopes:
Depth 2 is ok but could take too long

PR #1411, opened 2024-10-02
chore(consensus): blanket impl of `BlockHeader` for deref

emhane:
## Motivation

Simplifies plugging in a new type that derefs to an existing type that impl trait `BlockHeader`

## Solution

Add blanket impl of `BlockHeader` for all types that deref to a type that impl `BlockHeader`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Rjected (commented):
is this behavior not covered by deref coercion?

emhane:
> is this behavior not covered by deref coercion?

when you call a trait method yes, but not when you use it as param or function return type, if I'm not mistaken

PR #1412, opened 2024-10-02
chore(consensus): blanket impl of `Transaction` for deref

emhane:
## Motivation

Simplifies plugging in a new type that derefs to an existing type that impl trait `Transaction`

## Solution

Add blanket impl of `Transaction` for all types that deref to a type that impl `Transaction`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
hmm, this could be problematic because then you can't have both deref for the wrapper inner tx and a modified transaction trait impl

imo we should leave this up to the user

PR #1414, opened 2024-10-02
chore(consensus): move `Header::size` into trait `BlockHeader`

emhane:
## Motivation

`Header::size` must be accessible via abstraction `BlockHeader` in reth sdk

## Solution

Move `Header::size` into trait method of `BlockHeader`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/header.rs
+++ b/crates/consensus/src/header.rs
@@ -1,4 +1,6 @@
 use alloc::vec::Vec;
+use core::mem;
+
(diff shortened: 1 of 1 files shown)


mattsse (commented):
hmm, this feels a bit odd to make it part of the blockheader trait because this is something entirely unrelated.

PR #1442, opened 2024-10-07
Add default to payload id

cody-wang-cb:
## Motivation

Add `Default` to payload id derive so `EthPayloadBuilderAttributes` can also be derived as well.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
seems reasonable

PR #1443, opened 2024-10-09
fix: proposer_index rustdoc

liamaharon:
`proposer_index` is not "The execution block number of the parent block".


DaniPopes (approved):

mattsse (approved):

mattsse:
failures unrelated

PR #1451, opened 2024-10-11
fix: remove signature assoc type from tx response trait

prestwich:
## Motivation

#1328 broke network abstraction by adding a signature assoc type to tx response. Some networks have multiple (or missing like in Optimism) signature types. Adding an assoc type here would require adding a signature abstraction enum on a per-network basis (e.g. optimism would require `enum OpSigs { Ecdsa(Signature), Deposit }`

In general, not all transactions on a network will have the same signature type, so mandating a specific associated type prevents accurately representing Optimism. The current `fn signature(&self) -> Option<Self::Signature>` (which handles the OP representation effectively), doesn't accurately model Ethereum in which all txns must have a signature.

## Solution

Remove the signature associated type, which is unused within the library

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


klkvr (commented):
agreed that `Option` here is not great 

going from `TransactionResponse` to signature would be trivial once we enforce `AsRef<TxEnvelope>` in https://github.com/alloy-rs/alloy/pull/1460, so this should be fine

cc @emhane wondering if we're already using this in reth?

emhane (approved):
should be fine to get the signature via new reth trait `SignedTransaction::Signature`. Op needs a specific transaction to remove optimism feature gated code, see https://github.com/paradigmxyz/reth/pull/11573

mattsse (approved):
makes sense, this was perhaps premature

PR #1455, opened 2024-10-12
chore: use pending for next initial nonce

mattsse:
imo we should use pending here so that the this doesn't pick an already pending nonce in the pool.

@klkvr not exactly sure how this impacts things like

https://github.com/foundry-rs/foundry/pull/9096


klkvr (approved):
lgtm

it shouldn't affect foundry as we don't use nonce filler there

DaniPopes:
What about `CachedNonceManager`?

PR #1457, opened 2024-10-12
feat: add missing eth_getTransaction methods

### LINKED ISSUE #1450: [Feature] add support for eth_getTransactionByBlockNumberAndIndex
### Component

provider, pubsub

needs new method, currently missing

### Additional context

_No response_

mattsse:
closes #1450

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -534,7 +534,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         self.client().request("eth_getTransactionByHash", (hash,)).into()
     }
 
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -534,6 +534,50 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         self.client().request("eth_getTransactionByHash", (hash,)).into()
     }
 
+    /// Returns the transaction by block hash and transaction index position.
```
onbjerg, line 537: nit: change the docs to be in line with the other ones, e.g. "Get a transaction by a block number and its position in the block"

onbjerg (changes requested):

PR #1458, opened 2024-10-13
fix: Make RecommendedFillers generic over Network

popzxc:
Makes `RecommendedFillers` generic over `Network`.

## Motivation

Associated type in the trait requires `TxFiller`, which is generic over `Network`; it just has `Ethereum` as default.
So it's impossible to implement custom fillers for a specific network, if they can't be implemented for `Ethereum` as well.

The solution is a bit awkward since now you can implement `RecommendedFillers<NetworkA>` and `RecommendedFillers<NetworkB>` for `NetworkA`, but the pro is that it's not a breaking change.
The alternative would be to introduce an associated type, but it would be breaking, since associated type defaults are unstable.

## Solution

Make `RecommendedFillers` generic over `Network` 😅 .

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -310,9 +310,9 @@ where
 }
 
 /// A trait which may be used to configure default fillers for [Network] implementations.
-pub trait RecommendedFillers {
+pub trait RecommendedFillers<N: Network = Ethereum> {
     /// Recommended fillers for this network.
-    type RecomendedFillers: TxFiller;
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

klkvr:
couldn't we just change the AT bound to `TxFiller<Self>`?

popzxc:
@klkvr indeed! Fixed in https://github.com/alloy-rs/alloy/pull/1466

PR #1459, opened 2024-10-13
fix: Add network generic to RecommendedFillers bound in ProviderBuilder

popzxc:
Follow-up to #1458; I initially didn't notice that `RecommendedFillers` is used as a bound in `ProviderBuilder`.
Without this, `with_recommended_fillers` method is not available for custom networks.


popzxc:
A better approach was proposed, will open a new PR soon.

PR #1460, opened 2024-10-13
Embed TxEnvelope into `rpc-types-eth::Transaction`

### LINKED ISSUE #1165: [Feature] embed consensus transaction types in eth_getTransaction

prestwich:
closes #1165. Alternative smaller approach than #1169

**This is a breaking change**

cc @klkvr 

## Solution

changes:
- Embed `T` in `alloy_rpc_types_eth::Transaction`, default to `TxEnvelope`
- Modify implementation of `TransactionTrait for Transaction<T>`
- delete `rpc_types_eth::Signature` as in #1169 
- delete tests that tested invalid responses, as those are no longer deserialized
- introduce `AnyTxEnvelope`, `AnyTypedTransaction` and `AnyReceiptEnvelope` abstractions
- add `TransactionResponse: AsRef<Self::TxEnvelope>` bound to network assoc type

drive-by
- mark tokio as a non-wasm dependency to silence warnings in `cargo hack --target wasm32-unknown-unknown`

is this correct?
- remove legacy `gasPrice` field from `TxPoolContent`. My belief is that this is now only produced for legacy transactions, so removing it is correct behavior and will not break deserializers

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


prestwich:
> this would break AnyNetwork because rpc_types_eth::Transaction would not be able to deserialize non-ethereum transaction types

`<AnyNetwork as Network>::TxEnvelope` is already set to `TxEnvelope` so this would not break it any worse than it is already broken. It would make the type somewhat more internally consistent too

klkvr:
> > this would break AnyNetwork because rpc_types_eth::Transaction would not be able to deserialize non-ethereum transaction types
> 
> `<AnyNetwork as Network>::TxEnvelope` is already set to `TxEnvelope` so this would not break it any worse than it is already broken. It would make the type somewhat more internally consistent too

`TransactionResponse` is still an rpc type. changing transaction respose type to an envelope would break eg `get_block(..., true)` requests deserialization on OP because of deposit transaction

prestwich:
> TransactionResponse is still an rpc type. changing transaction respose type to an envelope would break eg get_block(..., true) requests deserialization on OP because of deposit transaction

yes, we agree that this is a breaking change. however, it is not going to break the `AnyNetwork` type, as it is already broken

a more-complete fix might look like this:

```
#[serde(untagged)]
pub enum AnyTxEnvelope { 
    Known(TxEnvelope), 
    Unknown { type: u8, fields: HashMap<String, Value> } 
}
```

klkvr:
> current CI error seems to indicate that anvil sometimes doesn't return a `type` flag?
> 
> ```
> "{\"hash\":\"0x4b56f1a6bdceb76d1b843e978c70ab88e38aa19f1a67be851b10ce4eec65b7d4\",\"nonce\":\"0x0\",\"blockHash\":\"0xc3d639df51a3a7af3c0587b16d4a59c66bb8c2724d288fca3d3d0b3a505689e0\",\"blockNumber\":\"0x1\",\"transactionIndex\":\"0x0\",\"from\":\"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266\",\"to\":\"0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045\",\"value\":\"0x64\",\"gasPrice\":\"0x4a817c800\",\"gas\":\"0x5208\",\"input\":\"0x\",\"r\":\"0x71575df226771c8a9c18eab02457d65bc02ab1b978c78ec4b291f903305291b6\",\"s\":\"0xd04224f8696f449af09a1851bedcaeae3c5836cedb8ccbe489622a34278b0da\",\"v\":\"0xf4f6\",\"chainId\":\"0x7a69\"}"
> ```

yep, anvil does not return type flag for legacy transactions atm, I think we should fix this though I believe this should be handled during deserialization as well

klkvr:
> ```
> #[serde(untagged)]
> pub enum AnyTxEnvelope { 
>     Known(TxEnvelope), 
>     Unknown { type: u8, fields: HashMap<String, Value> } 
> }
> ```

this won't work because of rlp bounds on envelope

klkvr:
> yes, we agree that this is a breaking change. however, it is not going to break the AnyNetwork type, as it is already broken

right now we are using `AnyNetwork` in foundry to fetch and display/re-execute arbitrary transactions through `AnyNetwork`. including this PR as is would break this

klkvr:
it would be great if we'd figure out a way to have `TransactionResponse: Into<Self::TxEnvelope>` bound on `Network` but this is just not possible if we want to have a catch-all network implementation (which `AnyNetwork` is supposed to be)

prestwich:
> this won't work because of rlp bounds on envelope

Envelope doesn't have RLP bounds, only 2718 bounds, which can be satisfied with a dummy implementation. It's impossible to correctly model unknown 2718 encodings so incorrect behavior here should be acceptable

prestwich:
latest commit includes `AnyTxEnvelope` and `AnyTypedTransaction`

prestwich:
okay @klkvr this should be ready for re-review with above concerns addressed. I also put a summary of changes in the PR description

> this would break AnyNetwork because rpc_types_eth::Transaction would not be able to deserialize non-ethereum transaction types

now handled by the new `Any_____` types

> yep, anvil does not return type flag for legacy transactions atm, I think we should fix this though I believe this should be handled during deserialization as well

handled by `MaybeTagged` deser

> this won't work because of rlp bounds on envelope

handled by the dummy 2718 impls

>  it would be great if we'd figure out a way to have TransactionResponse: Into<Self::TxEnvelope> bound on Network but this is just not possible if we want to have a catch-all network implementation (which AnyNetwork is supposed to be)

I added this as `AsRef<Self::TxEnvelope>`. An `Into` bound is blocked by `WithOtherFields`, as we can't generically `impl<T, U> Into<U> for WithOtherFields where T: Into<U>` because it conflicts with std lib impls

klkvr:
nice! thanks for spending your time on this

> handled by the dummy 2718 impls

this is kind of hacky but I don't think there's a way to fit `AnyNetwork` into current abstraction without this hack or another one. wdyt @mattsse @emhane

crates/network/src/any/mod.rs
```diff
@@ -46,6 +59,480 @@ impl From<TxType> for AnyTxType {
...
+        #[serde(rename = "type")]
+        ty: AnyTxType,
+        /// Additional fields.
+        fields: std::collections::BTreeMap<String, serde_json::Value>,
+
+        /// Memoization for deserialization.
+        #[serde(skip, default)]
+        memo: DeserMemo,
+    },
+}
```
klkvr, line 91: I've considered such catch-all envelope variant but imo it has a couple issues
1. `fields` would consume all extra fields, leaving nothing for outer `other`. that way, if network includes some additional fields (like unusual block context or something else) for all transactions, users will need to look for them in _either_ inner `Other` variant or in outer `WithOtherFields` which is incosistent.
2. Such envelope would make it harder to access common transaction fields (like `to`, `data`, `chain_id`, etc).

In https://github.com/alloy-rs/alloy/pull/1169 I've added a `AnyTxEnvelope` type duplicating current `rpc-types-eth::Transaction` to solve those concerns. It does not have any catch-all fields and provides relatively easy access to all common fields without need to manually deserialize them. wdyt on doing something similar here?

prestwich: > fields would consume all extra fields, leaving nothing for outer other. that way, if network includes some additional fields (like unusual block context or something else) for all transactions, users will need to look for them in either inner Other variant or in outer WithOtherFields which is incosistent.

This is an unavoidable problem of having a type be a catch-all. It's an inherent problem for a serializable `Any______`. We can never distinguish between transaction data and non-transaction data in a flattened serialization, and yes it sucks, but absorbing metadata in the inner scope is better than absorbing tx data in the outer scope

> Such envelope would make it harder to access common transaction fields (like to, data, chain_id, etc).

see the implementation of the `Transaction` trait 🫡 

> In https://github.com/alloy-rs/alloy/pull/1169 I've added a AnyTxEnvelope type duplicating current rpc-types-eth::Transaction to solve those concerns. It does not have any catch-all fields and provides relatively easy access to all common fields without need to manually deserialize them. wdyt on doing something similar here?

I think it's less consistent than the style in this PR. It also doesn't capture all conceivable transaction types, e.g. ones with non-secp256k1 signatures. and a signature field of some non-`Signature` type will cause deser errors. Supporting networks with other and mixed signature types has always been a goal (actually can you review #1451 too 🙏 ). the tx envelope in #1169 supports Optimism as the trivial example (signatures may be `Signature` or `None`), but would not support hypothetical ed25519 transaction types (which is actually the motivating use case for this PR and for #1451. We implemented ed25519 as a tx type and want to try it out). You could also envision a network that change the semantics of `to` to allow named system contracts in addition to addresses and null. This would also be broken by the #1169 version

long paragraph short, I think it doesn't quite meet the goals of the network abstraction because it would conflict with transactions that change the type of transaction fields while keeping the name the same

prestwich:
also, agree the 2718 is pretty jank >.> i thought about making it panic instead

klkvr:
> also, agree the 2718 is pretty jank >.> i thought about making it panic instead

something we could do is divide `Network` into 2 traits — one which would just contain the ATs we need for provider without much bounds (basically we just need serde for responses and 2718 for tx envelope). could call it `ProviderNetwork`

and second trait which would represent a normal network with infallible rpc<->consensus conversions which would have a blanket impl for all `ProviderNetwork` implementations satisfying stronger bounds

however doing all of this just for `AnyNetwork` doesn't make much sense

prestwich:
> something we could do is divide `Network` into 2 traits — one which would just contain the ATs we need for provider without much bounds (basically we just need serde for responses and 2718 for tx envelope). could call it `ProviderNetwork`
> 
> and second trait which would represent a normal network with infallible rpc<->consensus conversions which would have a blanket impl for all `ProviderNetwork` implementations satisfying stronger bounds. `AnyNetwork` would just not implement it
> 
> however doing all of this just for `AnyNetwork` doesn't make much sense

I like the idea generally, but i think you're right that doing it just for `AnyNetwork` doesn't make much sense. It would introduce a lot of mental overhead for bounds-writing, which is common, while using `AnyNetwork` is somewhat uncommon

prestwich:
updated the docs for `AnyNetwork` to reflect some known rough edges :)

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -31,382 +28,224 @@ pub use receipt::AnyTransactionReceipt;
...
-        )
-    )]
-    #[doc(alias = "tx_type")]
-    pub transaction_type: Option<u8>,
-    /// The signed authorization list is a list of tuples that store the address to code which the
-    /// signer desires to execute in the context of their EOA and their signature.
-    #[cfg_attr(feature = "serde", serde(default, skip_serializing_if = "Option::is_none"))]
-    pub authorization_list: Option<Vec<SignedAuthorization>>,
+    // /// Transaction hash
+    // pub hash: B256,
```
prestwich, line 61: This is the last outstanding issue. For some reason, uncommenting this makes inner deserialization fail. Probably a subtle and annoying serde interaction 😮‍💨

prestwich: ah, it's because 
- `Signed` has an unskipped `hash` field
- the outer `Transaction` consumes the `"hash"` in deserialization
- there is no "hash" field left for `Signed` to deserialize

prestwich: so the problem becomes
- let `Signed` deserialize the hash
- make `hash` available on `Transaction` in a generic way

prestwich: OR 

- manually impl deserialize for Transaction to avoid consuming the hash

this seems easier and doesn't require introducing new bounds 👍

klkvr: should we just remove the `hash` from Transaction?

prestwich: that breaks the `TransactionResponse` trait. which would (e.g.) make it so that we can't index txns by hash without knowing the structure of the internal transaction. Which would suck

prestwich: recap:

the three options here:
1. remove hash from `Transaction` and `TransactionResponse`
2. Introduce a new trait bound in `Network` like this: `type TxEnvelope: NewTrait + Encodable2718`. This would expose the `tx_hash` and allow for `impl<T: NewTrait> TransactionResponse for Transaction<T>`
3. Modify the deserialize impl to not consume the `"hash"` key

#3 has no impact on our trait system and downstream users, and is non-breaking. so it seemed the best choice

prestwich: current problem is because `hash` is doubled when serializing 😮‍💨

prestwich: hmmmm serde and serde_json have no way to tell whether a map actually already contains a key while serializing, which means that we'd need to write our own serializer which we do _not_ want to do lol 

I'm now coming around to option 2 🤔

klkvr: > remove hash from Signed<T> is not practical because the method of calculating the hash relies on information about the tx encoding, which is underneath the abstraction. i.e. we don't have a trait that generically exposes the tx_hash calculation. the only way we calculate it is when constructing a Signed

I believe transaction hash is just `keccak256(tx_envelope.encode_2718())`? which should be possible to do for any envelope

prestwich: > I believe transaction hash is just keccak256(tx_envelope.encode_2718())? which should be possible to do for any envelope

will look into this today. at first blush, it looks like `trie_hash` is provided on the `Encodable2718` trait, so we should be able to remove from `Signed`. We still strongly want memoization tho so need to figure out how to keep that without interfering with serialization 🤔 

it also looks like 0c588bfeb31557680aa1bafaf06b7b60d7c9c9af lightly broke eip2718 abstraction by making type flags tx associated instead of envelope associated. We would now behave badly for networks where the type flag is different (e.g. `TxEip1559` is flag `17`, or where both `4` and `5` are `TxEip7702`, so we should probably open a separate issue for that

prestwich: > I believe transaction hash is just keccak256(tx_envelope.encode_2718())? which should be possible to do for any envelope

this is actually not the case for eip4844 with sidecar 😮‍💨  but it's straightforward to work around i think

klkvr: I'm wondering if we could solve this by adding a `TxEnvelope` trait which would have a `tx_hash` method? all normal variants already keep `hash` in `Signed` and we can just add a mandatory `hash` field to `Other` variant

I think https://github.com/alloy-rs/alloy/pull/1485 is reasonable as well because sealable and signable are 2 valid distinct properties of transaction (eg deposit transaction is only sealable). but would like to keep scope here smaller if possible as PR is already pretty big

prestwich: > I'm wondering if we could solve this by adding a TxEnvelope trait which would have a tx_hash method? all normal variants already keep hash in Signed and we can just add a mandatory hash field to Other variant

the `trie_hash` method already exists on `Encodable2718`, so we are covered there :)

prestwich:
> but would like to keep scope here smaller if possible as PR is already pretty big

yes. I'd like to merge #1485 and then see how much of this i can trim while still keeping functionality right

crates/rpc-types-eth/src/transaction/mod.rs
klkvr: > the trie_hash method already exists on Encodable2718, so we are covered there :)

yeah but our impl of `Encodable2718` is not correct for `Other` variant?

crates/rpc-types-txpool/src/txpool.rs
```diff
@@ -204,7 +204,6 @@ mod tests {
         "blockNumber": null,
         "from": "0x00000000863b56a3c1f0f1be8bc4f8b7bd78f57a",
         "gas": "0x2af9e",
-        "gasPrice": "0x218711a00",
```
prestwich, line 207: need someone to confirm if this is correct. I believe these can be deleted as returning this field for eip1559 txns is deprecated

klkvr: yep, it should be fine to remove non-legacy

crates/network/src/any/mod.rs
```diff
@@ -46,11 +59,515 @@ impl From<TxType> for AnyTxType {
     }
 }
 
+/// Memoization for deserialization of [`AnyTxEnvelope`] and [`AnyTypedTransaction`].
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+#[allow(unnameable_types)]
+pub struct DeserMemo {
```
klkvr, line 65: I think instead of memoization we could just add `AnyTx` type which would just deserialize all of those if present?

crates/network/src/any/mod.rs
```diff
@@ -46,11 +59,515 @@ impl From<TxType> for AnyTxType {
...
+            Self::Ethereum(t) => t.encode_2718_len(),
+            Self::Other { .. } => 1,
+        }
+    }
+
+    fn encode_2718(&self, out: &mut dyn alloy_primitives::bytes::BufMut) {
+        match self {
+            Self::Ethereum(t) => t.encode_2718(out),
+            Self::Other { ty, .. } => {
+                out.put_u8(ty.into());
```
klkvr, line 112: I think a panic might be more appropriate here. this shouldn't be something users would hit accidentally, and a panic would give more context than a mismatched tx hash or rlp encoding

wdyt @mattsse @prestwich

prestwich: i'm concerned that this would turn common user actions like `Provider::send_raw_transaction` into panics. Users are not generally very aware of how transaction construction and serialization works (by design), so unless we can head off calling `send_raw_transaction` in the `AnyNetwork` entirely, I'm wary of making this panic

prestwich: retrieve -> modify -> rebroadcast is a reasonably common user behavior. this could also happen for e.g. searcher bundle, where the txns must be 2718 encoded before being added to a bundle. It is pretty reasonable to expect a searcher to attempt to retrieve a tx from RPC and then re-encode that tx. I'm not sure if it's easier to diagnose as a panic or as an RPC error at that point tho

klkvr: > this could also happen for e.g. searcher bundle, where the txns must be 2718 encoded before being added to a bundle

I'd expect such users to understand the rlp semantics and use a safe network implementation. also not sure whether there are any networks with both custom transaction types and bundle submision support?

> I'm not sure if it's easier to diagnose as a panic or as an RPC error at that point tho

I would say that debugging panic would be much easier than a decoding error from RPC. also imo both scenarios are showstoppers for such projects, so it feels more correct to detect on our side

prestwich: i don't have a strong opinion here. should i change this to a panic with an informative message?

mattsse: we could chose to not do anything which would definitely lead to issues somewhere else
but I think a panic would be more helpful for now and we can consider not writing anything at all if users somehow actually reach this

prestwich: this now panics 👍

crates/network/src/any/mod.rs
prestwich: i feel like lazy deserialization is more appropriate here, as we are making absolutely no assumptions about the contents of the tx other than that it is some sort of JSON

klkvr: we are making assumptions on both fields name and types they deserialize into in `Transaction` trait implementation, and overhead of always trying to deserialize it wouldn't be that much

imo users of `AnyNetwork` are very unlikely to use it for networks with transaction types which are significantly different from Etheruem transactions and thus require a different deserialization methods. instead, they mostly don't care about the consensus semantics and just want to capture common transaction fields (from,to,data,gas) which are preset for all transactions. and providing easy access to those fields by always trying deserialize them into struct fields would be handy imo

I've took similar approach when embedding header recently in #1573 as it wouldn't make sense to force users deal with raw serde_json values just to get header fields of networks which are missing `mixHash` in it

prestwich: > we are making assumptions on both fields names and types they deserialize into in Transaction trait implementation, and overhead of always trying to deserialize it wouldn't be that much

not exactly. The API in this PR makes no assumptions that the key exists or that its contents are what we expect. If we eagerly deserialize, we have to make some form of those assumptions. Lazy deserialization means we can continue to not assume that, and to allow users to (e.g.) modify the format of an authorization list because they're on the older EIP-7702 specification. If we deserialize eagerly, we end up either consuming that key (and the user is no longer allowed to modify that format at all) or going through a ton of complexity implementing `Deserialize` to preserve the user's access to that key. We want to avoid the case where we render some transaction contents inaccessible due to eager deser attempts

> I've took similar approach when embedding header recently in https://github.com/alloy-rs/alloy/pull/1573 as it wouldn't make sense to force users deal with raw serde_json values just to get header fields of networks which are missing mixHash in it

the approach here with memoization is the same as your approach, except that deserialization is done lazily instead of eagerly. To avoid having users mess around with `serde_json::Value` would it make sense to you to add a convenience function like this:

```
fn deser_key<T>(key: &str) -> Option<T> { ... }
```

klkvr: > the approach here with memoization is the same as your approach, except that deserialization is done lazily instead of eagerly. To avoid having users mess around with `serde_json::Value` would it make sense to you to add a convenience function like this:
> 
> ```
> fn deser_key<T>(key: &str) -> Option<T> { ... }
> ```

I would be fine with that as long as we still have common transaction fields present. imo it would make sense to attempt deserialization of fields of transactions up to 1559 as those are well defined, commonly used and unlikely to change semantics

prestwich: we have deserialization fo fields up to 1559 as `AnyTxEnvelope` accepts the Ethereum tx format. If users are using a custom tx type, we cover access to common fields via the transaction trait

prestwich: `deser_by_key` added in 51eb19a

klkvr: I don't think we have a strong reason to force users to import a trait just to access those fields

crates/rpc-types-eth/src/block.rs
```diff
@@ -40,6 +40,18 @@ pub struct Block<T = Transaction, H = Header> {
     pub withdrawals: Option<Withdrawals>,
 }
 
+// cannot derive, as the derive impl would constrain `where T: Default`
+impl Default for Block<Transaction, Header> {
```
klkvr, line 44:
```suggestion
impl<T, H: Default> Default for Block<T, H> {
```

crates/network/src/any/mod.rs
```diff
@@ -46,11 +59,537 @@ impl From<TxType> for AnyTxType {
...
+    }
+
+    fn trie_hash(&self) -> B256 {
+        match self {
+            Self::Ethereum(tx) => tx.trie_hash(),
+            Self::Other { fields, .. } => fields
+                .get("hash")
+                .and_then(|v| v.as_str())
+                .and_then(|v| v.parse().ok())
+                .unwrap_or_else(B256::default),
```
klkvr, line 147: let's require `Other` variant to deserialize `hash: B256`

klkvr (changes requested):
overall lgtm, just a couple comments

we don't have any more deserialization issues, right?

crates/network/src/any/mod.rs
prestwich: in 0ba14955

crates/rpc-types-eth/src/block.rs
prestwich: in 0ba14955

prestwich:
> we don't have any more deserialization issues, right?

Not that I know of

This is rebased and ready for review

prestwich:
side note, because the type aliases for `AnyRpcBlock` have to reference `AnyTxEnvelope` they've been pushed into the `alloy-network` crate and re-exported from the `any` module

crates/network/src/any/mod.rs
```diff
@@ -1,14 +1,30 @@
...
 
+pub use alloy_consensus::{AnyHeader, AnyReceiptEnvelope};
+
+/// A catch-all header type for handling headers on multiple networks.
+pub type AnyRpcHeader = alloy_rpc_types_eth::Header<alloy_consensus::AnyHeader>;
+
+/// A catch-all block type for handling blocks on multiple networks.
+pub type AnyRpcBlock =
+    WithOtherFields<Block<WithOtherFields<Transaction<AnyTxEnvelope>>, AnyRpcHeader>>;
+
```
klkvr, line 25: we already have these here https://github.com/alloy-rs/alloy/blob/273d7849790106bb17f134a56d9725e6e3944f50/crates/rpc-types-eth/src/lib.rs#L26

prestwich: https://github.com/alloy-rs/alloy/pull/1460#issuecomment-2449771075

Those can't be in the `rpc-types-eth` crate as they need to import `AnyTxEnvelope` now. In addition, it's somewhat unclean to put `AnyRpc____` objects into `rpc-types-eth`. Do you thinkwe should add `rpc-types-any` as a crate?

klkvr:
> side note, because the type aliases for `AnyRpcBlock` have to reference `AnyTxEnvelope` they've been pushed into the `alloy-network` crate and re-exported from the `any` module. I also re-exported `AnyHeader` and `AnyReceipt` there, so all `Any____` can be accessed from the same module

should we move `AnyTxEnvelope` definition to alloy-consensus then? as it is done for AnyReceiptEnvelope and AnyHeader

crates/network/src/any/mod.rs
klkvr: AnyRecipt and AnyHeader are already in alloy-consensus

can't AnyTxEnvelope be there too?

prestwich: i don't think it was a good idea to put them there in the first place. `alloy-consensus` has always been specifically `eth` consensus. It's also much more invasive to move a large type downwards instead of moving 2 aliases upwards

prestwich: followup work:
https://github.com/alloy-rs/alloy/issues/1598

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+        /// Transaction hash.
+        hash: B256,
+
+        /// Additional fields.
+        fields: std::collections::BTreeMap<String, serde_json::Value>,
+
+        /// Memoization for deserialization.
+        #[serde(skip, default)]
+        memo: DeserMemo,
+    },
```
mattsse, line 103: imo this should be it's own Standalone type, dealing with a lot of named fields is not that convenient

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+    Other {
+        /// Transaction type.
+        #[serde(rename = "type")]
+        ty: AnyTxType,
+
+        /// Transaction hash.
+        hash: B256,
+
+        /// Additional fields.
+        fields: std::collections::BTreeMap<String, serde_json::Value>,
```
mattsse, line 98: without flatten this doesn't really work I think

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+            Self::Ethereum(inner) => inner.chain_id(),
+            Self::Other { fields, .. } => fields.get("chainId").and_then(|v| v.as_u64()),
+        }
+    }
+
+    fn nonce(&self) -> u64 {
+        match self {
+            Self::Ethereum(inner) => inner.nonce(),
+            Self::Other { fields, .. } => {
+                fields.get("nonce").and_then(|v| v.as_u64()).expect("missing nonce in tx response")
```
mattsse, line 183: this will panic for types that don't have a nonce.
should just return a default here

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+    fn typed_decode(ty: u8, buf: &mut &[u8]) -> alloy_eips::eip2718::Eip2718Result<Self> {
+        TxEnvelope::typed_decode(ty, buf).map(Self::Ethereum)
+    }
+
+    fn fallback_decode(buf: &mut &[u8]) -> alloy_eips::eip2718::Eip2718Result<Self> {
+        TxEnvelope::fallback_decode(buf).map(Self::Ethereum)
+    }
+}
+
+impl alloy_consensus::Transaction for AnyTxEnvelope {
```
mattsse, line 171: do we need this to satisfy the the `Network` trait constraint

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+    }
+}
+
+impl From<AnyTxEnvelope> for WithOtherFields<TransactionRequest> {
+    fn from(value: AnyTxEnvelope) -> Self {
+        AnyTypedTransaction::from(value).into()
+    }
+}
+
+impl alloy_consensus::Transaction for AnyTypedTransaction {
```
mattsse, line 411: this duplicates a lot of things that we also have on the anyevelope.
we could simplify this if we convert type + fields + memo into it's own type.

the it also becomes possible to impl TryFrom conversions, e.g. for OpDeposit tx

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+///   non-Ethereum transaction types. It will encode the type for any unknown transaction type, but
+///   will not encode any other fields. This is symmetric with the decoding behavior, but still
+///   erroneous.
+/// - The [`TransactionRequest`] will build ONLY Ethereum types. It will error when attempting to
+///   build any unknown type.
+/// - The [`Network::TransactionResponse`] may deserialize unknown metadata fields into the inner
+///   [`AnyTxEnvelope`], rather than into the outer [`WithOtherFields`].
 #[derive(Clone, Copy, Debug)]
 pub struct AnyNetwork {
     _private: (),
```
mattsse, line 604: I'd prefer to have this top of file

crates/network/src/any/mod.rs
```diff
@@ -46,11 +68,537 @@ impl From<TxType> for AnyTxType {
...
+            Self::Other { fields, .. } => {
+                fields.get("maxFeePerBlobGas").and_then(|v| v.as_u64()).map(|v| v as u128)
+            }
+        }
+    }
+
+    fn priority_fee_or_price(&self) -> u128 {
+        self.max_priority_fee_per_gas()
+            .or_else(|| self.gas_price())
+            .expect("missing maxPriorityFeePerGas or gasPrice in tx response")
```
mattsse, line 478: panicking on missing fee fields also feels wrong,
e.g. this would also panic on deposittxs

mattsse (changes requested):

crates/network/src/any/mod.rs
mattsse: ref https://github.com/alloy-rs/op-alloy/blob/b60f6430a9644d93702aba9de1c783366e1aac84/crates/consensus/src/transaction/deposit.rs#L184-L184

crates/network/src/any/mod.rs
prestwich: fixed

crates/network/src/any/mod.rs
prestwich: we don't, but it seems sufficiently useful to users to have this

crates/network/src/any/mod.rs
prestwich: done

crates/network/src/any/mod.rs
prestwich: done

crates/network/src/any/mod.rs
prestwich: went with breaking out files to cut down on clutter

crates/network/src/any/mod.rs
prestwich: no more expects at all 👍

crates/network/Cargo.toml
```diff
@@ -32,6 +32,8 @@ alloy-sol-types.workspace = true
 auto_impl.workspace = true
 async-trait.workspace = true
 futures-utils-wasm.workspace = true
+serde.workspace = true
+serde_json = "1.0.128"
```
mattsse, line 36:
```suggestion
serde_json.workspace = true
```

crates/network/src/any/mod.rs
```diff
@@ -1,56 +1,52 @@
...
+use crate::Network;
+use alloy_rpc_types_eth::{AnyTransactionReceipt, Block, Transaction, TransactionRequest};
+use alloy_serde::WithOtherFields;
 
-    fn try_from(value: AnyTxType) -> Result<Self, Self::Error> {
-        value.0.try_into()
-    }
-}
+/// A catch-all header type for handling headers on multiple networks.
+pub type AnyRpcHeader = alloy_rpc_types_eth::Header<alloy_consensus::AnyHeader>;
```
mattsse, line 16: moving these aliases here makes sense imo

mattsse (approved):
cool, I think this is a pretty big UX improvement.

pending @klkvr

klkvr (approved):
lgtm! will try to plug this into reth and see if anything breaks

PR #1462, opened 2024-10-13
feat(eip4895): implement `Withdrawals`

tcoratger:
In order to be able to centralize withdrawal definitions in alloy and remove this implementation https://github.com/paradigmxyz/reth/blob/661b260f6172c047e8530e2331b2c84141e03c2b/crates/primitives-traits/src/withdrawal.rs#L35 from reth, we need to properly have the `Withdrawals` struct in alloy. 

We can improve it as needed in the future.


tcoratger:
@mattsse nice bump should be ready to review/merge

crates/eips/src/eip4895.rs
```diff
@@ -37,6 +38,63 @@ impl Withdrawal {
     }
 }
 
+/// Represents a collection of Withdrawals.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, RlpEncodable, RlpDecodable)]
```
mattsse, line 42: this doesn't match what we have here:

https://github.com/paradigmxyz/reth/blob/ff90079f9186a1e4e4f755a363e51b92f0b0e9ad/crates/primitives-traits/src/withdrawal.rs#L26-L27

tcoratger: Fixed thanks for pointing out

mattsse (approved):

PR #1463, opened 2024-10-13
chore: refactor some match with same arms

tcoratger:


tcoratger:
@mattsse nice bump should be ready to review/merge :)

mattsse (approved):

PR #1473, opened 2024-10-14
fix(types/filter): treat empty filter address as non-matching 

### LINKED ISSUE #1470: [Feature] Treat empty address filters as non matching if both are not empty
### Component

rpc

ref https://github.com/paradigmxyz/reth/issues/11711

this will always be true if empty:

https://github.com/alloy-rs/alloy/blob/440ddcb872a47105ec18e470c8750258d863a610/crates/rpc-types-trace/src/filter.rs#L170-L170

which wasn't the case pre https://github.com/alloy-rs/alloy/commit/338b5b94daddbf8a5bedff3dfc5f4f5c4fcdf177

and is also not how https://github.com/erigontech/erigon/blob/72ab70be8b28964a8446e52b7a9975958c6726bf/turbo/jsonrpc/trace_filtering.go#L238-L289 behaves

### Additional context

_No response_

jsvisa:
## Motivation

fix #1470 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/filter.rs
+++ b/crates/rpc-types-trace/src/filter.rs
@@ -118,11 +118,11 @@ impl From<Vec<Address>> for AddressFilter {
 impl AddressFilter {
     /// Returns `true` if the given address is in the filter or the filter address set is empty.
     pub fn matches(&self, addr: &Address) -> bool {
-        self.matches_all() || self.0.contains(addr)
+        self.is_empty() || self.0.contains(addr)
     }
 
     /// Returns `true` if the address set is empty.
-    pub fn matches_all(&self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.0.is_empty()
     }
 }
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/filter.rs
```diff
@@ -175,19 +177,27 @@ impl TraceFilterMatcher {
...
-            TraceFilterMode::Union => from_matches || to_matches,
+            TraceFilterMode::Union => {
+                if self.from_addresses.is_empty() {
+                    to_matches
+                } else if self.to_addresses.is_empty() {
+                    from_matches
+                } else {
+                    from_matches || to_matches
+                }
+            }
```
mattsse, line 200: this looks correct now, ty

mattsse (approved):
awesome, tysm.

pending pedantic clippy

PR #1475, opened 2024-10-14
eips: update fn encoded_2718

nkysg:
use Vec::with_capacity avoid realloc

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -208,7 +208,7 @@ pub trait Encodable2718: Sized + Send + Sync + 'static {
     /// This is a convenience method for encoding into a vec, and returning the
     /// vec.
     fn encoded_2718(&self) -> Vec<u8> {
-        let mut out = vec![];
+        let mut out = Vec::with_capacity(self.encode_2718_len());
         self.encode_2718(&mut out);
         out
     }


crates/eips/src/eip2718.rs
```diff
@@ -208,7 +208,7 @@ pub trait Encodable2718: Sized + Send + Sync + 'static {
     /// This is a convenience method for encoding into a vec, and returning the
     /// vec.
     fn encoded_2718(&self) -> Vec<u8> {
-        let mut out = vec![];
+        let mut out = Vec::with_capacity(self.encode_2718_len());
```
mattsse, line 211: I believe this is problematic because this could end up doing encoding twice @klkvr ?

but idk

klkvr: we already do this in `alloy_rlp::encode`: https://github.com/alloy-rs/rlp/blob/44606e88cb7ca9397a5b5e57a3299d4ae0305e01/crates/rlp/src/encode.rs#L292

and we don't even provide deafult `encode_2718_len` here so I think it's OK to expect implementations to provide a smarter impl

should be fine imo

mattsse (approved):

PR #1477, opened 2024-10-14
chore: unify use Option ref

nkysg:
## Motivation
I see most places use Option<&T>,  I think maybe Option<&T> is better than &Option<T>. If it's not, please close it.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
reasonable

PR #1478, opened 2024-10-14
consensus: add more constraints to `TxReceipt` trait

tcoratger:
Small step toward https://github.com/paradigmxyz/reth/pull/11643#discussion_r1797592811

### DIFF
--- a/crates/consensus/src/receipt/any.rs
+++ b/crates/consensus/src/receipt/any.rs
@@ -2,6 +2,7 @@ use crate::{Eip658Value, ReceiptWithBloom, TxReceipt};
 use alloy_eips::eip2718::{Decodable2718, Eip2718Result, Encodable2718};
 use alloy_primitives::{bytes::BufMut, Bloom, Log};
 use alloy_rlp::{Decodable, Encodable};
+use core::fmt;
 
 /// Receipt envelope, as defined in [EIP-2718].
 ///
@@ -86,7 +87,7 @@ impl<T> AnyReceiptEnvelope<T> {
     }
 }
 
-impl<T> TxReceipt<T> for AnyReceiptEnvelope<T> {
+impl<T: Clone + fmt::Debug + PartialEq + Eq> TxReceipt<T> for AnyReceiptEnvelope<T> {
     fn status_or_post_state(&self) -> Eip658Value {
         self.inner.status_or_post_state()
     }
--- a/crates/consensus/src/receipt/any.rs
+++ b/crates/consensus/src/receipt/any.rs
@@ -87,7 +87,7 @@ impl<T> AnyReceiptEnvelope<T> {
     }
 }
 
-impl<T: Clone + fmt::Debug + PartialEq + Eq> TxReceipt<T> for AnyReceiptEnvelope<T> {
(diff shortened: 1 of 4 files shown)


crates/consensus/src/receipt/receipts.rs
```diff
@@ -69,7 +69,7 @@ where
     }
 }
 
-impl<T> TxReceipt<T> for Receipt<T>
+impl<T: Clone + fmt::Debug + PartialEq + Eq> TxReceipt<T> for Receipt<T>
```
mattsse, line 72: are these now necessary because of the Log type t?

crates/consensus/src/receipt/mod.rs
```diff
@@ -15,7 +16,7 @@ pub use status::Eip658Value;
 /// Receipt is the result of a transaction execution.
 #[doc(alias = "TransactionReceipt")]
 #[auto_impl::auto_impl(&, Arc)]
-pub trait TxReceipt<T = Log> {
+pub trait TxReceipt<T = Log>: Clone + fmt::Debug + PartialEq + Eq {
```
mattsse, line 19: I think those are reasonable

mattsse (approved):
I think these make sense

wdyt @klkvr ?

crates/consensus/src/receipt/receipts.rs
tcoratger: Yes exactly

crates/consensus/src/receipt/mod.rs
emhane: `Send` and `Sync` would also be helpful

emhane (approved):

klkvr (approved):

crates/consensus/src/receipt/any.rs
```diff
@@ -86,7 +87,7 @@ impl<T> AnyReceiptEnvelope<T> {
     }
 }
 
-impl<T> TxReceipt<T> for AnyReceiptEnvelope<T> {
+impl<T: Clone + fmt::Debug + PartialEq + Eq + Send + Sync> TxReceipt<T> for AnyReceiptEnvelope<T> {
```
mattsse, line 90: this shouldn't be squeezed into the `<>` looks bad, needs where clause

crates/consensus/src/receipt/receipts.rs
```diff
@@ -69,7 +69,7 @@ where
     }
 }
 
-impl<T> TxReceipt<T> for Receipt<T>
+impl<T: Clone + fmt::Debug + PartialEq + Eq + Send + Sync> TxReceipt<T> for Receipt<T>
```
mattsse, line 72: same

mattsse (changes requested):
pedantic style nit

crates/consensus/src/receipt/receipts.rs
tcoratger: fixed

crates/consensus/src/receipt/any.rs
tcoratger: fixed

mattsse (approved):

Issue #142, opened 2024-01-22
[Feature] Adding EIP-712 Transaction support


dutterbutter:
### Component

consensus, eips, genesis

## Objective
Enhance Alloy to support EIP-712 transactions, enabling account abstraction / deployment for zkSync users that want to make use of alloy for their projects and tooling. In order to do so, alloy would need to support 712-transactions. Example:

```
pub struct Eip712Transaction {
    pub tx_type: U256,
    pub from: Address,
    pub to: Address,
    pub gas_limit: U256,
    pub gas_per_pubdata_byte_limit: U256,
    pub max_fee_per_gas: U256,
    pub max_priority_fee_per_gas: U256,
    pub paymaster: Address,
    pub nonce: U256,
    pub value: U256,
    pub data: Bytes,
    pub factory_deps: Vec<Bytes>,
    pub paymaster_input: Bytes,
    pub chain_id: U256,
}
```

### Resources

There is an initial implementation in zkSync `rust-web3-rs sdk` of 712-transactions which can be viewed here: [**712-tx**](https://github.com/lambdaclass/zksync-web3-rs/tree/main/src/eip712). It would be great to be able to extend alloy to include this functionality / support. 

- `rust-web3-rs sdk`: https://github.com/lambdaclass/zksync-web3-rs/
- Relevant docs: https://docs.zksync.io/zk-stack/concepts/transaction-lifecycle.html#eip-712-0x71

prestwich:
This sort of functionality should be added as `ProviderExt` traits in separate crates, similar to #84

nbaztec:
@prestwich is there an existing implementation that we can use as an example?

PR #1479, opened 2024-10-14
fix(signer-ledger): use `SIGN_ETH_EIP_712` instruction

Karrq:
## Motivation
Signing EIP712 messages thru `alloy_network::TxSigner`'s `sign_transaction` results in the usage of `SIGN` instruction which does not support EIP712 messages.

## Solution
Detect if the incoming tx is indeed a EIP712 transaction and dispatch to the appropriate method

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
not that familiar with this api

wdyt @DaniPopes

DaniPopes:
There's no such thing as a EIP-712 transaction, nor does any of the existing transaction ever return EIP-712-encoded data, so I don't know what this is trying to accomplish.

DaniPopes:
Nevermind, apparently there is https://github.com/alloy-rs/alloy/issues/142
Please add a comment for this
cc @prestwich

Karrq:
This allows foundry-zksync to sign transactions using ledger, since we make use of `sign_transaction` to sign the [encoded eip712 transacion](https://github.com/matter-labs/foundry-zksync/blob/b513e390bf349481e4009c2906c2a13986d3fa4d/crates/zksync/core/src/lib.rs#L166).

Karrq:
Hey guys, any update on this? Please let me know if the bugfix is not clear.

The main issue that this addresses is the correct dispatch of EIP712 encoded tx, since `TxSigner::sign_transaction` is the only entrypoint available from foundry. 
`alloy-signer-ledger` already supports EIP712 signing using `sign_typed_data`, but this method was just not accessible from `sign_transaction`, which is why the PR was made.

Please let me know if you any concerns so I may address them :)

crates/signer-ledger/src/signer.rs
```diff
@@ -43,7 +46,33 @@ impl alloy_network::TxSigner<Signature> for LedgerSigner {
         &self,
         tx: &mut dyn SignableTransaction<Signature>,
     ) -> Result<Signature> {
-        sign_transaction_with_chain_id!(self, tx, self.sign_tx_rlp(&tx.encoded_for_signing()).await)
+        let encoded = tx.encoded_for_signing();
+
+        match encoded.as_slice() {
+            // Ledger requires passing EIP712 data to a separate instruction
+            #[cfg(feature = "eip712")]
+            [0x19, 0x1, data @ ..] => {
```
mattsse, line 54: I don't this would clash with the other txs, so this seems okay to me.

crates/signer-ledger/src/signer.rs
```diff
@@ -43,7 +46,33 @@ impl alloy_network::TxSigner<Signature> for LedgerSigner {
...
+                let domain_sep = data
+                    .get(..32)
+                    .ok_or_else(|| {
+                        Error::other("eip712 encoded data did not have a domain separator")
```
mattsse, line 58: could you use the full path here so we don't need to feature gate the error import

crates/signer-ledger/src/signer.rs
```diff
@@ -43,7 +46,33 @@ impl alloy_network::TxSigner<Signature> for LedgerSigner {
...
+                    })
+                    .map(B256::from_slice)?;
+
+                let hash = data[32..]
+                    .get(..32)
+                    .ok_or_else(|| Error::other("eip712 encoded data did not have hash struct"))
```
mattsse, line 64: same

crates/signer-ledger/src/signer.rs
Karrq: I think [`ERC-191`](https://github.com/ethereum/EIPs/issues/191) describes the format and mentions that `0x19` is an invalid RLP header so no valid Ethereum txs will ever start with `0x19`, and then the `0x1` is the version specifier for `EIP712`, so yeah shouldn't ever conflict

mattsse (approved):

PR #1480, opened 2024-10-15
Add eip1559 params to executable payload

cody-wang-cb:
## Motivation

Required for Holocene EL changes
https://github.com/ethereum-optimism/specs/pull/428

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -332,6 +332,8 @@ pub struct ExecutionPayloadV3 {
     /// See <https://github.com/ethereum/execution-apis/blob/fe8e13c288c592ec154ce25c534e26cb7ce0530d/src/engine/cancun.md#ExecutionPayloadV3>
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub excess_blob_gas: u64,
+    /// Array of hex[`u64`] representing the EIP-1559 parameters, enabled with V3
+    pub eip_1559_params: B64,
 }
 
 impl ExecutionPayloadV3 {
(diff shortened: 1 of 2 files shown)


mattsse (approved):

crates/rpc-types-engine/src/payload.rs
```diff
@@ -332,6 +332,8 @@ pub struct ExecutionPayloadV3 {
     /// See <https://github.com/ethereum/execution-apis/blob/fe8e13c288c592ec154ce25c534e26cb7ce0530d/src/engine/cancun.md#ExecutionPayloadV3>
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub excess_blob_gas: u64,
+    /// Array of hex[`u64`] representing the EIP-1559 parameters, enabled with V3
+    pub eip_1559_params: B64,
```
mattsse, line 336: how will this be rolled out, will this just be a default value until this is activated?

mattsse, line 336: we can't do this, this is an op specific change.
this would break existing engine API because this field doesn't exist.
looks like we need to roll an op specific v3 type

mattsse:
I believe this should also be optional 

https://github.com/ethereum-optimism/specs/pull/428#discussion_r1800153482

crates/rpc-types-engine/src/payload.rs
cody-wang-cb: Yeah I just realized as well, I should add this in `op-alloy`

mattsse:
closing per https://github.com/ethereum-optimism/specs/pull/428#issuecomment-2418259114

PR #1481, opened 2024-10-15
fix(wasm): support ws

ethernomad:
## Motivation

When attempting to make a browser dapp with Dioxus I found that alloy would not build with WebSocket support:

```
error[E0599]: no function or associated item named `new` found for struct `WsConnect` in the current scope
  --> /home/jbrown/alloy/crates/rpc-client/src/builtin.rs:84:64
   |
84 |             Self::Ws(url, _) => alloy_transport_ws::WsConnect::new(url.clone())
   |                                                                ^^^ function or associated item not found in `WsConnect`
```

It is critical that Alloy supports building browser dapps with WebSocket support.

## Solution

All that is required to fix this is adding missing function new() to WsConnect in `crates/transport-ws/src/wasm.rs`.

With this fix I verified that Alloy could connect to a local node from the browser via WS.

Additionally I updated `scripts/check_no_std.sh` to check affected crates and `ws` feature. The target was changed from `riscv32imac-unknown-none-elf` to `wasm32-unknown-unknown to facilitate this`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/scripts/check_no_std.sh
+++ b/scripts/check_no_std.sh
@@ -1,18 +1,21 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-target=riscv32imac-unknown-none-elf
+target=wasm32-unknown-unknown
 crates=(
     alloy-eips
     alloy-genesis
     alloy-serde
     alloy-consensus
     alloy-network-primitives
+    alloy-rpc-client
     alloy-rpc-types-eth
     alloy-rpc-types-engine
+    alloy-transport
+    alloy-transport-ws
 )
 
-cmd=(cargo +stable hack check --no-default-features --target "$target")
+cmd=(cargo +stable hack check --no-default-features --ignore-unknown-features --features ws --target "$target")
 for crate in "${crates[@]}"; do
     cmd+=(-p "$crate")
 done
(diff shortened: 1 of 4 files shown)


scripts/check_no_std.sh
```diff
@@ -1,18 +1,21 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-target=riscv32imac-unknown-none-elf
+target=wasm32-unknown-unknown
```
mattsse, line 4: wasm32 should be even more restrictive than the riscv target so this makes sense

mattsse (approved):
this seems okay

scripts/check_no_std.sh
DaniPopes: It's not, wasm32 has std

ethernomad: Should there be a separate check_wasm.sh script?

DaniPopes: You can update the workflow `cargo hack` to also check for wasm feature https://github.com/alloy-rs/alloy/blob/f37ac3051c46248542d350c33f8c529fc59e8de3/.github/workflows/ci.yml#L92

ethernomad: Okay I moved the build check to ci.yml

DaniPopes (approved):
Thanks!

PR #1484, opened 2024-10-15
consensus: replace `to` by `kind` in Transaction trait

tcoratger:
`to` naming convention can be confusing because we already have a `to` function in most implementations and this `to` function designates the receiving address while here we are going to get the kind of the transaction.

For example here it could be really confusing with potential double `to().to()` to catch the final address:

https://github.com/paradigmxyz/reth/blob/c4d7b591834055bdf3afed96b696e0a2cef0f383/crates/primitives/src/transaction/mod.rs#L226-L247

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -296,7 +296,7 @@ impl Transaction for TxEip1559 {
         self.max_priority_fee_per_gas
     }
 
-    fn to(&self) -> TxKind {
+    fn kind(&self) -> TxKind {
         self.to
     }
 
--- a/crates/consensus/src/transaction/eip2930.rs
+++ b/crates/consensus/src/transaction/eip2930.rs
@@ -260,7 +260,7 @@ impl Transaction for TxEip2930 {
         self.gas_price
     }
 
-    fn to(&self) -> TxKind {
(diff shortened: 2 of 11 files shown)


crates/consensus/src/transaction/eip1559.rs
```diff
@@ -296,7 +296,7 @@ impl Transaction for TxEip1559 {
         self.max_priority_fee_per_gas
     }
 
-    fn to(&self) -> TxKind {
+    fn kind(&self) -> TxKind {
         self.to
     }
```
mattsse, line 301: can we add a to function that does `kind().to()`

https://github.com/paradigmxyz/reth/blob/5e386130da0b123e107f5eb5a9a3a3622bc50129/crates/primitives/src/transaction/mod.rs#L241-L247

mattsse (changes requested):
I was about to open an issue for exactly this

crates/consensus/src/transaction/eip1559.rs
tcoratger: sure, good idea, done

tcoratger: In next release, we will be able to rm the two old functions in reth :)

Issue #1483, opened 2024-10-15
Fix: move type flag encoding back into `encode_2718`


prestwich:
it looks like d0794e586df374bcdd4a56729f8bdfb1ccdaafc7 lightly broke eip2718 abstraction by making type flags tx-associated instead of envelope-associated. We would now behave badly for networks where the type flag is different (e.g. `TxEip1559` is flag `17`, or where both `4` and `5` are `TxEip7702`

It also means that if we have a tx type with two different on different networks, they can't re-use the same codepaths

Need to move the flag encoding back into the envelope-level `encode_2718` implementation

_Originally posted by @prestwich in https://github.com/alloy-rs/alloy/pull/1460#discussion_r1800987185_

prestwich:
cc @Rjected

prestwich:
this also affects `encode_for_signing` in `SignableTransaction`

klkvr:
> We would now behave badly for networks where the type flag is different (e.g. `TxEip1559` is flag `17`, or where both `4` and `5` are `TxEip7702`

I am not sure we should optimize for this edge case at cost of forcing users to route all encoding through envelope and introducing additional layering on envelope variants

prestwich:
> I am not sure we should optimize for this edge case at cost of forcing users to route all encoding through envelope 

The 2718 edge case aside, we already force users to route encoding through envelope, it was one of the main goals of introducing it as a type and the only reason for making all the encoding methods hidden in #529. There was a lot of discussion on this in issues and the telegram group at the time and we collectively agreed that running coding via the transaction types has proven to result in bad user behavior

Basically, from ethers we learned we can't rely on users to understand the distinction between eip2718, rlp, and network format. As a result, having those be a public API is a footgun for users. We saw this a lot with users RLP encoding 1559 transactions to submit to `sendRawTransaction`, which failed and resulted in an awful lot of confused messages in telegram. Making incorrect or confusing actions inaccessible to users was a primary goal of introducing the envelope type in the first place

prestwich:
closing in favor of solution in #1496

PR #1485, opened 2024-10-15
refactor: move hash memoization out of signed

### LINKED ISSUE #1537: [Bug] trie_hash impl incorrect for tx envelope
### Component

consensus, eips, genesis

_No response_

### Operating System

None

### Describe the bug

from #1496

`impl Encodable2718 for TxEnvelope` produces incorrect `trie_hash` when it contains a sidecar. It computes the hash of the 2718 serialization WITH the sidecar, when instead it should compute it WITHOUT the sidecar, as the trie-hash commits only to the on-chain version of the tx

prestwich:
This is a breaking change

## Motivation

See discussion in #1460 and #1483

Memoizing the hash in `Signed` is inappropriate, as the transaction does not know its full 2718 serialization without being embedded in an envelope. This breaks the hash out into a `Sealed` type that is produced when the `Envelope` is constructed. The sealed type is transparent to the user

## Solution

- Introduce `Enveloped` type alias
- ensure that `From<>` impls properly `Seal`
- ensure that `impl Encodable2718 for TxEnvelope` produces the correct `trie_hash` by returning the memoized hash. This previously produced incorrect hashes for 4844 txns with sidecars

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -395,6 +395,10 @@ impl Encodable2718 for TxEnvelope {
             }
         }
     }
+
+    fn trie_hash(&self) -> B256 {
+        self.tx_hash()
+    }
 }
 
 impl Transaction for TxEnvelope {
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -580,15 +580,15 @@ mod serde_from {
     #[serde(tag = "type")]
     pub(crate) enum TaggedTxEnvelope {
         #[serde(rename = "0x0", alias = "0x00")]
-        Legacy(Enveloped<TxLegacy>),
+        Legacy(#[serde(with = "alloy_serde::sealed::flat_hash")] Enveloped<TxLegacy>),
         #[serde(rename = "0x1", alias = "0x01")]
-        Eip2930(Enveloped<TxEip2930>),
+        Eip2930(#[serde(with = "alloy_serde::sealed::flat_hash")] Enveloped<TxEip2930>),
         #[serde(rename = "0x2", alias = "0x02")]
-        Eip1559(Enveloped<TxEip1559>),
+        Eip1559(#[serde(with = "alloy_serde::sealed::flat_hash")] Enveloped<TxEip1559>),
         #[serde(rename = "0x3", alias = "0x03")]
-        Eip4844(Enveloped<TxEip4844Variant>),
+        Eip4844(#[serde(with = "alloy_serde::sealed::flat_hash")] Enveloped<TxEip4844Variant>),
         #[serde(rename = "0x4", alias = "0x04")]
-        Eip7702(Enveloped<TxEip7702>),
+        Eip7702(#[serde(with = "alloy_serde::sealed::flat_hash")] Enveloped<TxEip7702>),
     }
 
     impl From<MaybeTaggedTxEnvelope> for TxEnvelope {
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -2,6 +2,7 @@ use crate::{
     transaction::eip4844::{TxEip4844, TxEip4844Variant, TxEip4844WithSidecar},
     Signed, Transaction, TxEip1559, TxEip2930, TxEip7702, TxLegacy,
 };
+use alloc::vec::Vec;
 use alloy_eips::{
     eip2718::{Decodable2718, Eip2718Error, Eip2718Result, Encodable2718},
     eip2930::AccessList,
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1,14 +1,21 @@
-use core::fmt;
-
-use crate::{Signed, Transaction, TxEip1559, TxEip2930, TxEip7702, TxLegacy};
+use crate::{
+    transaction::eip4844::{TxEip4844, TxEip4844Variant, TxEip4844WithSidecar},
+    Signed, Transaction, TxEip1559, TxEip2930, TxEip7702, TxLegacy,
+};
 use alloy_eips::{
     eip2718::{Decodable2718, Eip2718Error, Eip2718Result, Encodable2718},
     eip2930::AccessList,
 };
-use alloy_primitives::{Bytes, TxKind, B256};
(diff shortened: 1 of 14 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -108,53 +115,79 @@ pub enum TxEnvelope {
...
-    Eip7702(Signed<TxEip7702>),
+    Eip7702(Enveloped<TxEip7702>),
 }
 
 impl From<Signed<TxLegacy>> for TxEnvelope {
     fn from(v: Signed<TxLegacy>) -> Self {
-        Self::Legacy(v)
+        let mut buf = Vec::with_capacity(v.tx().encoded_len_with_signature(v.signature()));
+        v.tx().encode_with_signature_fields(v.signature(), &mut buf);
+        let h = keccak256(buf);
```
klkvr, line 128: should all of those be `Sealable` impls?

prestwich: no, only the envelope knows the correct serialization for its components (tx types don't know their flag as it may be network dependent c.f. issue #1483)

prestwich:
this is ready for re-review

crates/consensus/src/transaction/envelope.rs
```diff
@@ -95,64 +103,90 @@ impl TryFrom<u8> for TxType {
...
     /// historical transactions..
     ///
     /// 2 - The transaction with a sidecar, which is the form used to
     /// send transactions to the network.
-    Eip4844(Signed<TxEip4844Variant>),
+    Eip4844(Enveloped<TxEip4844Variant>),
     /// A [`TxEip7702`] tagged with type 4.
-    Eip7702(Signed<TxEip7702>),
+    Eip7702(Enveloped<TxEip7702>),
 }
```
emhane, line 121: excellent, **but**, what's the plan for `alloy_consensus::TypedTransaction` now? it's essentially the same type

emhane (approved):

crates/consensus/src/transaction/envelope.rs
```diff
@@ -95,64 +103,90 @@ impl TryFrom<u8> for TxType {
...
 
 impl From<Signed<TxLegacy>> for TxEnvelope {
     fn from(v: Signed<TxLegacy>) -> Self {
-        Self::Legacy(v)
+        let mut buf = Vec::with_capacity(v.tx().encoded_len_with_signature(v.signature()));
+        v.tx().encode_with_signature_fields(v.signature(), &mut buf);
```
mattsse, line 126: is there no other way to get the tx hash?

if the only solution is to go through the envelope type this will be very inconvenient

crates/consensus/src/transaction/envelope.rs
prestwich: `TypedTransaction` is the unsigned version. As such neither signature or hash is known, and it just includes a signable version of the inner transaction and is used in the `NetworkWallet::<N>::sign_transaction_from` API

actually now that you mention it, we need to apply #1489 to `TypedTransaction` as well

klkvr:
At current state of alloy transactions do know their encoding and their type byte. I've seen a lot of examples of people using `encode_with_signature` and I don't think there's a reason to break this API (which would be required to make tx type envelope-associated)

IMO this edge case (Ethereum transactions with different type byte) is very unlikely and already can be solved by introducing separate types

crates/consensus/src/transaction/envelope.rs
prestwich: c.f. #1483. Technically it's legal for networks to make envelopes that do the following:
- use the same inner type, but a different serialization (e.g. a TxLegacy serialized borsht instead of RLP)
- use a different tx flag for the same inner type (e.g. a `TxEnvelope` containing `TxEip1559` at `23` instead of `2`
- use the same inner type at multiple tx flags (e.g. a `TxEnvelope` containing `TxEip1559` at both `2` and `23`)

So there is no way to correctly get the hash of a transaction without wrapping it in the specific network's envelope first

klkvr: all of those usecases can be solved by introducing a separate inner transaction type (e.g. network can provide its own `TxLegacy`)

right now it makes total sense for separate transaction types to know their full encoding because we don't yet have a case of network rolling out encoding of transaction conflicting with existing tx types. also given that most of the networks target for compatibility with ethereum I'd consider such case very unlikely

klkvr:
> This API is broken right now, and their usage is almost certainly incorrect.

could you please elaborate a bit on that? is it broken for 4844 case?

crates/consensus/src/transaction/envelope.rs
prestwich: I also consider it unlikely. But it's still the only correct abstraction of an EIP-2718 envelope if we want our code to be reusable.

> because the type id is even part of the eip itself

This is a fair reading of the spec. And @klkvr is right that people downstream can just duplicate 500 lines of code per tx type to change a single constant. However, it costs us very little to have the envelope drive serialization and hashing

For another interesting example, consider the extension sentinel value in eip-2718 `0xff` 🙃  This is also unlikely ofc, but tx-association falls down in extended envelopes

prestwich:
> could you please elaborate a bit on that? is it broken for 4844 case?

There's that, yes. 4844 transaction with sidecars produce incorrect `trie_hash` output right now. So that needs to be solved regardless. 

- The docs for `encode_with_signature` are also wrong 
- `encode_with_signature` produces 2718 output when `with_header` is `false`, and RLP when `with_header` is true. All other `encode_` methods produce RLP output
- `decode_with_signature` is missing entirely
- `EncodableSignature` seems to now be redundant as reth migrated to `alloy-primitives::Signature`
- sometimes header presence is determined by method name `encode_fields`, while sometimes it's by arg `with_header: bool`

I can continue, but basically this is a broken API that confuses multiple serialization schemes and is almost impossible to correctly use

crates/consensus/src/transaction/envelope.rs
prestwich: #1494

prestwich:
actually this is a good time to just make a cleanup PR. I'll go do that

prestwich:
holding off on rebasing this as it should be blocked by #1496

prestwich:
post-#1496 would y'all prefer a PR that does not break out memoization, and instead just fixes #1537 using the new tx_hash api?

prestwich:
closing per discussion in slack

PR #1491, opened 2024-10-16
fix: remove `Clone` from nonce-related APIs

Daniel-Aaron-Bloom:
This is technically a breaking change as it modifies the public signatures of multiple traits and structs (removing clone, switching to `self` by value instead of reference), but it should have minimal impact on most downstream users.

## Motivation

tl;dr is cloning a `NonceManager` (particularly `CachedNonceManager` because `DashMap` is not an `Arc`) is a big foot gun. It's actually even worse than that because the various layers automatically `clone` stuff on your behalf. The why is because cloning often turns numbers that should only be used once into numbers that are used twice.

## Solution

Remove `Clone` from the relevant APIs, and patch up other APIs to remove their dependency on clonability of `TxFiller`s. This does result in some duplicate `ProviderLayer` impls (by-ref and by-val).

## PR Checklist

- [x] ~Added Tests~ Not applicable
- [x] ~Added Documentation~ Not applicable
- [x] Breaking changes


crates/provider/src/fillers/nonce.rs
```diff
@@ -54,7 +54,7 @@ impl NonceManager for SimpleNonceManager {
 ///
 /// There is also an alternative implementation [`SimpleNonceManager`] that does not store the
 /// transaction count locally.
-#[derive(Clone, Debug, Default)]
+#[derive(Debug, Default)]
 pub struct CachedNonceManager {
     nonces: DashMap<Address, Arc<Mutex<u64>>>,
```
mattsse, line 59: I see, unsure how we missed this

can we instead replace this with an Arc mutex instead? not a fan of dashmap anyway

mattsse: this change we could do this immediately, not sure I fully get the other parts of the pr yet, so I suggest to submit that separately

mattsse:
marking this as resolved via #1624

PR #1497, opened 2024-10-16
eip7251: add unit tests for `ConsolidationRequest`

tcoratger:


onbjerg (approved):
I think these are fine since they are just round trip tests, even if the encoding was just changed for devnet 4 (and will be in alloy).

Issue #1502, opened 2024-10-16
[Feature] Add serde helpers for handling maps with quantity keys


mattsse:
### Component

serde

### Describe the feature you would like

we already have helpers for handling vec

https://github.com/alloy-rs/alloy/blob/8857646d0851ab49252ed4478f8264b63804c63d/crates/serde/src/quantity.rs#L61-L64

it can be useful to have the same for hashmaps/btreemaps with quantity key types

## TODO
* add mods with manual map serde; take over this wip https://github.com/alloy-rs/alloy/pull/1503

PR #1503, opened 2024-10-16
feat: quantity map serde

mattsse:
towards #1502


mattsse:
https://github.com/alloy-rs/alloy/pull/1579

PR #1505, opened 2024-10-16
feat: make it possible to configure Ws config

### LINKED ISSUE #1504: [Bug] websocket cannot return a long response
### Component

transports

any

### Operating System

Linux

### Describe the bug

`PubSubConnect` uses [default](https://github.com/snapview/tungstenite-rs/blob/7d338cbb277c1f53bb476355bdd1d6c5390d9a10/src/protocol/mod.rs#L75) [WebSocketConfig](https://github.com/snapview/tungstenite-rs/blob/7d338cbb277c1f53bb476355bdd1d6c5390d9a10/src/protocol/mod.rs#L34) ([set](https://github.com/snapview/tungstenite-rs/blob/7d338cbb277c1f53bb476355bdd1d6c5390d9a10/src/protocol/mod.rs#L334)) for `tungstenite`.
```rust
impl Default for WebSocketConfig {
    fn default() -> Self {
        #[allow(deprecated)]
        WebSocketConfig {
            max_send_queue: None,
            write_buffer_size: 128 * 1024,
            max_write_buffer_size: usize::MAX,
            max_message_size: Some(64 << 20), // 64 MiB
            max_frame_size: Some(16 << 20), // 16 MiB
            accept_unmasked_frames: false,
        }
    }
}
```
But these sizes are not enough, because, for example, [in Geth you can disable the maximum message size](https://github.com/ethereum/go-ethereum/blob/18a591811f2afb68adeb978d69ca443cbb252fd8/rpc/client_opt.go#L38).

Can be encountered when receiving a large number of logs. (for example, through alchemy, eth)
```rust
tracing_subscriber::fmt::init();

    .on_ws(WsConnect::new("wss://eth-mainnet.g.alchemy.com/v2/..."))
    .await
    .expect("Failed to connect to provider");

let block = 20980952;
let filter = Filter::new()
    .event_signature(vec![
        b256!("c42079f94a6350d7e6235f29174924f928cc2ac818eb64fed8004e115fbcca67"),
        b256!("d78ad95fa46c994b6551d0da85fc275fe613ce37657fb8d5e3d130840159d822"),
        b256!("0c396cd989a39f4459b5fa1aed6a9a8dcdbc45908acfd67e028cd568da98982c"),
        b256!("7a53080ba414158be7ec69b987b5fb7d07dee101fe85488f0853ae16239d0bde"),
    ])
    .from_block(block - 500)
    .to_block(block);

provider.get_logs(&filter).await.unwrap();
```

Problem is that we cannot process the result as an error, because we only get an `error trace` and reconnecting.
```shell
2024-10-16T22:13:40.232414Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
2024-10-16T22:13:42.158094Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
2024-10-16T22:13:43.944266Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
2024-10-16T22:13:45.790928Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
2024-10-16T22:13:47.575576Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
2024-10-16T22:13:49.414130Z ERROR alloy_transport_ws::native: WS connection error err=Space limit exceeded: Message too long: 19563458 > 16777216
```

1makarov:
closes #1504

### DIFF
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -44,7 +44,7 @@ impl WsConnect {
     }
 
     /// Sets the websocket config.
-    pub fn with_config(mut self, config: WebSocketConfig) -> Self {
+    pub const fn with_config(mut self, config: WebSocketConfig) -> Self {
         self.config = Some(config);
         self
     }
@@ -72,13 +72,9 @@ impl PubSubConnect for WsConnect {
     async fn connect(&self) -> TransportResult<alloy_pubsub::ConnectionHandle> {
         let request = self.clone().into_client_request();
         let req = request.map_err(TransportErrorKind::custom)?;
-        let (socket, _) = tokio_tungstenite::connect_async_with_config(
-            req,
-            Some(self.config.unwrap_or_default()),
-            false,
-        )
-        .await
-        .map_err(TransportErrorKind::custom)?;
+        let (socket, _) = tokio_tungstenite::connect_async_with_config(req, self.config, false)
+            .await
+            .map_err(TransportErrorKind::custom)?;
 
         let (handle, interface) = alloy_pubsub::ConnectionHandle::new();
         let backend = WsBackend { socket, interface };
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -43,7 +43,7 @@ impl WsConnect {
         self
     }
 
-    /// Sets the websocket configuration.
+    /// Sets the websocket config.
     pub fn with_config(mut self, config: WebSocketConfig) -> Self {
         self.config = Some(config);
         self
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -4,8 +4,9 @@ use alloy_transport::{utils::Spawnable, Authorization, TransportErrorKind, Trans
 use futures::{SinkExt, StreamExt};
 use serde_json::value::RawValue;
 use std::time::Duration;
+pub use tokio_tungstenite::tungstenite::protocol::WebSocketConfig;
 use tokio_tungstenite::{
-    tungstenite::{self, client::IntoClientRequest, protocol::WebSocketConfig, Message},
+    tungstenite::{self, client::IntoClientRequest, Message},
     MaybeTlsStream, WebSocketStream,
 };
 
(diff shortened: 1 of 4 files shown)


crates/transport-ws/src/native.rs
```diff
@@ -63,8 +63,17 @@ impl PubSubConnect for WsConnect {
...
-        let (socket, _) =
-            tokio_tungstenite::connect_async(req).await.map_err(TransportErrorKind::custom)?;
+        let (socket, _) = tokio_tungstenite::connect_async_with_config(
+            req,
+            Some(WebSocketConfig {
+                max_frame_size: Some(usize::MAX),
+                max_message_size: Some(usize::MAX),
+                ..Default::default()
+            }),
+            false,
```
mattsse, line 73: this config should be moved to WsConnect so that the user can control this

mattsse (changes requested):

crates/transport-ws/src/native.rs
```diff
@@ -63,8 +72,13 @@ impl PubSubConnect for WsConnect {
     async fn connect(&self) -> TransportResult<alloy_pubsub::ConnectionHandle> {
         let request = self.clone().into_client_request();
         let req = request.map_err(TransportErrorKind::custom)?;
-        let (socket, _) =
-            tokio_tungstenite::connect_async(req).await.map_err(TransportErrorKind::custom)?;
+        let (socket, _) = tokio_tungstenite::connect_async_with_config(
+            req,
+            Some(self.config.unwrap_or_default()),
```
mattsse, line 77: I've navigated like 6 layers in tungstenite to check that they also do unwrap_or_default so this is okay

mattsse (approved):

crates/transport-ws/src/native.rs
mattsse, line 77: In case that changes we should stick to providing an option here
```suggestion
            self.config,
```

mattsse:
needs fmt and clippy, can't apply myself because opened from main branch

mattsse:
only that this breaks exhaustive `WsConnect {}` usage, by default this still uses the default tungs. config but now this can be configured on WsConnect manually

1makarov:
> Any downsides to this?

Those who use the `WsConnect::with_auth(url, header)` construct will have to update to `WsConnect::new(url).with_auth(header)`

PR #1526, opened 2024-10-19
feat(eips): Indexed Blob Hash

refcell:
### Description

Alloy's EIP 4844 Blob Sidecar provided a way to verify blobs given the indexed blob hash. Previously, [this argument was the EIP 1898 NumHash](https://github.com/alloy-rs/alloy/blob/main/crates/eips/src/eip4844/sidecar.rs#L109) construct, which doesn't translate to beacon. The indexed blob hash we actually want to accept is a net new type with an index and blob hash, instead of the block number and hash from the `NumHash` type.

This PR introduces that type, and changes the `verify_blob` method to accept the `IndexedBlobHash`.


clabby (approved):
+1 to this being a separate type.

mattsse (commented):
I see, we need to be actually compatible with beacon json API here because this is where the value is pulled from?

mattsse (approved):

PR #1496, opened 2024-10-16
fix: make a sensible encoding api

prestwich:
Spawned from the depths of #1485 

## Motivation

Fix transaction coding APIs to be usable and clear

- #1510 output for `encode_with_signature`
- Codebase generally confuses the 2 uses of RLP encoding (tx payload and network format)
- The docs for `encode_with_signature` are wrong
- `encode_with_signature` produces 2718 output when `with_header` is false, and RLP in network format when `with_header` is `true`. All other `encode_` methods produce RLP output in non-network format
- `decode_with_signature` is missing entirely
- `EncodableSignature` seems to now be redundant as reth migrated to `alloy-primitives::Signature`

## Solution

create a simple API in which encoding and decoding function names specify which format they produce/accept

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

PR #1528, opened 2024-10-19
fix: correct implementations of Encodable and Decodable for sidecars

### LINKED ISSUE #1499: [Bug] RLP is incorrectly implemented for Sidecars
### Component

consensus, eips, genesis

_No response_

### Operating System

None

### Describe the bug

it does not code or decode headers

prestwich:
Closes #1499 

## Motivation

current `Encodable` and `Decodable` implementations do not RLP encode or RLP decode the sidecar. Instead they encode only the fields, without the header, producing incorrect RLP output

## Solution

Current behavior has been moved into more-explicitly named functions and is used to construct correct RLP implementations

This is a breaking change as things that previously (incorrectly) deserialized now will not

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip4844.rs
+++ b/crates/consensus/src/transaction/eip4844.rs
@@ -1085,7 +1085,6 @@ mod tests {
         let len = expected_envelope.length();
         let mut buf = Vec::with_capacity(len);
         expected_envelope.encode(&mut buf);
-        dbg!(alloy_primitives::hex::encode(&buf[0..16]));
         assert_eq!(buf.len(), len);
 
         // ensure it's also the same size that `actual` claims to be, since we just changed the
--- a/crates/consensus/src/transaction/eip4844.rs
+++ b/crates/consensus/src/transaction/eip4844.rs
@@ -138,7 +138,7 @@ impl TxEip4844Variant {
             Self::TxEip4844WithSidecar(tx) => {
                 let payload_length = tx.tx().fields_len() + signature.rlp_vrs_len();
                 let inner_header = Header { list: true, payload_length };
-                inner_header.length() + payload_length + tx.sidecar().fields_len()
+                inner_header.length() + payload_length + tx.sidecar().rlp_encoded_fields_length()
             }
         };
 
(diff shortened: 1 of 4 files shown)


crates/consensus/src/transaction/eip4844.rs
```diff
@@ -1085,6 +1085,7 @@ mod tests {
         let len = expected_envelope.length();
         let mut buf = Vec::with_capacity(len);
         expected_envelope.encode(&mut buf);
+        dbg!(alloy_primitives::hex::encode(&buf[0..16]));
```
onbjerg, line 1088: should rm this i think?

onbjerg (commented):
smol nit,

do we not use this in reth right now? @Rjected or @mattsse 

if so, any reason for the encoding being the way it is? i assume we use this for network

prestwich:
nit fixed 🫡 

> if so, any reason for the encoding being the way it is? i assume we use this for network

It appears to have been a hack to deal with the eip4844 2718 encoding when the sidecar is present. See #1496 for more cleanup around these APIs (this bug was found while working on that). The sketch is that [4844 specifies](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md#networking) that the 2718 payload is `rlp([tx_payload_body, blobs, commitments, proofs])` in the network. Each field is encoded.

More interestingly, do we have any use case in which a sidecar itself should be RLP encoded at all? We have to RLP encode it's fields, but when is a standalone blob rlp encoded?

onbjerg:
>More interestingly, do we have any use case in which a sidecar itself should be RLP encoded at all?

not that I can recall, overall this PR looks good to me, but I'm wary of merging it mostly because I don't have a full overview of 4844 in reth, so I'd like to get either @mattsse or @Rjected to take a look since they worked on this and might know if there is such a usecase

prestwich:
> not that I can recall, overall this PR looks good to me, but I'm wary of merging it mostly because I don't have a full overview of 4844 in reth, so I'd like to get either @mattsse or @Rjected to take a look since they worked on this and might know if there is such a usecase

In the interest of fixing the logic bug, can we merge this as is, without deleting the impls? Reth's tests will easily flag reliance on the current faulty coding logic, and it's trivial to fix when reth bumps to a new alloy version

Rjected (commented):
Yeah, we should definitely make the no-header functions not `Encodable` / `Decodable` fns. Sidecars are basically never rlp encoded on their own. We can easily switch over the decoding fn to call, and we have blob tx encoding tests that would catch the behavior change

mattsse (approved):

mattsse:
yeah @Rjected the trait impls in the sidecar we should remove

PR #1530, opened 2024-10-19
chore(json-rpc): use a macro to check and set duplicate json-rpc fields

jsvisa:
## Motivation

Seems the fields in jsonrpc deserialize is similar, so we can use a macro to repeat the same works.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/json-rpc/src/lib.rs
```diff
@@ -135,3 +135,14 @@ impl<T> RpcReturn for T where T: DeserializeOwned + Debug + Send + Sync + Unpin
 pub trait RpcObject: RpcParam + RpcReturn {}
 
 impl<T> RpcObject for T where T: RpcParam + RpcReturn {}
+
+/// A macro to check if a field is present in a map and set it if it is.
+#[macro_export]
+macro_rules! check_and_set_field {
```
DaniPopes, line 141: I wouldn't export this, you can move it into a `macros` module and `#[macro_use]` on it

jsvisa: just raise a question, does this means we can only use this macro inside this crate? I want to use this macro also for the filter https://github.com/alloy-rs/alloy/blob/561d392d5c52352ffb8d3632602eb3ca21de1cee/crates/rpc-types-eth/src/filter.rs#L616-L648

jsvisa: Or maybe I misunderstood before and after. Is that what you meant to express?

```rs
# lib.rs
#[module_use]
mod macros;

# macros.rs
#[macro_export]
macro_rules! check_and_set_field {
    ($map:expr, $field:expr) => {{
        if $field.is_some() {
            return Err(serde::de::Error::duplicate_field(stringify!($field)));
        }
        $field = Some($map.next_value()?);
    }};
}
```

DaniPopes: If you want to re use it you might want to put it in alloy-serde if it doesnt add dependencies, otherwise duplicating it is also fine

mattsse (commented):
not a fan of using a macro for this, I prefer the non macro version because I think it's easier to follow even if more verbose

PR #1532, opened 2024-10-19
rpc: impl `From<Sealed<alloy_consensus::Header>>` for `Header`

tcoratger:
An easy to manage and maintain conversion util with the goal to replace https://github.com/paradigmxyz/reth/blob/2ae93682b4978bf80bf27c777334ead30d3e04f5/crates/rpc/rpc-types-compat/src/block.rs#L104


tcoratger:
@emhane is there something missing to be merged?

PR #1538, opened 2024-10-20
feat(rpc-types-trace/prestate): support disable_{code,storage}

jsvisa:
## Motivation

In some cases, we only need to track the ether balance/nonce changes, so no need to return the lonng code/storage back, so propose to add two new options: `disableCode` and `disableStorage` to indicate wheather need to track code/storage or not.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


mattsse (approved):
I think these additional settings make sense, but there is no reference impl for these, right?

jsvisa:
> I think these additional settings make sense, but there is no reference impl for these, right?

Yeah, erigon's changes are also in the way https://github.com/erigontech/erigon/pull/12384, and I'll file the similar changes for geth later

PR #1540, opened 2024-10-20
feat: integrate signature with boolean parity

klkvr:
## Motivation

Integrates signature from https://github.com/alloy-rs/core/pull/776

Depends on https://github.com/alloy-rs/eips/pull/12 and https://github.com/alloy-rs/core/pull/796, needs minor releases for core and eips

## Solution

Changes in this PR:
- Transaction encoding is made a bit more verbose as `signature.v()` is now encoded and accounted in length separately
- Validation/adjustments of signature parity value are removed
- Legacy transaction decoding now handles decoding of parity manually by decoding it into `u64` and extracting chain_id when possible. Encoding manually encodes chain_id into boolean parity as per EIP-155 spec.
- `sign_transaction_with_chain_id` macro no more adjusts the signature. as long as signer signed the correct payload with chain_id we are fine with parity being boolean, and legacy transaction encoding methods will figure out how to encode it into integer
- `signed_legacy_serde` module is added which should be used for serialization of `Signed<TxLegacy>` to respect the specification requiring parity being serialized as EIP-155 compliant `v` key.

This is not perfect from our pov because we now have more boilerplate logic for parity encoding. However, this makes encoding API safer without any changes to it.

One limitation here is that `Signed<TxLegacy>` on its own would be serialized incorrectly with `yParity` key. We can't implement serialize directly on individual `Signed<_>` types because this would make it impossible to implement serde for `Signed<_>` with custom transaction. For now I've added `with = "signed_legacy_serde"` to envelope legacy variants which should ensure that enveloped transactions are always (de)serialized correctly

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types-eth/src/transaction/signature.rs
```diff
@@ -74,27 +74,21 @@ where
...
+impl TryFrom<Signature> for alloy_primitives::PrimitiveSignature {
     type Error = alloy_primitives::SignatureError;
 
     fn try_from(value: Signature) -> Result<Self, Self::Error> {
         let parity = if let Some(y_parity) = value.y_parity {
-            alloy_primitives::Parity::Parity(y_parity.0)
+            y_parity.0
         } else {
-            value.v.to::<u64>().try_into()?
+            match value.v.to::<u64>() {
```
mattsse, line 84: this could panic, we should keep using try_into here

mattsse (approved):

crates/rpc-types-eth/src/transaction/signature.rs
klkvr: this type and conversion are now gone with #1460 :)

mattsse (approved):

PR #1541, opened 2024-10-20
consensus: impl `recover_signers` for `BlockBody<TxEnvelope>`

tcoratger:
Related https://github.com/paradigmxyz/reth/issues/11208

The ideal for the reth issue is to restrict the functions to be implemented as much as possible to the alloy `Transaction` trait however it doesn't make much sense to add a `recover_signer` or `recover_signers` method to the `Transaction` trait because we can only fetch the signer in the `Signed<Tx>` case, which is almost never the case when trying to implement `Transaction`.

So I think the simplest approach to replace https://github.com/paradigmxyz/reth/blob/0270128d4f7a9f6fad27dff69273095abdfa7452/crates/primitives/src/block.rs#L564-L566 is to do the similar implementation here in alloy using what already exists in `TxEnvelope` to recover the signers from an iterator of transactions `TxEnvelope`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -146,6 +146,7 @@ derive_more = { version = "1.0.0", default-features = false }
 strum = { version = "0.26", default-features = false }
 http = "1.1.0"
 jsonwebtoken = "9.3.0"
+rayon = "1.7"
 
 ## serde
 serde = { version = "1.0", default-features = false, features = [
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -42,6 +42,7 @@ derive_more = { workspace = true, features = [
     "into_iterator"
 ], default-features = false }
 auto_impl.workspace = true
+rayon = { workspace = true, optional = true }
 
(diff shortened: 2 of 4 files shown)


tcoratger:
@emhane Maybe we should keep this for a follow up PR right? My reasoning was as follows:
- Currently in reth, the `BlockBody` contains a vector of `TransactionSigned`
- In alloy we don't have a `TransactionSigned` type but we have the `TxEnvelope` type which is equivalent because it is an enum that contains all the `Signed` variants.
     - In addition we already have the `recover_signer` method which is already implemented for `TxEnvelope`.
     - So I thought it would be rather logical to make an implementation of `BlockBody<TxEnvelope>` to use it in reth. But maybe if we want to be even more generic we could open a follow up issue to move the `recover_signer` and `recover_signers` methods to the appropriate traits right?

mattsse (commented):
we've changed this a bunch on reth
we still want to eventually upstream this but before we lock in an API for this here we need to stabilize on reth

PR #1542, opened 2024-10-20
bump rust msrv to 1.80

tcoratger:


DaniPopes:
Why? I'd rather not bump without a reason

tcoratger:
> Why? I'd rather not bump without a reason

@DaniPopes it's due to LazyLock in std only available in 1.80 for https://github.com/alloy-rs/alloy/pull/1541 but not 100% sure this is the approach @mattsse wanted to take here so maybe let us wait for the review of https://github.com/alloy-rs/alloy/pull/1541 and then merge this if the review is positive only wdyt?

zerosnacks:
Appears that https://github.com/alloy-rs/alloy/pull/1552#issuecomment-2432509825 would supersede this PR, closing in favor of https://github.com/alloy-rs/alloy/pull/1552

PR #1543, opened 2024-10-21
consensus: add `Debug` trait bound for `Transaction` trait

tcoratger:
Nice to have at an alloy level to remove the bound here https://github.com/paradigmxyz/reth/blob/20dc0c7da0a12e31ab0c896ccf4208585658ad07/crates/primitives-traits/src/transaction/mod.rs#L19


mattsse (approved):
this is reasonable

PR #1544, opened 2024-10-21
eip1898: add unit tests to secure all conversions and impl

tcoratger:
- Full path under serde is used instead of import under cfg flag.
- Some functions are simplified.
- Unit tests are added to protect all implementations and conversions that can be tricky to mix.


zerosnacks (approved):
Proposed changes are sensible and does not appear to contain breaking changes. Tests look good, ran tests against `main` and verified there were no regressions or change in behavior.

tcoratger:
> Proposed changes are sensible and does not appear to contain breaking changes. Tests look good, ran tests against `main` and verified there were no regressions.

@zerosnacks Yes exactly, that's exactly why I was doing this PR. My constant was that there were a lot of conversions in this file and that simplifications could be made in order to end up with exactly the same code while having a less wordy and error-prone implementation.
- So I simplified the places that could be simplified by using all the available conversions as much as possible.
- I took care to normally add all the unit tests covering all situations because these unit tests are very quick to run and they allow two things:
    - Check that this PR doesn't break anything and that everything is correct (conversions, string manipulations...)
    - Be sure that we don't break anything in the future as soon as we touch this EIP.

zerosnacks:
cc @yash-atreya pending final review, LGTM

yash-atreya (approved):

PR #1547, opened 2024-10-22
chore: bump alloy-eip7702

klkvr:


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -481,7 +481,7 @@ mod tests {
...
+            r#"{"hash":"0x0000000000000000000000000000000000000000000000000000000000000001","nonce":"0x2","blockHash":"0x0000000000000000000000000000000000000000000000000000000000000003","blockNumber":"0x4","transactionIndex":"0x5","from":"0x0000000000000000000000000000000000000006","to":"0x0000000000000000000000000000000000000007","value":"0x8","gasPrice":"0x9","gas":"0xa","maxFeePerGas":"0x15","maxPriorityFeePerGas":"0x16","input":"0x0b0c0d","r":"0xe","s":"0xe","v":"0xe","chainId":"0x11","type":"0x14","authorizationList":[{"chainId":"0x1","address":"0x0000000000000000000000000000000000000006","nonce":"0x1","yParity":"0x0","r":"0x48b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d495a36649353","s":"0xefffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab949f53faa07bd2c804"}]}"#
```
mattsse, line 484: this is an rpc change, I believe this is correct and this was only `v` before because we used a fake v value for testing?

mattsse (approved):

crates/rpc-types-eth/src/transaction/mod.rs
klkvr: yep, the test sig had v as 27 which also resulted in invalid rlp for auth

PR #1549, opened 2024-10-23
fix: sidecar rlp decoding

klkvr:
## Motivation

Right now this encodes as list and expects string during decoding which is incorrect. 

I am wondering if we should remove those impls completely? Given that sidecar encoding on its own is not defined

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


prestwich (approved):

prestwich:
> I am wondering if we should remove those impls completely? Given that sidecar encoding on its own is not defined

sidecar encoding on its own is defined, because RLP is well-defined for any set of RLP objects. It is not used anywhere that we're aware of https://github.com/alloy-rs/alloy/pull/1528#issuecomment-2424298961

mattsse (approved):

PR #1552, opened 2024-10-23
chore(general): bump MSRV to 1.81, use `core::error::Error` on `no-std` compatible crates

Evalir:
## Motivation

For `no-std` compatible crates, we can use `core::error::Error` instead of `std::error::Error` to avoid gating the Error impl under `std` for the impl blocks.

## Solution

replace `std::error::Error` with `core::error::Error` on the impl blocks.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


DaniPopes:
I'm ok with bumping MSRV for this since it's in line with our policy. Do you want to do this in alloy-core too?

Evalir:
Yep, happy to do it on alloy core too. Sounds good!

Evalir:
@DaniPopes i've not removed any `std` error feature gates where we reuse errors from alloy-rlp/const-hex/ruint. should we bump there too, or is leaving it like it OK for now?

README.md
```diff
@@ -124,7 +124,7 @@ When updating this, also update:
 - .github/workflows/ci.yml
 -->
 
-The current MSRV (minimum supported rust version) is 1.79.
+The current MSRV (minimum supported rust version) is 1.81.
 
 Alloy will keep a rolling MSRV policy of **at least** two versions behind the
```
tcoratger, line 129: Maybe we should remove this comment as it is not true anymore no?

zerosnacks:
Added a temporary block label as it relies on `1.83` being released: https://releases.rs/docs/1.83.0/

zerosnacks:
Yay! Unblocked by https://blog.rust-lang.org/2024/11/28/Rust-1.83.0.html

DaniPopes (approved):

Evalir:
Awesome! Thanks y'all 😄

PR #1554, opened 2024-10-23
rpc: add `uncle_block_from_header` impl and test

tcoratger:
The goal here is similar to https://github.com/alloy-rs/alloy/pull/1532

In reth we have: 

https://github.com/paradigmxyz/reth/blob/2ae93682b4978bf80bf27c777334ead30d3e04f5/crates/rpc/rpc-types-compat/src/block.rs#L178-L192

And I have found myself several times in the situation where I was forced to go through the `reth-rpc-types-compat` crate just for this small conversion.

Seems like more logical and smooth to have it directly in alloy since the target type is an alloy rpc type.

I did a test against reth implementation to check the results are the same

image

### DIFF
--- a/crates/rpc-types-eth/src/block.rs
+++ b/crates/rpc-types-eth/src/block.rs
@@ -2,11 +2,12 @@
 
 use crate::{ConversionError, Transaction, Withdrawal};
 use alloc::collections::BTreeMap;
-use alloy_consensus::Sealed;
+use alloy_consensus::{Sealed, TxEnvelope};
 use alloy_network_primitives::{
     BlockResponse, BlockTransactions, HeaderResponse, TransactionResponse,
 };
 use alloy_primitives::{Address, BlockHash, Bloom, Bytes, B256, B64, U256};
+use alloy_rlp::Encodable;
 
 use alloc::vec::Vec;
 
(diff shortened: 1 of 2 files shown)


crates/rpc-types-eth/src/block.rs
```diff
@@ -51,6 +52,32 @@ impl<T: TransactionResponse, H> Block<T, H> {
...
 
+impl<T> Block<T> {
+    /// Constructs an "uncle block" from the provided header.
+    ///
+    /// This function creates a new [`Block`] structure for uncle blocks (ommer blocks),
+    /// using the provided [`alloy_consensus::Header`].
+    pub fn uncle_block_from_header(header: alloy_consensus::Header) -> Self {
+        Self {
+            uncles: vec![],
+            header: header.clone().seal(header.hash_slow()).into(),
```
mattsse, line 63: isn't this just sealed().into()

crates/rpc-types-eth/src/block.rs
```diff
@@ -51,6 +52,32 @@ impl<T: TransactionResponse, H> Block<T, H> {
...
+            transactions: BlockTransactions::Uncle,
+            withdrawals: Some(vec![]),
```
mattsse, line 65: this should actually be None I believe

crates/rpc-types-eth/src/block.rs
```diff
@@ -51,6 +52,32 @@ impl<T: TransactionResponse, H> Block<T, H> {
...
+                alloy_consensus::Block {
+                    header,
+                    body: alloy_consensus::BlockBody::<TxEnvelope> {
+                        transactions: vec![],
+                        ommers: vec![],
+                        withdrawals: None,
+                    },
+                }
+                .length(),
+            )),
```
mattsse, line 76: the size is also based on the entire block so we need to create the block first, then get it's size

mattsse (changes requested):

crates/rpc-types-eth/src/block.rs
tcoratger: just fixed

crates/rpc-types-eth/src/block.rs
tcoratger: With https://github.com/paradigmxyz/reth/pull/12038, sounds ok

crates/rpc-types-eth/src/block.rs
tcoratger: @mattsse Normally it's good. I just quickly implemented what you did in https://github.com/alloy-rs/alloy/pull/1561 (we can merge the two one after the other it should work) so that I can use it directly here. So the logic is for this line:
- Create a consensus primitive block
- Calculate its length
- Insert it here in the block size

mattsse (approved):

Issue #1560, opened 2024-10-24
JSON-RPC types do not agree with Ethereum's OpenRPC spec


0xaatif:
As an example, take `debug_getBadBlocks` (alphabetically, the first method in the spec)

```yaml
- name: debug_getBadBlocks
  summary: Returns an array of recent bad blocks that the client has seen on the network.
  params: []
  result:
    name: Blocks
    schema:
      title: Bad block array
      type: array
      items:
        $ref: '#/components/schemas/BadBlock'
```
https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/debug/getters.yaml#L82-L91

`BadBlock` looks like
```yaml
BadBlock:
  title: Bad block
  type: object
  required:
    - block
    - hash
    - rlp
```
https://github.com/ethereum/execution-apis/blob/v1.0.0-beta.4/src/schemas/block.yaml#L140-L146

But in `alloy`, we return a `Vec<Block>`:
https://github.com/alloy-rs/alloy/blob/b621af07b7bd970d256c523c90ec0c91ec61cea1/crates/provider/src/ext/debug.rs#L30

This is the wrong struct (there's no `BadBlock` struct anyway)

EdwardJES:
Keen to take the the `BadBlock` task.

mattsse:
cool @EdwardJES for this we need a new helper type and then simply swap out the return type

yash-atreya:
Closed by #1566

PR #1566, opened 2024-10-25
rpc: add `BadBlock` type to `debug_getbadblocks` return type

EdwardJES:
This PR adds `BadBlock` as a new type and adds it to the return type to `debug_getBadBlocks`. Addresses https://github.com/alloy-rs/alloy/issues/1560

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types-eth/src/block.rs
```diff
@@ -105,11 +104,11 @@ pub struct Header {
     /// Timestamp
     #[cfg_attr(feature = "serde", serde(default, with = "alloy_serde::quantity"))]
     pub timestamp: u64,
+    /// Extra data
```
EdwardJES, line 107: Reordered to allow deriving of `Rlp` traits.

crates/rpc-types-eth/src/block.rs
```diff
@@ -70,9 +68,10 @@ impl<T> Block<T> {
 
 /// Block header representation.
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
-#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash, RlpDecodable, RlpEncodable)]
```
EdwardJES, line 71: Will these have any conflicts with https://github.com/alloy-rs/alloy/blob/bc2040d20216a8d14ff87c31c34eff2a74635ce0/crates/consensus/src/header.rs#L351?

crates/rpc-types-eth/src/block.rs
```diff
@@ -462,6 +461,19 @@ impl<T: TransactionResponse, H: HeaderResponse> BlockResponse for Block<T, H> {
     }
 }
 
+/// Bad block representation.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct BadBlock {
```
EdwardJES, line 468: Unsure if this deserved it's own file or if this is the most suitable location.

crates/rpc-types-eth/src/block.rs
mattsse, line 71: no rlp for rpc types

mattsse (changes requested):
this pr should be limited to the new badblock type and not change any existing types.

crates/rpc-types-eth/src/block.rs
EdwardJES: Cool, got it. Will change. 

What would be the best way to calculate the rlp encoded header for the `rlp` field? 

Ty 🙏

EdwardJES: Good point, my mistake. Fixed.

EdwardJES:
Thank you, @mattsse. Have addressed the comments.

mattsse (approved):

EdwardJES:
Fixing clippy now.

PR #1567, opened 2024-10-25
consensus: rm `Receipts` `root_slow` unused method

tcoratger:
Follow up on https://github.com/alloy-rs/alloy/pull/1563:
- Full coverage with unit tests
- Small refactor

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -355,7 +355,7 @@ mod test {
             Receipt { status: Eip658Value::Eip658(false), cumulative_gas_used: 2000, logs: vec![] };
 
         // Create a `Receipts` instance with a single set of receipts
-        let receipts = Receipts { receipt_vec: vec![vec![receipt1.clone(), receipt2.clone()]] };
+        let receipts = Receipts { receipt_vec: vec![vec![receipt1, receipt2]] };
 
         // Calculate the root hash of the receipts at index 0
         //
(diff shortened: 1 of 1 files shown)


DaniPopes:
I don't see the point of this function, after changing `&[&T]` to `&[T]` because that collect is unnecessary, it becomes essentially `receipts.map(f)`, which doesn't need to be a function

tcoratger:
> I don't see the point of this function, after changing `&[&T]` to `&[T]` because that collect is unnecessary, it becomes essentially `receipts.map(f)`, which doesn't need to be a method

@DaniPopes Yes fair point. I think it requires changing some function definitions in reth but I think it can only be beneficial. Here is what I propose:
- I have a PR open in draft on reth to work on it (and more generally the replacement of `Receipts`) https://github.com/paradigmxyz/reth/pull/12059/
- Let's do a review on it without using the `root_slow` function as you propose here.
- If it is valid then we transform this PR on alloy into a PR for the definitive removal of this function.

What do you think?

emhane:
agree with @DaniPopes , this only makes sense if this type is the l1 type, and it implements a new trait `ReceiptsRoot`, which should then also be implemented on a similar `op-alloy` type `OpReceipts`

tcoratger:
@emhane https://github.com/paradigmxyz/reth/pull/12059 is ready for review, if everything is fine and merged then we can delete the function here in alloy :)

tcoratger:
@DaniPopes @emhane Given the path taken by https://github.com/paradigmxyz/reth/pull/12162, just removed this unused method

DaniPopes (approved):

mattsse (approved):

PR #1572, opened 2024-10-26
consensus: implement `validate_against_parent_4844` for `Header`

tcoratger:
Right now in Reth, we have a `validate_against_parent_4844` that is implemented inside the consensus crate. 

https://github.com/paradigmxyz/reth/blob/1bdf429af5092a0b737e900977c1418bc070ec59/crates/consensus/common/src/validation.rs#L245-L274

Since it affects consensus and applies to the `Header`, it seems to me that it makes sense to move it here (I don't know if this proposal will be accepted). So I also created a dedicated `ConsensusError` type in alloy.


crates/consensus/src/header.rs
```diff
@@ -346,6 +375,50 @@ impl Header {
...
     }
+
+    /// Validates that the EIP-4844 fields in the current block header are correct compared to the
+    /// parent block header.
+    ///
+    /// Specifically, this function checks two things:
+    /// 1. The current block header contains the `blob_gas_used` and `excess_blob_gas` fields.
+    /// 2. The `excess_blob_gas` field matches the expected value calculated from the parent header
+    ///    fields.
+    pub fn validate_against_parent_4844(&self, parent: &Self) -> Result<(), ConsensusError> {
```
tcoratger, line 386: Wanted to create a follow up PR with unit tests to validate all the scenarios, but first I wanted to know if this proposal of migration would be accepted

crates/consensus/src/header.rs
```diff
@@ -12,6 +12,35 @@ use alloy_primitives::{
...
+    #[display(
+        "invalid excess blob gas, got: {got}, expected: {expected}; \
+            parent excess blob gas: {parent_excess_blob_gas}, \
+            parent blob gas used: {parent_blob_gas_used}"
+    )]
+    ExcessBlobGasDiff {
+        /// Expected excess blob gas.
+        expected: u64,
+        /// Actual excess blob gas.
+        got: u64,
```
tcoratger, line 36: I've seen several places in alloy where the got/expected error pattern is present, maybe it would be good to move the reth primitive

https://github.com/paradigmxyz/reth/blob/ab07fcfb113cb0b579e1f9f55a5dd9511b576687/crates/primitives-traits/src/error.rs#L9

into alloy-primitives to be able to use it everywhere

clippy.toml
```diff
@@ -1 +1 @@
-msrv = "1.79"
+msrv = "1.81"
```
tcoratger, line 1: I had to bump Rust due to the derive_more error stuff. I think this goes against 

> Alloy will keep a rolling MSRV policy of at least two versions behind the latest stable release (so if the latest stable release is 1.58, we would support 1.56).

I think that Rust 1.81 is also needed for https://github.com/alloy-rs/alloy/pull/1552 so maybe we should wait depending on the prio level

mattsse:
we shouldn't embedd this into the header because OP also uses the same type and should treat these consensus checks separately

PR #1573, opened 2024-10-26
feat: embed consensus header into RPC

klkvr:
Changes in this PR:
- `size` is moved from block body to header. we don't have a way to distinct which fields in RPC belong to block body vs header but moving size to header leaves RPC block fields as header + flattened body which makes sense imo
- `rpc_types_eth::Header` now wraps an inner consensus header and additionaly stores `hash`, `totalDifficulty` and `size` which are RPC-specific
- the previous header type without RPC-specific fields is now `AnyHeader` in alloy-consensus crate which is the same as `alloy_consensus::Header` with `mixHash` and `nonce` being optional.
- `HeaderResponse` trait is removed and `Network::HeaderResponse` is now required to implement `AsRef<Network::Header>` making it possible to simply use methods from consensus `BlockHeader` trait everywhere


crates/consensus/src/block/any.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use super::BlockHeader;
+
+/// Block header representation with certain fields made optional to account for possible
+/// differencies in network implementations.
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[derive(Clone, Debug, Default, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct AnyHeader {
```
mattsse, line 11: ah yeah we def need this for foundry

crates/network/src/any/mod.rs
```diff
@@ -65,15 +65,15 @@ impl Network for AnyNetwork {
...
+    type Header = alloy_consensus::AnyHeader;
 
     type TransactionRequest = WithOtherFields<TransactionRequest>;
 
     type TransactionResponse = WithOtherFields<Transaction>;
 
     type ReceiptResponse = AnyTransactionReceipt;
 
-    type HeaderResponse = Header;
+    type HeaderResponse = alloy_rpc_types_eth::Header<Self::Header>;
```
mattsse, line 76: nice

mattsse (approved):
this will unlock a __ton__ of simplifications

Issue #1557, opened 2024-10-24
[Feature] Make prague field and enum


mattsse:
### Component

rpc

### Describe the feature you would like

starting with prague the CL sends a list of requests alongside the actual payload:

https://github.com/alloy-rs/alloy/blob/b621af07b7bd970d256c523c90ec0c91ec61cea1/crates/rpc-types-engine/src/sidecar.rs#L16-L18

the requests are never stored be the EL because not part of the block.
to mimick consensus, mostly for testing this makes it difficult and we want a way to provide just the requests hash instead

## TODO
* introduce an enum that has either a list of requests or an hash
* expose a requests_hash function on that enum

PR #1574, opened 2024-10-27
feat(eips): Make prague field an enum

DoTheBestToGetTheBest:
closes https://github.com/alloy-rs/alloy/issues/1557


mattsse (changes requested):
cool, can we also integrate this into the ExecutionPayloadSidecar type?

crates/eips/src/eip7685.rs
```diff
@@ -76,7 +76,29 @@ impl Requests {
         self.0.extend(other.take());
     }
 }
+/// A list of requests or a precomputed requests hash.
```
tcoratger, line 79:
```suggestion
}

/// A list of requests or a precomputed requests hash.
```

crates/eips/src/eip7685.rs
```diff
@@ -76,7 +76,29 @@ impl Requests {
...
+    /// Stores a precomputed requests hash.
+    Hash(B256),
+}
 
+impl PragueRequests {
+    /// Returns the requests hash for the enum instance.
+    ///
+    /// - If the instance contains a list of requests, this function calculates the hash using
+    ///   `requests_hash` of the `Requests` struct.
+    /// - If it contains a precomputed hash, it returns that hash directly.
```
tcoratger, line 93:
```suggestion
    /// Returns the requests hash for the enum instance.
    ///
    /// - If the instance contains a list of requests, this function calculates the hash using
    ///   `requests_hash` of the [`Requests`] struct.
    /// - If it contains a precomputed hash, it returns that hash directly.
```

DoTheBestToGetTheBest:
> cool, can we also integrate this into the ExecutionPayloadSidecar type?

done thanks for the review

crates/eips/src/eip7685.rs
DoTheBestToGetTheBest: thanks @tcoratger

crates/eips/src/eip7685.rs
DoTheBestToGetTheBest: ty!

crates/eips/src/eip7685.rs
```diff
@@ -77,6 +77,30 @@ impl Requests {
     }
 }
 
+/// A list of requests or a precomputed requests hash.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum PragueRequests {
```
onbjerg, line 83: please find a better name like `RequestsOrHash`, this is very vague

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -51,6 +51,10 @@ impl ExecutionPayloadSidecar {
 
     /// Returns the EIP-7685 requests
     pub const fn requests(&self) -> Option<&Requests> {
-        self.prague.as_ref()
+        if let Some(PragueRequests::Requests(ref requests)) = self.prague {
```
onbjerg, line 54: please update the doc comment to reflect that this might not be requests anymore

onbjerg (changes requested):

crates/eips/src/eip7685.rs
onbjerg: additionally, we should note that the hash is for testing purposes only, so please add more doc comments on why this enum exists

crates/rpc-types-engine/src/sidecar.rs
onbjerg: this also isn't super helpful for the motivation of the issue; we should probably add a `requests_hash` method as well on the sidecar.

DoTheBestToGetTheBest:
@onbjerg hey sorry for the delay i'm having internet problem at home, does the change look good to you ?

yash-atreya (approved):
lgtm! cc @mattsse

crates/eips/src/eip7685.rs
```diff
@@ -77,6 +77,35 @@ impl Requests {
     }
 }
 
+/// A list of requests or a precomputed requests hash.
+///
+/// For testing purposes, the `Hash` variant stores a precomputed requests hash. This can be useful
+///   when the exact contents of the requests are unnecessary, and only a consistent hash value is
+/// needed   to simulate the presence of requests without holding actual data.
```
onbjerg, line 84: the whitespaces in this comment are a bit funky, please reformat it :D

crates/eips/src/eip7685.rs
```diff
@@ -77,6 +77,35 @@ impl Requests {
...
+
```
onbjerg, line 85:
```suggestion
```

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -51,6 +51,20 @@ impl ExecutionPayloadSidecar {
...
+
+    /// Calculates or retrieves the requests hash.
+    ///
+    /// - If the `prague` field contains a list of requests, it calculates the requests hash
+    ///   dynamically.
+    /// - If it contains a precomputed hash (used for testing), it returns that hash directly.
+
+    pub fn requests_hash(&self) -> B256 {
+        self.prague.as_ref().unwrap().requests_hash()
     }
```
onbjerg, line 69: Please make this an `Option<B256>`, we don't want to unwrap here. In the case `self.prague` is `None`, we should just return `None`.

crates/rpc-types-engine/src/sidecar.rs
```diff
@@ -51,6 +51,20 @@ impl ExecutionPayloadSidecar {
...
+            None
+        }
+    }
+
+    /// Calculates or retrieves the requests hash.
+    ///
+    /// - If the `prague` field contains a list of requests, it calculates the requests hash
+    ///   dynamically.
+    /// - If it contains a precomputed hash (used for testing), it returns that hash directly.
+
```
onbjerg, line 66:
```suggestion
```

onbjerg (dismissed):

crates/rpc-types-engine/src/sidecar.rs
DoTheBestToGetTheBest: > Please make this an `Option<B256>`, we don't want to unwrap here. In the case `self.prague` is `None`, we should just return `None`.

done

crates/eips/src/eip7685.rs
DoTheBestToGetTheBest: > the whitespaces in this comment are a bit funky, please reformat it :D

done

klkvr (approved):

PR #1577, opened 2024-10-28
fix: relay types

klkvr:
## Motivation

actualize relay types based on https://github.com/flashbots/builder/blob/03ee71cf0a344397204f65ff6d3a917ee8e06724/eth/block-validation/api.go and https://github.com/flashbots/rbuilder/blob/7ff6fe15d0a108f16b2301fcc13e534d7fbd3967/crates/rbuilder/src/mev_boost/mod.rs

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-beacon/src/relay.rs
+++ b/crates/rpc-types-beacon/src/relay.rs
@@ -5,7 +5,7 @@
 use crate::{BlsPublicKey, BlsSignature};
 use alloy_primitives::{Address, Bytes, B256, U256};
 use alloy_rpc_types_engine::{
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this seems correct:

https://github.com/attestantio/go-builder-client/blob/master/api/deneb/submitblockrequest.go

I think we can also remove SubmitBlockRequest to avoid confusion?

klkvr:
> I think we can also remove SubmitBlockRequest to avoid confusion?

yep it's removed

mattsse (approved):

mattsse (approved):

PR #1578, opened 2024-10-28
feat(consensus): Protected Legacy Signature

refcell:
### Description

Adds a method to the `TxEnvelope` type that returns if the associated signature is protected.

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -190,6 +190,22 @@ impl TxEnvelope {
         matches!(self, Self::Eip7702(_))
     }
 
+    /// Returns true if the signature of the transaction is protected.
+    #[inline]
+    pub const fn is_protected(&self) -> bool {
+        match self {
+            Self::Legacy(tx) => {
+                let v = tx.signature().v().to_u64();
+                if 64 - v.leading_zeros() <= 8 {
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -190,6 +190,22 @@ impl TxEnvelope {
...
+                let v = tx.signature().v().to_u64();
+                if 64 - v.leading_zeros() <= 8 {
+                    return v != 27 && v != 28 && v != 1 && v != 0;
+                }
+                // anything not 27 or 28 is considered protected
+                true
+            }
+            _ => true,
+        }
+    }
```
klkvr, line 207: should this be `false` for non-legacy?

refcell: No, see [the reference impl in kona](https://github.com/anton-rs/kona/blob/main/crates/derive/src/batch/span_batch/utils.rs#L62-L75)

refcell: [Here](https://github.com/ethereum-optimism/optimism/blob/0cb0cd6f44ed4fb5b8cc312c8e4981b03788bb42/op-node/rollup/derive/span_batch_txs.go#L395-L402) is the reference impl in optimism's `op-node`

klkvr (approved):

PR #1581, opened 2024-10-29
fix: replay protection checks

prestwich:
followup to #1578 

## Motivation

#1578 is poorly documented, doesn't specify what's protected in the API, and includes math that is already implemented elsewhere

## Solution

Clean up
- add documentation with relevant EIP link
- make naming unambiguous ("protected from what?")
- use functions math from the existing `Parity` object

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
much better, ty

PR #1582, opened 2024-10-29
fix(alloy-provider): `get_block_by_number` arg

steph-rs:
## Motivation
Different args in `alloy-provider` crate: 
```
    /// Gets a block by its [BlockHash], with full transactions or only hashes.
    async fn get_block_by_hash(
        &self,
        hash: BlockHash,
        kind: BlockTransactionsKind,    <--
    )
    ...
    /// Get a block by its number.
    // TODO: Network associate
    async fn get_block_by_number(
        &self,
        number: BlockNumberOrTag,
        hydrate: bool,                  <--
    )
```

## Solution

Changing `hydrate` argument to `kind` to make methods symmetrical.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (approved):
makes sense to me. noting that this is a breaking change.

PR #1590, opened 2024-10-30
eip4895: make `Vec<Withdrawal>` pub in `Withdrawals`

tcoratger:
Related https://github.com/paradigmxyz/reth/pull/12119#discussion_r1822621079

### DIFF
--- a/crates/eips/src/eip4895.rs
+++ b/crates/eips/src/eip4895.rs
@@ -39,7 +39,7 @@ impl Withdrawal {
     }
 }
 
-/// Represents a collection of Withdrawals.
+/// Represents a collection of withdrawals.
 #[derive(
     Debug,
     Clone,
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip4895.rs
```diff
@@ -57,32 +57,32 @@ impl Withdrawal {
 )]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-pub struct Withdrawals(Vec<Withdrawal>);
+pub struct Withdrawals(pub Vec<Withdrawal>);
```
onbjerg, line 60: I personally think this makes the API a bit confused, either we make this `pub` and remove the `iter`/`into_inner` or we continue as the API is today and add `as_ref`. there's no reason to have eg `iter` if this is `pub` as you can now just do `withdrawals.0.iter()`

tcoratger: I agree and I think that we can remove iter/into_inner, will make everything more generic with less functions. Let me know what you think and I can adapt

klkvr: actually we already have `AsRef` for this (through derive), so I believe we can use it in reth and avoid any changes here?

PR #1597, opened 2024-10-31
feat(`serde`): StorageKeyKind

### LINKED ISSUE #1553: [Feature] Support truncated key for JsonStorageKey
### Component

rpc, serde

https://github.com/alloy-rs/alloy/blob/b2278c40b2693908e4e5108d65ade26e8d716765/crates/serde/src/storage.rs#L31-L31

ref https://github.com/ethereum/go-ethereum/blob/a5fe7353cff959d6fcfcdd9593de19056edb9bdb/internal/ethapi/api.go#L741-L744

this exists for backwards compatiblity, which I think we don't need to focus on but if the input is a 32bytes hex string then this should be preserved in the serde roundtrip

this can be solved with a helper enum that is similar to:

https://github.com/alloy-rs/alloy/blob/b2278c40b2693908e4e5108d65ade26e8d716765/crates/eips/src/eip1898.rs#L665-L674

but with B256,U256, B256 must be decoded first

## TODO
* introduce helper enum
* replace wrapped value of JsonStorageKey

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #1553 

## Solution

Introduce the enum `StorageKeyKind` with variants `Hash(B256)` and `Number(U256)`. Used in `JsonStorageKey`.
Tries to deserialize a string into the `Hash` variant first.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/serde/src/storage.rs
+++ b/crates/serde/src/storage.rs
@@ -1,27 +1,9 @@
 use core::str::FromStr;
 
-use alloc::{
-    collections::BTreeMap,
-    fmt::Write,
-    string::{String, ToString},
-};
-use alloy_primitives::{Bytes, B256, U256};
+use alloc::collections::BTreeMap;
+use alloy_primitives::{ruint::ParseError, Bytes, B256, U256};
+use core::fmt::{Display, Formatter};
 use serde::{Deserialize, Deserializer, Serialize};
-
-/// A storage key kind that can be either [B256] or [U256].
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub enum StorageKeyKind {
-    /// A full 32-byte key (tried first during deserialization)
-    Hash(B256),
-    /// A number (fallback if B256 deserialization fails)
-    Number(U256),
-}
-
-impl Default for StorageKeyKind {
-    fn default() -> Self {
-        Self::Number(U256::ZERO)
-    }
-}
 
 /// A storage key type that can be serialized to and from a hex string up to 32 bytes. Used for
 /// `eth_getStorageAt` and `eth_getProof` RPCs.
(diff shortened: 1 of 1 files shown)


crates/serde/src/storage.rs
```diff
@@ -26,13 +43,31 @@ use serde::{Deserialize, Deserializer, Serialize};
 ///
 /// The contained [B256] and From implementation for String are used to preserve the input and
 /// implement this behavior from geth.
-#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(from = "U256", into = "String")]
-pub struct JsonStorageKey(pub B256);
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize)]
+#[serde(into = "String")]
```
mattsse, line 47: @Rjected isn't all of this now just serde(untagged) ?

crates/serde/src/storage.rs
```diff
@@ -26,13 +43,31 @@ use serde::{Deserialize, Deserializer, Serialize};
...
+pub struct JsonStorageKey(pub StorageKeyKind);
```
mattsse, line 48: this type needs a function to get the value as B256

mattsse (changes requested):

crates/serde/src/storage.rs
yash-atreya: Addressed in [c6e85f9](https://github.com/alloy-rs/alloy/pull/1597/commits/c6e85f97d6df336a761804042f8793971d6fe397)

crates/serde/src/storage.rs
```diff
@@ -6,6 +8,21 @@ use alloc::{
 use alloy_primitives::{Bytes, B256, U256};
 use serde::{Deserialize, Deserializer, Serialize};
 
+/// A storage key kind that can be either [B256] or [U256].
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum StorageKeyKind {
```
mattsse, line 13: I think this is now just #[serde(untagged)] then we no longer need any serialization on JsonStorageKey

mattsse (changes requested):

crates/serde/src/storage.rs
Rjected: yeah I think so

mattsse (approved):
I simplified this and converted the jsonstoragekey type to an enum instead

PR #1600, opened 2024-10-31
feat: EIP-7742

klkvr:
## Motivation

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7742.md
https://github.com/ethereum/execution-apis/pull/574
https://github.com/ethereum/EIPs/pull/8994

- Added new mod to alloy-eips with utilities for blob fee calculation.
- Extended `PayloadAttributes` with `target_blobs_per_block` and `max_blobs_per_block`.

Some PRs are not merged and field names are different per different specs for now, but besides that this should be complete

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -221,21 +221,18 @@ impl Header {
     ///
     /// Returns `None` if `excess_blob_gas` is None.
     ///
-    /// If `next_block_target_blobs_per_block` is [`Some`], uses EIP-7742 formula for calculating
+    /// If [`Self::target_blobs_per_block`] is [`Some`], uses EIP-7742 formula for calculating
     /// the blob gas price, otherwise uses EIP-4844 formula.
     ///
     /// See also [Self::next_block_excess_blob_gas]
-    pub fn next_block_blob_fee(
-        &self,
-        next_block_target_blobs_per_block: Option<u64>,
-    ) -> Option<u128> {
+    pub fn next_block_blob_fee(&self) -> Option<u128> {
         let next_block_excess_blob_gas = self.next_block_excess_blob_gas()?;
-        Some(next_block_target_blobs_per_block.map_or_else(
-            || eip4844::calc_blob_gasprice(next_block_excess_blob_gas),
-            |target_blobs_per_block| {
-                eip7742::calc_blob_gasprice(next_block_excess_blob_gas, target_blobs_per_block)
-            },
-        ))
+
+        if self.target_blobs_per_block().is_none() {
+            Some(eip4844::calc_blob_gasprice(next_block_excess_blob_gas))
+        } else {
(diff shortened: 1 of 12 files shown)


crates/eips/src/eip7742.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Controls the update rate of the blob base fee based on `target_blob_count`.
+pub const BLOB_BASE_FEE_UPDATE_FRACTION_PER_TARGET_BLOB: u64 = 1112825;
+
+/// Calculates the `excess_blob_gas` from the parent header's `blob_gas_used`, `excess_blob_gas` and
+/// `target_blob_count`.
+///
+/// Similar to [crate::eip4844::calc_excess_blob_gas], but derives the target blob gas from
+/// `parent_target_blob_count`.
+#[inline]
+pub const fn calc_excess_blob_gas(
```
mattsse, line 14: I'm a bit concerned that this will be easy to misuse if the functions have the same name, but since this takes an additional param this is okay imo

mattsse (approved):
cool, the integration in reth will be slightly more tricky because we now ran out of bits 

https://github.com/paradigmxyz/reth/blob/eaac2aa2cfb6ea5a61eea2d9a36a8b66b7854ce4/crates/storage/codecs/src/alloy/header.rs#L58-L58

klkvr:
pushed latest changes as per https://github.com/ethereum/EIPs/pull/9047

crates/eips/src/eip7742.rs
klkvr: those are now completely identical :/ EIP-7742 just expects a normalized excess gas

should we rename the new one?

crates/consensus/src/block/header.rs
```diff
@@ -622,24 +679,54 @@ pub trait BlockHeader {
...
     fn extra_data(&self) -> &Bytes;
 
     /// Calculate excess blob gas for the next block according to the EIP-4844
     /// spec.
     ///
+    /// If [`BlockHeader::target_blobs_per_block`] is [`Some`], uses EIP-7742 formula for
+    /// calculating the excess blob gas, otherwise uses EIP-4844 formula.
+    ///
+    /// Note: this function will return incorrect (unnormalized, lower) value at EIP-7742 activation
+    /// block. If this is undesired, consider using [`eip7742::calc_excess_blob_gas_at_transition`].
```
klkvr, line 695: I think this should be OK asI'd expect all consumers of this to not know whether the next block activates 7742 if it's not an EL

mattsse (approved):
lgtm

crates/consensus/src/block/header.rs
klkvr: yep

PR #1602, opened 2024-11-01
chore: move txtype-specific builders to network-primitives

klkvr:
## Motivation

ref https://github.com/alloy-rs/op-alloy/pull/215

We need to import those traits from op-alloy-rpc-types, however, right now this breaks no-std support due to dependency on alloy-network.

## Solution

Move those traits to alloy-network-primitives which is no-std compatible

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense

PR #1608, opened 2024-11-02
fix: receipt status serde

klkvr:
## Motivation

ref https://github.com/foundry-rs/foundry/issues/9239#issuecomment-2453005357

Right now we will fail to deserialize receipt if both `status` and `root` are present.

## Solution

Update serde impl to deserialize `Eip658Value` as untagged enum with `status` and `root`. We prefer `status` as it is more common and adds useful context.

This also allowed to just use `serde(flatten)` for status field and remove manual serde for receipt.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -7,7 +7,7 @@ use derive_more::{DerefMut, From, IntoIterator};
 
 /// Receipt containing result of transaction execution.
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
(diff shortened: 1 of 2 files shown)


crates/consensus/src/receipt/receipts.rs
```diff
@@ -7,15 +7,15 @@ use derive_more::{DerefMut, From, IntoIterator};
...
+#[cfg_attr(feature = "serde", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 #[doc(alias = "TransactionReceipt", alias = "TxReceipt")]
 pub struct Receipt<T = Log> {
     /// If transaction is executed successfully.
     ///
     /// This is the `statusCode`
-    #[cfg_attr(feature = "serde", serde(alias = "root"))]
+    #[cfg_attr(feature = "serde", serde(flatten))]
```
mattsse, line 18: ah right alias doesn't quite work here

mattsse (approved):

PR #1610, opened 2024-11-03
rpc: derive `Default` for `JwtSecret`

tcoratger:
Should be useful for:
- Testing configurations
- Be able to derive `Default` for https://github.com/paradigmxyz/reth/blob/21d911abb2cf594834c1818d2f723718b5109f8b/crates/rpc/rpc-builder/src/auth.rs#L89-L97

### DIFF
--- a/crates/rpc-types-engine/src/jwt.rs
+++ b/crates/rpc-types-engine/src/jwt.rs
@@ -160,7 +160,7 @@ impl Default for Claims {
 /// for the JWT, which is included in the JWT along with its payload.
 ///
 /// See also: [Secret key - Engine API specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#key-distribution)
-#[derive(Copy, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/jwt.rs
```diff
@@ -160,7 +160,7 @@ impl Default for Claims {
 /// for the JWT, which is included in the JWT along with its payload.
 ///
 /// See also: [Secret key - Engine API specs](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#key-distribution)
-#[derive(Copy, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, PartialEq, Eq, Default)]
```
zerosnacks, line 163: One potential concern with defaulting a secret value is that it could make it easier for users to forget to fill it as it falls through

tcoratger: Yes I agree, as you prefer, feel free to merge it or not... I think rn could be a blocker for deriving Default in some objects using this, like I mention in the pr description. But I understand your concern.

mattsse (commented):
yeah I'd like to prevent any misuse of default for this

PR #1620, opened 2024-11-05
fix(`provider`): make `Caller` `EthCall` specific

### LINKED ISSUE #1471: provider: make `Caller` more `EthCall` specific
given we're now only using this for `EthCall`, I think we should make it more call-specific, eg make it accept `TransactionRequest`, block and overrides directly to allow implementations to directly call underlying evm with them.

_Originally posted by @klkvr in https://github.com/alloy-rs/alloy/pull/788#discussion_r1748862509_

yash-atreya:
## Motivation

Closes #1471 

## Solution

- Change the `data` and `overrides` fields in `EthCallParams` to be `Cow`. 
- Accept `EthCallParams` as arg in `Caller.call(&self, method, params: EthCallParams<'_, N>)`.
- `ProviderCall` should be used over `EthCallParams`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/caller.rs
+++ b/crates/provider/src/provider/caller.rs
@@ -20,10 +20,10 @@ where
     ///
     /// This method handles serialization of the params and sends the request to relevant data
     /// source and returns a `ProviderCall`.
-    fn call<'req>(
+    fn call(
         &self,
         method: Cow<'static, str>,
-        params: EthCallParams<'req, N>,
+        params: EthCallParams<'_, N>,
     ) -> TransportResult<ProviderCall<T, serde_json::Value, Resp>>;
 }
 
--- a/crates/provider/src/provider/caller.rs
+++ b/crates/provider/src/provider/caller.rs
@@ -15,8 +15,7 @@ where
 {
     /// Method that needs to be implemented to convert to a `ProviderCall`.
     ///
-    /// This method handles serialization of the params and sends the request to relevant data
-    /// source and returns a `ProviderCall`.
+    /// This method sends the request to relevant data source and returns a `ProviderCall`.
     fn call(
         &self,
         method: Cow<'static, str>,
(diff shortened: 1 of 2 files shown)


crates/provider/src/provider/caller.rs
```diff
@@ -1,46 +1,42 @@
...
+    N: Network,
     Resp: RpcReturn,
 {
     fn call(
         &self,
         method: Cow<'static, str>,
-        params: Params,
-    ) -> TransportResult<ProviderCall<T, serde_json::Value, Resp>> {
+        params: EthCallParams<'_, N>,
+    ) -> TransportResult<ProviderCall<T, EthCallParams<'static, N>, Resp>> {
```
klkvr, line 36: would be nice to have either enum for method or just separate `call`/`estimate_gas` methods so that custom implementations doesn't have to match on strings

klkvr (approved):
approach lgtm, this is already pretty useful as is, just left a couple comments

mattsse (approved):

crates/provider/src/provider/caller.rs
```diff
@@ -1,47 +1,62 @@
...
-    Params: RpcParam,
+    N: Network,
     Resp: RpcReturn,
 {
     /// Method that needs to be implemented to convert to a `ProviderCall`.
     ///
-    /// This method handles serialization of the params and sends the request to relevant data
-    /// source and returns a `ProviderCall`.
+    /// This method sends the request to relevant data source and returns a `ProviderCall`.
     fn call(
```
klkvr, line 18: @yash-atreya could we make `Caller::call` always output `Bytes` and `estimate_gas` would output u64? `Resp` generic might be tricky to deal with in custom implementations

PR #1621, opened 2024-11-05
feat(rpc-types-trace/parity): add creationMethod for create action

jsvisa:
## Motivation

align with geth's flatcall trace, ref ethereum/go-ethereum#30539

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-types-trace/src/parity.rs
```diff
@@ -300,8 +300,23 @@ pub struct CallAction {
     pub value: U256,
 }
 
+/// Creation method.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum CreationMethod {
+    /// None
+    #[default]
+    None,
```
mattsse, line 309: is this actually a valid variant?

can't see any `"none"` in the referenced pr

should this be an Option instead and serialize null?

jsvisa: I copied this field from `CallType`, as it also has the `None` variant

```rust
pub enum CallType {
    /// None
    #[default]
    None,
    /// Call
    Call,
```

mattsse (approved):

PR #1579, opened 2024-10-29
feat: serde helpers for hashmaps and btreemaps with quantity key types

stevencartavia:
Resolves #1502 

## Motivation
Existing `mod vec` implementation

## Solution
Created the same for hashmaps/btreemaps with quantity key types and tests.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #1622, opened 2024-11-05
eip5792: use `ChainId` in `WalletCapabilities`

tcoratger:
After https://github.com/alloy-rs/alloy/pull/1579 we can directly use `ChainId` (u64) instead of U64 big int in `WalletCapabilities`.


crates/consensus/Cargo.toml
```diff
@@ -23,6 +23,7 @@ alloy-eips = { workspace = true, features = ["kzg-sidecar"] }
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-serde = { workspace = true, optional = true }
+alloy-genesis.workspace = true
```
tcoratger, line 26:
```suggestion
```

mattsse (approved):

PR #1629, opened 2024-11-06
fix(`rpc-types`): `FeeHistory` deser

yash-atreya:
## Motivation

Currently we don't correctly handle a `eth_feeHistory` response in which `gasUsedRatio` is `null`.

## Solution

Handle this case by deserializing to an empty vec. 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/fee.rs
+++ b/crates/rpc-types-eth/src/fee.rs
@@ -45,6 +45,7 @@ pub struct FeeHistory {
     pub base_fee_per_gas: Vec<u128>,
     /// An array of block gas used ratios. These are calculated as the ratio
     /// of `gasUsed` and `gasLimit`.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "gas_used_ratio_deser::deserialize"))]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/fee.rs
```diff
@@ -45,6 +45,7 @@ pub struct FeeHistory {
     pub base_fee_per_gas: Vec<u128>,
     /// An array of block gas used ratios. These are calculated as the ratio
     /// of `gasUsed` and `gasLimit`.
+    #[cfg_attr(feature = "serde", serde(deserialize_with = "gas_used_ratio_deser::deserialize"))]
```
mattsse, line 48: I believe we have this in alloy-serde:

https://github.com/alloy-rs/alloy/blob/caa380d188da17d3795b7d4f540609c1b64ff3c5/crates/serde/src/optional.rs#L7-L7

mattsse (approved):
makes sense, 

since this is just an rpc response type we can do this, but we have this helper already

PR #1630, opened 2024-11-07
fix: serde for transactions

klkvr:
## Motivation

Makes all transactions to always serialize `to: null` instead of omitting the field.

Adds deprecated `gaaPrice` field to RPC transaction which is only serialized if inner tx does not have its own `gasPrice`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/mod.rs
+++ b/crates/rpc-types-eth/src/transaction/mod.rs
@@ -36,24 +36,30 @@ pub use alloy_consensus::{
 #[derive(Clone, Debug, Default, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(all(any(test, feature = "arbitrary"), feature = "k256"), derive(arbitrary::Arbitrary))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(
+    feature = "serde",
+    serde(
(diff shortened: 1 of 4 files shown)


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -259,6 +265,108 @@ impl<T: TransactionTrait + Encodable2718> TransactionResponse for Transaction<T>
...
+mod tx_serde {
+    //! Helper module for serializing and deserializing OP [`Transaction`].
+    //!
+    //! This is needed because we might need to deserialize the `gasPrice` field into both
+    //! [`crate::Transaction::effective_gas_price`] and [`alloy_consensus::TxLegacy::gas_price`].
+    use super::*;
+    use serde::{Deserialize, Serialize};
+
+    /// Helper struct which will be flattened into the transaction and will only contain `gasPrice`
+    /// field if inner [`TxEnvelope`] did not consume it.
```
mattsse, line 278: is this a serde flatten hack?

mattsse (approved):
lgtm, could use some additional test cases from hive

crates/rpc-types-eth/src/transaction/mod.rs
klkvr: yep, it relies on it being deserialized after the inner tx. if it was just a field it would always be deserialized and consumed before

PR #1631, opened 2024-11-07
fix(`network`): impl `NetworkWallet<AnyNetwork>`

yash-atreya:
## Motivation

In https://github.com/alloy-rs/alloy/pull/1460, `AnyTxEnvelope` and `AnyTypedTransaction` were introduced. These are now being used in `AnyNetwork`. 

This made initializing an `AnyNetwork` provider with the `EthereumWallet` wallet filler incompatible as it signs `TypedTransaction` and returns the `TxEnvelope`.

## Solution

Implmenent `NetworkWallet<AnyNetwork>` for `EthereumWallet` that sign `AnyTypedTransactions`

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


klkvr (changes requested):
I don't think we need a separate type here given that signable transactions set is still the same

In op-alloy I've done this https://github.com/alloy-rs/op-alloy/blob/01f11b4747f3b533f947be19188099cb21182a1a/crates/network/src/lib.rs#L198 which allows to reuse `EthereumWallet` for OP as well

yash-atreya:
> I don't think we need a separate type here given that signable transactions set is still the same
> 
> In op-alloy I've done this https://github.com/alloy-rs/op-alloy/blob/01f11b4747f3b533f947be19188099cb21182a1a/crates/network/src/lib.rs#L198 which allows to reuse `EthereumWallet` for OP as well

@klkvr addressed in [55144c2](https://github.com/alloy-rs/alloy/pull/1631/commits/55144c20dbf4c801702271298ca8dff26613364f)

crates/network/src/ethereum/wallet.rs
```diff
@@ -139,3 +139,56 @@ where
...
+                TypedTransaction::Eip4844(mut t) => {
+                    let sig = self.sign_transaction_inner(sender, &mut t).await?;
+                    let signed = t.into_signed(sig);
+                    Ok(AnyTxEnvelope::Ethereum(signed.into()))
+                }
+                TypedTransaction::Eip7702(mut t) => {
+                    let sig = self.sign_transaction_inner(sender, &mut t).await?;
+                    let signed = t.into_signed(sig);
+                    Ok(AnyTxEnvelope::Ethereum(signed.into()))
+                }
```
klkvr, line 189:
```suggestion
                NetworkWallet::<Ethereum>::sign_transaction_from(self, sender, t).await?
```

crates/network/src/ethereum/wallet.rs
```diff
@@ -139,3 +139,56 @@ where
...
+            },
+            _ => unimplemented!("cannot sign UnknownTypedTransaction"),
```
klkvr, line 191:
```suggestion
            _ => Err(alloy_signer::Error::other("cannot sign UnknownTypedTransaction")),
```

klkvr (approved):

PR #1638, opened 2024-11-12
Add trait method `Transaction::is_dynamic_fee`

emhane:
## Motivation

Migrate reth utility method https://github.com/paradigmxyz/reth/pull/12437#discussion_r1837882648

## Solution

Adds trait method `alloy_consensus::Transaction::is_dynamic_fee`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/any/unknowns.rs
+++ b/crates/network/src/any/unknowns.rs
@@ -146,6 +146,11 @@ impl alloy_consensus::Transaction for UnknownTypedTransaction {
         self.gas_price().or(self.max_priority_fee_per_gas()).unwrap_or_default()
     }
 
+    fn is_dynamic_fee(&self) -> bool {
+        self.fields.get_deserialized::<U128>("maxFeePerGas").is_some()
+            || self.fields.get_deserialized::<U128>("maxFeePerBlobGas").is_some()
+    }
+
     fn kind(&self) -> TxKind {
         self.fields
             .get("to")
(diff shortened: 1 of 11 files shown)


crates/network/src/any/unknowns.rs
```diff
@@ -146,6 +146,11 @@ impl alloy_consensus::Transaction for UnknownTypedTransaction {
         self.gas_price().or(self.max_priority_fee_per_gas()).unwrap_or_default()
     }
 
+    fn is_dynamic_fee(&self) -> bool {
+        self.fields.get_deserialized::<U128>("maxFeePerGas").is_some()
+            || self.fields.get_deserialized::<U128>("maxFeePerBlobGas").is_some()
```
mattsse, line 151: can we use the `self.ty` here?

mattsse (changes requested):
smol nit

crates/network/src/any/unknowns.rs
emhane: how? I thought ab it too, but it would be incorrect for eg deposit tx to do smthg like `self.ty() > 1`

mattsse: oh, yeah we'd only catch known types, which defeats the purpose here

mattsse (approved):

PR #1640, opened 2024-11-12
Add trait method `Transaction::effective_gas_price`

emhane:
## Motivation

Port utility method from reth https://github.com/paradigmxyz/reth/blob/3a337cd7d4ad8c14569d20141fdcc2b1c54dbdf3/crates/primitives/src/transaction/mod.rs#L329-L342

## Solution

Adds trait method `Transaction::effective_gas_price`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -185,20 +185,17 @@ impl Transaction for TxEip1559 {
     }
 
     fn effective_gas_price(&self, base_fee: Option<u64>) -> u128 {
-        match base_fee {
-            None => self.max_fee_per_gas,
-            Some(base_fee) => {
-                // if the tip is greater than the max priority fee per gas, set it to the max
-                // priority fee per gas + base fee
-                let tip = self.max_fee_per_gas.saturating_sub(base_fee as u128);
-                if tip > self.max_priority_fee_per_gas {
-                    self.max_priority_fee_per_gas + base_fee as u128
-                } else {
-                    // otherwise return the max fee per gas
-                    self.max_fee_per_gas
-                }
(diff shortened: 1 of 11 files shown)


crates/network/src/any/unknowns.rs
```diff
@@ -146,6 +146,28 @@ impl alloy_consensus::Transaction for UnknownTypedTransaction {
         self.gas_price().or(self.max_priority_fee_per_gas()).unwrap_or_default()
     }
 
+    fn effective_gas_price(&self, base_fee: Option<u64>) -> u128 {
+        self.gas_price().unwrap_or(match base_fee {
```
mattsse, line 150: I'd rather use if let here than sequeezing the match in an `or`

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -202,6 +202,10 @@ impl Transaction for TxEip1559 {
         self.max_priority_fee_per_gas
     }
 
+    fn effective_gas_price(&self, base_fee: Option<u64>) -> u128 {
+        self.effective_gas_price(base_fee)
+    }
```
mattsse, line 207: this looks weird

can be unified?

crates/consensus/src/transaction/eip4844.rs
```diff
@@ -607,6 +614,10 @@ impl Transaction for TxEip4844 {
         self.max_priority_fee_per_gas
     }
 
+    fn effective_gas_price(&self, base_fee: Option<u64>) -> u128 {
+        self.effective_gas_price(base_fee)
```
mattsse, line 618: same

mattsse (changes requested):
nits

mattsse (approved):

PR #1642, opened 2024-11-13
Inline getters in impl of `Transaction`

emhane:
## Motivation

Avoid call overhead for transaction getters, also for compilation without LTO. Getters are expected to be hot paths with generalisation of data primitives in reth. To my understanding, that may make profiling rather cluttered and less meaningful (since releases are built with maxperf `lto = fat`).

## Solution

Inline getters in impl of `alloy_consensus::Transaction`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):

harunfox72s:
@harunfox72..

https://github.com/BWS-fox

PR #1649, opened 2024-11-14
feat(debug): add debug_executionWitness to debug api

Rjected:
## Motivation

I noticed that `debug_executionWitness` only existed on the server side, and not on the alloy client side.

## Solution

Add the `debug_executionWitness` method to the `DebugApi`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/debug.rs
+++ b/crates/provider/src/ext/debug.rs
@@ -2,6 +2,7 @@
 use crate::Provider;
 use alloy_network::Network;
 use alloy_primitives::{hex, Bytes, TxHash, B256};
+use alloy_rpc_types_debug::ExecutionWitness;
 use alloy_rpc_types_eth::{
     BadBlock, BlockId, BlockNumberOrTag, Bundle, StateContext, TransactionRequest,
 };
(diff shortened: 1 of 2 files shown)


crates/provider/src/ext/debug.rs
```diff
@@ -128,6 +128,23 @@ pub trait DebugApi<N, T>: Send + Sync {
...
+    /// The first argument is the block number or block hash. The second argument is a boolean
+    /// indicating whether to include the preimages of keys in the response.
+    ///
+    /// # Note
+    ///
+    /// Not all nodes support this call.
+    async fn debug_execution_witness(
+        &self,
+        block: BlockNumberOrTag,
+        include_preimages: bool,
```
mattsse, line 146: we removed the additional bool

https://github.com/paradigmxyz/reth/blob/870ffae9094e411343781f7a1dc29d503740f168/crates/rpc/rpc-api/src/debug.rs#L142-L142

mattsse (changes requested):

crates/provider/src/ext/debug.rs
Rjected: whoops, removed, looks like we need to update the reth doc comment as well

Rjected: https://github.com/paradigmxyz/reth/pull/12559

mattsse (approved):

PR #1650, opened 2024-11-15
consensus: add `BlockWithParent`

tcoratger:
As per:
- https://github.com/paradigmxyz/reth/issues/12493
- https://github.com/paradigmxyz/reth/pull/12514

We need to introduce `BlockWithParent` struct in alloy to use it directly in reth.


mattsse (changes requested):
I think we can add this to where BlockNumHash is defined, in alloy-eips, doesn't quite fit, but seems appropriate

tcoratger:
> I think we can add this to where BlockNumHash is defined, in alloy-eips, doesn't quite fit, but seems appropriate

@mattsse fixed

mattsse (approved):

PR #1654, opened 2024-11-17
fix(`consensus`): serde aliases to avoid breaking changes

yash-atreya:
## Motivation

Recent changes in `consensus` types brought breaking changes to serialization/deserialization usecases. 

See https://github.com/foundry-rs/foundry/pull/9280/files/999eea97817168b67318f15b3e58f33d378e8c96#r1844004048 and https://github.com/foundry-rs/foundry/issues/9324.

## Solution

Add aliases to avoid breaking changes

- Alias `sha3Uncles` with `ommersHash` and `miner` with `beneficiary` in `Header`.
- Alias `gas` with `gasLimit` in tx types. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #1655, opened 2024-11-17
fix(consensus): transaction types should be able to deserialize old name

dbeal-eth:
## Motivation

the `gas_limit` field in the transaction types was recently renamed to serialize to `gas` instead of `gasLimit`

However, this causes old state dumps prior to this change to fail to load (in fact, downstream `foundry` tests were modified in order to accomodate for this, although I really don't think this should have happened) https://github.com/foundry-rs/foundry/commit/7b118faeff4848be480f9c30c11237b9e9e6eb31 .

## Solution

added `alias` so that `serde` knows it can also read from `gasLimit` if it doesn't find `gas` field.

manually verified that downstream foundry seems to work better with the legacy tests reverted to previous after making these changes.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## Other

Regarding tests, the downstream foundry tests should be reverted to their previous state (https://github.com/foundry-rs/foundry/commits/7b118faeff4848be480f9c30c11237b9e9e6eb31/crates/anvil/test-data/state-dump-legacy.json and https://github.com/foundry-rs/foundry/commits/7b118faeff4848be480f9c30c11237b9e9e6eb31/crates/anvil/test-data/state-dump-legacy-stress.json) once this change gets downstreamed to verify the change. However, in this repo, it seems we don't yet have serialization tests that verify decoding from json or other similar language. is there a reccomended way to add this?

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -22,7 +22,7 @@ pub struct TxEip1559 {
     /// this transaction. This is paid up-front, before any
     /// computation is done and may not be increased
     /// later; formally Tg.
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity", rename = "gas"))]
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity", alias = "gasLimit", rename = "gas"))]
     pub gas_limit: u64,
     /// A scalar value equal to the maximum
     /// amount of gas that should be used in executing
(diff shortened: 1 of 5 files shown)


yash-atreya:
Resolved by #1654

dbeal-eth:
> Resolved by #1654

@yash-atreya  thanks for your attention on this!

do you know if the tests in the downstream foundry repo will be reverted back to their previous state? This issue would not have been propogated in the first place had the tests not been modified.

PR #1661, opened 2024-11-17
impl ConvertRuint for U256 to be compatible with alloy_serde

DougieHauser:
Adding the ConvertRuint trait to apply to the non-primitive type of U256,
this is a first step towards fixing [issue 85 in op-alloy](https://github.com/alloy-rs/op-alloy/issues/85)

This is my 1st contribution ever.. be gentle :)


DaniPopes:
Thanks but this is not needed, you can simply remove the `serde(with = "quantity")` attribute since `quantity` should only used for non-ruint integers

PR #1671, opened 2024-11-20
fix(`signer`): export PrimitiveSignature instead of deprecated sig

yash-atreya:
## Motivation

Currently we are still re-exporting deprecated `Signature` from `alloy-signer`

## Solution

Export `PrimitiveSignature` instead

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer/src/lib.rs
+++ b/crates/signer/src/lib.rs
@@ -14,7 +14,7 @@ pub use signer::{Signer, SignerSync};
 
 pub mod utils;
 
-pub use alloy_primitives::Signature;
+pub use alloy_primitives::PrimitiveSignature;
 pub use k256;
 
 /// Utility to get and set the chain ID on a transaction and the resulting signature within a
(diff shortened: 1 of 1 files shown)


crates/signer/src/lib.rs
```diff
@@ -14,7 +14,7 @@ pub use signer::{Signer, SignerSync};
 
 pub mod utils;
 
-pub use alloy_primitives::Signature;
+pub use alloy_primitives::PrimitiveSignature;
```
klkvr, line 17:
```suggestion
pub use alloy_primitives::PrimitiveSignature as Signature;
```

`PrimitiveSignature` is a temporary name. we wanted to avoid breaking changes in core so didn't replace the sig type right away, but this will happen on the next minor release

mattsse (approved):

PR #1672, opened 2024-11-21
feat: modifiy ReceiptWithBloom and associated impls to use with Reth

royvardhan:
Unblocks: https://github.com/paradigmxyz/reth/issues/12701

### DIFF
--- a/crates/rpc-types-eth/src/transaction/receipt.rs
+++ b/crates/rpc-types-eth/src/transaction/receipt.rs
@@ -1,6 +1,6 @@
 use crate::Log;
 use alloc::vec::Vec;
-use alloy_consensus::{Receipt, ReceiptEnvelope, TxReceipt, TxType};
+use alloy_consensus::{ReceiptEnvelope, TxReceipt, TxType};
 use alloy_eips::eip7702::SignedAuthorization;
 use alloy_network_primitives::ReceiptResponse;
 use alloy_primitives::{Address, BlockHash, TxHash, B256};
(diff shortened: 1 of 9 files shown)


royvardhan:
@mattsse As discussed, before updating `ReceiptWithBloom` type, I have updated TxReceipt trait and also changed the associated impls. Wdyt?

crates/rpc-types-eth/src/transaction/receipt.rs
```diff
@@ -95,7 +95,14 @@ impl TransactionReceipt {
...
     pub const fn transaction_type(&self) -> TxType {
-        self.inner.tx_type()
+        match &self.inner {
+            ReceiptEnvelope::Eip1559(_) => TxType::Eip1559,
+            ReceiptEnvelope::Eip2930(_) => TxType::Eip2930,
+            ReceiptEnvelope::Eip4844(_) => TxType::Eip4844,
+            ReceiptEnvelope::Eip7702(_) => TxType::Eip7702,
+            ReceiptEnvelope::Legacy(_) => TxType::Legacy,
+            _ => TxType::Legacy,
+        }
```
mattsse, line 105: why can we no longer use the function?

mattsse (commented):
looks good so far,

what's missing?

cc @klkvr

royvardhan:
@mattsse I think we are good.

mattsse (approved):
lgtm, 
this should be simple to port to op-alloy, I believe

royvardhan:
> lgtm, 
> 
> this should be simple to port to op-alloy, I believe

On it.

PR #1683, opened 2024-11-25
feat(consensus-tx): enable fast `is_create`

hai-rise:
## Motivation

`alloy-consensus::Transaction` implementors didn't have a cheap way to check if the transaction was a `Create`. They had to go through `kind` which copies the 21-byte `TxKind`. This may scale poorly for code that processes (hundreds of) thousands of transactions per second: https://github.com/paradigmxyz/reth/issues/12838.

## Solution

Add `is_create` to the trait that allows fast non-allocating implementation. I considered making `kind` returns `&TxKind` but transactions like EIP-4844, 7702, etc. only store an `Address` and it would be bad if we forced them to promote it to `TxKind`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -220,6 +220,11 @@ impl Transaction for TxEip1559 {
     }
 
     #[inline]
+    fn is_create(&self) -> bool {
+        self.to.is_create()
+    }
+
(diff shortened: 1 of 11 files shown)


mattsse (approved):
makes sense

Issue #1681, opened 2024-11-23
[Feature] Missing JSON-RPC methods


eshaan7:
### Component

network, json-rpc, rpc

### Describe the feature you would like

The following 4 methods are missing from RPC's `Provider` trait:
- [x] `eth_getFilterLogs` 
- [x] `eth_getBlockTransactionCountByHash`
- [x] `eth_getBlockTransactionCountByNumber`
- [x] `eth_uninstallFilter`

PR #1686, opened 2024-11-25
feat: get_block_transaction_count_by_hash in Provider trait

eshaan7:
Related: #1681.

## Motivation

Right now cannot call `get_block_transaction_count_by_hash` JSON-RPC method without using `raw_request` which is not ideal.

## Solution

Adds a new `get_block_transaction_count_by_hash` method to `Provider` trait.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -343,6 +343,7 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         Ok(block)
     }
 
+    /// Returns the number of transactions in a block from a block matching the given block hash.
     async fn get_block_transaction_count_by_hash(
         &self,
         hash: BlockHash,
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -343,6 +343,16 @@ pub trait Provider<T: Transport + Clone = BoxTransport, N: Network = Ethereum>:
         Ok(block)
     }
```
mattsse, line 345: needs docs:
 /// Returns the number of transactions in a block from a block matching the given block hash.

mattsse (changes requested):

crates/provider/src/provider/trait.rs
eshaan7: ah, good catch. ty.

eshaan7: fixed

mattsse (approved):

PR #1705, opened 2024-11-29
block_number_index added to callBundle reference type

ciaranmcveigh5:
## Motivation

Ability to quickly identify via an API call why a bundle did not land

https://github.com/paradigmxyz/reth/issues/12353

## Solution

Allow callBundle simulations to occur from a block index rather than a state_block_number, if a bundle has a target transaction that lands in block x this enables you to simulate from the exact index in block x where the target transaction landed to see if there were transactions higher in the block that caused a state conflict.

## PR Checklist

- Ran cargo test --doc - no tests present for alloy_rpc_types_mev
- No documentation present on alloy_rpc_types_mev outside of "Types for the MEV bundle JSON-RPC namespace."

### DIFF
--- a/crates/rpc-types-mev/src/eth_calls.rs
+++ b/crates/rpc-types-mev/src/eth_calls.rs
@@ -18,9 +18,9 @@ pub struct EthCallBundle {
     pub block_number: u64,
     /// Either a hex encoded number or a block tag for which state to base this simulation on
     pub state_block_number: BlockNumberOrTag,
-    /// The block number index to use for which state to base this simulation on (optional - overrides state_block_number)
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub block_number_index: Option<u64>,
+    /// Inclusive number of tx to replay in block. -1 means replay all 
+    #[serde(default, skip_serializing_if = "Option::is_none")] 
+    pub transaction_index: Option<TransactionIndex>, 
     /// the coinbase to use for this bundle simulation
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub coinbase: Option<Address>,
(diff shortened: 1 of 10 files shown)


crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -18,6 +18,9 @@ pub struct EthCallBundle {
     pub block_number: u64,
     /// Either a hex encoded number or a block tag for which state to base this simulation on
     pub state_block_number: BlockNumberOrTag,
+    /// The block number index to use for which state to base this simulation on (optional - overrides state_block_number)
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub block_number_index: Option<u64>,
```
mattsse, line 23: unclear to me what `index` refers to here, is this the transaction index in the block?

ciaranmcveigh5:
One thing to note, there are 2 options naming wise, which likely have implications on the implementation

Don't have a strong opinion on which is better


Option 1

- Naming: block_number_index
- Fields that must be present: block_number & block_number_index


Option 2

- Naming: state_block_number_index
- Fields that must be present: block_number, state_block_number & state_block_number_index

This requires you to update the state_block_number to be the same as the block_number and include all 3 fields

crates/rpc-types-mev/src/eth_calls.rs
ciaranmcveigh5: yep is the transaction index you will simulate from

ciaranmcveigh5: Say you wanted to simulate the bundle from index 3 in the block, you would need to update the top of block state with transactions at index 0, 1 & 2 to get the state aligned with that index

ciaranmcveigh5: variety of options

tx_index
state_block_index
state_block_position
block_position

Let me know what makes most sense

mattsse: so this is basically eth_callMany's:

https://github.com/alloy-rs/alloy/blob/67576f9f6e70026116443851a43e4c805902c3b3/crates/rpc-types-eth/src/call.rs#L35-L38

could we use the same instead?

mattsse (approved):
since this is just an extra setting, this should be fine

ciaranmcveigh5:
Just updated the imports for TransactionIndex

PR #1708, opened 2024-11-29
wip: receipt root

mattsse:
tried to make this work because would be useful for reth, but there are some conflicting designs.

we need 2718 for the root, the alloy `ReceiptEnvelope` already wraps `ReceiptWithBloom<Receipt>` reth's equivalent is `enum Receipt {EIP155(Receipt)}` and then we have 

https://github.com/paradigmxyz/reth/blob/b09c345257cda4a88e8e347654e946a20f9e5cb7/crates/primitives/src/receipt.rs#L418-L418

not quite sure yet, how to unify this on alloy


mattsse (commented):
lgtm

klkvr (approved):

Issue #1598, opened 2024-10-31
[Feature] Move `AnyNetwork` types into their own crate(s)


prestwich:
### Component

consensus, eips, genesis, network, json-rpc

### Operating System

None

### Describe the bug

per discussion here: https://github.com/alloy-rs/alloy/pull/1460#discussion_r1824406025

Currently `alloy-consensus` contains `Any` types for `Receipt` and `Header` that are then bubbled up through `alloy-rpc-types-eth` to `alloy-network`. These types are not `eth` types and it is somewhat messy to put them alongside the eth types. We have a standard pattern for adding networks to alloy, that  `AnyNetwork` is currently not following

Approximate steps to take:

- create `alloy-consensus-any`, depending on `alloy-consensus`
- create `alloy-rpc-types-any`, depending on `alloy-rpc-types-eth`
- move `AnyReceipt` and `AnyHeader` from `alloy-consensus` to `alloy-consensus-any`
- move `AnyTxEnvelope` from `alloy-network` to `alloy-consensus-any`
- create type aliases for `AnyRpcHeader` and `AnyRpcBlock` in `alloy-rpc-types-any`
- change `alloy-network` to depend on `alloy-consensus-any` and `alloy-rpc-types-any`

nadtech-hub:
I would like to take on it

TechFusionData:
I'd like to pick this up — the previous claim from Oct 2024 hasn't produced a PR. I can see that `alloy-consensus-any` and `alloy-rpc-types-any` crates already exist with `AnyHeader` and `AnyReceiptEnvelope` moved, so the remaining work appears to be moving `AnyTxEnvelope` from `alloy-network` to `alloy-consensus-any` and `AnyRpcBlock` from `alloy-network` to `alloy-rpc-types-any`, with re-exports maintained for backwards compatibility. Will submit a PR shortly.

PR #1609, opened 2024-11-03
feat: Move `AnyReceipt` and `AnyHeader` to `alloy-consensus-any`

moricho:
## Motivation

ref https://github.com/alloy-rs/alloy/issues/1598

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #1715, opened 2024-11-30
Move AnyNetwork related types into their own crates

moricho:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/1598
Continued from https://github.com/alloy-rs/alloy/pull/1609

> Currently `alloy-consensus` contains Any types for `Receipt` and `Header` that are then bubbled up through `alloy-rpc-types-eth` to `alloy-network`. These types are not eth types and it is somewhat messy to put them alongside the eth types. We have a standard pattern for adding networks to alloy, that `AnyNetwork` is currently not following, which is to create `consensus-*` and `rpc-types-*` crates

### `consensus-any`
- Moved `AnyTypedTransaction`, `AnyTxEnvelope` and `AnyTxType` to `consensus-any` from `network`
- Moved `UnknownTypedTransaction`, `UnknownTxEnvelope` to `consensus-any/unknown` from `network`

### `rpc-types-any`
- Moved `AnyRpcTransaction` and `AnyRpcBlock` to `rpc-types-any` from `network`
- Introduced a catch-all transaction request type `struct AnyTransactionRequest(pub WithOtherFields<TransactionRequest>)`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/consensus-any/src/transaction/unknown/typed.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// Transaction type.
+    pub ty: AnyTxType,
+
+    /// Additional fields.
+    #[cfg_attr(feature = "serde", serde(flatten))]
+    pub fields: OtherFields,
+
+    /// Memoization for deserialization.
+    #[cfg_attr(feature = "serde", serde(skip, default))]
+    pub memo: DeserMemo,
```
moricho, line 40: Due to the migration of `UnknownTypedTransaction` to `consensus-any`, a std-based implementation is now required in `consensus-any`, but should we change OnceCell to a no_std implementation

moricho: Changed to export this only if `std` feature is enabled: https://github.com/alloy-rs/alloy/pull/1715/commits/fd43c12537adceaf5c606dcb4631a42f6c0dfc8b

mattsse (commented):
we resolved the cyclic issue here differently by introducing dedicated Any types so this is now a non issue

PR #1717, opened 2024-11-30
feat: Add Account helpers

moricho:
## Motivation
`TrieAccount` in Reth will be replaced with `Account` from alloy, so it would be better for `TrieAccount` related helpers to be exported to alloy.
https://github.com/paradigmxyz/reth/issues/12318

https://github.com/paradigmxyz/reth/blob/c8e5b233ef242c4c5eaebe6dc5ae66b98184d8a8/crates/trie/common/src/root.rs#L10-L36

```rs
/// Hashes and sorts account keys, then proceeds to calculating the root hash of the state
/// represented as MPT.
/// See [`state_root_unsorted`] for more info.
pub fn state_root_ref_unhashed<'a, A: Into<TrieAccount> + Clone + 'a>(
    state: impl IntoIterator<Item = (&'a Address, &'a A)>,
) -> B256 {
    state_root_unsorted(
        state.into_iter().map(|(address, account)| (keccak256(address), account.clone())),
    )
}

/// Hashes and sorts account keys, then proceeds to calculating the root hash of the state
/// represented as MPT.
/// See [`state_root_unsorted`] for more info.
pub fn state_root_unhashed<A: Into<TrieAccount>>(
    state: impl IntoIterator<Item = (Address, A)>,
) -> B256 {
    state_root_unsorted(state.into_iter().map(|(address, account)| (keccak256(address), account)))
}

/// Sorts the hashed account keys and calculates the root hash of the state represented as MPT.
/// See [`state_root`] for more info.
pub fn state_root_unsorted<A: Into<TrieAccount>>(
    state: impl IntoIterator<Item = (B256, A)>,
) -> B256 {
    state_root(state.into_iter().sorted_unstable_by_key(|(key, _)| *key))
}
```

Also implements `From<GenesisAccount>` for `Account` since Reth needs it in `fn state_root_ref_unhashed<'a, A: Into<Account> + Clone + 'a>`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -20,7 +20,7 @@ workspace = true
 
 [dependencies]
 alloy-eips = { workspace = true, features = ["kzg-sidecar"] }
-alloy-genesis.workspace = true
+alloy-genesis = { workspace = true, optional = true }
 alloy-primitives = { workspace = true, features = ["rlp"] }
 alloy-rlp.workspace = true
 alloy-trie.workspace = true
@@ -48,7 +48,7 @@ derive_more = { workspace = true, features = [
     "into_iterator",
 ], default-features = false }
 auto_impl.workspace = true
-itertools.workspace = true
+itertools = { workspace = true, optional = true }
 
 [dev-dependencies]
 alloy-eips = { workspace = true, features = ["arbitrary"] }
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
great, I like this,

could we also add these tests:

https://github.com/paradigmxyz/reth/blob/5d7115035585395d77934f0bc4bac281daf47dff/crates/trie/common/src/account.rs#L76-L76

moricho:
Got it, I'll add them 👍

mattsse (commented):
actually, now that I looked at this,
it would be easier if we upstream this to alloy-trie entirely

the trieaccount type + state root functions

could you please open these changes as a pr to alloy-trie
then we import alloy-trie in alloy-genesis and to the from impl there

moricho:
Ah, that makes sense. I'll work on it!

moricho:
@mattsse 
Opened a PR to export this to alloy-trie. Could you take a look at it when you have time?: https://github.com/alloy-rs/trie/pull/65

PR #1720, opened 2024-11-30
feat(consensus): public RlpEcdsaTx

htiennv:
# Description

Public trait `RlpEcdsaTx` for implementation of `OpTypedTransaction`.

Ref: https://github.com/alloy-rs/op-alloy/issues/316


mattsse (changes requested):
I think we can re-export but I'd like to use a separete `pub use` and mark this as `doc(hidden)` because this is considered private API and should only be used by other network impl like op-alloy, if we really need it

htiennv:
Hi @mattsse, i was updated. Can you check it? Thanks

mattsse (approved):

mattsse (commented):
this is actually already exported

https://github.com/alloy-rs/alloy/blob/b0909d4650308918b2d552a641915ddb9c810f06/crates/consensus/src/transaction/mod.rs#L39-L40

transaction::RlpEcdsaTx

PR #1719, opened 2024-11-30
feat: impl `Encodable2718` for `ReceiptWithBloom`

klkvr:
Divides `RlpReceipt` into `RlpEncodableReceipt` and `RlpDecodableReceipt`, implementing `RlpEncodableReceipt` for references.

Provides `impl Encodable2718 for ReceiptWithBloom` when inner receipt knows its type.
 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #1721, opened 2024-11-30
fix: remove `Borrow` impl for RPC receipt

klkvr:
## Motivation

Based on #1719 

I've added this impl to support a more generic `AnyReceiptEnvelope` https://github.com/alloy-rs/alloy/blob/67576f9f6e70026116443851a43e4c805902c3b3/crates/rpc-types-eth/src/log.rs#L156-L160

However, it is in fact incorrect as `Borrow` shouldn't be implemented in this case

## Solution

This PR reverts changes including this impl and changes `AnyReceiptEnvelope` back to always holding a `Receipt<T>`

If we want to still support `AnyReceiptEnvelope` being generic over arbitrary receipt types, we'd need to either add `AsRef<Self> for alloy_primitives::Log` and use it instead or introduce a separate `Log` trait to replace such bounds with: https://github.com/alloy-rs/alloy/blob/67576f9f6e70026116443851a43e4c805902c3b3/crates/consensus/src/receipt/receipts.rs#L27-L29

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus-any/src/receipt/envelope.rs
+++ b/crates/consensus-any/src/receipt/envelope.rs
@@ -1,7 +1,8 @@
-use alloy_consensus::{Eip658Value, Receipt, ReceiptWithBloom, RlpEncodableReceipt, TxReceipt};
+use alloy_consensus::{Eip658Value, Receipt, ReceiptWithBloom, TxReceipt};
 use alloy_eips::eip2718::{Decodable2718, Eip2718Result, Encodable2718};
 use alloy_primitives::{bytes::BufMut, Bloom, Log};
 use alloy_rlp::{Decodable, Encodable};
+use core::fmt;
 
 /// Receipt envelope, as defined in [EIP-2718].
 ///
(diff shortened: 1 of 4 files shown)


crates/consensus-any/src/receipt/envelope.rs
```diff
@@ -16,23 +17,23 @@ use alloy_rlp::{Decodable, Encodable};
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[doc(alias = "AnyTransactionReceiptEnvelope", alias = "AnyTxReceiptEnvelope")]
-pub struct AnyReceiptEnvelope<T = Receipt<Log>> {
+pub struct AnyReceiptEnvelope<T = Log> {
```
mattsse, line 20: this could potentially be problematic for use cases where the any network should support different receipt types, for example op.

but this doesn't change the defaults and is fine imo.

we should also add

> AsRef<Self> for alloy_primitives::Log

to alloy_primitives

mattsse (approved):

PR #1744, opened 2024-12-03
feat(engine): Forkchoice Version

refcell:
### Description

Introduces a forkchoice update version type.

Context: https://github.com/alloy-rs/op-alloy/pull/321#discussion_r1866125516

### DIFF
--- a/crates/rpc-types-engine/src/version.rs
+++ b/crates/rpc-types-engine/src/version.rs
@@ -1,0 +1,16 @@
+//! Versions for the engine api.
+
+/// The version of the engine api.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[repr(u32)]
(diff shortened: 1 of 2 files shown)


crates/rpc-types-engine/src/version.rs
```diff
@@ -0,0 +1,16 @@
...
+#[repr(u32)]
+pub enum ForkchoiceUpdateVersion {
+    /// Version 1 of the engine api.
+    V1 = 1,
+    /// Version 2 of the engine api.
+    V2 = 2,
+    /// Version 3 of the engine api.
+    V3 = 3,
+    /// Version 4 of the engine api.
+    V4 = 4,
```
mattsse, line 15: there's actually no fcu 4 for prague anymore, only newPayloadv4, so we either remove this or rename it to something like

https://github.com/paradigmxyz/reth/blob/9d5e159968737950473a572cdce3881af033f9fc/crates/payload/primitives/src/lib.rs#L324-L342

mattsse (changes requested):

mattsse (approved):
nvm, we now have fcu 4

Issue #1735, opened 2024-12-03
[Feature] Make `Transaction: Typed2718`


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

the transaction trait currently has its own `fn ty` function.

we can unify this with the newer `Typed2718` trait by 

1. remove `fn ty` from `trait Transaction`
2. require `trait Transaction: Typed2718`

nadtech-hub:
:raised_back_of_hand:

PR #1746, opened 2024-12-03
feat(consensus): require typed2718 for transaction

nuts-rice:
## Motivation
Address #1735 ,  unifies tx types around eip2718

## Solution
Implements the Typed2718 for transaction types with `fn ty` 

## PR Checklist
- [X] cargo fmt
- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,4 +1,4 @@
-use crate::{transaction::RlpEcdsaTx, SignableTransaction, Signed, Transaction, TxType};
+use crate::{transaction::RlpEcdsaTx, SignableTransaction, Signed, Transaction, TxType, Typed2718};
 use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization};
 use alloy_primitives::{Bytes, ChainId, PrimitiveSignature as Signature, TxKind, B256, U256};
 use alloy_rlp::{BufMut, Decodable, Encodable};
@@ -235,11 +235,6 @@ impl Transaction for TxEip1559 {
     }
 
     #[inline]
-    fn ty(&self) -> u8 {
-        TxType::Eip1559 as u8
-    }
-
-    #[inline]
     fn access_list(&self) -> Option<&AccessList> {
         Some(&self.access_list)
     }
@@ -252,6 +247,39 @@ impl Transaction for TxEip1559 {
     #[inline]
     fn authorization_list(&self) -> Option<&[SignedAuthorization]> {
         None
+    }
+}
+
+impl Typed2718 for TxEip1559 {
+    fn ty(&self) -> u8 {
+        TxType::Eip1559 as u8
+    }
+    fn is_type(&self, ty: u8) -> bool {
+        self.ty() == ty
+    }
+
+    /// Returns true if the type is a legacy transaction.
+    fn is_legacy(&self) -> bool {
+        self.ty() == 0
+    }
+    /// Returns true if the type is an EIP-2930 transaction.
+    fn is_eip2930(&self) -> bool {
+        self.ty() == 1
+    }
+
+    /// Returns true if the type is an EIP-1559 transaction.
+    fn is_eip1559(&self) -> bool {
+        self.ty() == 2
+    }
+
+    /// Returns true if the type is an EIP-4844 transaction.
+    fn is_eip4844(&self) -> bool {
+        self.ty() == 3
+    }
+
+    /// Returns true if the type is an EIP-7702 transaction.
+    fn is_eip7702(&self) -> bool {
+        self.ty() == 4
     }
 }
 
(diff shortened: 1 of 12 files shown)


crates/consensus/src/transaction/eip1559.rs
```diff
@@ -255,6 +250,39 @@ impl Transaction for TxEip1559 {
...
+
+    /// Returns true if the type is an EIP-4844 transaction.
+    fn is_eip4844(&self) -> bool {
+        self.ty() == 3
+    }
+
+    /// Returns true if the type is an EIP-7702 transaction.
+    fn is_eip7702(&self) -> bool {
+        self.ty() == 4
+    }
```
mattsse, line 283: all of those should have a default impl, so you can remove these

crates/consensus/src/transaction/eip2930.rs
```diff
@@ -213,6 +208,20 @@ impl Transaction for TxEip2930 {
...
+        TxType::Eip2930 as u8
+    }
+    fn is_type(&self, ty: u8) -> bool {
+        self.ty() == ty
+    }
+
+    /// Returns true if the type is an EIP-2930 transaction.
+    fn is_eip2930(&self) -> bool {
+        self.ty() == 1
+    }
```
mattsse, line 222: same

mattsse (changes requested):
thanks, 
if you look at:

https://github.com/alloy-rs/alloy/blob/36b09310beb27e514df03faf4269aaf55d90a49f/crates/consensus/src/transaction/mod.rs#L344-L372

then all of those already have a default impl, so the only thing we need to impl is the `fn type` fn
(and optionally override the corresponding is_ fn and return a `true`, but imo we don't need to)

nuts-rice:
alright removed those

crates/consensus/src/transaction/envelope.rs
```diff
@@ -656,6 +645,46 @@ impl Transaction for TxEnvelope {
...
+
+    /// Returns true if the type is an EIP-4844 transaction.
+    fn is_eip4844(&self) -> bool {
+        self.ty() == 3
+    }
+
+    /// Returns true if the type is an EIP-7702 transaction.
+    fn is_eip7702(&self) -> bool {
+        self.ty() == 4
+    }
```
mattsse, line 685: one more

crates/consensus/src/transaction/legacy.rs
```diff
@@ -349,6 +346,20 @@ impl SignableTransaction<Signature> for TxLegacy {
...
+        TxType::Legacy as u8
+    }
+    fn is_type(&self, ty: u8) -> bool {
+        self.ty() == ty
+    }
+
+    /// Returns true if the type is a legacy transaction.
+    fn is_legacy(&self) -> bool {
+        self.ty() == 0
+    }
```
mattsse, line 360: same

crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -179,7 +180,7 @@ impl From<Transaction> for TxEnvelope {
     }
 }
 
-impl<T: TransactionTrait> TransactionTrait for Transaction<T> {
+impl<T: TransactionTrait + Typed2718> TransactionTrait for Transaction<T> {
```
mattsse, line 183: this should be redundant because guaranteed
```suggestion
impl<T: TransactionTrait> TransactionTrait for Transaction<T> {
```

crates/consensus/src/transaction/mod.rs
```diff
@@ -55,7 +55,7 @@ pub mod serde_bincode_compat {
 /// Represents a minimal EVM transaction.
 #[doc(alias = "Tx")]
 #[auto_impl::auto_impl(&, Arc)]
-pub trait Transaction: fmt::Debug + any::Any + Send + Sync + 'static {
+pub trait Transaction: fmt::Debug + any::Any + Send + Sync + 'static + Typed2718 {
```
mattsse, line 58: let's put Typed2718 first

```suggestion
pub trait Transaction: Typed2718 + fmt::Debug + any::Any + Send + Sync + 'static {
```

crates/network/src/any/either.rs
```diff
@@ -199,6 +191,24 @@ impl TransactionTrait for AnyTypedTransaction {
...
+}
+
+impl Typed2718 for AnyTxEnvelope {
+    fn ty(&self) -> u8 {
+        match self {
+            Self::Ethereum(inner) => inner.ty(),
+            Self::Unknown(inner) => inner.ty(),
+        }
+    }
+}
```
mattsse, line 210: move this below the struct def

mattsse (changes requested):
last nits

nuts-rice:
okay dokie

mattsse (approved):
lgtm

this is technically a breaking change so I'm holding off merging this

PR #1752, opened 2024-12-04
feat: impl Encodable / Decodable for Receipts

Rjected:
## Motivation
The `Deref` / `DerefMut` derives were kind of an antipattern, giving the illusion that `Encodable` was implemented, when it was only due to deref coercion. Decodable was also not implemented. This is required if we ever want to use the `Receipts` type in reth.

## Solution

This instead this removes the derived derefs, and directly implements the encodable / decodable traits.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -1,12 +1,12 @@
-use crate::receipt::{Eip658Value, RlpDecodableReceipt, RlpEncodableReceipt, TxReceipt};
+use crate::receipt::{
+    Eip2718EncodableReceipt, Eip658Value, RlpDecodableReceipt, RlpEncodableReceipt, TxReceipt,
+};
 use alloc::{vec, vec::Vec};
 use alloy_eips::eip2718::Encodable2718;
 use alloy_primitives::{Bloom, Log};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use core::{borrow::Borrow, fmt};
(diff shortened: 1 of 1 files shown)


crates/consensus/src/receipt/receipts.rs
```diff
@@ -186,6 +184,22 @@ impl<T> FromIterator<Vec<T>> for Receipts<T> {
...
+    fn length(&self) -> usize {
+        self.receipt_vec.length()
+    }
+}
+
+impl<T: Decodable> Decodable for Receipts<T> {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        Ok(Self { receipt_vec: Decodable::decode(buf)? })
+    }
+}
```
mattsse, line 201: aren't those derivable?

mattsse (commented):
this mimics the eth rlpx format, right?

crates/consensus/src/receipt/receipts.rs
Rjected: Yes this should mimic the rlpx format. I'm not sure how we can derive them, because it's a struct and not a newtype, we can't use `RlpEncodableWrapper` etc. If we derive `RlpEncodable`, it will insert an additional header.

mattsse: ah because RlpEncodableWrapper only works with unnamed and can't handle a single field?

mattsse (approved):

PR #1762, opened 2024-12-06
feat: EIP-7691

### LINKED ISSUE #1712: [Feature] EIP-7691
### Component

consensus, eips, genesis

EIP-7691 is mostly a CL change, however, it will also change `BLOB_BASE_FEE_UPDATE_FRACTION` in the EL to change the fee market a bit. This is a patchwork solution until we gather more info about blob fee markets for Fusaka.

See the relevant PR that adds these EL changes: https://github.com/ethereum/EIPs/pull/9060/files

TL;DR: For < Prague we use the current `BLOB_BASE_FEE_UPDATE_FRACTION`, for >= Prague use `BLOB_BASE_FEE_UPDATE_FRACTION_ELECTRA`.

Relevant snippets: https://github.com/alloy-rs/alloy/blob/20d13faff78325dda42de8952e0addad31d11db7/crates/eips/src/eip4844/mod.rs#L70-L71

https://github.com/alloy-rs/alloy/blob/20d13faff78325dda42de8952e0addad31d11db7/crates/eips/src/eip4844/mod.rs#L131-L142

### Additional context

_No response_

klkvr:
## Motivation

Closes #1712 
Implements https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7691.md

## Solution

The current API might be a bit confusing given that the fn has same signature as `eip4844::calc_blob_gasprice`. Unsure how we can make this nicer, something similar to `BaseFeeParams` could probably work but it's unclear how callers would set it

Also I don't not how we should handle this in `BlockHeader::next_block_blob_fee` and `BlobGasFiller`. Given variable update fraction we no longer can estimate the blob fee for next block without knowing the hardfork activation times

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7691.rs
+++ b/crates/eips/src/eip7691.rs
@@ -1,0 +1,21 @@
+use crate::eip4844::BLOB_TX_MIN_BLOB_GASPRICE;
+
+/// CL-enforced target blobs per block after Pectra hardfork activation.
+pub const TARGET_BLOBS_PER_BLOCK_ELECTRA: u64 = 6;
+
+/// CL-enforced maximum blobs per block after Pectra hardfork activation.
+pub const MAX_BLOBS_PER_BLOCK_ELECTRA: u64 = 9;
+
+/// Determines the maximum rate of change for blob fee after Pectra hardfork activation.
+pub const BLOB_GASPRICE_UPDATE_FRACTION_PECTRA: u128 = 5007716;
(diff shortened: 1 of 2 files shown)


mattsse (approved):
I think this is fine, ideally this is something we only have to get right once

PR #1764, opened 2024-12-06
fix: use `feeHistory` when estimating blob fee

klkvr:
With EIP-7691, we can no longer rely on `BlockHeader::next_block_blob_fee` as it will start producing invalid values once the update fraction changes.

This PR changes blob fee filler to use `feeHistory` instead to obtain those values from RPC.

I think we should also consider either deprecating `BlockHeader::next_block_excess_blob_gas`/`BlockHeader::next_block_blob_fee` methods or making them accept an update fraction parameter

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -6,11 +6,9 @@ use crate::{
     utils::Eip1559Estimation,
     Provider,
 };
-use alloy_consensus::BlockHeader;
 use alloy_eips::eip4844::BLOB_TX_MIN_BLOB_GASPRICE;
 use alloy_json_rpc::RpcError;
 use alloy_network::{Network, TransactionBuilder, TransactionBuilder4844};
-use alloy_network_primitives::{BlockResponse, BlockTransactionsKind};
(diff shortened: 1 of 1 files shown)


crates/provider/src/fillers/gas.rs
```diff
@@ -226,17 +224,13 @@ where
...
-            .ok_or(RpcError::NullResp)?;
-
-        let latest_header = latest_block.header().as_ref();
-
-        latest_header
-            .next_block_blob_fee()
-            .map(Into::into)
-            .ok_or(RpcError::UnsupportedFeature("eip4844"))
+            .base_fee_per_blob_gas
+            .last()
```
mattsse, line 231: makes sense, the last entry should be the estimated next one

mattsse (approved):

mattsse:
> BlockHeader::next_block_excess_blob_gas/BlockHeader::next_block_blob_fee methods or making them accept an update fraction parameter

yeah let's mark them as deprecated and introduce one with the update fraction

PR #1767, opened 2024-12-07
feat: port reth pooled tx type

mattsse:
migrates the reth `PooledTransactionElement` type and renames it to just `PooledTransaction`.

@klkvr I assume we can simplify the decode/encode impls now

### DIFF
--- a/crates/consensus/src/transaction/pooled.rs
+++ b/crates/consensus/src/transaction/pooled.rs
@@ -1,15 +1,17 @@
-//! Defines the exact transaction variant that are allowed to be propagated over the eth p2p protocol.
+//! Defines the exact transaction variant that are allowed to be propagated over the eth p2p
+//! protocol.
 
-use crate::{transaction::{TxEip1559, TxEip2930, TxEip4844, TxLegacy}, SignableTransaction, Signed, TxEip4844WithSidecar, TxEnvelope};
-use crate::{Transaction, TxEip7702};
+use crate::{
+    transaction::{RlpEcdsaTx, TxEip1559, TxEip2930, TxEip4844, TxLegacy},
+    SignableTransaction, Signed, Transaction, TxEip4844WithSidecar, TxEip7702, TxEnvelope, TxType,
+};
 use alloy_eips::{
-    eip2718::{Decodable2718, Eip2718Result, Encodable2718},
+    eip2718::{Decodable2718, Eip2718Error, Eip2718Result, Encodable2718},
     eip2930::AccessList,
     eip7702::SignedAuthorization,
 };
-use alloy_primitives::bytes;
 use alloy_primitives::{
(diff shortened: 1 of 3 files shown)


klkvr (commented):
yep, decode impls would just need to be same as TxEnvelope

wondering if it would also make sense to just make `TxEnvelope` generic over Eip4844 variant? then we can just do `type PooledTransaction = TxEnvelope<TxEIp4844WithSidecar>` and won't need to duplicate any code

mattsse:
hmm, good question.

The `TxEnvelope` is an integral part of the entire alloy client side API, 

imo this extra type isn't too bad, and I'd rather have it separate

crates/consensus/src/transaction/pooled.rs
```diff
@@ -0,0 +1,558 @@
...
+
+impl Decodable2718 for PooledTransaction {
+    fn typed_decode(ty: u8, buf: &mut &[u8]) -> Eip2718Result<Self> {
+        match ty.try_into().map_err(|_| alloy_rlp::Error::Custom("unexpected tx type"))? {
+            TxType::Eip2930 => Ok(TxEip2930::rlp_decode_signed(buf)?.into()),
+            TxType::Eip1559 => Ok(TxEip1559::rlp_decode_signed(buf)?.into()),
+            TxType::Eip4844 => Ok(TxEip4844WithSidecar::rlp_decode_signed(buf)?.into()),
+            // TxType::Eip4844 =>
+            // TxEip4844WithSidecar::rlp_decode_signed(buf)?.try_into_4844_with_sidecar().
+            // map_err(|_| alloy_rlp::Error::Custom("unexpected tx type")),
```
klkvr:
```suggestion
```

crates/consensus/src/transaction/pooled.rs
```diff
@@ -0,0 +1,555 @@
...
+    /// An untagged [`TxLegacy`].
+    Legacy(Signed<TxLegacy>),
+    /// A [`TxEip2930`] tagged with type 1.
+    Eip2930(Signed<TxEip2930>),
+    /// A [`TxEip1559`] tagged with type 2.
+    Eip1559(Signed<TxEip1559>),
+    /// A [`TxEip7702`] tagged with type 4.
+    Eip7702(Signed<TxEip7702>),
+    /// A EIP-4844 transaction, which includes the transaction, blob data, commitments, and proofs.
+    Eip4844(Signed<TxEip4844WithSidecar>),
```
klkvr, line 39:
```suggestion
     /// A EIP-4844 transaction, which includes the transaction, blob data, commitments, and proofs.
    Eip4844(Signed<TxEip4844WithSidecar>),
    /// A [`TxEip7702`] tagged with type 4.
    Eip7702(Signed<TxEip7702>),
```

klkvr (approved):
just a nit

PR #1768, opened 2024-12-07
chore: introduce recovered and recoverable

mattsse:
ports the reth `RecoveredTx` to alloy as `Recovered`.

this is similar to `Signed`.

we don't have `recover_signer` as part of any API rn, so this introduces this as a new trait.

once we're happy with this we can integrate this like: `Signed -> Recovered<Signed>`

wdyt @Rjected @klkvr

### DIFF
--- a/crates/consensus/src/transaction/recovered.rs
+++ b/crates/consensus/src/transaction/recovered.rs
@@ -38,8 +38,8 @@ impl<T> Recovered<T> {
     }
 
     /// Dissolve Self to its component
-    pub fn to_components(self) -> (T, Address) {
-        (self.signed_transaction, self.signer)
+    pub fn into_parts(self) -> (T, Address) {
+        (self.tx, self.signer)
     }
 
     /// Create [`Recovered`] from the given transaction and [`Address`] of the signer.
--- a/crates/consensus/src/transaction/recovered.rs
+++ b/crates/consensus/src/transaction/recovered.rs
@@ -1,0 +1,123 @@
+use alloy_eips::eip2718::Encodable2718;
+use alloy_primitives::{bytes, Address, B256};
+use alloy_rlp::{Decodable, Encodable};
+use derive_more::{AsRef, Deref};
+
+/// Signed transaction with recovered signer.
+#[derive(Debug, Clone, PartialEq, Hash, Eq, AsRef, Deref)]
+pub struct Recovered<T> {
+    /// Signer of the transaction
+    signer: Address,
+    /// Signed transaction
+    #[deref]
+    #[as_ref]
+    signed_transaction: T,
+}
+
+// === impl RecoveredTx ===
+
+impl<T> Recovered<T> {
+    /// Signer of transaction recovered from signature
+    pub const fn signer(&self) -> Address {
+        self.signer
+    }
+
+    /// Reference to the signer of transaction recovered from signature
+    pub const fn signer_ref(&self) -> &Address {
+        &self.signer
+    }
+
+    /// Returns a reference to [`TransactionSigned`]
+    pub const fn as_signed(&self) -> &T {
+        &self.signed_transaction
+    }
+
+    /// Transform back to [`TransactionSigned`]
+    pub fn into_signed(self) -> T {
+        self.signed_transaction
+    }
+
+    /// Dissolve Self to its component
+    pub fn to_components(self) -> (T, Address) {
+        (self.signed_transaction, self.signer)
+    }
+
+    /// Create [`Recovered`] from the given transaction and [`Address`] of the signer.
+    ///
+    /// Note: This does not check if the signer is the actual signer of the transaction.
+    #[inline]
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+
+/// Signed transaction with recovered signer.
+#[derive(Debug, Clone, PartialEq, Hash, Eq, AsRef, Deref)]
+pub struct Recovered<T> {
+    /// Signer of the transaction
+    signer: Address,
+    /// Signed transaction
+    #[deref]
+    #[as_ref]
+    signed_transaction: T,
```
klkvr, line 14:
```suggestion
    tx: T,
```

crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Reference to the signer of transaction recovered from signature
+    pub const fn signer_ref(&self) -> &Address {
+        &self.signer
+    }
+
+    /// Returns a reference to [`TransactionSigned`]
+    pub const fn as_signed(&self) -> &T {
+        &self.signed_transaction
+    }
```
klkvr, line 33:
```suggestion
    /// Returns a reference to [`TransactionSigned`]
    pub const fn tx(&self) -> &T {
        &self.tx
    }
```

crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Transform back to [`TransactionSigned`]
+    pub fn into_signed(self) -> T {
+        self.signed_transaction
+    }
+
+    /// Dissolve Self to its component
+    pub fn to_components(self) -> (T, Address) {
+        (self.signed_transaction, self.signer)
+    }
```
klkvr, line 43:
```suggestion
    /// Dissolve Self to its component
    pub fn into_parts(self) -> (T, Address) {
        (self.tx, self.signer)
    }
```

crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Returns a reference to [`TransactionSigned`]
+    pub const fn as_signed(&self) -> &T {
+        &self.signed_transaction
+    }
+
+    /// Transform back to [`TransactionSigned`]
+    pub fn into_signed(self) -> T {
+        self.signed_transaction
+    }
```
klkvr, line 38:
```suggestion
    /// Transform back to [`TransactionSigned`]
    pub fn into_tx(self) -> T {
        self.tx
    }
```

crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+        (self.signed_transaction, self.signer)
+    }
+
+    /// Create [`Recovered`] from the given transaction and [`Address`] of the signer.
+    ///
+    /// Note: This does not check if the signer is the actual signer of the transaction.
+    #[inline]
+    pub const fn new_unchecked(signed_transaction: T, signer: Address) -> Self {
+        Self { signed_transaction, signer }
+    }
```
klkvr, line 51:
```suggestion
    pub const fn new_unchecked(tx: T, signer: Address) -> Self {
        Self { tx, signer }
    }
```

klkvr (changes requested):
makes sense to me, this likely makes should be required to be implemented by consensus transaction

left some suggestions for renaming things to match `Signed` API/structure

crates/consensus/src/transaction/recovered.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// This can fail for some early ethereum mainnet transactions pre EIP-2, use
+    /// [`Self::recover_signer_unchecked`] if you want to recover the signer without ensuring that
+    /// the signature has a low `s` value.
+    fn recover_signer(&self) -> Result<Address, alloy_primitives::SignatureError>;
+
+    /// Recover signer from signature and hash _without ensuring that the signature has a low `s`
+    /// value_.
+    ///
+    /// Returns an error if the transaction's signature is invalid.
+    fn recover_signer_unchecked(&self) -> Result<Address, alloy_primitives::SignatureError>;
```
klkvr, line 122: btw was recently thinking about how we could avoid embedding the `_unchecked` stuff into the abstraction

could this be a part of tx validation which is aware of the timestamp and can reject transactions outside of the actual recovery logic?

klkvr (approved):

crates/consensus/src/transaction/recovered.rs
mattsse: hmm, this could likely be a bit tricky because we also have this a ton in rpc, but would actually be nicer.

let's proceed with unchecked and think how we could phase that out?

klkvr: sg!

Issue #1766, opened 2024-12-07
[Feature] Make next_block_blob_fee functions accept an update fraction parameter


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

https://github.com/alloy-rs/alloy/pull/1764

> I think we should also consider either deprecating BlockHeader::next_block_excess_blob_gas/BlockHeader::next_block_blob_fee methods or making them accept an update fraction parameter

for prague these will no longer work, and need updating

PR #1769, opened 2024-12-08
feat: Make next_block_blob_fee functions accept an update fraction parameter 

DoTheBestToGetTheBest:
closes https://github.com/alloy-rs/alloy/issues/1766

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -15,6 +15,8 @@ use alloy_primitives::{
 use alloy_rlp::{length_of_length, BufMut, Decodable, Encodable};
 use core::mem;
 
+const BLOB_BASE_FEE_UPDATE_FRACTION_ELECTRA: f64 = 0.145;
+
 /// Ethereum Block header
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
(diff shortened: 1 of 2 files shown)


crates/consensus/src/block/header.rs
```diff
@@ -214,10 +216,41 @@ impl Header {
...
+    ///
+    /// # Returns
+    ///
+    /// - `Some(u128)` containing the calculated blob fee if `next_block_excess_blob_gas` is `Some`.
+    /// - `None` if `next_block_excess_blob_gas` is `None`.
+    ///
+    /// # Panics
+    ///
+    /// - Panics if `update_fraction` is provided and not within the range [0.0, 1.0].
+    pub fn next_block_blob_fee_with_fraction(&self, update_fraction: Option<f64>) -> Option<u128> {
```
klkvr, line 241: I'd prefer this to not be optional

klkvr: and we also need this on `BlochHeader` trait

crates/consensus/src/block/header.rs
```diff
@@ -214,10 +216,41 @@ impl Header {
...
+        let fraction = update_fraction.unwrap_or(BLOB_BASE_FEE_UPDATE_FRACTION_ELECTRA);
+
+        if !(0.0..=1.0).contains(&fraction) {
+            panic!("update_fraction must be between 0.0 and 1.0");
+        }
+
+        let excess_blob_gas = self.next_block_excess_blob_gas()?;
+        let adjusted_blob_gas = (excess_blob_gas as f64 * fraction) as u64;
+
+        Some(eip4844::calc_blob_gasprice(adjusted_blob_gas))
```
klkvr, line 251: we should introduce an additional helper like `calc_blob_gasprice_with_update_fraction` and use it here

klkvr (changes requested):

crates/consensus/src/block/header.rs
```diff
@@ -214,10 +214,39 @@ impl Header {
     /// Returns `None` if `excess_blob_gas` is None.
     ///
     /// See also [Self::next_block_excess_blob_gas]
+    #[deprecated(
+        since = "0.8.1",
```
DoTheBestToGetTheBest, line 218: not sure if  you are okay with this version the current one is 0.8.0 so for the next one is 0.8.1;

lmk if you want me to change or remove this

crates/consensus/src/block/header.rs
DoTheBestToGetTheBest: > we should introduce an additional helper like `calc_blob_gasprice_with_update_fraction` and use it here

thanks for your review!

crates/consensus/src/block/header.rs
DoTheBestToGetTheBest: > and we also need this on `BlochHeader` trait

thanks, everything done!

DoTheBestToGetTheBest:
ping @klkvr

mattsse:
@klkvr with recent 77840 changes https://github.com/alloy-rs/alloy/pull/1828 this is no longer required?

klkvr:
yeah sorry @DoTheBestToGetTheBest, this is superseded by https://github.com/alloy-rs/alloy/pull/1828

PR #1775, opened 2024-12-09
feat: reth's block body fns

stevencartavia:
## Motivation
https://github.com/paradigmxyz/reth/issues/12519

## Solution
Brought BlockBody fns from reth to replace reth's BlockBody type with alloy's

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/mod.rs
+++ b/crates/consensus/src/block/mod.rs
@@ -93,22 +93,16 @@ impl<T: Transaction> BlockBody<T> {
         self.transactions.iter().any(|tx| tx.is_eip7702())
     }
 
-    /// Helper for blob_transactions_iter function
-    #[inline]
-    fn is_eip4844(tx: &T) -> bool {
-        tx.is_eip4844()
-    }
-
     /// Returns an iterator over all blob transactions of the block
     #[inline]
     pub fn blob_transactions_iter(&self) -> impl Iterator<Item = &T> + '_ {
-        self.transactions.iter().filter(|tx| Self::is_eip4844(tx))
+        self.transactions.iter().filter(|tx| tx.is_eip4844())
     }
 
     /// Returns only the blob transactions, if any, from the block body.
     #[inline]
-    pub fn blob_transactions(&self) -> Vec<&T> {
-        self.blob_transactions_iter().collect()
+    pub fn blob_transactions(&self) -> impl Iterator<Item = &T> + '_ {
+        self.blob_transactions_iter()
     }
 }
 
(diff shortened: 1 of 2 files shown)


stevencartavia:
I just brought these fns at the moment since the rest explicitly use `T: TransactionSigned`

crates/consensus/src/transaction/mod.rs
```diff
@@ -177,6 +180,18 @@ pub trait Transaction: fmt::Debug + any::Any + Send + Sync + 'static {
...
+    #[inline]
+    fn is_eip4844(&self) -> bool {
+        self.ty() == EIP4844_TX_TYPE_ID
+    }
+
+    /// Returns true if the transaction is an EIP-7702 transaction.
+    #[inline]
+    fn is_eip7702(&self) -> bool {
+        self.ty() == EIP7702_TX_TYPE_ID
+    }
```
mattsse, line 194: these are no longer necessary with #1773

mattsse (changes requested):
cool, makes sense

one suggestion

crates/consensus/src/block/mod.rs
```diff
@@ -58,6 +60,56 @@ impl<T> BlockBody<T> {
...
+    #[inline]
+    pub fn has_eip7702_transactions(&self) -> bool {
+        self.transactions.iter().any(|tx| tx.is_eip7702())
+    }
+
+    /// Helper for blob_transactions_iter function
+    #[inline]
+    fn is_eip4844(tx: &T) -> bool {
+        tx.is_eip4844()
+    }
```
mattsse, line 100: this function is not needed and can be removed

crates/consensus/src/block/mod.rs
```diff
@@ -58,6 +60,56 @@ impl<T> BlockBody<T> {
...
+    #[inline]
+    pub fn blob_transactions_iter(&self) -> impl Iterator<Item = &T> + '_ {
+        self.transactions.iter().filter(|tx| Self::is_eip4844(tx))
+    }
+
+    /// Returns only the blob transactions, if any, from the block body.
+    #[inline]
+    pub fn blob_transactions(&self) -> Vec<&T> {
+        self.blob_transactions_iter().collect()
+    }
```
mattsse, line 112: this should be an impl Iterator instead because this is mostly useful when iterating over blob txs

mattsse (changes requested):
nits

mattsse (approved):

Issue #1763, opened 2024-12-06
[Bug] `SignedBidSubmissionV4` needs to use "JSON" encoding of execution requests


ryanschneider:
### Component

serde

### What version of Alloy are you on?

v0.7.3

### Operating System

None

### Describe the bug

So it turns out that even after EIP-7865 the builder APIs still need to use the "JSON" encoding for `execution_requests`:

https://github.com/ethereum/builder-specs/blob/main/types/electra/bid.yaml#L18-L20

I tried changing this with https://github.com/ethereum/builder-specs/pull/107 but it looks like that would had follow-on impacts on the beacon API itself, so I think we need to go forward with reintroducing JSON serialization in alloy's `Requests` type.

I discussed this with @mattsse and his suggestion was to:

- Reintroduce the `Requests` enum originally deleted here: https://github.com/alloy-rs/alloy/pull/1515/files#diff-db23d6583d0c8ba369cdc39fa3ab17440cca32dbf24f27423f1e668dc56dd616L20
- But call it something like `RequestsV4` and have a conversion between `Requests` and `RequestsV4`

PR #1515, opened 2024-10-18
chore: remove 7685 request variants

mattsse:
these got removed and replaced with a general purpose request blob

PR #1786, opened 2024-12-11
feat(relay): Re-add {Deposit,Withdrawal,Consolidation}Request objects for relay submissions

ryanschneider:
## Motivation

As reviewing the builder-specs (and unsuccessfully attempting to change them with https://github.com/ethereum/builder-specs/pull/107) I determined that the `V4` relay submission actually needs to include the requests in their JSON "object" form rather than the encoded EIP-7865 `Vec<Bytes>`.

## Solution

So, this restores some of the structs removed in https://github.com/alloy-rs/alloy/pull/1515 and adds a new `RequestsV4` struct to `rpc-types-beacon`.

A couple things to note:

- In restoring the `eip` structs I removed the "EL-specific" encoding helpers in favor of the CL flavors (e.g. integers are encoded using `DisplayFromStr` instead of `alloy_serde::quantity`.  I think this makes sense since these structs are no longer used by the EL/engine API, but I could be convinced to either move these structs in `rpc-types-beacon` or restore them with "EL-encoding" and add wrapper/custom serde functionality in `rpc-types-beacon`.
- I also replaced the `quantity` encoding on `SignedBidSubmissionV4::target_blobs_per_block` with `DisplayFromStr` as I'm pretty sure this should be "decimal" encoded.
- The example JSON in `crates/rpc-types-beacon/src/examples/relay_builder_block_validation_request_v4.json` is a bit of frankenstein, it's a payload I captured from my locally running `rbuilder` inside our `builder-playground` with `execution_requests` and `target_blobs_per_block` manually edited, but it's the best I could find at the moment.

Note that after this PR is accepted, we'll also need to add functionality for converting `eip7865::Requests` to `rpc-types-beacon::requests::RequestsV4`, for use for example https://github.com/flashbots/rbuilder/issues/267.  So this PR is only the first step in fully closing https://github.com/alloy-rs/alloy/issues/1763

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


ryanschneider:
Closing in favor of https://github.com/alloy-rs/alloy/pull/1787 which is a cleaned up version of this that also includes the necessary conversions.

PR #1787, opened 2024-12-12
feat(relay):  ExecutionRequestsV4 with eip7685::Requests conversion

ryanschneider:
This is a further cleaned up version of #1786 which also includes the necessary `ExecutionRequestsV4 <-> Requests` conversions.  However, the `Requests -> ExecutionRequestsV4` conversion requires use of the `ssz_types` crate, which was added as a feature.

Fixes https://github.com/alloy-rs/alloy/issues/1763.

## Motivation

As reviewing the builder-specs (and unsuccessfully attempting to change them with https://github.com/ethereum/builder-specs/pull/107) I determined that the `V4` relay submission actually needs to include the requests in their JSON "object" form rather than the encoded EIP-7865 `Vec<Bytes>`.

## Solution

So, this restores some of the structs removed in https://github.com/alloy-rs/alloy/pull/1515 and adds a new `ExecutionRequestsV4` struct to `rpc-types-beacon`.

A couple things to note:

- In restoring the `eip` structs I removed the "EL-specific" encoding helpers in favor of the CL flavors, adding a new `alloy_serde::ssz::json::uint` serde helper for encoing integers in SSZ "canonical JSON" format.  I think this makes sense since these structs are no longer used by the EL/engine API, but I could be convinced to either move these structs in `rpc-types-beacon` or restore them with "EL-encoding" and add wrapper/custom serde functionality in `rpc-types-beacon`.
- I also replaced the `quantity` encoding on `SignedBidSubmissionV4::target_blobs_per_block` with `DisplayFromStr` as I'm pretty sure this should be "decimal" encoded.  NOTE: With 7742 pulled from PEctra I think this can be fully removed instead.
- I also removed (or didn't readd) RLP encoding/decoding from these types since that should never be necessary.
- The example JSON in `crates/rpc-types-beacon/src/examples/relay_builder_block_validation_request_v4.json` is a bit of frankenstein, it's a payload I captured from my locally running `rbuilder` inside our `builder-playground` with `execution_requests` and `target_blobs_per_block` manually edited, but it's the best I could find at the moment.
- This should give us everything we need to round trip the requests created by the system contracts into `ExecutionRequestsV4` objects that can be included in relay submissions, for example to close https://github.com/flashbots/rbuilder/issues/267.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-types-beacon/src/relay.rs
```diff
@@ -158,9 +159,9 @@ pub struct SignedBidSubmissionV4 {
     /// The Electra block bundle for this bid.
     pub blobs_bundle: BlobsBundleV1,
     /// The Pectra execution requests for this bid.
-    pub execution_requests: Vec<Bytes>,
+    pub execution_requests: ExecutionRequestsV4,
     /// The EIP-7742 blobs per block for this bid.
-    #[serde(with = "alloy_serde::quantity")]
+    #[serde_as(as = "DisplayFromStr")]
     pub target_blobs_per_block: u64,
```
ryanschneider, line 165: Aside: I saw today that EIP-7742 was pulled from Pectra, my PR might conflict with removing that since I changed `SignedBidSubmissionV4::target_blobs_per_block` to `DisplayFromStr` as well.

https://ethereum-magicians.org/t/all-core-devs-consensus-acdc-147-december-12-2024/22161/2

crates/rpc-types-beacon/src/examples/relay_builder_block_validation_request_v4.json
```diff
@@ -0,0 +1,164 @@
...
+    ],
+    "consolidations": [
+      {
+        "source_address": "0xabcf8e0d4e9587369b2301d0790347320302cc09",
+        "source_pubkey": "0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a",
+        "target_pubkey": "0x93247f2209abcacf57b75a51dafae777f9dd38bc7053d1af526f220a7489a6d3a2753e5f3e8b1cfe39b56f43611df74a"
+      }
+    ]
+  },
+  "target_blobs_per_block": "6",
```
ryanschneider, line 162: EIP7742 will need to be removed here as well.

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    impl TryFrom<&Requests> for ExecutionRequestsV4 {
+        type Error = Error;
+        fn try_from(value: &Requests) -> Result<Self, Self::Error> {
+            use alloy_eips::{
+                eip6110::DEPOSIT_REQUEST_TYPE, eip7002::WITHDRAWAL_REQUEST_TYPE,
+                eip7251::CONSOLIDATION_REQUEST_TYPE,
+            };
+            use ssz::Decode;
+            use ssz_types::{typenum, VariableList};
```
klkvr, line 37: let's move those imports above

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,150 @@
...
+                                MaxDepositRequestsPerPayload,
+                            > = VariableList::from_ssz_bytes(&request[1..])?;
+                            acc.1.extend(Vec::<WithdrawalRequest>::from(list));
+                        }
+                        CONSOLIDATION_REQUEST_TYPE => {
+                            let list: VariableList<
+                                ConsolidationRequest,
+                                MaxConsolidationRequestsPerPayload,
+                            > = VariableList::from_ssz_bytes(&request[1..])?;
+                            acc.2.extend(Vec::<ConsolidationRequest>::from(list));
```
klkvr, line 70: I believe there are ssz impls for `Vec<T>` so we shouldn't need ssz_types here?

ryanschneider: Oh perfect, I'll give that a shot, and then remove the `ssz_types` dep and merge the two conditional modules into one dependent on just `ssz`.

crates/rpc-types-beacon/src/requests.rs
ryanschneider: Yup!  Moved as part of the larger changes in https://github.com/alloy-rs/alloy/pull/1787/commits/cac4fab81059957caea3ff13e9236be4f8d6566c.

crates/rpc-types-beacon/src/requests.rs
ryanschneider: Ok thanks @klkvr I was able to remove the ssz_types dep and just go with ssz.  I also added `const` for the payload limits to the appropriate EIP crates and moved `cfg_eval` to a workspace dep that's conditional on `serde` feature to clean up some warnings w/ `--no-default-features`.  PTAnotherL!

crates/eips/src/eip6110.rs
```diff
@@ -1,12 +1,79 @@
-//! Contains Deposit request constants, first introduced in the [Prague hardfork](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md).
+//! Contains Deposit request types, first introduced in the [Prague hardfork](https://github.com/ethereum/execution-apis/blob/main/src/engine/prague.md).
 //!
 //! See also [EIP-6110](https://eips.ethereum.org/EIPS/eip-6110): Supply validator deposits on chain
 
-use alloy_primitives::{address, Address};
+use alloy_primitives::{address, Address, FixedBytes, B256};
+#[cfg(feature = "serde")]
+use cfg_eval::cfg_eval;
```
mattsse, line 7: why do we need this?
can we do without? not familiar with this

crates/eips/src/eip7002.rs
```diff
@@ -17,3 +21,65 @@ pub static WITHDRAWAL_REQUEST_PREDEPLOY_CODE: Bytes = bytes!("   3373fffffffffff
...
+        }"#;
+
+        // Deserialize the JSON into a WithdrawalRequest struct
+        let withdrawal_request: WithdrawalRequest =
+            serde_json::from_str(json_data).expect("Failed to deserialize");
+
+        // Verify the deserialized content
+        assert_eq!(
+            withdrawal_request.source_address,
+            Address::from_str("0xAE0E8770147AaA6828a0D6f642504663F10F7d1E").unwrap()
```
mattsse, line 69: can use address! macro

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,148 @@
+use alloy_eips::{
+    eip6110::DepositRequest, eip7002::WithdrawalRequest, eip7251::ConsolidationRequest,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+#[serde(rename_all = "snake_case")]
+#[cfg_attr(feature = "ssz", derive(ssz_derive::Decode, ssz_derive::Encode))]
+/// An Electra-compatible execution requests payload.
```
mattsse, line 10: swap order, docs then derives

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,148 @@
...
+            let deposit_bytes = val.deposits.as_ssz_bytes();
+            let withdrawals_bytes = val.withdrawals.as_ssz_bytes();
+            let consolidations_bytes = val.consolidations.as_ssz_bytes();
+
+            let mut requests = Self::default();
+            requests.push_request_with_type(DEPOSIT_REQUEST_TYPE, deposit_bytes);
+            requests.push_request_with_type(WITHDRAWAL_REQUEST_TYPE, withdrawals_bytes);
+            requests.push_request_with_type(CONSOLIDATION_REQUEST_TYPE, consolidations_bytes);
+            requests
+        }
```
mattsse, line 121: I'd prefer to have this as a native function of `ExecutionRequestsV4` as well

`ExecutionRequestsV4::to_requests(&self) {self.into()} `

mattsse (changes requested):
some nits and qs

crates/eips/src/eip6110.rs
klkvr: iiuc we need this because of how `serde_as` works. we could solve this with a helper like `serde(with = "from_str")` I guess

similarly to https://github.com/foundry-rs/foundry/blob/b4a47336038cd4d1342eedb3f9555772585e745f/crates/config/src/lib.rs#L2514-L2534 but without lowercase

klkvr (changes requested):

crates/eips/src/eip6110.rs
ryanschneider: Its necessary to make `#[serde_as]` conditional on the `serde` feature: https://docs.rs/serde_with/latest/serde_with/guide/serde_as/index.html#gating-serde_as-on-features

Without it we get compilation failures running e.g. `cargo +nightly check --no-default-features --target riscv32imac-unknown-none-elf --manifest-path crates/consensus/Cargo.toml` in the no_std GHA.

ryanschneider: FWIW @klkvr mentioned in telegram that we could move from `serde_as` to `serde_with` to remove the dep but IMO thats outside the scope of this PR.

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,148 @@
...
+                                    size,
+                                ));
+                            }
+                            acc.0.extend(list);
+                        }
+                        WITHDRAWAL_REQUEST_TYPE => {
+                            let list: Vec<WithdrawalRequest> = Vec::from_ssz_bytes(&request[1..])?;
+                            let size = list.len();
+                            if size > MAX_WITHDRAWAL_REQUESTS_PER_BLOCK {
+                                return Err(TryFromRequestsError::RequestPayloadSizeExceeded(
```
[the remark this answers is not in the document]
mattsse: those are probably fine, the request object is now versioned and I assume if this changes then we'd need a V5/V6 where the other rules are enforced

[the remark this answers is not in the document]
ryanschneider: My arugment for including is that `ExecutionRequestsV4` is specific to Electra, so if for example these values change (or new request types are added) it'll only happen during a hard fork and we'll need a `V5` to accommodate them.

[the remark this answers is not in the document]
mattsse: yeah, that makes sense

crates/eips/src/eip6110.rs
ryanschneider: Ah we cross-posted so I didn't see the foundry link, I'll take a stab at this later this afternoon, thanks!

ryanschneider:
Thanks for the latest feedback, I got caught in some other tasks but will address this round later this afternoon.

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,148 @@
...
+        SszDecodeError(DecodeError),
+        /// Requests of request_type exceeds Electra size limits
+        RequestPayloadSizeExceeded(u8, usize),
+    }
+
+    impl From<DecodeError> for TryFromRequestsError {
+        fn from(value: DecodeError) -> Self {
+            Self::SszDecodeError(value)
+        }
+    }
```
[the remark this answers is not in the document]
ryanschneider: Added in https://github.com/alloy-rs/alloy/pull/1787/commits/03ff69eb92352848df1fe6394e997f5d243cb250

crates/eips/src/eip6110.rs
ryanschneider: Ok, I added a new `alloy_serde::ssz::json::uint` (de)serializer that if we're happy with I can switch other places to start using as well (in a follow-on PR).  This let me get rid of `cfg_eval` and stop using `serde_with` in `alloy-eips`.

```rust
use alloy_serde;

#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
pub struct Container {
    #[serde(with = "alloy_serde::ssz::json::uint")]
    value: u64,
}

let val = Container { value: 18112749083033600 };
let s = serde_json::to_string(&val).unwrap();
 assert_eq!(s, "{\"value\":\"18112749083033600\"}");

let deserialized: Container = serde_json::from_str(&s).unwrap();
assert_eq!(val, deserialized);
```

ryanschneider:
Ok, I think all new feedback has been addressed, let me know what you think, especially around https://github.com/alloy-rs/alloy/pull/1787#discussion_r1884556448.

Cargo.toml
```diff
@@ -136,6 +136,7 @@ serde_with = { version = "3", default-features = false, features = ["macros"] }
 auto_impl = "1.2"
 base64 = "0.22"
 bimap = "0.6"
+cfg_eval = "0.1.2"
```
ryanschneider, line 139: oops will remove!

crates/serde/src/ssz.rs
```diff
@@ -0,0 +1,50 @@
+//! Serde functions for encoding SSZ primitives.
```
ryanschneider, line 1: We could consider naming this module `ssz_json` and remove the inner `json` mod.

ryanschneider:
Also Claude recommended a pretty good refactor on the `try_from` implementation, if you think this is worth doing let me know and I'll try it out: https://claude.site/artifacts/e420d91a-8c2d-41ef-a40a-51666a8fa0e6

ryanschneider:
^ Actually I went ahead and nerd sniped myself into integrating it. 😊

crates/serde/src/ssz.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// }
+    ///
+    /// let val = Container { value: 18112749083033600 };
+    /// let s = serde_json::to_string(&val).unwrap();
+    /// assert_eq!(s, "{\"value\":\"18112749083033600\"}");
+    ///
+    /// let deserialized: Container = serde_json::from_str(&s).unwrap();
+    /// assert_eq!(val, deserialized);
+    /// ```
+    pub mod uint {
```
mattsse, line 26: this isn't limited to uint

crates/serde/src/ssz.rs
mattsse, line 1: isn't this just display+fromstr, not exclusive to ssz?

mattsse, line 1: can we rename this entire module to displayfromstr ?

crates/serde/src/ssz.rs
```diff
@@ -0,0 +1,50 @@
+//! Serde functions for encoding SSZ primitives.
+
+/// Serde functions for encoding SSZ primitives using the "canonical JSON mapping" described
+/// in the consensus-specs here: <https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#json-mapping>
+pub mod json {
```
mattsse, line 5: these nested mods aren't necessary then, we can keep this note though

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,181 @@
...
+            requests
+        }
+    }
+
+    impl ExecutionRequestsV4 {
+        /// Convert the [ExecutionRequestsV4] into a [Requests].
+        pub fn to_requests(&self) -> Requests {
+            self.into()
+        }
+    }
```
mattsse, line 155: can we move this top of mod?

crates/rpc-types-beacon/src/requests.rs
```diff
@@ -0,0 +1,181 @@
...
+                fn accumulate(mut self, request: &[u8]) -> Result<Self, TryFromRequestsError> {
+                    if request.is_empty() {
+                        return Err(TryFromRequestsError::EmptyRequest);
+                    }
+
+                    let (request_type, payload) =
+                        request.split_first().expect("already checked for empty");
+
+                    match *request_type {
+                        DEPOSIT_REQUEST_TYPE => {
```
mattsse, line 79: this doesn't enforce any ordering, but this is fine because this is validated via the requests hash anyway

mattsse (changes requested):
last pedantic nits

crates/serde/src/ssz.rs
ryanschneider: Ya will do, personally I have a slight preference for the current naming since it says _why_ the encoding is used instead instead of _how_ the encoding works, but I'll defer to you all.

ryanschneider: Renamed in https://github.com/alloy-rs/alloy/pull/1787/commits/daf77039c827bb5491f570fb7b64158339b45682

crates/rpc-types-beacon/src/requests.rs
ryanschneider: Yup, moved in https://github.com/alloy-rs/alloy/pull/1787/commits/daf77039c827bb5491f570fb7b64158339b45682.  Required some addiitonal `#[cfg(feature = "ssz")` pragmas though.

crates/rpc-types-beacon/src/requests.rs
ryanschneider: Ya, technically this'll accept an "out of order" `Requests` (and then reorder it if converted back).  Agree it's fine since the hashes wouldn't match, if it ends up coming up often we can add a check that the types are in order.

mattsse (approved):
lgtm

PR #1790, opened 2024-12-13
fix(`transport`): allow `RetryPolicy` to be set via layer

yash-atreya:
## Motivation

Currently, we cannot set custom retry policies (`RetryPolicy`) as the `RetryBackoffService` defaults to the `RateLimitRetryPolicy` type here: https://github.com/alloy-rs/alloy/blob/663e926b1b721184c78433631e4c86647d9e2222/crates/transport/src/layers/retry.rs#L81.

## Solution

Make the `RetryBackoffLayer` and `RetryBackoffService` generic over `RetryPolicy` with `RateLimitRetryPolicy` as the default. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm

PR #1793, opened 2024-12-13
feat: add info tx types

mattsse:
upstream's reth's `TransactionMeta` type

and moves the TxInfo type to consensus from rpc.

these two types are almost identical, but info has optional fields because this is primarily used in rpc where the block context is optional.

naming is a bit horrible, but we can break this if we can come up with a better way to deal with this, perhaps introducing BlockContext or something similar


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -80,6 +79,26 @@ where
...
+    /// This contains various metadata about the transaction and block context if available.
+    pub fn info(&self) -> TransactionInfo {
+        TransactionInfo {
+            hash: Some(self.tx_hash()),
+            index: self.transaction_index,
+            block_hash: self.block_hash,
+            block_number: self.block_number,
+            // We don't know the base fee of the block when we're constructing this from
+            // `Transaction`
+            base_fee: None,
```
klkvr, line 97: should it take `base_fee` as an argument?

mattsse: we set this via the additional builder fn: `into().with_base_fee()`

PR #1795, opened 2024-12-13
chore: reuse v3 envelope in v4 envelope

klkvr:
## Motivation

This is similar to how we handle `ExecutionPayload`s and makes it clear which fields were actually added in v4

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
yep this is better

PR #1797, opened 2024-12-13
chore: add arbitrary for block

mattsse:
ported from reth

### DIFF
--- a/crates/consensus/src/block/mod.rs
+++ b/crates/consensus/src/block/mod.rs
@@ -34,6 +34,26 @@ impl<T, H> Block<T, H> {
     }
 }
 
+#[cfg(any(test, feature = "arbitrary"))]
+impl<'a, T> arbitrary::Arbitrary<'a> for Block<T>
+where
+    T: arbitrary::Arbitrary<'a>,
+{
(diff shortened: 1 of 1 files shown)


crates/consensus/src/block/mod.rs
```diff
@@ -34,6 +34,26 @@ impl<T, H> Block<T, H> {
...
+where
+    T: arbitrary::Arbitrary<'a>,
+{
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+        // first generate up to 100 txs
+        let transactions =
+            (0..100).map(|_| T::arbitrary(u)).collect::<arbitrary::Result<Vec<_>>>()?;
+
+        // then generate up to 2 ommers
+        let ommers = (0..2).map(|_| Header::arbitrary(u)).collect::<arbitrary::Result<Vec<_>>>()?;
```
klkvr, line 48: wouldn't this always generate exactly 2 ommers?

mattsse: actually yes, copy pasted this from reth, fixing

klkvr (approved):

PR #1807, opened 2024-12-16
feat(pectra): Revert EIP-7742

ryanschneider:
## Motivation

With EIP-7742 pulled from Pectra this undos most of the changes in #1600, #1697, and #1713.

## Solution

I kept the actual `eip7742` module but am happy to delete that too.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
-.-

makes sense, mind taking a look at this @klkvr re the next blob fee calc?

klkvr (approved):
lgtm, thank you!

> I kept the actual eip7742 module but am happy to delete that too.

yeah, we can remove this as well

PR #1808, opened 2024-12-17
Add edge case tests for `extract_value` and fix the newly discovered bug

hthuz:
## Motivation

Add two edge case tests for `extract_value` in https://github.com/alloy-rs/alloy/blob/main/crates/node-bindings/src/utils.rs#L26, which extract values given keys from the line of text. 
When there is no comma (like extract `value` from string `"key: value"`), the function will fail the test.
Also add doc to better explain this function.

## Solution

The code uses different patterns for finding values for keys ending with '=' and ': ', as shown below. By changing them into the same pattern, the issue is fixed. 10 tests in `utils.rs` are all passed.

```
    // Try to find the key with '='
    if let Some(pos) = line.find(key_equal.as_ref()) {
        let start = pos + key_equal.len();
        let end = line[start..].find(' ').map(|i| start + i).unwrap_or(line.len());
        if start <= line.len() && end <= line.len() {
            return Some(line[start..end].trim());
        }
    }

    // If not found, try to find the key with ': '
    if let Some(pos) = line.find(key_colon.as_ref()) {
        let start = pos + key_colon.len();
       // NOTE that there's no map here. If end of line is met, start + end will definited > line.len()
        let end = line[start..].find(',').unwrap_or(line.len()); // Assuming comma or end of line
        if start <= line.len() && start + end <= line.len() {
            return Some(line[start..start + end].trim());
        }
    }
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/utils.rs
+++ b/crates/node-bindings/src/utils.rs
@@ -22,7 +22,7 @@ pub(crate) fn unused_port() -> u16 {
 }
 
 /// Extracts the value for the given key from the line of text.
-/// 
+///
 /// It supports keys that end with '=' or ': '.
 /// For keys end with '=', find value until ' ' is encountered or end of line
 /// For keys end with ':', find value until ',' is encountered or end of line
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ty

PR #1809, opened 2024-12-18
fix: relax legacy chain id check

mattsse:
closes https://github.com/succinctlabs/kona/issues/31
closes https://github.com/anton-rs/kona/issues/874

tenderly apparently always returns the chainid for legacy transactions

imo we relax this check because we set the correct (optional) chainid after deserialization


klkvr (approved):
lgtm

crates/consensus/src/transaction/legacy.rs
```diff
@@ -516,9 +516,11 @@ pub mod signed_legacy_serde {
         let SignedLegacy { tx, signature, hash } = SignedLegacy::deserialize(deserializer)?;
         let (parity, chain_id) = from_eip155_value(signature.v.to())
             .ok_or_else(|| serde::de::Error::custom("invalid EIP-155 signature parity value"))?;
-        if let Some(tx_chain_id) = tx.chain_id() {
-            // Some nodes respond with 0 chain ID for legacy transactions when it is missing.
-            if tx_chain_id > 0 && chain_id != Some(tx_chain_id) {
+
+        // Note: some implementations always set the chain id in the response, so we only check if
+        // the differ if both are set.
```
DaniPopes, line 521:
```suggestion
        // they differ if both are set.
```

PR #1818, opened 2024-12-22
fix(alloy-eips): `SimpleCoder::decode_one()` should return `Ok(None)`

### LINKED ISSUE #1388: [Bug] SidecarBuilder fails to encode anything bigger than 1 blob of data
### Component

consensus, eips, genesis

├── alloy v0.3.6 │   ├── alloy-consensus v0.3.6 │   │   ├── alloy-eips v0.3.6 │   │   │   ├── alloy-eip2930 v0.1.0 │   │   │   │   ├── alloy-primitives v0.8.5 │   │   │   │   │   ├── alloy-rlp v0.3.8 │   │   │   │   │   │   ├── alloy-rlp-derive v0.3.8 (proc-macro) │   │   │   │   │   │   ├── alloy-rlp v0.3.8 (*) │   │   │   │   ├── alloy-rlp v0.3.8 (*) │   │   │   ├── alloy-eip7702 v0.1.0 │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-rlp v0.3.8 (*) │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   ├── alloy-rlp v0.3.8 (*) │   │   │   ├── alloy-serde v0.3.6 │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   ├── alloy-primitives v0.8.5 (*) │   │   ├── alloy-rlp v0.3.8 (*) │   │   ├── alloy-serde v0.3.6 (*) │   ├── alloy-contract v0.3.6 │   │   ├── alloy-dyn-abi v0.8.5 │   │   │   ├── alloy-json-abi v0.8.5 │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-sol-type-parser v0.8.5 │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   ├── alloy-sol-type-parser v0.8.5 (*) │   │   │   ├── alloy-sol-types v0.8.5 │   │   │   │   ├── alloy-json-abi v0.8.5 (*) │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-sol-macro v0.8.5 (proc-macro) │   │   │   │   │   ├── alloy-sol-macro-expander v0.8.5 │   │   │   │   │   │   ├── alloy-json-abi v0.8.5 │   │   │   │   │   │   │   ├── alloy-primitives v0.8.5 │   │   │   │   │   │   │   ├── alloy-sol-type-parser v0.8.5 │   │   │   │   │   │   ├── alloy-sol-macro-input v0.8.5 │   │   │   │   │   │   │   ├── alloy-json-abi v0.8.5 (*) │   │   │   │   │   ├── alloy-sol-macro-input v0.8.5 (*) │   │   ├── alloy-json-abi v0.8.5 (*) │   │   ├── alloy-network v0.3.6 │   │   │   ├── alloy-consensus v0.3.6 (*) │   │   │   ├── alloy-eips v0.3.6 (*) │   │   │   ├── alloy-json-rpc v0.3.6 │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-sol-types v0.8.5 (*) │   │   │   ├── alloy-network-primitives v0.3.6 │   │   │   │   ├── alloy-eips v0.3.6 (*) │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-serde v0.3.6 (*) │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   ├── alloy-rpc-types-eth v0.3.6 │   │   │   │   ├── alloy-consensus v0.3.6 (*) │   │   │   │   ├── alloy-eips v0.3.6 (*) │   │   │   │   ├── alloy-network-primitives v0.3.6 (*) │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-rlp v0.3.8 (*) │   │   │   │   ├── alloy-serde v0.3.6 (*) │   │   │   │   ├── alloy-sol-types v0.8.5 (*) │   │   │   ├── alloy-serde v0.3.6 (*) │   │   │   ├── alloy-signer v0.3.6 │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   ├── alloy-sol-types v0.8.5 (*) │   │   ├── alloy-network-primitives v0.3.6 (*) │   │   ├── alloy-primitives v0.8.5 (*) │   │   ├── alloy-provider v0.3.6 │   │   │   ├── alloy-chains v0.1.34 │   │   │   ├── alloy-consensus v0.3.6 (*) │   │   │   ├── alloy-eips v0.3.6 (*) │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   ├── alloy-network v0.3.6 (*) │   │   │   ├── alloy-network-primitives v0.3.6 (*) │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   ├── alloy-pubsub v0.3.6 │   │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-transport v0.3.6 │   │   │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   ├── alloy-rpc-client v0.3.6 │   │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   │   ├── alloy-primitives v0.8.5 (*) │   │   │   │   ├── alloy-pubsub v0.3.6 (*) │   │   │   │   ├── alloy-transport v0.3.6 (*) │   │   │   │   ├── alloy-transport-http v0.3.6 │   │   │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   │   │   ├── alloy-transport v0.3.6 (*) │   │   │   │   ├── alloy-transport-ipc v0.3.6 │   │   │   │   │   ├── alloy-json-rpc v0.3.6 (*) │   │   │   │   │   ├── alloy-pubsub v0.3.6 (*) │   │   │   │   │   ├── alloy-transport v0.3.6 (*) │   │   │   │   ├── alloy-transport-ws v0.3.6 │   │   │   │   │   ├── alloy-pubsub v0.3.6 (*) │   │   │   │   │   ├── alloy-transport v0.3.6 (*) │   │   │   ├── alloy-rpc-types-eth v0.3.6 (*) │   │   │   ├── alloy-transport v0.3.6 (*) │   │   │   ├── alloy-transport-http v0.3.6 (*) │   │   │   ├── alloy-transport-ipc v0.3.6 (*) │   │   │   ├── alloy-transport-ws v0.3.6 (*) │   │   ├── alloy-pubsub v0.3.6 (*) │   │   ├── alloy-rpc-types-eth v0.3.6 (*) │   │   ├── alloy-sol-types v0.8.5 (*) │   │   ├── alloy-transport v0.3.6 (*) │   ├── alloy-core v0.8.5 │   │   ├── alloy-dyn-abi v0.8.5 (*) │   │   ├── alloy-json-abi v0.8.5 (*) │   │   ├── alloy-primitives v0.8.5 (*) │   │   └── alloy-sol-types v0.8.5 (*) │   ├── alloy-eips v0.3.6 (*) │   ├── alloy-network v0.3.6 (*) │   ├── alloy-provider v0.3.6 (*) │   ├── alloy-pubsub v0.3.6 (*) │   ├── alloy-rpc-client v0.3.6 (*) │   ├── alloy-rpc-types v0.3.6 │   │   ├── alloy-rpc-types-eth v0.3.6 (*) │   │   ├── alloy-serde v0.3.6 (*) │   ├── alloy-signer v0.3.6 (*) │   ├── alloy-signer-local v0.3.6 │   │   ├── alloy-consensus v0.3.6 (*) │   │   ├── alloy-network v0.3.6 (*) │   │   ├── alloy-primitives v0.8.5 (*) │   │   ├── alloy-signer v0.3.6 (*) │   ├── alloy-transport v0.3.6 (*) │   ├── alloy-transport-http v0.3.6 (*) │   ├── alloy-transport-ipc v0.3.6 (*) │   └── alloy-transport-ws v0.3.6 (*)

### Operating System

Linux

### Describe the bug

I've noticed that all of my blob transactions have data only in the last blob, while the preceding ones are filled with 0s. 

Upon closer inspection it turns out that the `SidecarBuilder` behaves weirdly once you approach the 1 blob boundary. 

Reproduction:
```rust
    #[test]
    fn reproduction() {
        use alloy::consensus::SidecarBuilder;
        use alloy::consensus::SidecarCoder;
        use alloy::{consensus::SimpleCoder, eips::eip4844};
        use rand::RngCore;
        use rand::SeedableRng;

        let mut rng = rand::rngs::SmallRng::from_seed([0; 32]);

        let mut random_data = |num_input_bytes| {
            let mut data = vec![0; num_input_bytes];
            rng.fill_bytes(&mut data);

            data
        };

        let encode_blobs = |data: Vec<u8>| {
            let builder = SidecarBuilder::from_coder_and_data(SimpleCoder::default(), &data);

            builder.build().unwrap().blobs
        };

        let decode_blobs = |blobs: Vec<eip4844::Blob>| {
            let chunks = SimpleCoder::default().decode_all(&blobs)?;

            Some(chunks.into_iter().flatten().collect::<Vec<u8>>())
        };

        // This works for data sizes in the range (0,126915)
        {
            let small_amount_of_data = random_data(50_000);
            let blobs = encode_blobs(small_amount_of_data.clone());

            let recreated_data = decode_blobs(blobs).unwrap();

            assert_eq!(small_amount_of_data, recreated_data);
        }

        // for the range [126915, 126946) we get back None when decoding the data
        {
            let around_the_blob_boundary = random_data(126915);
            let blobs = encode_blobs(around_the_blob_boundary.clone());
            assert_eq!(blobs.len(), 1);

            let recreated_data = decode_blobs(blobs);

            assert!(recreated_data.is_none());
        }

        // and anything in [126946, +Inf) we get back empty data
        {
            let bigger_than_a_blob_when_encoded = random_data(126946);
            let blobs = encode_blobs(bigger_than_a_blob_when_encoded.clone());
            assert_eq!(blobs.len(), 2);
            // every blob, except the last one, is filled with 0s
            assert!(blobs[0].iter().all(|byte| *byte == 0));

            let recreated_data = decode_blobs(blobs).unwrap();

            assert!(recreated_data.is_empty());
        }
    }
```

I think this might be happening due to the way `SidecarBuilder` is allocating and giving blobs to the underlying coder:
[preallocation of empty blobs](https://github.com/alloy-rs/alloy/blob/754431d1cc935021ae3acd3ad3e88406bab81ac1/crates/eips/src/eip4844/builder.rs#L48)
```rust
    /// Create a new builder, preallocating room for `capacity` blobs, and push
    /// an empty blob to it.
    pub fn with_capacity(capacity: usize) -> Self {
        let mut blobs = Vec::with_capacity(capacity);
        blobs.push(Blob::new([0u8; BYTES_PER_BLOB]));
        Self { blobs, fe: 0 }
    }
``` 

this makes sure that for the last example there are two blobs which the coder can fill. 

The issue (probably) is due to the coder [asking for the current blob](https://github.com/alloy-rs/alloy/blob/754431d1cc935021ae3acd3ad3e88406bab81ac1/crates/eips/src/eip4844/builder.rs#L101) and that is always the last blob that the builder preallocated.

What ends up happening is that out of all of the preallocated blobs only the last one is filled.

Since the [docstring](https://github.com/alloy-rs/alloy/blob/754431d1cc935021ae3acd3ad3e88406bab81ac1/crates/eips/src/eip4844/builder.rs#L266) says this should be used for large amounts of data;
```
/// Build a [`BlobTransactionSidecar`] from an arbitrary amount of data.
///
/// This is useful for creating a sidecar from a large amount of data,
/// which is then split into blobs. It delays KZG commitments and proofs
/// until all data is ready.
///
/// [`BlobTransactionSidecar`]: crate::eip4844::BlobTransactionSidecar
```

these issues caught me by surprise. 

<s>Giving the input data byte by byte to the builder solves the issues (but is inefficient, goes without saying)</s>
```rust
        let encode_blobs = |data: Vec<u8>| {
            let mut builder = SidecarBuilder::<SimpleCoder>::new();
            for byte in &data {
                builder.ingest(std::slice::from_ref(byte));
            }

            builder.build().unwrap().blobs
        };
```

Nope, byte-by-byte doesn't  solve it, just pushes the limit to 256000B where we again get `None` from `decode_all`.

StackOverflowExcept1on:
## Motivation

Resolves #1388 

For example you want to decode 4096 FEs. In this case the first time `decode_one()` will return data, but the second time (because of `loop`) it will return `Err(())` and it will fail the whole `decode_all()`.

https://github.com/alloy-rs/alloy/blob/24ab11e86866f6fbf6984adb8ff3f7737e398db4/crates/eips/src/eip4844/builder.rs#L246-L259

## Solution

Returns `Ok(None)` if there is no data (empty iterator, length prefix is 0).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip4844/builder.rs
```diff
@@ -434,6 +436,17 @@ mod tests {
         assert_eq!(decoded, data);
     }
 
+    #[test]
+    fn big_ingestion_strategy() {
+        let data = vec![1u8; 126_945];
```
StackOverflowExcept1on, line 441: Blob has size of `128 * 1024` bytes. First 32 bytes are used as length, thus leaving 4095 FEs / 4096 FEs. This constant was calculated as `128 * 1024 - 32 - 4095` (`4095` zeros in each FE).

crates/eips/src/eip4844/builder.rs
```diff
@@ -244,8 +247,21 @@ impl SidecarCoder for SimpleCoder {
...
+            .iter()
+            .flat_map(|blob| blob.chunks(FIELD_ELEMENT_BYTES).map(WholeFe::new))
+            .any(|fe| fe.is_none())
+        {
+            return None;
+        }
+
+        let mut fes = blobs
+            .iter()
+            .flat_map(|blob| blob.chunks(FIELD_ELEMENT_BYTES).map(WholeFe::new_unchecked));
```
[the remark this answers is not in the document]
StackOverflowExcept1on: I decided to do this in a non-lazy way to do not pass `Option<Option<Option<Option<WholeFe>>` to `decode_one()`.

crates/eips/src/eip4844/mod.rs
```diff
@@ -34,7 +34,7 @@ pub const BLS_MODULUS_BYTES: B256 =
 pub const BLS_MODULUS: U256 = U256::from_be_bytes(BLS_MODULUS_BYTES.0);
 
 /// Size a single field element in bytes.
-pub const FIELD_ELEMENT_BYTES: u64 = 32;
```
StackOverflowExcept1on, line 37: It's breaking change, but I don't think it should break a lot of user-space code.

mattsse: can we keep this as u64 and avoid the breaking change, this would be more consistent with the other constant

crates/eips/src/eip4844/utils.rs
mattsse, line 1: no functional changes here, right?

crates/eips/src/eip4844/builder.rs
```diff
@@ -96,7 +97,8 @@ impl PartialSidecar {
 
     /// Get a mutable reference to the current blob.
     fn current_blob_mut(&mut self) -> &mut Blob {
-        self.blobs.last_mut().expect("never empty")
+        let last_unused_blob_index = self.fe / FIELD_ELEMENTS_PER_BLOB as usize;
+        self.blobs.get_mut(last_unused_blob_index).expect("never empty")
```
mattsse, line 101: this makes sense

crates/eips/src/eip4844/builder.rs
```diff
@@ -192,10 +195,12 @@ impl SimpleCoder {
...
     ///
     /// Returns `Ok(Some(data))` if there is some data.
-    /// Returns `Ok(None)` if there is no data (length prefix is 0).
+    /// Returns `Ok(None)` if there is no data (empty iterator, length prefix is 0).
     /// Returns `Err(())` if there is an error.
     fn decode_one<'a>(mut fes: impl Iterator<Item = WholeFe<'a>>) -> Result<Option<Vec<u8>>, ()> {
-        let first = fes.next().ok_or(())?;
+        let Some(first) = fes.next() else {
+            return Ok(None);
+        };
```
mattsse, line 203: this also makes sense, because before this would always return an error for iters len > 1, right?

crates/eips/src/eip4844/builder.rs
```diff
@@ -244,8 +249,21 @@ impl SidecarCoder for SimpleCoder {
     fn finish(self, _builder: &mut PartialSidecar) {}
 
     fn decode_all(&mut self, blobs: &[Blob]) -> Option<Vec<Vec<u8>>> {
-        let mut fes =
-            blobs.iter().flat_map(|blob| blob.chunks(32).map(WholeFe::new)).map(Option::unwrap);
+        if !matches!(blobs.len(), 1..=MAX_BLOBS_PER_BLOCK) {
+            return None;
+        }
```
mattsse, line 254: this restriction will be problematic soon, because this constant should no longer be used, so I'd rather remove this check entirely

mattsse (changes requested):
tysm for looking into this,

all of this looks fine to me, although not an expert when it comes to the internals, but these changes match with what I discovered when I looked into the issue earlier

some nits

crates/eips/src/eip4844/builder.rs
StackOverflowExcept1on: what's the problem? if the number of blobs is increased, we just update the constants and everything will work.

crates/eips/src/eip4844/utils.rs
StackOverflowExcept1on: I could leave it `u64`, but then I'd have to make `as usize` in all usages, which isn't very convenient. Maybe it makes sense to introduce `FIELD_ELEMENT_BYTES_USIZE: usize = FIELD_ELEMENT_BYTES as usize`? Yes, there are no other functional changes.

crates/eips/src/eip4844/mod.rs
StackOverflowExcept1on: dbaa6efd0c87c035c39c1b38ea67025d36846197

crates/eips/src/eip4844/builder.rs
mattsse: the max number of blobs will be decoupled from this constant https://github.com/ethereum/EIPs/pull/9129

@klkvr should we use the `eip7840::BlobParams::cancun()` for now or remove this entirely, I guess the proper way of doing this would be to pass around the max_block setting but this seems a bit much...

klkvr: yeah don't think we need this check

custom implementations can store the constant in coder if needed

StackOverflowExcept1on: Ok, I'll remove the use of `MAX_BLOBS_PER_BLOCK` in this place and here for now:
https://github.com/alloy-rs/alloy/blob/8ab278c57ed76a3fb37947bc1bbde62e5ee70d90/crates/eips/src/eip4844/builder.rs#L206-L209

StackOverflowExcept1on: We could then do a check that `blobs: &[Blob]` is not empty and that `num_bytes` does not exceed 2 MiB.

crates/eips/src/eip4844/builder.rs
```diff
@@ -244,8 +249,21 @@ impl SidecarCoder for SimpleCoder {
...
+            .iter()
+            .flat_map(|blob| blob.chunks(FIELD_ELEMENT_BYTES_USIZE).map(WholeFe::new))
+            .any(|fe| fe.is_none())
+        {
+            return None;
+        }
+
+        let mut fes = blobs
+            .iter()
+            .flat_map(|blob| blob.chunks(FIELD_ELEMENT_BYTES_USIZE).map(WholeFe::new_unchecked));
```
mattsse, line 266: yeah as pointed out, this overhead should be fine

mattsse (approved):

PR #1822, opened 2024-12-23
chore: replace derive_more with thiserror

mattsse:
with thiserror we can have this for no-std now

### DIFF
--- a/crates/rpc-types-eth/src/block.rs
+++ b/crates/rpc-types-eth/src/block.rs
@@ -275,24 +275,14 @@ impl<H: BlockHeader> HeaderResponse for Header<H> {
 }
 
 /// Error that can occur when converting other types to blocks
-#[derive(Clone, Copy, Debug, derive_more::Display)]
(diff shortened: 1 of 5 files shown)


klkvr (approved):

crates/rpc-types-eth/src/block.rs
```diff
@@ -275,26 +275,16 @@ impl<H: BlockHeader> HeaderResponse for Header<H> {
...
+#[derive(Clone, Copy, Debug, thiserror::Error)]
 pub enum BlockError {
     /// A transaction failed sender recovery
-    #[display("transaction failed sender recovery")]
+    #[error("transaction failed sender recovery")]
     InvalidSignature,
     /// A raw block failed to decode
-    #[display("failed to decode raw block {_0}")]
+    #[error("failed to decode raw block {0}")]
     RlpDecodeRawBlock(alloy_rlp::Error),
```
DaniPopes, line 285: missing `#[source]`

PR #1827, opened 2024-12-23
feat: add map transactions fn

mattsse:
adds helpers to map the tx type

### DIFF
--- a/crates/consensus/src/block/mod.rs
+++ b/crates/consensus/src/block/mod.rs
@@ -43,6 +43,42 @@ impl<T, H> Block<T, H> {
     pub fn into_body(self) -> BlockBody<T> {
         self.body
     }
+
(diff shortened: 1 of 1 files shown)


crates/consensus/src/block/mod.rs
```diff
@@ -43,6 +43,42 @@ impl<T, H> Block<T, H> {
     pub fn into_body(self) -> BlockBody<T> {
         self.body
     }
+
+    /// Converts the block's transaction type by applying a function to each transaction.
+    ///
+    /// Returns the block with the new transaction type.
+    pub fn map_transactions<U>(self, f: impl Fn(T) -> U) -> Block<U, H> {
```
DaniPopes, line 50: FnMut

mattsse: done

DaniPopes (approved):

PR #1828, opened 2024-12-23
feat: EIP-7840

klkvr:
## Motivation

Implements EIP-7840 https://github.com/ethereum/EIPs/pull/9129

## Solution

`alloy_eips::eip7840` has 2 structs: `BlobScheduleItem` which is the one present in genesis config as per EIP, and `BlobParams` which is a new aggregated configuration for blob constants, including an update fraction which is going to change in Prague as well with EIP-7691. All methods of header which need to perform blobfee-related calculations are now accepting `BlobParams` similarly to methods performing eip1559 calculations.

`ChainConfig` is extended with `blob_schedule` which is a mapping from `String` to `BlobScheduleItem`. It is not great because this means that we'd need to perform parsing of strings into the actual hardforks names on demand and won't have type-safety during parsing.

One way to improve this could be to make `Genesis` and `ChainConfig` generic over hardforks enum but this might be quite invasive, wdyt @mattsse 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -6,10 +6,12 @@ use crate::{eip4844, eip7691};
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct BlobScheduleItem {
+    /// Target blob count for the block.
     #[cfg_attr(feature = "serde", serde(rename = "target"))]
-    target_blob_count: u64,
+    pub target_blob_count: u64,
+    /// Max blob count for the block.
     #[cfg_attr(feature = "serde", serde(rename = "max"))]
-    max_blob_count: u64,
+    pub max_blob_count: u64,
 }
 
 /// Configuration for the blob-related calculations.
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -1,0 +1,56 @@
+//! Contains constants and utility functions for [EIP-7840](https://github.com/ethereum/EIPs/tree/master/EIPS/eip-7840.md)
+
+use crate::{eip4844, eip7691};
+
+/// Configuration for the blob-related calculations.
+#[derive(Debug)]
+pub struct BlobParams {
+    /// Target blob count for the block.
+    pub target_blob_count: u64,
+    /// Max blob count for the block.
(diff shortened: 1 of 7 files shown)


crates/eips/src/eip7840.rs
```diff
@@ -0,0 +1,66 @@
+//! Contains constants and utility functions for [EIP-7840](https://github.com/ethereum/EIPs/tree/master/EIPS/eip-7840.md)
+
+use crate::{eip4844, eip7691};
+
+/// A single item of `blobSchedule` defined in EIP-7840.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct BlobScheduleItem {
+    #[cfg_attr(feature = "serde", serde(rename = "target"))]
+    target_blob_count: u64,
```
mattsse, line 10: can we make these pub?

crates/genesis/src/lib.rs
```diff
@@ -479,6 +480,10 @@ pub struct ChainConfig {
     /// The deposit contract address
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub deposit_contract_address: Option<Address>,
+
+    /// The blob schedule for the chain.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub blob_schedule: BTreeMap<String, BlobScheduleItem>,
```
mattsse, line 486: what's the key here? 
can we add to docs?

mattsse (changes requested):
nits

mattsse:
> It is not great because this means that we'd need to perform parsing of strings into the actual hardforks names on demand and won't have type-safety during parsing.

imo this is fine because genesis -> chainspec is an additional step anyway

klkvr:
> > It is not great because this means that we'd need to perform parsing of strings into the actual hardforks names on demand and won't have type-safety during parsing.
> 
> imo this is fine because genesis -> chainspec is an additional step anyway

rn we're keeping Genesis directly on chainspec so we would need to update this to ensure correct hardfork names https://github.com/paradigmxyz/reth/blob/dd4715f5d1b33c695783541fe1b32d07baff8bf5/crates/chainspec/src/spec.rs#L35-L36

mattsse (approved):

PR #1842, opened 2024-12-23
chore: rm non exhaustive for envelope

mattsse:
this is a footgun because this forces users to add a _ => unreachable on a match missing new tx types


klkvr (approved):
yeah this is not great

should we do the same for `ReceiptEnvelope`?

PR #1848, opened 2024-12-25
fix: use u64 for all gas values

klkvr:
## Motivation

Some types are still using u128 for gas which is inconsistent. With this PR only gas price values are stored as u128

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
nice, this a lot better

good timing

PR #1850, opened 2024-12-27
feat: add more builder style fns

mattsse:
adds functions that make it easier to create these rpc in a builder like API.

this will allow more simplifications

https://github.com/paradigmxyz/reth/blob/d644900a8078db976cdbfb310a19df5df9b33125/crates/rpc/rpc-types-compat/src/block.rs#L114-L114

### DIFF
--- a/crates/network-primitives/src/block.rs
+++ b/crates/network-primitives/src/block.rs
@@ -130,6 +130,11 @@ impl<T> BlockTransactions<T> {
 }
 
 impl<T: TransactionResponse> BlockTransactions<T> {
+    /// Creates a new [`BlockTransactions::Hashes`] variant from the given iterator of transactions.
+    pub fn new_hashes(txs: impl IntoIterator<Item = T>) -> Self {
+        Self::Hashes(txs.into_iter().map(|tx| tx.tx_hash()).collect())
+    }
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/block.rs
```diff
@@ -130,6 +130,11 @@ impl<T> BlockTransactions<T> {
 }
 
 impl<T: TransactionResponse> BlockTransactions<T> {
+    /// Creates a new [`BlockTransactions::Hashes`] variant from the given iterator of transactions.
+    pub fn new_hashes(txs: impl IntoIterator<Item = T>) -> Self {
+        Self::Hashes(txs.into_iter().map(|tx| tx.tx_hash()).collect())
+    }
```
DaniPopes, line 136: maybe
```suggestion
    /// Creates a new [`BlockTransactions::Hashes`] variant from the given iterator of transactions.
    pub fn new_hashes(txs: impl IntoIterator<Item = impl AsRef<T>>) -> Self {
        Self::Hashes(txs.into_iter().map(|tx| tx.as_ref().tx_hash()).collect())
    }
```

DaniPopes (approved):

PR #1859, opened 2024-12-29
feat!: remove T: Transport from public APIs

DaniPopes:
Currently, every high-level type derived from RPC client contains a `T: Transport + Clone = BoxTransport` generic parameter.

However, there is, AFAICT, no benefit to exposing this generic in high-level types. The dynamic dispatch performance impact is negligible compared to what a transport actually does (many syscalls, network calls, etc.), and it only complicates generic bounds where any of those types are used.

- Remove `T: Transport (+ Clone)` generic parameters from almost every public API by always boxing the transport internally.
- Add a public `IntoBoxTransport` for where this generic is still needed, mainly the builder/layer types.
- Remove `BoxTransportConnect` and rename (with deprecation) `connect_boxed` methods to simpler names.
  - driveby: Rename (with deprecation) `connect_builtin` and similar to just `connect`
- Deprecate `.boxed()` methods, as they just return `self` now

Breaking changes have been adjusted to not require any changes in alloy-core (`sol!` macro); mainly, `CallBuilder` and `Event` will keep the `T` generic, but the trait bounds are replaced with a fake `Transport` trait only exposed to `sol!`, so that the compiler is able to infer the type to be `()`, the only implementer of that fake trait.

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -7,7 +7,6 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
 use alloy_provider::{PendingTransactionBuilder, Provider};
 use alloy_rpc_types_eth::{state::StateOverride, AccessList, BlobTransactionSidecar, BlockId};
 use alloy_sol_types::SolCall;
-use alloy_transport::Transport;
 use std::{
     future::{Future, IntoFuture},
     marker::PhantomData,
@@ -16,13 +15,13 @@ use std::{
 
 /// [`CallBuilder`] using a [`SolCall`] type as the call decoder.
 // NOTE: please avoid changing this type due to its use in the `sol!` macro.
-pub type SolCallBuilder<T, P, C, N = Ethereum> = CallBuilder<T, P, PhantomData<C>, N>;
+pub type SolCallBuilder<P, C, N = Ethereum> = CallBuilder<P, PhantomData<C>, N>;
 
 /// [`CallBuilder`] using a [`Function`] as the call decoder.
-pub type DynCallBuilder<T, P, N = Ethereum> = CallBuilder<T, P, Function, N>;
+pub type DynCallBuilder<P, N = Ethereum> = CallBuilder<P, Function, N>;
 
 /// [`CallBuilder`] that does not have a call decoder.
-pub type RawCallBuilder<T, P, N = Ethereum> = CallBuilder<T, P, (), N>;
+pub type RawCallBuilder<P, N = Ethereum> = CallBuilder<P, (), N>;
 
 /// A builder for sending a transaction via `eth_sendTransaction`, or calling a contract via
 /// `eth_call`.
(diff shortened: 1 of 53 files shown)


crates/provider/src/lib.rs
```diff
@@ -6,21 +6,23 @@
...
+pub type ReqwestProvider<N = alloy_network::Ethereum> = crate::RootProvider<N>;
 
 /// Type alias for a [`RootProvider`] using the [`Http`] transport and a hyper
 /// client.
 ///
 /// [`Http`]: alloy_transport_http::Http
 #[cfg(feature = "hyper")]
-pub type HyperProvider<N = alloy_network::Ethereum> =
-    crate::RootProvider<alloy_transport_http::HyperTransport, N>;
+pub type HyperProvider<N = alloy_network::Ethereum> = crate::RootProvider<N>;
```
DaniPopes, line 25: Should these be deprecated as well?

crates/transport/src/boxed.rs
```diff
@@ -1,32 +1,57 @@
...
+    fn into_box_transport(self) -> BoxTransport;
+}
+
+impl<T: Transport + Clone> IntoBoxTransport for T {
+    fn into_box_transport(self) -> BoxTransport {
+        // "specialization" to re-use `BoxTransport`.
+        if TypeId::of::<T>() == TypeId::of::<BoxTransport>() {
+            // SAFETY: `self` is `BoxTransport`. This is a no-op.
+            let this = std::mem::ManuallyDrop::new(self);
+            return unsafe { std::mem::transmute_copy(&this) };
```
mattsse, line 27: What's the difference between this and just a regular transmute?

mattsse (approved):
removing the (useless) transport generic is long overdue.

only have some questions and a smol ask to describe the alloy-provider <-> sol! macro relationship

we also need to migrate https://github.com/alloy-rs/examples should we do this against this branch or later?

this requires a minor release, I'm fine with doing that whenever but would prefer if we had the examples migrated already

crates/transport/src/boxed.rs
DaniPopes: transmute doesn't allow size mismatch, T has generic size so it never works

mattsse: I see, so we effectively just copy the pointer here? hence the manual drop

DaniPopes: yeah

crates/contract/src/call.rs
```diff
@@ -7,13 +7,15 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
...
 use alloy_rpc_types_eth::{state::StateOverride, AccessList, BlobTransactionSidecar, BlockId};
 use alloy_sol_types::SolCall;
-use alloy_transport::Transport;
 use std::{
     future::{Future, IntoFuture},
     marker::PhantomData,
     pin::Pin,
 };
 
+// NOTE: The `T` generic here is kept to mitigate breakage with the `sol!` macro.
```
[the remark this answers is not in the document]
DaniPopes: yes

crates/contract/src/lib.rs
```diff
@@ -35,6 +35,13 @@ pub use call::*;
...
-    pub use alloy_transport::Transport;
+
+    // Fake traits to mitigate `sol!` macro breaking changes.
+    pub trait Provider<T, N: Network>: alloy_provider::Provider<N> {}
+    impl<N: Network, P: alloy_provider::Provider<N>> Provider<(), N> for P {}
+
+    // This is done so that the compiler can infer the `T` type to be `()`, which is the only type
+    // that implements this fake `Transport` trait.
+    pub trait Transport {}
+    impl Transport for () {}
```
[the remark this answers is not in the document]
DaniPopes: the docs is don't do it, if you have to, message me

DaniPopes:
- https://github.com/foundry-rs/foundry-fork-db/pull/40
- https://github.com/alloy-rs/examples/pull/170

mattsse (approved):

PR #1869, opened 2024-12-31
feat: add conversions from rpc block to consensus

mattsse:
these conversions were missing.

the rpc block to consensus has some caveats, but should be fine

### DIFF
--- a/crates/network-primitives/src/block.rs
+++ b/crates/network-primitives/src/block.rs
@@ -109,7 +109,7 @@ impl<T> BlockTransactions<T> {
     /// Consumes the type and returns the transactions as a vector.
     ///
     /// Note: if this is an uncle or hashes, this will return an empty vector.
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/block.rs
```diff
@@ -106,6 +106,16 @@ impl<T> BlockTransactions<T> {
         }
     }
 
+    /// Consumes the type and returns the transactions as a vector.
+    ///
+    /// Note: if this is an uncle or hashes, this will return an empty vector.
+    pub fn into_vec(self) -> Vec<T> {
```
DaniPopes, line 112: into_transactions_vec? match fn above

mattsse: renamed

DaniPopes (approved):

PR #1875, opened 2024-12-31
feat: add block to payloadv1

mattsse:
https://github.com/paradigmxyz/reth/blob/b5536e4f2534b928254d10230f2b5cf1aa88f46a/crates/rpc/rpc-types-compat/src/engine/payload.rs#L52-L74

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -301,12 +301,12 @@ impl ExecutionPayloadV1 {
     /// Converts [`alloy_consensus::Block`] to [`ExecutionPayloadV1`].
     ///
     /// Note: This re-calculates the block hash.
-    pub fn from_block_slow<T: Encodable2718>(block: Block<T>) -> Self {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/payload.rs
```diff
@@ -297,6 +297,35 @@ impl ExecutionPayloadV1 {
...
+            fee_recipient: block.beneficiary,
+            state_root: block.state_root,
+            receipts_root: block.receipts_root,
+            logs_bloom: block.logs_bloom,
+            prev_randao: block.mix_hash,
+            block_number: block.number,
+            gas_limit: block.gas_limit,
+            gas_used: block.gas_used,
+            timestamp: block.timestamp,
+            extra_data: block.extra_data.clone(),
```
DaniPopes, line 323: take block by ref or remove clone

mattsse: actually, taking by ref is better here

DaniPopes (approved):

PR #1882, opened 2025-01-03
feat: add match_versioned_hashes

hoank101:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip4844/sidecar.rs
+++ b/crates/eips/src/eip4844/sidecar.rs
@@ -56,20 +56,24 @@ impl core::fmt::Debug for BlobTransactionSidecar {
 }
 
 impl BlobTransactionSidecar {
-    /// Matches versioned hashes and constructs `BlobAndProofV1` objects.
-    pub fn match_versioned_hashes(&self, versioned_hashes: &[B256]) -> Vec<Option<BlobAndProofV1>> {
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip4844/sidecar.rs
```diff
@@ -54,6 +55,24 @@ impl core::fmt::Debug for BlobTransactionSidecar {
     }
 }
 
+impl BlobTransactionSidecar {
+    /// Matches versioned hashes and constructs `BlobAndProofV1` objects.
+    pub fn match_versioned_hashes(&self, versioned_hashes: &[B256]) -> Vec<Option<BlobAndProofV1>> {
+        let mut result = vec![None; versioned_hashes.len()];
```
mattsse, line 61: this alloc is a bit unfortunate because this isn't very cheap because we need to alloc `len(versioned_hashes)` for each `BlobTransactionSidecar`

I think what we can do here instead is make this return an iterator of `(j, BlobAndProofV1)` because we're only interested in the index of the matching hash in the versioned_hashes slice

mattsse (changes requested):

mattsse (approved):

PR #1888, opened 2025-01-04
chore: shift std::error impls to core

mattsse:
more core error impls

all of those could now use thiserror but I'd like to open a good first issue for this instead

### DIFF
--- a/crates/eips/src/eip1898.rs
+++ b/crates/eips/src/eip1898.rs
@@ -247,8 +247,8 @@ pub enum ParseBlockNumberError {
 impl core::error::Error for ParseBlockNumberError {
     fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         match self {
-            Self::ParseIntErr(err) => core::error::Error::source(err),
-            Self::MissingPrefix(err) => core::error::Error::source(err),
+            Self::ParseIntErr(err) => Some(err),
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip1898.rs
```diff
@@ -600,12 +599,12 @@ impl fmt::Display for ParseBlockIdError {
...
-impl std::error::Error for ParseBlockIdError {
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+impl core::error::Error for ParseBlockIdError {
+    fn source(&self) -> Option<&(dyn core::error::Error + 'static)> {
         match self {
-            Self::ParseIntError(err) => std::error::Error::source(err),
-            Self::FromHexError(err) => std::error::Error::source(err),
-            Self::ParseError(err) => std::error::Error::source(err),
+            Self::ParseIntError(err) => core::error::Error::source(err),
+            Self::FromHexError(err) => core::error::Error::source(err),
```
DaniPopes, line 606: these are wrong btw, the source is err itself, not the source of the source

DaniPopes (approved):

crates/eips/src/eip1898.rs
mattsse: right, updating

PR #1892, opened 2025-01-06
refactor: use the params struct in more places

prestwich:
## Motivation

DRY up some code a bit

## PR Checklist

- [n/a] Added Tests
- [n/a] Added Documentation
- [ ] Breaking changes

PR #1893, opened 2025-01-06
refactor: delete `Caller` trait

prestwich:
based on #1892 

## Motivation

`Caller` trait is unused, and results in extra indirection via structs like `Arc<Weak<Client>>`. Its abstraction over `"eth_call"` vs other methods is well-covered by custom ext traits on provider, or by custom transports if low-level rewriting is needed, so it seems safe to delete it.

## Solution

Remove the caller trait

This is a breaking change, as the `Caller` is part of the public API. it is unused in alloy, in op-alloy, and in every other user i've checked

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes


prestwich:
merge commit on #1892 makes it easier to start this from scratch now

PR #1899, opened 2025-01-08
fix(`transport`): use `HttpsConnector` in `HyperTransport`

yash-atreya:
## Motivation

Currently the `HyperTransport` uses the `HttpConnector` as the default connector making it impossible to use: `ProviderBuilder::new().on_builtin("https://..")` when the `hyper` feature is enabled as it will always throw "invalid URL, scheme is not http".

## Solution

Use `HttpsConnector` from the `hyper_tls` lib as default in `HyperTransport`. 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/Cargo.toml
+++ b/crates/transport-http/Cargo.toml
@@ -54,7 +54,6 @@ reqwest = [
 ]
 hyper = [
     "dep:hyper",
-    "dep:hyper-tls",
     "dep:hyper-util",
     "dep:http-body-util",
     "dep:alloy-json-rpc",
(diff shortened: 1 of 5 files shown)


crates/transport-http/Cargo.toml
```diff
@@ -37,6 +37,7 @@ tracing = { workspace = true, optional = true }
 http-body-util = { workspace = true, optional = true }
 hyper = { workspace = true, default-features = false, optional = true }
 hyper-util = { workspace = true, features = ["full"], optional = true }
+hyper-tls = { workspace = true, optional = true }
```
mattsse, line 40: this now adds native-tls by default, should this be configurable instead?

mattsse (commented):
I'm okay with making this the default personally but wonder if we should also feature gate this? like we do with reqwest

yash-atreya:
> I'm okay with making this the default personally but wonder if we should also feature gate this? like we do with request

@mattsse feature gated behind "hyper-tls"

mattsse (approved):

PR #1901, opened 2025-01-08
feat(`provider`): instantiate recommended fillers by default

yash-atreya:
## Motivation

Instantiate `RecommendedFillers` by default. 

## Solution

- `RecommededFillers` are added when the `ProviderBuilder` is instantiated using `new()` i.e `ProviderBuilder::new()` will now return `FillProvider` with the recommended fillers instantiated.
- Users can opt-out of this behaviour using `.disable_recommended_fillers()`. 

For example, if a user want just wants `cached_nonce_management`. It can be achieved like so: 

```rust
            .disable_recommended_fillers()
            .with_cached_nonce_management()
            .on_anvil();
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -570,7 +570,8 @@ mod tests {
     use alloy_consensus::Transaction;
     use alloy_primitives::{address, b256, bytes, hex, utils::parse_units, B256};
     use alloy_provider::{
-        layers::AnvilProvider, Provider, ProviderBuilder, RootProvider, WalletProvider,
+        fillers::FillProvider, layers::AnvilProvider, utils::JoinedRecommendedFillers, Provider,
+        ProviderBuilder, RootProvider, WalletProvider,
     };
     use alloy_rpc_types_eth::AccessListItem;
     use alloy_sol_types::sol;
@@ -619,10 +620,12 @@ mod tests {
         }
     }
 
+    type AnvilFillProvider =
+        FillProvider<JoinedRecommendedFillers, AnvilProvider<RootProvider>, Ethereum>;
     /// Creates a new call_builder to test field modifications, taken from [call_encoding]
     #[allow(clippy::type_complexity)]
     fn build_call_builder(
-    ) -> CallBuilder<(), AnvilProvider<RootProvider>, PhantomData<MyContract::doStuffCall>> {
+    ) -> CallBuilder<(), AnvilFillProvider, PhantomData<MyContract::doStuffCall>> {
         let provider = ProviderBuilder::new().on_anvil();
         let contract = MyContract::new(Address::ZERO, provider);
         let call_builder = contract.doStuff(U256::ZERO, true).with_cloned_provider();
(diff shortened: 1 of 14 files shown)


crates/contract/src/call.rs
```diff
@@ -619,10 +620,12 @@ mod tests {
         }
     }
 
+    type AnvilFillProvider =
+        FillProvider<JoinedRecommendedFillers, AnvilProvider<RootProvider>, Ethereum>;
```
gakonst, line 624: Should from now on this just be `impl Provider` in the return type?

crates/contract/src/event.rs
```diff
@@ -322,8 +327,15 @@ mod tests {
 
         let poller = event.watch().await.unwrap();
 
-        let _receipt =
-            contract.doEmit().send().await.unwrap().get_receipt().await.expect("no receipt");
+        let _receipt = contract
+            .doEmit()
+            .from(from)
```
gakonst, line 332: this should be automatically set?

yash-atreya: It won't be in this case because `wallet` filler hasn't been attached

crates/provider/src/builder.rs
```diff
@@ -125,6 +139,13 @@ impl<N> Default for ProviderBuilder<Identity, Identity, N> {
     }
 }
 
+impl<L, F, N> ProviderBuilder<L, F, N> {
+    /// Opt-out of the recommended fillers by reseting the fillers stack in the [`ProviderBuilder`].
+    pub fn disable_recommended_fillers(self) -> ProviderBuilder<L, Identity, N> {
```
DaniPopes, line 144: I think this should be under `fn new` above

crates/provider/src/builder.rs
```diff
@@ -112,10 +112,29 @@ pub struct ProviderBuilder<L, F, N = Ethereum> {
...
+    ProviderBuilder<
+        Identity,
+        JoinFill<Identity, <alloy_network::Ethereum as RecommendedFillers>::RecommendedFillers>,
+        Ethereum,
+    >
+{
+    /// Create a new [`ProviderBuilder`] with the recommended filler enabled.
+    ///
+    /// Recommended fillers are preconfigured set of fillers that handle gas estimation, nonce
+    /// management, and chain-id fetching.
```
DaniPopes, line 125: let's add a paragraph to the struct's docs too, explaining the difference between new() and default()

DaniPopes (approved):

crates/provider/src/builder.rs
yash-atreya: [8e098a8](https://github.com/alloy-rs/alloy/pull/1901/commits/8e098a8eba3244259c703c9b10cea62baba8c349)

yash-atreya:
@gakonst @mattsse bump

crates/provider/src/builder.rs
```diff
@@ -112,10 +119,29 @@ pub struct ProviderBuilder<L, F, N = Ethereum> {
...
+    /// Building a provider with this setting enabled will return a [`crate::fillers::FillProvider`]
+    /// with [`crate::utils::JoinedRecommendedFillers`].
+    ///
+    /// You can opt-out of using these fillers by using the `.disable_recommended_fillers()` method.
+    pub fn new() -> Self {
+        ProviderBuilder::default().with_recommended_fillers()
+    }
+
+    /// Opt-out of the recommended fillers by reseting the fillers stack in the [`ProviderBuilder`].
+    pub fn disable_recommended_fillers(self) -> ProviderBuilder<Identity, Identity, Ethereum> {
```
DaniPopes, line 143: doc nit: + another line saying "this is equivalent to creating the builder with ::default()"

DaniPopes (approved):

mattsse (approved):
lgtm!

PR #1909, opened 2025-01-11
feat: add rlp to numHash

mattsse:
this exact type is also present on eth p2p if we reorder the fields


mattsse:
on second though, probably a bad idea to outsource p2p bindings to this

PR #1910, opened 2025-01-12
Define trait `TransactionEnvelope`

emhane:
Moves part of implementation for `TxEnvelope` into new trait `TransactionEnvelope`.


emhane:
blocked by https://github.com/alloy-rs/alloy/pull/1913

crates/consensus/src/transaction/envelope.rs
```diff
@@ -658,6 +658,67 @@ impl Typed2718 for TxEnvelope {
     }
 }
 
+/// Transaction envelope.
+#[auto_impl::auto_impl(&, Arc)]
+pub trait TransactionEnvelope: Typed2718 {
+    /// Returns ref to [`Signed<TxLegacy>`] if transaction is a legacy transaction.
+    fn as_legacy(&self) -> Option<&Signed<TxLegacy>>;
```
mattsse, line 665: this bakes in the signed type, which I think is fine because I think this will not be used by reth.

~~but the function name could be problematic, could we introduce `as_signed_*` instead?~~

ah I guess this isn't necessary:

https://github.com/alloy-rs/alloy/blob/f390d442b11a667f18c93cfff68ec081dab54c63/crates/consensus/src/transaction/envelope.rs#L282-L282

mattsse (changes requested):
what's the main motiviation for this?

we haven't seen a usecase where we'd need to deal with various txenvelope types, although imo this trait seems reasonable.
but this currently clashes with txenvelope's native functions which we could then remove

emhane:
relying on this trait instead of hardcoded `TxEnvelope` type, allows for rollups to define their own `TxEnvelope` type to use in block bodies. it has same benefits as generalising the hardcoded transaction type in reth to `T: alloy_consensus::Transaction`.

crates/consensus/src/transaction/envelope.rs
emhane: I guess these constant functions are likely not needed anywhere

mattsse:
this doesn't seem necessary

PR #1912, opened 2025-01-12
feat: Add Blob type that behaves like Bytes

DoTheBestToGetTheBest:
currently docs are missing but lmk if changes are needed, if not, i'll add docs @mattsse


crates/eips/src/eip4844/mod.rs
```diff
@@ -87,8 +92,169 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
 /// How many bytes are in a proof
 pub const BYTES_PER_PROOF: usize = 48;
 
-/// A Blob serialized as 0x-prefixed hex string
-pub type Blob = FixedBytes<BYTES_PER_BLOB>;
+#[derive(Clone, Debug)]
+pub struct Blob {
+    inner: Arc<Mutex<Bytes>>,
```
mattsse, line 97: this arc mutex isn't necessary because Bytes itself is already immutable.

crates/eips/src/eip4844/builder.rs
mattsse, line 1: let's ignore all of these for now and undo and only focus on the additional blob type.

crates/eips/src/eip4844/mod.rs
```diff
@@ -87,8 +92,169 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
...
+    fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+    
+        let data: Vec<u8> = u.arbitrary::<[u8; BYTES_PER_BLOB]>()?.to_vec();
+
+        Ok(Self { inner: Arc::new(Mutex::new(Bytes::from(data))) })
+    }
+}
+impl Eq for Blob {}
+
+impl PartialEq for Blob {
```
mattsse, line 136: these can be derived now

crates/eips/src/eip4844/mod.rs
```diff
@@ -87,8 +92,169 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
...
+    fn default() -> Self {
+        let default_bytes = Bytes::from(vec![0u8; BYTES_PER_BLOB]); 
+        Self { inner: Arc::new(Mutex::new(default_bytes)) }
+    }
+}
+
+impl From<[u8; 131072]> for Blob {
+    fn from(array: [u8; BYTES_PER_BLOB]) -> Self {
+        Self { inner: Arc::new(Mutex::new(Bytes::from(array.to_vec()))) }
+    }
```
mattsse, line 109: we also want some constructors for this type

crates/eips/src/eip4844/mod.rs
```diff
@@ -87,8 +92,169 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
...
+}
+impl Encodable for Blob {
+    fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
+        let lock = self.inner.lock().unwrap();
+
+        lock.encode(out);
+    }
+}
+impl Decodable for Blob {
```
mattsse, line 118: these can also be derived via the rlpencodablewrapper derives

crates/eips/src/eip4844/mod.rs
```diff
@@ -87,8 +92,169 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
 /// How many bytes are in a proof
 pub const BYTES_PER_PROOF: usize = 48;
 
-/// A Blob serialized as 0x-prefixed hex string
-pub type Blob = FixedBytes<BYTES_PER_BLOB>;
+#[derive(Clone, Debug)]
+pub struct Blob {
```
mattsse, line 96: we also need serde support.

deserialize we need to implement manually because we need an additional check that the decoded blob has the correct len

mattsse (changes requested):

crates/eips/src/eip4844/mod.rs
DoTheBestToGetTheBest: > this arc mutex isn't necessary because Bytes itself is already immutable.

so you actually you only wants ``` struct Blob( reth_primitives::Bytes)``` ?

DoTheBestToGetTheBest:
Two tests are currently failling

crates/eips/src/eip4844/builder.rs
mattsse, line 1: let's undo all of the changes here

crates/eips/src/eip4844/sidecar.rs
mattsse, line 1: same

I want to focus only on the additional type first

crates/eips/src/eip4844/mod.rs
```diff
@@ -86,9 +91,146 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
 
 /// How many bytes are in a proof
 pub const BYTES_PER_PROOF: usize = 48;
+/// A fixed-size container for binary data, designed to hold exactly 131,072 bytes.
+#[derive(Clone, Debug, RlpEncodable, RlpDecodable, Hash, PartialEq, Eq)]
```
mattsse, line 95: RlpDecodable we can't derive but must impl manually because we also need to enforce length

crates/eips/src/eip4844/mod.rs
```diff
@@ -86,9 +91,146 @@ pub const BYTES_PER_COMMITMENT: usize = 48;
...
+    }
+}
+
+impl From<[u8; BYTES_PER_BLOB]> for Blob {
+    fn from(array: [u8; BYTES_PER_BLOB]) -> Self {
+        Self { inner: Bytes::from(array.to_vec()) }
+    }
+}
+#[cfg(feature = "serde")]
+impl serde::Serialize for Blob {
```
mattsse, line 120: actually this we can derive

mattsse (changes requested):

PR #1913, opened 2025-01-13
chore(eips): Add super trait `Typed2718` to `Encodable2718`

emhane:
Adds `Typed2718` as super trait to `Encodable2718` to remove redundancy

### DIFF
--- a/crates/consensus/src/constants.rs
+++ b/crates/consensus/src/constants.rs
@@ -1,6 +1,10 @@
 //! Ethereum protocol-related constants
 use alloy_primitives::{b256, B256};
 
+pub use alloy_eips::eip2718::{
+    EIP1559_TX_TYPE_ID, EIP2930_TX_TYPE_ID, EIP4844_TX_TYPE_ID, EIP7702_TX_TYPE_ID,
+    LEGACY_TX_TYPE_ID,
+};
 pub use alloy_trie::{EMPTY_ROOT_HASH, KECCAK_EMPTY};
 
 /// The first four bytes of the call data for a function call specifies the function to be called.
--- a/crates/consensus/src/constants.rs
+++ b/crates/consensus/src/constants.rs
@@ -61,19 +61,3 @@ pub const EMPTY_TRANSACTIONS: B256 = EMPTY_ROOT_HASH;
 
 /// Withdrawals root of empty withdrawals set.
 pub const EMPTY_WITHDRAWALS: B256 = EMPTY_ROOT_HASH;
-
-/// Identifier for legacy transaction, however a legacy tx is technically not
-/// typed.
-pub const LEGACY_TX_TYPE_ID: u8 = 0;
-
-/// Identifier for an EIP2930 transaction.
-pub const EIP2930_TX_TYPE_ID: u8 = 1;
-
-/// Identifier for an EIP1559 transaction.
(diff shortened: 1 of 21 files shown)


crates/eips/src/eip2718.rs
```diff
@@ -162,16 +178,16 @@ pub trait Decodable2718: Sized {
 /// over the accepted transaction types.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
-pub trait Encodable2718: Sized + Send + Sync {
+pub trait Encodable2718: Typed2718 + Sized + Send + Sync {
```
mattsse, line 181: I think this is reasonable
because this merely unifies the type_flag behaviour

crates/eips/src/eip2718.rs
```diff
@@ -10,6 +10,22 @@ use core::fmt;
...
+pub const EIP2930_TX_TYPE_ID: u8 = 1;
+
+/// Identifier for an EIP1559 transaction.
+pub const EIP1559_TX_TYPE_ID: u8 = 2;
+
+/// Identifier for an EIP4844 transaction.
+pub const EIP4844_TX_TYPE_ID: u8 = 3;
+
+/// Identifier for an EIP7702 transaction.
+pub const EIP7702_TX_TYPE_ID: u8 = 4;
```
mattsse, line 27: moving these is reasonable

crates/consensus/src/transaction/pooled.rs
```diff
@@ -79,7 +79,7 @@ impl PooledTransaction {
     /// length of the header + the length of the type flag and inner encoding.
     fn network_len(&self) -> usize {
         let mut payload_length = self.encode_2718_len();
-        if !Encodable2718::is_legacy(self) {
+        if !self.is_legacy() {
```
mattsse, line 82: cool

crates/consensus/src/lib.rs
```diff
@@ -37,12 +37,15 @@ pub mod transaction;
...
 
-pub use alloy_eips::eip4844::{
-    builder::{SidecarBuilder, SidecarCoder, SimpleCoder},
-    utils, Blob, BlobTransactionSidecar, Bytes48,
+pub use alloy_eips::{
+    eip4844::{
+        builder::{SidecarBuilder, SidecarCoder, SimpleCoder},
+        utils, Blob, BlobTransactionSidecar, Bytes48,
+    },
+    Typed2718,
```
mattsse, line 48: yep we def need to re-export this

crates/consensus/src/constants.rs
```diff
@@ -61,19 +61,3 @@ pub const EMPTY_TRANSACTIONS: B256 = EMPTY_ROOT_HASH;
 
 /// Withdrawals root of empty withdrawals set.
 pub const EMPTY_WITHDRAWALS: B256 = EMPTY_ROOT_HASH;
-
```
mattsse, line 64: could we re-export the constants here so this becomes non-breaking?

mattsse (changes requested):
this makes a lot of sense,

smol nit re re-exports because we import these constants in a few other places externally

mattsse (approved):
lgtm,

pending @klkvr or @DaniPopes

klkvr (approved):
lgtm!

DaniPopes (approved):

PR #1917, opened 2025-01-14
chore: reuse `BlockOverrides` in `SimBundleOverrides`

klkvr:
## Motivation

Those are pretty much identical, and it is beneficial to unify so that it's possible to have something like
```rust
trait OverridableBlockEnv {
     fn apply_overrides(&mut self, overrides: BlockOverrides);
}
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense

PR #1920, opened 2025-01-15
fix(`node-bindings`): reset `child.stdout` in `AnvilInstance`

yash-atreya:
## Motivation

Currently, `AnvilInstance.child.stdout` is not being set again after initial parsing, resulting in `child.stdout` being always `None`. 

## Solution

Set `child.stdout` after initial parsing has completed in `.spawn()`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


zerosnacks (approved):
Makes sense, for reference `reth.rs` does set this correctly

PR #1921, opened 2025-01-16
refactor: change json-rpc trait names, relax bounds

prestwich:
## Motivation

Make the RPC trait system more clear and flexible, provide non-owned options requirements baked into the traits

## Solution

* Rename `RpcParam` to `RpcSend` and `RpcReturn` to `RpcReceive` to make them clearer
* Add lifetimed versions of `RpcReturn` and `RpcObject` to allow borrowing from deserializers
* relax many bounds from `RpcObject` to either `RpcBorrow` or `RpcSend`, e.g. in `ResponsePayload::serialize_payload` we can relax to `RpcSend`

## Note

The implementation of `RpcRecv` is still **more** strict than `DeserializeOwned`, as it requires `+ 'static` rather than `for<'de> Deserialize<'de>`. This means that it is not iplemented on (e.g.) things that borrow from some non-deserializer context or environment during deserialization. Some parts of our future and streaming systems rely on this, and it seems like just marginal benefit to change

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/json-rpc/src/error.rs
+++ b/crates/json-rpc/src/error.rs
@@ -1,4 +1,4 @@
-use crate::{ErrorPayload, RpcReturn};
+use crate::{ErrorPayload, RpcRecv};
 use serde_json::value::RawValue;
 
 /// An RPC error.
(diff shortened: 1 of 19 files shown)


mattsse (approved):
seems okay to me, I actually prefer send/recv over param/return

any objections @DaniPopes @klkvr ?

all of this is internal, so I expect the impact of this to be minimal

DaniPopes (approved):

PR #1923, opened 2025-01-18
test: derive serde for test

mattsse:
also derive serde for test


DaniPopes:
what is the point of this? I assume that you know that cfg(test) doesn't propagate to dependencies

mattsse:
this way we derived serde traits in tests, makes `cargo t` actually compile

DaniPopes:
dont really like doing this at definitions, if a test needs serde then the test should be cfg serde

PR #1926, opened 2025-01-18
feat(consensus): make fn tx_type() public

lgingerich:
## Motivation

The `tx_type` method for `TxEip1559` is available only within the `consensus` crate, while it is available to external crates for all other `TxEip...` structs.

## Solution

Change method to be available to external crates.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -75,7 +75,7 @@ pub struct TxEip1559 {
 impl TxEip1559 {
     /// Get the transaction type
     #[doc(alias = "transaction_type")]
-    pub(crate) const fn tx_type() -> TxType {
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/eip1559.rs
```diff
@@ -75,7 +75,7 @@ pub struct TxEip1559 {
 impl TxEip1559 {
     /// Get the transaction type
     #[doc(alias = "transaction_type")]
-    pub(crate) const fn tx_type() -> TxType {
+    pub const fn tx_type() -> TxType {
         TxType::Eip1559
     }
```
mattsse, line 80: this makes sense since we also do it for the other tx

mattsse (approved):

PR #1935, opened 2025-01-22
wip: feat: delegator macro

mattsse:
adds a proc macro that generates delegate macros for a given type

https://github.com/paradigmxyz/reth/blob/6d2648dc1b786c1177c947395e4cd2d03d3adad9/crates/ethereum/primitives/src/transaction.rs#L30-L40

this is a common pattern for transaction types, for example:

https://github.com/alloy-rs/alloy/blob/9aadefa5c47c44d775a57a42e39268fd5d5306dc/crates/consensus/src/transaction/pooled.rs#L326-L335

we can simplify this boilerplate with 
1. a helper that generates the delegator macro
2. add builtin support for deriving alloy traits by delegating

1. could even be a derive I believe

still experimenting with this


mattsse:
actually not worth it, only takes a few lines to impl this, and derive macro for this isn't worth it

PR #1937, opened 2025-01-22
feat(`provider`): `eth_callMany`

### LINKED ISSUE #1274: [Feature] Integrate `eth_callMany` support into `EthCall` builder
### Component

provider, pubsub, rpc

Rather than adding `eth_callMany` as an RPC method it should be integrated into the `EthCall` builder.

A previous PR proposed adding this to the `debug` namespace which is incorrect: https://github.com/alloy-rs/alloy/pull/1085

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #1274  - Provider lacks implementation for `eth_callMany` using `EthCall` builder.

## Solution

Intergrate it into `EthCall` builder and `EthCallParams`. 

1. enum based params.
```rust
pub enum EthCallParams<'req, N: Network> {
    /// Parameters used for `"eth_call"` and `"eth_estimateGas"` RPC requests.
    Call(CallParams<'req, N>),
    /// Parameters used for `"eth_callMany"` RPC requests.
    CallMany(CallManyParams<'req>),
}
```

2. Propagate builder methods such as `context(c: StateContext)` for `eth_callMany` to `EthCall`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (commented):
should we really squeeze this into the existing `EthCallParams`

now a user can call eth_call with call_many params?

I don't think there is a usecase where we either want ethcall or callmany, so combining those two variants doesn't feel quite right.

I assume the overhead for rolling a standalone ethcallmany wouldn't be too bad

yash-atreya:
Closing in favour of #1944

PR #1945, opened 2025-01-24
fix(`contract`): rm IntoFuture for CallBuilder

yash-atreya:
## Motivation

Contract calls like `contract.balanceOf` cannot be directly awaited despite the `IntoFuture` impl as it requires `Self: 'static`. Ref: https://github.com/rust-lang/rust/issues/63063.
 
Awaiting calls results in an ambiguous error for users. See https://t.me/ethers_rs/43016. 

```rust
let bal = contract.balanceOf(addr).await.unwrap()
// Error: argument requires that borrow lasts for `'static`
```

The appropriate way to call a contract is:

```rust
let bal = contract.balanceOf(addr).call().await.unwrap();
```
The above error doesn't suggest this. 

## Solution

Remove the `IntoFuture` impl for `CallBuilder`. 

Users now *have* to use `.call()`. Moreover, using `.await` results in a slightly better error: `CallBuilder` is not a future.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -7,11 +7,7 @@ use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
 use alloy_provider::{PendingTransactionBuilder, Provider};
 use alloy_rpc_types_eth::{state::StateOverride, AccessList, BlobTransactionSidecar, BlockId};
 use alloy_sol_types::SolCall;
-use std::{
-    future::{Future, IntoFuture},
-    marker::PhantomData,
-    pin::Pin,
-};
+use std::{self, marker::PhantomData};
 
 // NOTE: The `T` generic here is kept to mitigate breakage with the `sol!` macro.
 // It should always be `()` and has no effect on the implementation.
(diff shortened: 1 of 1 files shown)


mattsse:
what does this type borrow, do we have an example here?

DaniPopes:
call takes &self

DaniPopes (changes requested):
let me have one last look to see if i can make this work

mattsse:
ngl the lifetimes of the ethcall type are a bit confusing

DaniPopes:
It could be fixable with implementing for &CallBuilder but then .await does not auto borrow so you would have to do (&callbuilder).await

DaniPopes (approved):
lgtm, let's open an issue to track re-adding this impl

mattsse (approved):

smanilov:
Note: this has outdated the documentation for CallBuilder. It still claims

> The builder can be .awaited directly, which is equivalent to invoking call. Prefer using call when possible, as awaiting the builder directly will consume it, and currently also boxes the future due to type system limitations.

Issue #1939, opened 2025-01-22
[Feature] Simplify conversion from `TransactionReceipt` for RLP-encoding


Wollac:
### Component

network, json-rpc

### Describe the feature you would like

It would be beneficial to have a more straightforward way to RLP-encode a `TransactionReceipt` or to easily convert it into a `ReceiptEnvelope`, which already implements the necessary encoding traits. Currently, the only process I could find requires manually converting internal structures like `alloy_rpc_types::Log` to `alloy_primitives::Log`, which is very cumbersome and error-prone.

Here's an example of the current workaround:

```rust
match tx_receipt.inner {
  ReceiptEnvelope::Legacy(t) => ReceiptEnvelope::Legacy(
        ReceiptWithBloom::new(
            Receipt {
                status: t.receipt.status,
                cumulative_gas_used: t.cumulative_gas_used(),
                logs: t.receipt.logs.into_iter().map(|log| log.inner).collect(),
            },
            t.logs_bloom,
        )),
  // ... (Handle other ReceiptEnvelope variants)
})
```

While I understand that issues #623 and #854 aim to provide a more generic solution, a simple utility method or trait implementation in the interim would significantly improve developer experience and potentially save time for those facing similar use cases.

**Context:**

A primary motivation for this request is the need to validate receipts returned for a block. While calculating the transaction root is straightforward using the existing features (as shown below), verifying the receipt root for a generic `Provider` connected to a `Network` does not currently appear to be possible at all, and even for `Ethereum` it requires the manual conversion outlined above.

```rust
let block = provider
    .get_block_by_number(BlockNumberOrTag::Latest, BlockTransactionsKind::Full)
    .await?
    .unwrap();

let transactions = block.transactions().as_transactions().unwrap();
let transactions_root = alloy_trie::root::ordered_trie_root_with_encoder(transactions, |tx, buf| {
        Encodable2718::encode_2718(tx.as_ref(), buf)
    });
assert_eq!(block.header().transactions_root(), transactions_root);
```

Ideally, a similar approach should be possible for validating receipt roots. A utility or trait to simplify the conversion from `TransactionReceipt` to a readily RLP-encodable format would enable this functionality until the aforementioned issues are resolved.

mattsse:
I see, yeah the rpc embeddings still have some rough edges,

what we can do here is solving this via a few helper functions

[...]

PR #1949, opened 2025-01-25
chore: add receipt conversion fns

mattsse:
towards #1939

adds helper conversion fn for the various receipt types to map logs and convert to primitives logs.

adds a from impl for the rpc txreceipt to convert the wrapped receipt envelope into an envelope with primitives logs

cc @Wollac


klkvr (approved):

yash-atreya (approved):
lgtm!

PR #1953, opened 2025-01-27
feat: add helpers for tx conditional

mattsse:
adds missing helpers to check if any max parameter are exceeded

### DIFF
--- a/crates/rpc-types-eth/src/erc4337.rs
+++ b/crates/rpc-types-eth/src/erc4337.rs
@@ -87,7 +87,7 @@ impl TransactionConditional {
     }
 
     /// Returns true if the configured max timestamp is lower or equal to the given `timestamp`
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/erc4337.rs
```diff
@@ -71,6 +71,27 @@ pub struct TransactionConditional {
...
+
+    /// Returns true if the configured max block number is lower or equal to the given
+    /// `block_number`
+    pub const fn has_exceeded_block_number(&self, block_number: BlockNumber) -> bool {
+        let Some(max_num) = self.block_number_max else { return false };
+        block_number >= max_num
+    }
+
+    /// Returns true if the configured max timestamp is lower or equal to the given `timestamp`
+    pub const fn has_exceeded_timestamp(&self, timestamp: BlockNumber) -> bool {
```
DaniPopes, line 90: not a BlockNumber

klkvr (approved):

PR #1955, opened 2025-01-27
feat: couple convenience methods

prestwich:
Adds a couple convenience methods to check if hard forks are active in a given header

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -334,6 +334,18 @@ impl Header {
     pub const fn seal(self, hash: B256) -> Sealed<Self> {
         Sealed::new_unchecked(self, hash)
     }
+
+    /// True if the Cancun hardfork is active.
+    pub const fn cancun_active(&self) -> bool {
+        self.blob_gas_used.is_some()
+    }
+
+    /// True if the Prague hardfork is active.
+    ///
+    /// This function checks that the withdrawals root is present.
+    pub const fn prague_active(&self) -> bool {
+        self.withdrawals_root.is_some()
+    }
 }
 
 impl Encodable for Header {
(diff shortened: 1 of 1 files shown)


crates/consensus/src/block/header.rs
```diff
@@ -334,6 +334,18 @@ impl Header {
...
+
+    /// True if the Cancun hardfork is active.
+    pub const fn cancun_active(&self) -> bool {
+        self.blob_gas_used.is_some()
+    }
+
+    /// True if the Prague hardfork is active.
+    ///
+    /// This function checks that the withdrawals root is present.
+    pub const fn prague_active(&self) -> bool {
```
mattsse, line 346: since this is the ethereum header type, I think this makes sense because these fields were introduced with the corresponding hardforks. although this assumes that the header is valid, but I think we can do this.
the alternative would be `has_withdrawals_root` and then the caller is expected to know that this means hardfork active

crates/consensus/src/block/header.rs
```diff
@@ -334,6 +334,18 @@ impl Header {
...
     }
+
+    /// True if the Cancun hardfork is active.
+    pub const fn cancun_active(&self) -> bool {
+        self.blob_gas_used.is_some()
+    }
+
+    /// True if the Prague hardfork is active.
+    ///
+    /// This function checks that the withdrawals root is present.
```
mattsse, line 345: this should be shanghai, right?
and for prague we can check requests_hash

mattsse (changes requested):

crates/consensus/src/block/header.rs
prestwich: yeah, fixed :)

crates/consensus/src/block/header.rs
prestwich: these will also need to be updated if any field is removed in future hard forks, but that seems unlikely (and also fine)

mattsse (approved):

PR #1957, opened 2025-01-28
chore(deps): breaking bumps

DaniPopes:
Also remove schnellru as we already have lru

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ all = "warn"
 
 [workspace.lints.clippy]
 all = { level = "warn", priority = -1 }
-missing-const-for-fn = "warn"
+missing-const-for-fn = "allow" # TODO: https://github.com/rust-lang/rust-clippy/issues/14020
 use-self = "warn"
 option-if-let-else = "warn"
 redundant-clone = "warn"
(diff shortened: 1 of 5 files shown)


crates/provider/Cargo.toml
```diff
@@ -56,7 +56,6 @@ auto_impl.workspace = true
 dashmap = "6.0"
 futures-utils-wasm.workspace = true
 futures.workspace = true
-schnellru.workspace = true
```
mattsse, line 59: makes sense, no reason to import two lru impls

Cargo.toml
```diff
@@ -25,7 +25,7 @@ all = "warn"
 
 [workspace.lints.clippy]
 all = { level = "warn", priority = -1 }
-missing-const-for-fn = "warn"
+missing-const-for-fn = "allow" # TODO: https://github.com/rust-lang/rust-clippy/issues/14020
```
mattsse, line 28: this has been driving me insane for the last two days

mattsse (approved):

Cargo.toml
DaniPopes: unfortunate that it's very unstable

PR #1959, opened 2025-01-28
feat(eth_call): fix serializion to include default block to latest

0xfourzerofour:
## Motivation

In an `eth_call` the `block` parameter is always required. In the current serialization logic, it does not add a block if the state_overrides is None and the block is None. 

This results in a node returning an invalid parameter error.

## Solution

I have updated the code to always set the block as the second argument, if it is None it will use the default `latest` value. The third argument will only be added if `state_overrides` is Some.

## PR Checklist

- [y] Added Tests
- [y] Added Documentation
- [n] Breaking changes

### DIFF
--- a/crates/provider/src/provider/eth_call/mod.rs
+++ b/crates/provider/src/provider/eth_call/mod.rs
@@ -302,7 +302,7 @@ mod test {
         assert_eq!(params.overrides(), None);
         assert_eq!(
             serde_json::to_string(&params).unwrap(),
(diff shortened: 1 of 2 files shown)


DaniPopes:
the block parameter is not required as per the RPC spec

mattsse (commented):
this param is optional,
I think we don't need to set it and leave this up to the node to pick the default

0xfourzerofour:
damn, no worries I shall fix on my client side

PR #1962, opened 2025-01-29
feat: introduce maybe helpers for blob calc

mattsse:
now that we have configurable blobparams that are hardfork specific, this means we need to deal with a case where blobs are not active: `None`

https://github.com/paradigmxyz/reth/pull/14049

this adds helpers with default impl for this so that we don't need to use `maybe_params.map(|params| header.next_blob_fee(params))`


klkvr (approved):
lgtm

PR #1968, opened 2025-01-30
fix: map txcount resp

### LINKED ISSUE #1940: [Bug] CacheProvider does not deserialize `get_transaction_count` to `u64`
### Component

provider, pubsub

0.9.2

### Operating System

Linux

### Describe the bug

Node provider is Reth. Here is the code that encountred this bug:

```rust
use alloy_network::Ethereum;
use alloy_provider::{
    layers::{CacheLayer, CacheProvider},
    RootProvider,
};
use alloy_pubsub::PubSubFrontend;

pub use alloy_provider::Provider;

pub type NodeProvider = RootProvider<PubSubFrontend, Ethereum>;
pub type CacheNodeProvider = CacheProvider<NodeProvider, PubSubFrontend, Ethereum>;

pub fn create_cache_provider(
    provider: NodeProvider,
    cache_provider_limit: u32,
) -> CacheNodeProvider {
    let cache_layer = CacheLayer::new(cache_provider_limit);

    CacheNodeProvider::new(provider, cache_layer.cache())
}

#[cfg(test)]
mod test {
    use alloy_primitives::address;
    use alloy_provider::{Provider, ProviderBuilder, WsConnect};
    use anyhow::Result;

    use crate::node_provider::create_cache_provider;

    #[tokio::test]
    async fn should_get_nonce() -> Result<()> {
        let ws = todo!("add node ws");

        let provider = ProviderBuilder::new().on_ws(WsConnect::new(ws)).await?;

        let cache_provider = create_cache_provider(provider, 2000);

        let weth_address = address!("c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2");

        cache_provider.get_transaction_count(weth_address).await?;
        // Error: deserialization error: invalid type: string "0x1", expected u64 at line 1 column 5

        Ok(())
    }
}
```

mattsse:
closes #1940

inlines macro  and adds the required map

### DIFF
--- a/crates/provider/src/layers/cache.rs
+++ b/crates/provider/src/layers/cache.rs
@@ -1,4 +1,6 @@
-use crate::{ParamsWithBlock, Provider, ProviderCall, ProviderLayer, RootProvider, RpcWithBlock};
+use crate::{
+    utils, ParamsWithBlock, Provider, ProviderCall, ProviderLayer, RootProvider, RpcWithBlock,
+};
 use alloy_eips::BlockId;
 use alloy_json_rpc::{RpcError, RpcObject, RpcSend};
 use alloy_network::Network;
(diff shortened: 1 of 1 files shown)


chain-lover:
I see that there's no caching for get_transaction_count. Is it on purpose or only for fixing the bug?

https://github.com/alloy-rs/alloy/blob/main/crates/provider/src/layers/cache.rs#L339-L369

mattsse:
oh you're right, 
we don't want to cache txcount because this can change at any point, so we probably should just delete this override

PR #1973, opened 2025-01-30
chore!: remove Service impls for &T

DaniPopes:
Remove an unused and unusable layer of indirection, as we require `Transport: 'static`.


mattsse (approved):
yeah these are also clone, so impl for & not useful

PR #1977, opened 2025-01-30
fix: store pubsubfrontend clone in rpcinner

### LINKED ISSUE #1972: [Bug] `subscribe_blocks` could not downcast to pubsub when creating provider using `RetryBackoffLayer`
### Component

provider, pubsub

0.9.2

### Operating System

Linux

### Describe the bug

I'm trying to build provider using `RetryBackoffLayer` as it was in the example below but adding `IpcConnect`:
https://github.com/alloy-rs/examples/blob/9db483f94af1fb23141baa8e574f3c2ae12b90e6/examples/layers/examples/retry_layer.rs#L29

It doesn't downcast to `pubsub` when running `subscribe_blocks` method.

```rust
struct ProviderConfig {
    ipc: String,
    max_rate_limit_retries: u32,
    initial_backoff: u64,
    compute_units_per_second: u64,
}

#[cfg(test)]
mod test {
    use alloy_provider::{IpcConnect, Provider, ProviderBuilder};
    use alloy_rpc_client::RpcClient;
    use alloy_transport::layers::RetryBackoffLayer;
    use anyhow::Result;

    use crate::aggregate::ProviderConfig;

    #[tokio::test]
    async fn should_stream_block() -> Result<()> {
        let provider_config = ProviderConfig {
            ipc: "/tmp/reth.ipc".to_string(),
            max_rate_limit_retries: 1,
            initial_backoff: 50,
            compute_units_per_second: 1600,
        };

        let ipc_connect = IpcConnect::new(provider_config.ipc);
        let retry_layer = RetryBackoffLayer::new(
            provider_config.max_rate_limit_retries,
            provider_config.initial_backoff,
            provider_config.compute_units_per_second,
        );

        let rpc_client = RpcClient::builder()
            .layer(retry_layer)
            .ipc(ipc_connect)
            .await?;

            .disable_recommended_fillers()
            .on_client(rpc_client);

       let block_stream = provider.subscribe_blocks().await?;
        // Error: subscriptions are not available on this provider

        Ok(())
    }
}
```

mattsse:
closes #1972

this is a bit cursed.
It described this in the docs but due to how:

https://github.com/alloy-rs/alloy/blob/bb77d2f8a2f587c9b220b8af2cd8cbcdd749c97d/crates/rpc-client/src/builder.rs#L44-L44

creates the transport service object if there's an additional layer, this downcast does not work:

https://github.com/alloy-rs/alloy/blob/bb77d2f8a2f587c9b220b8af2cd8cbcdd749c97d/crates/rpc-client/src/client.rs#L211-L212

the solution to this is a workaround that captures the Pubsubfrontend before we collapse the service builder.

this can still be bypassed manually via

https://github.com/alloy-rs/alloy/blob/bb77d2f8a2f587c9b220b8af2cd8cbcdd749c97d/crates/rpc-client/src/client.rs#L163-L163

but all the builtin transport fn should now capture the pubsubfrontend.
pubsubfrontend is just a channel so capturing an additional clone seems fine

### DIFF
--- a/crates/provider/tests/it/ws.rs
+++ b/crates/provider/tests/it/ws.rs
@@ -1,9 +1,11 @@
-use alloy_provider::{IpcConnect, Provider, ProviderBuilder};
+use alloy_node_bindings::Anvil;
+use alloy_provider::{Provider, ProviderBuilder};
 use alloy_rpc_client::RpcClient;
 use alloy_transport::layers::RetryBackoffLayer;
+use alloy_transport_ws::WsConnect;
 
 struct ProviderConfig {
-    ipc: String,
+    endpoint: String,
     max_rate_limit_retries: u32,
     initial_backoff: u64,
     compute_units_per_second: u64,
@@ -11,22 +13,23 @@ struct ProviderConfig {
 
 // <https://github.com/alloy-rs/alloy/issues/1972>
 #[tokio::test]
-async fn ipc_retry_pubsub() -> Result<(), Box<dyn std::error::Error>> {
+async fn ws_retry_pubsub() -> Result<(), Box<dyn std::error::Error>> {
+    let anvil = Anvil::new().spawn();
     let provider_config = ProviderConfig {
-        ipc: "/tmp/reth.ipc".to_string(),
+        endpoint: anvil.ws_endpoint(),
         max_rate_limit_retries: 1,
         initial_backoff: 50,
         compute_units_per_second: 1600,
     };
 
-    let ipc_connect = IpcConnect::new(provider_config.ipc);
+    let ws = WsConnect::new(provider_config.endpoint);
     let retry_layer = RetryBackoffLayer::new(
         provider_config.max_rate_limit_retries,
         provider_config.initial_backoff,
         provider_config.compute_units_per_second,
     );
 
-    let rpc_client = RpcClient::builder().layer(retry_layer).ipc(ipc_connect).await?;
+    let rpc_client = RpcClient::builder().layer(retry_layer).ws(ws).await?;
 
     let provider = ProviderBuilder::new().disable_recommended_fillers().on_client(rpc_client);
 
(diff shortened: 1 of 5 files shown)


crates/provider/tests/it/ws.rs
```diff
@@ -0,0 +1,39 @@
...
+    let ws = WsConnect::new(provider_config.endpoint);
+    let retry_layer = RetryBackoffLayer::new(
+        provider_config.max_rate_limit_retries,
+        provider_config.initial_backoff,
+        provider_config.compute_units_per_second,
+    );
+
+    let rpc_client = RpcClient::builder().layer(retry_layer).ws(ws).await?;
+
+    let provider = ProviderBuilder::new().disable_recommended_fillers().on_client(rpc_client);
```
gakonst, line 34: nit: no need for this

```suggestion
    let provider = ProviderBuilder::new().on_client(rpc_client);
```

also i wish we had the ability to configure layers on the `ProviderBuilder` w/o having to drop to `RpcClient`, so you'd just do

```
Providerbuilder::new().layer(retry_layer).on_builtin(url).await?
```

gakonst (approved):

DaniPopes (commented):
lgtm, the only thing would be  that it creates a copy of the internal state instead of sharing it, so we might want to arc pubsubfrontend internally?

mattsse:
> hat it creates a copy of the internal state instead of sharing

can you elaborate on this? unclear what state you mean

DaniPopes:
the Pubsubfrontend struct itself is now both inside of transport and in the new field

mattsse:
this is fine because this is just another copy of the sender half

https://github.com/alloy-rs/alloy/blob/7fb7a51bbf45828f2513165d71fa5a144222c599/crates/pubsub/src/frontend.rs#L18-L22

which also gets dropped when the transport drops

PR #1985, opened 2025-01-31
fix: opt-in to keep stdout

### LINKED ISSUE #1983: [Bug] Test hangs when mining many blocks
### Component

node-bindings

latest

### Operating System

macOS (Apple Silicon)

### Describe the bug

I found a regression when updating alloy from `0.9.2` to `0.11.0`.

wrote a small test that reproduces the issue:

```rust

    #[tokio::test]
    async fn test_mine_many_blocks() {
        // 328 passes, 329 fails
        let blocks_to_mine = 329;

        let args =
            vec!["--base-fee", "0", "--gas-limit", "30000000", "--timestamp", "999", "--no-mining"];
        let anvil = alloy_node_bindings::Anvil::new().args(args).spawn();
        let provider = ProviderBuilder::new().on_http(anvil.endpoint_url());

        let head =
            provider.get_block(alloy_eips::BlockId::latest(), false.into()).await.unwrap().unwrap();
        assert_eq!(head.header.number, 0);
        let ts = head.header.timestamp;

        // Mine N blocks using anvil_mine_detailed
        for i in 0..blocks_to_mine {
            let mine_options =
                MineOptions::Options { timestamp: Some(ts + i + 1), blocks: Some(1) };
            provider.anvil_mine_detailed(Some(mine_options)).await.unwrap();
        }

        // Verify block number is at least 1000
        let final_block = provider.get_block_number().await.unwrap();
        assert!(
            final_block >= blocks_to_mine,
            "Expected at least {} blocks, got {}",
            blocks_to_mine,
            final_block
        );
    }
```

For some reason, on the latest version, that test will hang forever if I try to mine 329+ blocks. With less blocks, it passes.

I bissected this behavior and found the following:

```shell
70df381dd5ae09133137bd8c51b5c7f3530445c9 is the first bad commit
commit 70df381dd5ae09133137bd8c51b5c7f3530445c9 (HEAD)
Author: Yash Atreya <44857776+yash-atreya@users.noreply.github.com>
Date:   Thu Jan 16 00:00:21 2025 +0530

    fix(`node-bindings`): reset `child.stdout` in `AnvilInstance` (#1920)
    
    * feat(`node-bindings`): expose `stdout` reader
    
    * remove reader and reset child.stdout
    
    * Update crates/node-bindings/src/nodes/anvil.rs
    
    ---------
    
    Co-authored-by: DaniPopes <57450786+DaniPopes@users.noreply.github.com>

 crates/node-bindings/src/nodes/anvil.rs |  4 +++-
 crates/provider/src/provider/trait.rs   | 28 +++++++++++++++++++++++++++-
 2 files changed, 30 insertions(+), 2 deletions(-)
```

removing [this line](https://github.com/alloy-rs/alloy/blob/70df381dd5ae09133137bd8c51b5c7f3530445c9/crates/node-bindings/src/nodes/anvil.rs#L337) seems to fix the issue.

Please assist 🙏

mattsse:
fixes an issue introduced in #1920

closes #1983

what happens here is that, if stdout is not cleared the kernel buffer ends up blocking the program.

this makes keeping the stdout handle opt in

### DIFF
--- a/crates/node-bindings/src/nodes/anvil.rs
+++ b/crates/node-bindings/src/nodes/anvil.rs
@@ -264,7 +264,7 @@ impl Anvil {
 
     /// Keep the handle to anvil's stdout in order to read from it.
     ///
-    /// Caution: if the stdout handle is unused, this can end up blocking.
+    /// Caution: if the stdout handle isn't used, this can end up blocking.
     pub const fn keep_stdout(mut self) -> Self {
         self.keep_stdout = true;
         self
--- a/crates/node-bindings/src/nodes/anvil.rs
+++ b/crates/node-bindings/src/nodes/anvil.rs
@@ -136,6 +136,7 @@ pub struct Anvil {
     fork_block_number: Option<u64>,
     args: Vec<OsString>,
     timeout: Option<u64>,
+    keep_stdout: bool,
 }
 
 impl Anvil {
@@ -261,6 +262,14 @@ impl Anvil {
         self
     }
 
+    /// Keep the handle to anvil's stdout in order to read from it.
+    ///
+    /// Caution: if the stdout handle is unused, this can end up blocking.
+    pub const fn keep_stdout(mut self) -> Self {
+        self.keep_stdout = true;
+        self
+    }
+
     /// Consumes the builder and spawns `anvil`.
     ///
     /// # Panics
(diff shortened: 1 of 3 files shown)


jorgemmsilva:
noticed the same thing happens here:
https://github.com/alloy-rs/alloy/blob/main/crates/node-bindings/src/nodes/reth.rs#L513
not sure if relevant, just FYI

PR #1989, opened 2025-02-01
feat: add additional payloadbody conversion fn

mattsse:
this type doesn't need the entire block, only txs

this way we get a constructor decoupled from the block type which is more helpful for reth:

https://github.com/paradigmxyz/reth/blob/209b44829ee80a2e5a7b3e673c02b4fa1d81233e/crates/rpc/rpc-types-compat/src/engine/payload.rs#L9-L10

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -1080,18 +1080,29 @@ pub struct ExecutionPayloadBodyV1 {
 }
 
 impl ExecutionPayloadBodyV1 {
-    /// Converts a [`alloy_consensus::Block`] into an execution payload body.
-    pub fn from_block<T: Encodable2718, H>(block: Block<T, H>) -> Self {
+    /// Creates an [`ExecutionPayloadBodyV1`] from the given withdrawals and transactions
+    pub fn new<'a, T>(
+        withdrawals: Option<Withdrawals>,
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/payload.rs
```diff
@@ -1080,18 +1080,29 @@ pub struct ExecutionPayloadBodyV1 {
...
         Self {
-            transactions: block.body.transactions().map(|tx| tx.encoded_2718().into()).collect(),
-            withdrawals: block.body.withdrawals.map(Withdrawals::into_inner),
+            transactions: transactions.into_iter().map(|tx| tx.encoded_2718().into()).collect(),
+            withdrawals: withdrawals.map(Withdrawals::into_inner),
         }
     }
+
+    /// Converts a [`alloy_consensus::Block`] into an execution payload body.
+    pub fn from_block<T: Encodable2718, H>(block: &Block<T, H>) -> Self {
```
DaniPopes, line 1098: this is breaking

mattsse: yep, these were just introduced, dont believe anyone's using that besides us

mattsse: removed the breaking change, the recomended way would be the new fn anyway

Issue #296, opened 2024-03-13
[Bug] Provider subscriptions are only available for `T = PubSubFrontend`


DaniPopes:
### Component

provider, pubsub

### Operating System

None

### Describe the bug

Due to a current limitation in the pubsub provider implementation, `subscribe*` and other `pubsub`-feature-gated methods in `alloy_provider::Provider` are available only when the underlying transport `T` is `PubSubFrontend`.

mattsse:
I think the workaround in https://github.com/alloy-rs/alloy/pull/1977 "fixed" this

PR #1991, opened 2025-02-02
fix(docs): update outdated Provider doc comment

merklefruit:
## Motivation

While updating some stuff to 0.11.0 I noticed an outdated doc comment.

## Solution

Removed the reference to a non existing generic.

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -34,18 +34,10 @@ use super::EthCallMany;
 /// See [`PollerBuilder`] for more details.
 pub type FilterPollerBuilder<R> = PollerBuilder<(U256,), Vec<R>>;
 
-// todo: adjust docs
-// todo: reorder
-/// Provider is parameterized with a network and a transport. The default
-/// transport is type-erased, but you can do `Provider<Http, N>`.
+/// Provider is parameterized with a network. The default
+/// transport is type-erased, but you can do `Provider<N>`.
 ///
 /// # Subscriptions
-///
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -34,19 +34,11 @@ use super::EthCallMany;
...
+/// transport is type-erased, but you can do `Provider<N>`.
 ///
 /// # Subscriptions
 ///
-/// **IMPORTANT:** this is currently only available when `T` is
-/// `PubSubFrontend` or `BoxedClient` over `PubSubFrontend` due to an internal
-/// limitation. This means that layering transports will always disable
-/// subscription support. See
-/// [issue #296](https://github.com/alloy-rs/alloy/issues/296).
-///
```
merklefruit, line 49: I also removed these lines as #296 was closed recently.

crates/provider/src/provider/trait.rs
```diff
@@ -34,19 +34,11 @@ use super::EthCallMany;
 /// See [`PollerBuilder`] for more details.
 pub type FilterPollerBuilder<R> = PollerBuilder<(U256,), Vec<R>>;
 
-// todo: adjust docs
-// todo: reorder
-/// Provider is parameterized with a network and a transport. The default
-/// transport is type-erased, but you can do `Provider<Http, N>`.
+/// Provider is parameterized with a network. The default
+/// transport is type-erased, but you can do `Provider<N>`.
```
DaniPopes, line 38: we can change this to "Ethereum JSON-RPC interface." and remove the second sentence

PR #1994, opened 2025-02-03
feat(`node-bindings`): expose anvil wallet

yash-atreya:
## Motivation

Currently, `AnvilInstance` doesn't expose `EthereumWallet` of its dev private keys.
One can create an `EthereumWallet` for the instance via `.keys()`, like:

```rust
let pk: PrivateKeySigner = anvil.keys()[0].into();
let wallet = EthereumWallet::new(pk);
```

This can be made simpler 

```rust
// Returns a wallet consisting of signers associated with all anvil private keys
let wallet = anvil.wallet().ok_or_eyre("No Keys Available")?;
```

## Solution

Instantiate and store the `EthereumWallet` when anvil is spawned. 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/Cargo.toml
+++ b/crates/node-bindings/Cargo.toml
@@ -25,6 +25,9 @@ workspace = true
 [dependencies]
 alloy-primitives = { workspace = true, features = ["std", "k256", "serde"] }
 alloy-genesis.workspace = true
+alloy-network.workspace = true
+alloy-signer-local.workspace = true
+alloy-signer.workspace = true
 k256.workspace = true
 rand.workspace = true
 serde_json = { workspace = true, features = ["std"] }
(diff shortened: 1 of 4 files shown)


crates/node-bindings/Cargo.toml
```diff
@@ -25,6 +25,9 @@ workspace = true
 [dependencies]
 alloy-primitives = { workspace = true, features = ["std", "k256", "serde"] }
 alloy-genesis.workspace = true
+alloy-network.workspace = true
+alloy-signer-local.workspace = true
+alloy-signer.workspace = true
```
mattsse, line 30: this seems okay

mattsse (approved):

PR #1995, opened 2025-02-03
chore(provider): default to `Ethereum` network in `FillProvider`

merklefruit:
## Motivation

Although `FillProvider` isn't meant to be built directly, it's often used as return type of `ProviderBuilder::on_http`.
Since `Provider` uses this default, I think it makes sense to do the same for `FillProvider`.

## Solution

Set the default `N = Ethereum` with `N: Network`.


yash-atreya (commented):
lgtm!

mattsse (approved):
lgtm

PR #2002, opened 2025-02-03
feat(transport): made avg_cost to be configurable in retrybackoff

### LINKED ISSUE #1999: [Feature] Make average cost in retrybackoff configurable
### Component

rpc

this should be configurable:

https://github.com/alloy-rs/alloy/blob/0fbc37da5178681e51e93d5274681f85c874702a/crates/transport/src/layers/retry.rs#L192-L202

## TODO
* convert this to a field, but keep constructors as is and use the current value as the default
* add additional setter

### Additional context

_No response_

Dhruv-2003:
## Motivation

closes #1999 

## Solution

New `avg_cost` field added on the `RetryBackoffLayer` & `RetryBackoffService`. As issue suggests, `DEFAULT_AVG_COST` const was added to default to `17`
A setter `with_gas_cost` was added as well on `RetryBackoffLayer` to set the avg_cost if required

It made sense to me to create a const here for better maintainability and configure it on the layer level, lmk if another approach is to be followed here

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport/src/layers/retry.rs
+++ b/crates/transport/src/layers/retry.rs
@@ -20,6 +20,19 @@ use wasmtimer::tokio::sleep;
 #[cfg(not(target_arch = "wasm32"))]
 use tokio::time::sleep;
 
+/// The default average cost of a request in compute units (CU). The cost of
+/// requests are usually weighted and can vary from 10 CU to several 100 CU,
+/// cheaper requests are more common some example alchemy
+/// weights:
+/// - `eth_getStorageAt`: 17
+/// - `eth_getBlockByNumber`: 16
+/// - `eth_newFilter`: 20
+///
+/// (coming from forking mode) assuming here that storage request will be the
+/// driver for Rate limits we choose `17` as the average cost
+/// of any request
+const DEFAULT_AVG_COST: u64 = 17u64;
+
 /// A Transport Layer that is responsible for retrying requests based on the
 /// error type. See [`TransportError`].
 ///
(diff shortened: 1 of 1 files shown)


crates/transport/src/layers/retry.rs
```diff
@@ -47,9 +62,15 @@ impl RetryBackoffLayer {
             max_rate_limit_retries,
             initial_backoff,
             compute_units_per_second,
+            avg_cost: DEFAULT_AVG_COST,
             policy: RateLimitRetryPolicy,
         }
     }
+
+    /// Set the average cost of a request
+    pub fn with_avg_cost(mut self, avg_cost: u64) {
```
mattsse, line 71:
```suggestion
    pub fn with_avg_unit_cost(mut self, avg_cost: u64) {
```

crates/transport/src/layers/retry.rs
```diff
@@ -47,9 +62,15 @@ impl RetryBackoffLayer {
             max_rate_limit_retries,
             initial_backoff,
             compute_units_per_second,
+            avg_cost: DEFAULT_AVG_COST,
             policy: RateLimitRetryPolicy,
         }
     }
+
+    /// Set the average cost of a request
```
mattsse, line 70: This should mention, that by default this uses 17

we can move the docs from the const above to here

mattsse (changes requested):
nice, ty.

smol pedantic nit

mattsse (approved):

PR #2008, opened 2025-02-05
feat(provider): DynProvider added as a helper on provider

### LINKED ISSUE #1982: [Feature] Add helper Provider for dyn Provider
### Component

rpc

by installing fillers you get a `FillProvider` with a bunch of generics, this type is inconvenient to work with, especially because one usually doesn't care about any of this.

an easy workaround for this would be a `reqwest::Client` style wrapper:

```rust
struct P<N>(dyn Provider<N> + Unpin);
```

## TODO
* add wrapper struct
* implement provider for it
* add a function to Provider trait that converts that type into that wrapper

### Additional context

_No response_

Dhruv-2003:
## Motivation

closes #1982 

## Solution

`WrappedProvider` is added as a wrapper over `Arc dyn Provider` in helper.rs. Implemented the Provider trait for it and then added a method `wrapped` on Provider to convert into a wrapped one.

Had to use lifetime as it was tricky to create the wrapped instance on the provider trait. There is a warning on `WrappedProvider` which I couldn't resolve, it's around visibility, let me know how it should be

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/provider/src/provider/helper.rs
```diff
@@ -0,0 +1,30 @@
+use crate::{Provider, RootProvider};
+
+use alloy_network::Network;
+use std::sync::Arc;
+
+/// A wrapper struct around Arc dyn provider
+pub struct WrappedProvider<'a, N>(Arc<dyn Provider<N> + 'a>);
```
mattsse, line 7: we want this to be 'static here

crates/provider/src/provider/helper.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+impl<N> std::fmt::Debug for WrappedProvider<'_, N> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("WrappedProvider")
+            .field(&"<dyn Provider>") // Since we can't debug the trait object directly
+            .finish()
+    }
+}
```
mattsse, line 30: I'd like to add a few tests here that ensure that this type is `Clone + Unpin +'static `

like `fn assert_provider<P: Provider + Clone + Unpin +'static`()` and called with wrapped provider

crates/provider/src/provider/helper.rs
```diff
@@ -0,0 +1,30 @@
+use crate::{Provider, RootProvider};
+
+use alloy_network::Network;
+use std::sync::Arc;
+
+/// A wrapper struct around Arc dyn provider
+pub struct WrappedProvider<'a, N>(Arc<dyn Provider<N> + 'a>);
+
+impl<'a, N: Network> WrappedProvider<'a, N> {
+    pub fn new<P>(provider: &'a P) -> Self
```
mattsse, line 10: this should consume the provider because we don't want to use references for this

crates/provider/src/provider/helper.rs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        Self(Arc::new(provider))
+    }
+}
+
+impl<N: Network> Provider<N> for WrappedProvider<'_, N> {
+    fn root(&self) -> &RootProvider<N> {
+        self.0.root()
+    }
+}
```
mattsse, line 22: I assume we need override more functions because the wrapped provider could use non default fns

so this needs to delegate all fns

crates/provider/src/provider/helper.rs
Dhruv-2003: Yup, If I do that, then the wrapped implementation would need me to constrain Clone on provider

```
   fn wrapped(&self) -> WrappedProvider<N>
    where
        Self: Sized + Clone + 'static,
    {
        WrappedProvider::new(self.clone())
    }
``` 

and that seems to break a few other things in the codebase as well, in contract crate and here https://github.com/Dhruv-2003/alloy/blob/045371819661cba7c2b5ebb9509a38d2c8ec58fe/crates/provider/src/fillers/nonce.rs#L175

Let me know if another approach is to be taken here

crates/provider/src/provider/helper.rs
Dhruv-2003: delegated all the fns except `get_filter_changes` and `raw_request`, as they required `Sized`

we already had their dyn version

mattsse:
nice, taking a look at the errors shortly

Dhruv-2003:
thanks for making the changes @mattsse 

the `auto_impl` thing was new for me, lmk if there is more to do here

mattsse:
ty!

mattsse (approved):

PR #2010, opened 2025-02-05
feat(`provider`): multicall

yash-atreya:
## Motivation

Implementing a Multicall3 builder type

## Solution

Use a tuple based approach to add numerous calls to the builder with type inference.

```rust
sol! {
     #[sol(rpc)]
     #[derive(Debug)]
     interface ERC20 {
         function totalSupply() external view returns (uint256 totalSupply);
         function balanceOf(address owner) external view returns (uint256 balance);
     }
}

let weth = address!("C02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");
let provider = ProviderBuilder::new().on_anvil_with_config(|a| a.fork(FORK_URL));

let erc20 = ERC20::new(weth, &provider);
let owner = address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045");
let multicall = provider.multicall().add(erc20.totalSupply()).add(erc20.balanceOf(owner));

let (total_supply, balance) = multicall.aggregate().await.unwrap();
```

- [x] fix return type decoding - make it a flat tuple
- [x] `aggregate`
- [x] `aggregate3`
- [x] `aggregate3Value`
- [x] `try*` aggregate methods
- [x] other util methods for fetching block context

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


DaniPopes:
FYI https://github.com/alloy-rs/core/pull/492 contains possibly useful stuff like TuplePush to flatten the tuple return type

crates/contract/src/multicall.rs
```diff
@@ -0,0 +1,230 @@
...
+sol! {
+    function identity() public;
+}
+
+impl Iterator for Identity {
+    type Item = Call3;
+    fn next(&mut self) -> Option<Self::Item> {
+        None
+    }
+}
```
mattsse, line 197: if this is the default, then this will actually send that call, right?

crates/contract/src/multicall.rs
```diff
@@ -0,0 +1,230 @@
...
+impl DecodeReturns for Identity {
+    type Returns = ();
+    fn decode_returns(_data: &[Bytes]) -> ContractResult<Self::Returns> {
+        Ok(())
+    }
+}
+
+// Decode the stack recursively.
+impl<L: SolCall, R: DecodeReturns> DecodeReturns for Stack<L, R> {
+    type Returns = (R::Returns, L::Return); // Maintain call order.
```
mattsse, line 130: this will result in a big nested tuple.

since we know that there's an upper limit to how many calls can be batched we could approach this a bit differently by introducing helper types

`struct R2<A, B> {a: A, b: B}.... R16<A....>{...}`

into which we collect.

because this is sequential, we could also integrate these into builder stack.

so that R1 -> R2 -> R3 -> R4

like `Multicall<R1<A>.add -> MultiCall<R2<A,B>`.

this way we can flatten the output into tuples, and likely also add something like `collect` so that this can be collected into a user-defined type perhaps

mattsse (changes requested):

crates/contract/src/multicall.rs
DaniPopes: https://github.com/alloy-rs/alloy/pull/2010#issuecomment-2637863177

yash-atreya: @DaniPopes @mattsse ptal

Made several changes, using `TuplePush` based approach now. Looking a lot cleaner.

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+        let total_value = calls.iter().map(|c| c.value).fold(U256::ZERO, |acc, x| acc + x);
+        let call = aggregate3ValueCall { calls: calls.to_vec() };
+        let output = self.build_and_call(call, Some(total_value)).await?;
+        T::decode_return_results(&output.returnData)
+    }
+
+    async fn build_and_call<M: SolCall>(
+        &self,
+        call_type: M,
+        value: Option<U256>,
```
mattsse, line 175: why does this need a standalone value arg?

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+            state_override: self.state_override,
+            address: self.address,
+            _pd: Default::default(),
+        }
+    }
+
+    /// Call the `aggregate` function
+    ///
+    /// Requires that all calls succeed.
+    pub async fn call_aggregate(&self) -> ContractResult<(U256, T::SuccessReturns)> {
```
mattsse, line 137: what does this return?

I think we should consider a custom future for this so that we can also await into the tuple directly: `let (r1,r2,..)`

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,434 @@
...
+        self.add_call(call)
+    }
+
+    /// Add a call to get the last block hash
+    pub fn add_get_last_block_hash(self) -> MulticallBuilder<T::Pushed, P, N>
+    where
+        T: TuplePush<getLastBlockHashCall>,
+        T::Pushed: CallTuple,
+    {
+        let call = CallInfo::new(MULTICALL3_ADDRESS, getLastBlockHashCall {});
```
DaniPopes, line 300: self.address in all these here?

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,434 @@
...
+    provider: P,
+    /// The block to make the call at
+    block: Option<BlockId>,
+    /// The state overrides for this call
+    state_override: Option<StateOverride>,
+    /// This is the address of the [`IMulticall3`](crate::multicall::bindings::IMulticall3)
+    /// contract.
+    ///
+    /// If none, resolved to the default address for most chains: [`MULTICALL3_ADDRESS`].
+    address: Option<Address>,
```
DaniPopes, line 43: this should be just Address

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,436 @@
+//! A Multicall Builder
+
+use crate::{Error, MulticallError, Result as ContractResult};
+use alloy_network::{Network, TransactionBuilder};
+use alloy_primitives::{address, Address, BlockNumber, Bytes, U256};
+use alloy_provider::Provider;
+use alloy_rpc_types_eth::{state::StateOverride, BlockId};
+use alloy_sol_types::SolCall;
+
+mod bindings;
```
DaniPopes, line 10:
```suggestion
pub mod bindings;
```

crates/contract/src/multicall/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+pub use inner_types::CallInfo;
+use inner_types::CallInfoTrait;
+
+mod tuple;
+pub use tuple::{CallTuple, Failure, TuplePush};
+
+/// Default address for the Multicall3 contract on most chains. See: <https://www.multicall3.com/deployments>
+pub const MULTICALL3_ADDRESS: Address = address!("cA11bde05977b3631167028862bE2a173976CA11");
+
+/// A multicall builder
```
DaniPopes, line 28: link to the repo

crates/provider/src/ext/multicall.rs
```diff
@@ -0,0 +1,64 @@
...
+    P: Provider<N>,
+{
+    fn multicall<C: SolCall + 'static>(
+        &self,
+        call: C,
+        target: Address,
+    ) -> MulticallBuilder<(C,), &P, N> {
+        MulticallBuilder::new(self).add(call, target)
+    }
+}
```
mattsse: imo this should be directly part of provider and shouldn't require an extra trait import

crates/provider/src/ext/multicall.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl<N, P> MulticallApi<N> for P
+where
+    N: Network,
+    P: Provider<N>,
+{
+    fn multicall<C: SolCall + 'static>(
+        &self,
+        call: C,
+        target: Address,
```
mattsse, line 63: this shouldn't take a call generic if part of the provider trait

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+    ///         ProviderBuilder::new().on_anvil_with_config(|a| a.fork("https://eth.merkle.io"));
+    ///
+    ///     let multicall = MulticallBuilder::new(provider).add(ts_call, weth).add(balance_call, weth);
+    ///
+    ///     let (_block_num, (total_supply, balance)) = multicall.aggregate().await.unwrap();
+    ///
+    ///     println!("Total Supply: {:?}, Balance: {:?}", total_supply, balance);
+    /// }
+    /// ```
+    pub async fn aggregate(&self) -> Result<(U256, T::SuccessReturns)> {
```
mattsse, line 223: can we introduce a helper future for this,

in case user doesn't care about the block number

mattsse, line 223: block num should be u64?

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,727 @@
...
+        if let Some(overrides) = &self.state_override {
+            eth_call = eth_call.overrides(overrides);
+        }
+
+        let res = eth_call.await.map_err(MulticallError::TransportError)?;
+        M::abi_decode_returns(&res, true).map_err(MulticallError::DecodeError)
+    }
+
+    /// Add a call to get the block hash from a block number
+    pub fn add_get_block_hash(self, number: BlockNumber) -> MulticallBuilder<T::Pushed, P, N>
```
mattsse, line 393: should we consider dropping the `add_` prefix?

@gakonst

mattsse (changes requested):
more nits :)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [b145879](https://github.com/alloy-rs/alloy/pull/2010/commits/b1458792989a81815d5a5db8e6697a8fe91f0774)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [391bf46](https://github.com/alloy-rs/alloy/pull/2010/commits/391bf460d90410abc05720c6b4a4e4c6ca3d06bb)

crates/provider/src/ext/multicall.rs
yash-atreya: [f1ae06e](https://github.com/alloy-rs/alloy/pull/2010/commits/f1ae06e96c7359120b2abed42528aaa748257b06)

crates/provider/src/ext/multicall.rs
yash-atreya: [f1ae06e](https://github.com/alloy-rs/alloy/pull/2010/commits/f1ae06e96c7359120b2abed42528aaa748257b06)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: I think we should just return `T::SuccessReturns` and leave out the block number.  Users that want block context can use `block_and_aggregate` which returns `block_number` and `block_hash`

wdyt?

crates/contract/src/multicall/mod.rs
yash-atreya: https://github.com/alloy-rs/alloy/pull/2010#discussion_r1946425796

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+///
+///     let (_block_num, (total_supply, balance)) = multicall.aggregate().await.unwrap();
+///
+///     println!("Total Supply: {:?}, Balance: {:?}", total_supply, balance);
+/// }
+/// ```
+#[derive(Debug)]
+pub struct MulticallBuilder<T: CallTuple, P: Provider<N>, N: Network> {
+    /// Batched calls
+    calls: Vec<Box<dyn CallInfoTrait>>,
```
klkvr, line 72: seems that this trait can't really be used externally so I'm wondering if we should just store a `Vec<Call3Value>` here and abi-encode when pushing?

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,774 @@
...
+    async fn test_aggregate() {
+        let ts_call = ERC20::totalSupplyCall {};
+        let balance_call =
+            ERC20::balanceOfCall { owner: address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045") };
+
+        let weth = address!("C02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");
+        let provider = ProviderBuilder::new().on_anvil_with_config(|a| a.fork(FORK_URL));
+        let multicall = MulticallBuilder::new(provider)
+            .add(ts_call.clone(), weth)
+            .add(balance_call.clone(), weth)
```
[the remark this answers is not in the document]
yash-atreya: [1ed8e44](https://github.com/alloy-rs/alloy/pull/2010/commits/1ed8e449afb766d9a6c5335eaf99d3991a01d88a)

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+    state_override: Option<StateOverride>,
+    /// This is the address of the [`IMulticall3`](crate::bindings::IMulticall3)
+    /// contract.
+    ///
+    /// By default it is set to [`MULTICALL3_ADDRESS`].
+    address: Address,
+    _pd: std::marker::PhantomData<(T, N)>,
+}
+
+impl<P, N> MulticallBuilder<Empty, P, N>
```
DaniPopes, line 90: Let's also add `fn with_cloned_provider` like in `CallBuilder`

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,550 @@
...
+///
+///     let multicall = provider.multicall().add(ts_call).add(balance_call);
+///
+///     let (_block_num, (total_supply, balance)) = multicall.aggregate().await.unwrap();
+///
+///     println!("Total Supply: {:?}, Balance: {:?}", total_supply, balance);
+/// }
+/// ```
+#[derive(Debug)]
+pub struct MulticallBuilder<T: CallTuple, P: Provider<N>, N: Network> {
```
DaniPopes, line 73: Should we also support a dynamic version for e.g. fetching the balances of an address for `Vec<Erc20>`? Might be able to include it in the abstraction, if not it could be a different type

yash-atreya: You mean some like `provider.erc20_balances(addr) -> ERC20Balances` using Multicall underneath?

DaniPopes: yeah but thats just one example, I mean something like where you can .add() without changing the type, and the return type is also a Vec of the same length as the inputs

DaniPopes: could possibly be implemented with `T` being a marker type for the dynamic builder and implementing `fn add_dynamic(&mut self)`

crates/contract/src/call.rs
```diff
@@ -754,4 +769,271 @@ mod tests {
...
+    }
+
+    sol! {
+        // solc 0.8.25; solc DummyThatFails.sol --optimize --bin
+        #[sol(rpc, bytecode = "6080604052348015600e575f80fd5b5060a780601a5f395ff3fe6080604052348015600e575f80fd5b50600436106030575f3560e01c80630b93381b146034578063a9cc4718146036575b5f80fd5b005b603460405162461bcd60e51b815260040160689060208082526004908201526319985a5b60e21b604082015260600190565b60405180910390fdfea2646970667358221220c90ee107375422bb3516f4f13cdd754387c374edb5d9815fb6aa5ca111a77cb264736f6c63430008190033")]
+        #[derive(Debug)]
+        contract DummyThatFails {
+            function fail() external {
+                revert("fail");
+            }
```
zerosnacks, line 818: fyi: multicall3 does not support bubbling up of errors (or custom error forwarding for that matter) which will make it difficult to debug

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [47d5a73](https://github.com/alloy-rs/alloy/pull/2010/commits/47d5a734494e0c88cccc239e79bf3945a50aa826)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [57062d1](https://github.com/alloy-rs/alloy/pull/2010/commits/57062d16c131a7412eb89c3ecb7b64db27e5ada5)

crates/provider/src/provider/trait.rs
```diff
@@ -181,6 +181,51 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    ///
+    ///     let (_block_num, (total_supply, balance)) = multicall.aggregate().await.unwrap();
+    ///
+    ///     println!("Total Supply: {:?}, Balance: {:?}", total_supply, balance);
+    /// }
+    /// ```
+    #[auto_impl(keep_default_for(&, &mut, Rc, Arc, Box))]
+    fn multicall(&self) -> MulticallBuilder<Empty, &Self, N>
+    where
+        Self: Sized,
```
mattsse, line 224: hmm, this is the first time we do it like this and not use the client.

for mutlicall we def need the provider because we need its functions

mattsse (approved):
lgtm, pending @DaniPopes and @klkvr

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [fae92c6](https://github.com/alloy-rs/alloy/pull/2010/commits/fae92c6d41d4bfc581aff449fda643457cf160f0)

klkvr (approved):
lgtm!

zerosnacks (approved):

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+
+    /// Instantiate a new [`MulticallBuilder`] that restricts the calls to a specific call type.
+    ///
+    /// Multicalls made using this builder return a vector of the decoded return values.
+    ///
+    /// An example would be trying to fetch multiple ERC20 balances of an address.
+    ///
+    /// ## Example
+    ///
+    /// ```ignore
```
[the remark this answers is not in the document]
yash-atreya: Yep! Won't compile

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+    ///
+    ///    let balances: Vec<ERC20::balanceOfReturn> = erc20_balances.aggregate().await.unwrap();
+    ///
+    ///    let weth_bal = balances[0];
+    ///    let usdc_bal = balances[1];
+    ///    println!("WETH Balance: {:?}, USDC Balance: {:?}", weth_bal, usdc_bal);
+    /// }
+    pub fn new_dynamic<D: SolCall + 'static>(
+        provider: P,
+        item: impl MulticallItem<Decoder = D>,
```
DaniPopes, line 150: this arg is redundant, the generics should be able to be inferred by moving this method to a separate impl block for `impl<D> MulticallBuilder<Dynamic<D>> { ... }`

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+    ///    let weth = address!("C02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2");
+    ///    let usdc = address!("A0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48");
+    ///     
+    ///    let provider = ProviderBuilder::new().on_http("https://eth.merkle.io".parse().unwrap());
+    ///    let weth = ERC20::new(weth, &provider);
+    ///    let usdc = ERC20::new(usdc, &provider);
+    ///
+    ///    let owner = address!("0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045");
+    ///
+    ///    let erc20_balances = MulticallBuilder::new_dynamic(provider, weth.balanceOf(owner)).add_dynamic(usdc.balanceOf(owner));
```
DaniPopes, line 140: this should show that you can do it in a loop like `for token in [...] { builder = builder.add_dynamic(...); }`

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -0,0 +1,680 @@
...
+
+        let call = CallItem::<D>::new(target, input);
+
+        self.calls.push(call.to_call3_value());
+        self
+    }
+
+    /// Extend the builder with a sequence of calls
+    ///
+    /// Note: The MulticallBuilder has a call limit of 16.
```
DaniPopes, line 183: Not sure if this is right for dynamic

yash-atreya: [310526c](https://github.com/alloy-rs/alloy/pull/2010/commits/310526cba3917778b67b6ff5aaefab92a147642f)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [310526c](https://github.com/alloy-rs/alloy/pull/2010/commits/310526cba3917778b67b6ff5aaefab92a147642f)

crates/provider/src/provider/multicall/mod.rs
yash-atreya: [310526c](https://github.com/alloy-rs/alloy/pull/2010/commits/310526cba3917778b67b6ff5aaefab92a147642f)

Issue #2015, opened 2025-02-06
[Feature] custom deserde impl for executionpayload


siosw:
### Component

rpc

### Describe the feature you would like

as described in #2004 

https://github.com/alloy-rs/alloy/blob/0fbc37da5178681e51e93d5274681f85c874702a/crates/rpc-types-engine/src/payload.rs#L827-L832

> right now if we have a `ExecutionPayloadV3` json, we can still do:
> `let payload: ExecutionPayloadV2 = serde_json::from_str(json).unwrap();`
> 
> and extra fields just implicitly get ignored. but obv you can no longer roundtrip then.

 _Originally posted by @siosw in [#2004](https://github.com/alloy-rs/alloy/issues/2004#issuecomment-2637200967)_

a custom deserde impl could enforce that executionpayload can only be deserialized into the correct version

### Additional context

@mattsse i think i have a PR that fixes this but it adds a lot of boilerplate, maybe there is a better way to do it

siosw:
closed by #2017

PR #2017, opened 2025-02-06
custom deserde impl

siosw:
closing #2015

unsure about the following:
- maybe there is a way to do this in a less verbose way
- when adding a V4 struct in the future, a custom deserde for V2 will have to be added. maybe there is a solution that solves this more generally

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -1670,6 +1670,10 @@ mod tests {
 
         let payload: ExecutionPayload = serde_json::from_str(response_v3).unwrap();
         assert!(payload.as_v3().is_some());
+        assert_eq!(serde_json::to_string(&payload).unwrap(), response_v3);
+
+        let payload_v3: ExecutionPayloadV3 = serde_json::from_str(response_v3).unwrap();
+        assert_eq!(payload.as_v3().unwrap(), &payload_v3);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/payload.rs
```diff
@@ -373,6 +380,161 @@ impl ExecutionPayloadV1 {
     }
 }
 
+#[cfg(feature = "serde")]
+impl<'de> Deserialize<'de> for ExecutionPayloadV1 {
```
mattsse, line 384: I don't want to touch serde on the versioned types directly but we should implement it like this for the `ExecutionPayload` enum

deserializing all fields and then selecting the version based on the fields

siosw: ah fair, and then i guess remove the derived deserialize from the versioned types?

mattsse: no we need proper serde on ExecutionPayloadV1 ExecutionPayloadV2 ExecutionPayloadV3

we only need manual deserde on ExecutionPayload which then is implemented like this draft

mattsse (changes requested):

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1479,6 +1663,36 @@ mod tests {
...
+
+        let payload: ExecutionPayload = serde_json::from_str(response_v3).unwrap();
```
mattsse, line 1671: can we add roundtrips to all of the tests to ensure that payload serializes into the same `serde_json::Value` as response_v3

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1051,25 +1051,209 @@ impl<T: Decodable2718> TryFrom<ExecutionPayload> for Block<T> {
...
+                    gas_limit,
+                    gas_used,
+                    timestamp,
+                    extra_data,
+                    base_fee_per_gas,
+                    block_hash,
+                    transactions,
+                };
+
+                if let Some(withdrawals) = withdrawals {
```
mattsse, line 1211: you can also use 

```
let Some(Some(withdrawals) = withdrawals else {
  return ...
}
```

syntax here

mattsse (changes requested):
nice, tysm

one pedantic test nit

mattsse (approved):
lgtm, ty

PR #2022, opened 2025-02-07
feat!: add eip1559 estimator type

### LINKED ISSUE #2021: [Feature] Improve eip1559 estimation api
### Component

rpc, provider, pubsub

this is suboptimal

https://github.com/alloy-rs/alloy/blob/fb2f9971e93a1998c32d3d9f2c9a3c805b6ced3e/crates/provider/src/provider/trait.rs#L219-L219

this should be two functions, one that uses the default estimator and one that uses the default

### Additional context

_No response_

mattsse:
closes #2021

removes the weird `estimator: Option<EstimatorFunction>,` argument and instead introduces a helper enum instead for the usecase where this is required.

we can use a trait because we want this to be object safe.

this is breaking but we could also keep it for a bit but I like `estimate_eip1559_fees` and `estimate_eip1559_fees_with`


crates/provider/src/provider/trait.rs
```diff
@@ -212,11 +211,11 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
 
     /// Estimates the EIP1559 `maxFeePerGas` and `maxPriorityFeePerGas` fields.
     ///
-    /// Receives an optional [EstimatorFunction] that can be used to modify
+    /// Receives an optional [Eip1559Estimator] that can be used to modify
```
yash-atreya, line 214:
```suggestion
    /// Receives an [Eip1559Estimator] that can be used to modify
```

yash-atreya, line 214:
```suggestion
    /// Receives an [Eip1559Estimator] that can be used to modify
```

Issue #1954, opened 2025-01-27
[Feature] Add helpers for PrivateTransactionRequest


mattsse:
### Component

rpc

### Describe the feature you would like

https://github.com/alloy-rs/alloy/blob/ae7c342395db74bf01f5d4e82aa34232ecb7b14c/crates/rpc-types-mev/src/eth_calls.rs#L299-L312

for this type we could use some helper functions, like

`fn from_tx<T:Encodable2718>`

and builder style fns for setting the additional fields

nadtech-hub:
i would like to work on this

Ayushdubey86:
if this is still open and unsolved? can i take a look in this?

PR #2023, opened 2025-02-07
feat: builder fns for PrivateTransactionRequest and inner props (#1954)

nadtech-hub:
## Motivation

Builder style functions for PrivateTransactionRequest

## Solution

Added PrivateTransactionRequest's and inner properties implementation blocks containing builder fns 

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-mev/src/eth_calls.rs
+++ b/crates/rpc-types-mev/src/eth_calls.rs
@@ -314,12 +314,12 @@ pub struct PrivateTransactionRequest {
 impl<T: Encodable2718> From<T> for PrivateTransactionRequest {
     fn from(envelope: T) -> Self {
         Self {
-            tx: Bytes::from(envelope.encoded_2718()),
-            max_block_number: Option::None,
+            tx: envelope.encoded_2718().into(),
+            max_block_number: None,
             preferences: PrivateTransactionPreferences {
-                fast: Option::None,
-                validity: Option::None,
-                privacy: Option::None,
+                fast: None,
+                validity: None,
+                privacy: None,
             },
         }
     }
--- a/crates/rpc-types-mev/src/eth_calls.rs
+++ b/crates/rpc-types-mev/src/eth_calls.rs
@@ -311,6 +311,34 @@ pub struct PrivateTransactionRequest {
     pub preferences: PrivateTransactionPreferences,
 }
 
+impl<T: Encodable2718> From<T> for PrivateTransactionRequest {
+    fn from(envelope: T) -> Self {
+        PrivateTransactionRequest {
+            tx: Bytes::from(envelope.encoded_2718()),
+            max_block_number: Option::None,
+            preferences: PrivateTransactionPreferences {
+                fast: Option::None,
+                validity: Option::None,
+                privacy: Option::None,
+            },
+        }
+    }
+}
+
+impl PrivateTransactionRequest {
+    /// Sets private tx's max block number.
+    pub const fn max_block_number(mut self, num: Option<u64>) -> Self {
+        self.max_block_number = num;
+        self
+    }
+
+    /// Sets private tx's preferences.
+    pub fn with_preferences(mut self, preferences: PrivateTransactionPreferences) -> Self {
+        self.preferences = preferences;
+        self
+    }
+}
+
 /// Additional preferences for `eth_sendPrivateTransaction`
 #[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


nadtech-hub:
afaik, you lot don't cover such functionality with tests, do you?

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -311,6 +311,34 @@ pub struct PrivateTransactionRequest {
...
 
+impl<T: Encodable2718> From<T> for PrivateTransactionRequest {
+    fn from(envelope: T) -> Self {
+        Self {
+            tx: Bytes::from(envelope.encoded_2718()),
+            max_block_number: Option::None,
+            preferences: PrivateTransactionPreferences {
+                fast: Option::None,
+                validity: Option::None,
+                privacy: Option::None,
```
mattsse, line 322: we can just use `None` here

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -311,6 +311,34 @@ pub struct PrivateTransactionRequest {
     pub preferences: PrivateTransactionPreferences,
 }
 
+impl<T: Encodable2718> From<T> for PrivateTransactionRequest {
+    fn from(envelope: T) -> Self {
+        Self {
+            tx: Bytes::from(envelope.encoded_2718()),
```
mattsse, line 317:
```suggestion
            tx: envelope.encoded_2718().into(),
```

mattsse (changes requested):
cool, smol nit

nadtech-hub:
> cool, smol nit

I was struggling to understand usage of this request

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -311,6 +311,34 @@ pub struct PrivateTransactionRequest {
...
+impl<T: Encodable2718> From<T> for PrivateTransactionRequest {
+    fn from(envelope: T) -> Self {
+        Self {
+            tx: envelope.encoded_2718().into(),
+            max_block_number: None,
+            preferences: PrivateTransactionPreferences {
+                fast: None,
+                validity: None,
+                privacy: None,
+            },
```
mattsse, line 323: can we use a new function for this as well, from functions alone are more obfuscated

mattsse (changes requested):

crates/rpc-types-mev/src/eth_calls.rs
nadtech-hub: is empty Bytes allowed?

mattsse: no but all those fields are pub,

the only thing missing here is this from fn as 

```
impl PrivateTransactionRequest {

pub fn new<T>() -> Self {
Self {
            tx: envelope.encoded_2718().into(),
            max_block_number: None,
            preferences: PrivateTransactionPreferences {
                fast: None,
                validity: None,
                privacy: None,
            },
}
```

and then the from function would just call  `PrivateTransactionRequest::new(tx)`

nadtech-hub: makes some sense

nadtech-hub: @mattsse I tried to impl TryFrom in order to return custom error but stuck with conflict. Is error needed?

mattsse (approved):
tysm

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -311,23 +311,27 @@ pub struct PrivateTransactionRequest {
...
-                validity: None,
-                privacy: None,
-            }
+            preferences: Default::default(),
         }
     }
-    
+
+    /// Apply a function to the request, returning the modified request.
+    pub fn apply<F>(self, f: F) -> Self
```
nadtech-hub, line 325: what supplied fn can do to tx object?

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -311,6 +311,43 @@ pub struct PrivateTransactionRequest {
     pub preferences: PrivateTransactionPreferences,
 }
 
+impl PrivateTransactionRequest {
+    /// Creates new [`PrivateTransactionRequest`] from the given encodable transaction.
+    pub fn new<T: Encodable2718>(tx: &T) -> Self {
```
nadtech-hub, line 316: here, just byte array should be passed

PR #2025, opened 2025-02-07
feat: use `OnceCell` for `Signed::hash`

### LINKED ISSUE #1987: [Feature] Use lazy hashing for Signed
### Component

consensus, eips, genesis

currently the signed type mandates a hash:

https://github.com/alloy-rs/alloy/blob/dc3a7fe498de5a705b0e45f7b3bf7555626734b0/crates/consensus/src/signed.rs#L15-L16

ideally we can relax this and instead compute the hash on demand if missing via a OnceLock

https://github.com/paradigmxyz/reth/blob/36807928311a91db559db228c3d25615152b4c8b/crates/ethereum/primitives/src/transaction.rs#L298-L300

https://github.com/paradigmxyz/reth/blob/36807928311a91db559db228c3d25615152b4c8b/crates/ethereum/primitives/src/transaction.rs#L355-L359

the signed value is currently just a type T, to compute the hash on demand we'd either need to introduce a new trait (basically `Sealable`) or we require that `T: Transaction`
I think we could make `T: Transaction` because this type is already intended for transactions:

https://github.com/alloy-rs/alloy/blob/dc3a7fe498de5a705b0e45f7b3bf7555626734b0/crates/consensus/src/signed.rs#L6-L9

## TODO
* replace B256 with OnceLock<B256>

### Additional context

_No response_

klkvr:
## Motivation

closes #1987

Lazily computes hash for the transaction in `Signed`. Methods relying on hash are now bounded by `T: RlpEcdsaTx` which works for all of the transactions we have but would likely get trickier for custom signature usecases 
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,4 +1,4 @@
-use crate::{transaction::RlpEcdsaTx, SignableTransaction, Transaction, TxType};
+use crate::{SignableTransaction, Transaction, TxType};
 use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization, Typed2718};
 use alloy_primitives::{Bytes, ChainId, PrimitiveSignature as Signature, TxKind, B256, U256};
 use alloy_rlp::{BufMut, Decodable, Encodable};
@@ -272,10 +272,6 @@ impl SignableTransaction<Signature> for TxEip1559 {
 
     fn payload_len_for_signature(&self) -> usize {
         self.length() + 1
-    }
-
-    fn tx_hash_with_signature(&self, signature: &Signature) -> B256 {
-        self.tx_hash(signature)
     }
 }
 
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,4 +1,4 @@
-use crate::{SignableTransaction, Signed, Transaction, TxType};
+use crate::{transaction::RlpEcdsaTx, SignableTransaction, Transaction, TxType};
 use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization, Typed2718};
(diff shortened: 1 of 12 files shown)


mattsse (commented):
I like this,

we should make it very clear that this Signed type is specifically for transactions, it's already mentioned but I think we can make this even clearer

klkvr:
we can but we'd still need once_cell in no-std

mattsse:
@klkvr we can simply upstream this to alloy:

https://github.com/paradigmxyz/reth/blob/b955551f8b3b37b0ead6862e8c1bdae1e7ff69e9/crates/primitives-traits/src/sync.rs#L3-L9

mattsse (approved):

klkvr:
We need `Signed<TypedTransaction` to still be usable as it seems that there are people relying on it e.g https://github.com/alloy-rs/alloy/pull/2070

So this means we need to somehow have a helper `tx_hash(&self, signature) -> B256` on it. RIght now it's only preset on `RlpEcdsaTx` which has a transaction type AT which is irrelevant for `TypedTransaction`. For now I've extended `SignableTransaction` with a `tx_hash_with_signature` method doing the same thing

Also I've changed `into_signed` to have default impl using `new_unhashed`

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -270,9 +270,8 @@ impl SignableTransaction<Signature> for TxEip1559 {
         self.length() + 1
     }
 
-    fn into_signed(self, signature: Signature) -> Signed<Self> {
-        let tx_hash = self.tx_hash(&signature);
-        Signed::new_unchecked(self, signature, tx_hash)
```
mattsse, line 275: ah these we no longer need because part of SignableTransaction

mattsse (approved):
lgtm!

but blocking until we finalize #2138

mattsse (approved):

PR #2037, opened 2025-02-10
feat(provider): expose inner `AnvilInstance` from `AnvilProvider`

0xdapper:
Some questions:
1. Not sure what kind of test to add for this
2. Does returning `&Arc<T>` feel right or should i return a cloned `Arc<T>`?

## Motivation

There is no way to access some of the `AnvilInstance` methods like `keys()`, `addresses()` right now when you spawn the anvil provider with `ProviderBuilder::on_anvil_XXX` methods.

## Solution

Exposing a reference to inner `AnvilInstance` will allow to access those methods instead of writing a wrapper delegate method for all the underlying methods.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/anvil.rs
+++ b/crates/provider/src/layers/anvil.rs
@@ -70,6 +70,11 @@ where
     pub fn new(inner: P, _anvil: Arc<AnvilInstance>) -> Self {
         Self { inner, _anvil }
     }
+
+    /// Expose inner anvil instance.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
reasonable

mattsse:
`&Arc<T>` makes sense here

PR #2038, opened 2025-02-10
fix(eips): Eip2718Error Eq + PartialEq

refcell:
### Description

Adds the `PartialEq + Eq` bounds to the `Eip2718Error`.


crates/eips/src/eip2718.rs
```diff
@@ -29,7 +29,7 @@ pub const EIP7702_TX_TYPE_ID: u8 = 4;
 /// [EIP-2718] decoding errors.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
mattsse, line 32: what's the motivation for this?

hesitant to lock in support for PartialEq Eq in here in case we need to add an additional variant

Issue #2030, opened 2025-02-08
[Feature] Add future type for getBlock requests


mattsse:
### Component

rpc

### Describe the feature you would like

getBlockBy supports an additional setting for full/hashes:
https://github.com/alloy-rs/alloy/blob/ef23dae5414d0f233de1e1f203dd29127163ca45/crates/provider/src/provider/trait.rs#L299-L304

it be nice if we could set this on the return type of `get_block_by_hash`

```rust
let resp = provider. get_block_by_hash(hash).full().await
// or
let resp = provider. get_block_by_hash(hash).await

```

similar to

https://github.com/alloy-rs/alloy/blob/ef23dae5414d0f233de1e1f203dd29127163ca45/crates/provider/src/provider/trait.rs#L145-L145

## TODO
* add helper future type for this

yohkaz:
Hey, can I take a look at this one?

nadtech-hub:
✋

PanGan21:
> [@mattsse](https://github.com/mattsse) naive question: how can the provider crate be tested locally? when running `cargo test`, I get such errors:
> 
> ```
> ---- provider::wallet::test::bubbles_through_fillers stdout ----
> thread 'provider::wallet::test::bubbles_through_fillers' panicked at crates/provider/src/layers/anvil.rs:22:66:
> called `Result::unwrap()` on an `Err` value: SpawnError(Os { code: 2, kind: NotFound, message: "No such file or directory" })
> ```

Do you have Geth and Anvil installed locally as the contribution docs state?
https://github.com/alloy-rs/alloy/blob/main/CONTRIBUTING.md#cargo-commands

yohkaz:
Missed this line, thank you : )

yohkaz:
@mattsse: started implementing the requested feature (https://github.com/alloy-rs/alloy/pull/2044), which result in having the RPC request call happen outside the `get_block_by_{hash, number}` function (=when the caller performs `.await` on the result). I then noticed the `CacheProvider` implementation of `Provider`, which kind of goes agaisnt the idea of our feature, since it does some pre-processing and post-processing on the results. Thinking on what would be a best way to support this.

PR #2044, opened 2025-02-11
Add future type for getBlockBy requests

yohkaz:
Towards https://github.com/alloy-rs/alloy/issues/2030


yohkaz:
The requested feature results in having the RPC request call happen outside the get_block_by_{hash, number} function (i.e. the RPC request call happens only when the caller performs .await on the result returned). 
I was quite happy with my code til I noticed the `CacheProvider` implementation of the `Provider` trait, which performs some "pre-processing" (check cache), and "post-processing" (update cache) with the result of the RPC request call.

I tried to support this by adding possibility for pre-process / post-process methods to be provided to the `EthGetBlock` struct.
This adds some more complexity, but also this solution might be too "tight" to the cache case.
lmk wdyt

crates/provider/src/provider/eth_get_block.rs
```diff
@@ -0,0 +1,336 @@
...
+            Self::Hash(hash) => hash.serialize(serializer),
+            Self::Number(num) => num.serialize(serializer),
+        }
+    }
+}
+
+impl From<BlockId> for BlockIdParam {
+    fn from(block: BlockId) -> Self {
+        match block {
+            BlockId::Hash(RpcBlockHash { block_hash, .. }) => Self::Hash(block_hash),
```
mattsse, line 36: this type is just BlockId, so I don't think we need this?

crates/provider/src/provider/trait.rs
```diff
@@ -284,70 +286,19 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
-        &self,
-        block: BlockId,
-        kind: BlockTransactionsKind,
-    ) -> TransportResult<Option<N::BlockResponse>> {
-        match block {
-            BlockId::Hash(hash) => self.get_block_by_hash(hash.into(), kind).await,
-            BlockId::Number(number) => self.get_block_by_number(number, kind).await,
-        }
+    /// Gets a block by either its hash, tag, or number
+    fn get_block(&self, block: BlockId) -> EthGetBlock<N> {
```
mattsse, line 290: for this pr I'd like to keep the api unchanged otherwise this becomes a breaking change

mattsse (changes requested):
good start

some suggestions

crates/provider/src/provider/eth_get_block.rs
yohkaz: Indeed, but it's re-implementing `Serialize` differently.  
When doing the `client.request(..., (block, full))` call, internally it serializes the parameters `(block, full)`. The `BlockId` serialize implementation is not good for that purpose.

crates/provider/src/provider/trait.rs
yohkaz: sure : )

yohkaz: So, if we keep the API unchanged (including returned types), all the logic happen inside the `get_block_by_{hash, number}` functions, so we don't actually need the pre/post-processing callbacks support anymore.
I removed this extra code.

mattsse (commented):
cool, this looks complete, I'll need to take closer look before I merge it

ty

crates/provider/src/provider/eth_get_block.rs
```diff
@@ -0,0 +1,277 @@
...
+    fn kind(&self) -> BlockTransactionsKind {
+        self.kind
+    }
+}
+
+/// The [`EthGetBlockByFut`] future is the future type for an `eth_getBlockBy` RPC request.
+#[allow(unnameable_types)]
+#[doc(hidden)] // Not public API.
+#[derive(Debug)]
+pub struct EthGetBlockFut<N>
```
DaniPopes, line 79: Needs must_use

crates/provider/src/provider/trait.rs
```diff
@@ -348,70 +349,31 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
     }
 
-    /// Get a block by its number.
-    // TODO: Network associate
+    /// Gets a block by its [BlockNumberOrTag]
     async fn get_block_by_number(
         &self,
         number: BlockNumberOrTag,
         kind: BlockTransactionsKind,
     ) -> TransportResult<Option<N::BlockResponse>> {
```
DaniPopes, line 375: The signatures must be changed to return this type rather than awaiting it directly

yohkaz: wouldn't this considered a breaking API change?

DaniPopes: Yes, but that is the point of the linked issue as you can see by the example usage

yohkaz: @mattsse mentionned not wanting an API breaking change in a previous comment, so I reverted the signatures changes

DaniPopes: ah I didnt see that

I dont see the point of this change then if signatures arent changed then? @mattsse

yohkaz: My guess was to internally have the code ready for a bigger release in which it'll be ok to have an API breaking change 
🤔

crates/provider/src/provider/eth_get_block.rs
```diff
@@ -0,0 +1,277 @@
...
+use futures::FutureExt;
+use std::{future::Future, marker::PhantomData, task::Poll};
+
+type BlockResult<N> = TransportResult<Option<N>>;
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum BlockIdParam {
+    Hash(BlockHash),
+    Number(BlockNumberOrTag),
+}
```
mattsse, line 16: this is just BlockId?

crates/provider/src/provider/trait.rs
mattsse: @yash-atreya  now that we already merged breaking changes we can also do this here

crates/provider/src/provider/eth_get_block.rs
yohkaz: Answered here: https://github.com/alloy-rs/alloy/pull/2044#discussion_r1953286713 : )

mattsse: I see

crates/provider/src/provider/trait.rs
mattsse: @yohkaz let's change `fn get_block_by_number(` in this pr as well

yohkaz: @mattsse
to clarify
1. we can now introduce the breaking change in which `get_block_by_{number, hash}` returns `EthGetBlock` ?
2. should `get_block_by_{number, hash}` methods receive the `BlockTransactionsKind` as argument, or should the caller perform another call on the `EthGetBlock` object returned, as originally asked in the issue example https://github.com/alloy-rs/alloy/issues/2030 ?

yash-atreya: [2125249](https://github.com/alloy-rs/alloy/pull/2044/commits/2125249ebb5624d4ea173de633ab33787469b927)

crates/network-primitives/src/block.rs
```diff
@@ -12,10 +12,10 @@ use core::slice;
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum BlockTransactions<T> {
-    /// Full transactions
-    Full(Vec<T>),
     /// Only hashes
     Hashes(Vec<B256>),
+    /// Full transactions
+    Full(Vec<T>),
```
yash-atreya, line 18: Changed the ordering of the variants here to fix this test: https://github.com/alloy-rs/alloy/blob/cbbb69820f7640b23d61b00974ce8500133f90c5/crates/provider/src/provider/trait.rs#L2015 

Now empty transactions vec is deserialized to the `Hashes` variant

cc @mattsse

crates/provider/src/provider/get_block.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn kind(mut self, kind: BlockTransactionsKind) -> Self {
+        self.kind = kind;
+        self
+    }
+
+    /// Set the `full:bool` argument in RPC calls
+    pub fn full(mut self) -> Self {
+        self.kind = BlockTransactionsKind::Full;
+        self
+    }
```
mattsse, line 82: can we also add hashes for completeness?

crates/provider/src/provider/trait.rs
```diff
@@ -354,70 +355,42 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
-    async fn get_block(
-        &self,
-        block: BlockId,
-        kind: BlockTransactionsKind,
-    ) -> TransportResult<Option<N::BlockResponse>> {
-        match block {
-            BlockId::Hash(hash) => self.get_block_by_hash(hash.into(), kind).await,
-            BlockId::Number(number) => self.get_block_by_number(number, kind).await,
+    /// Gets a block by either its hash, tag, or number
+    fn get_block(&self, block: BlockId) -> EthGetBlock<BlockId, Option<N::BlockResponse>> {
```
mattsse, line 359: why do we need to define the Option< as the type here?

mattsse, line 359: do we also need the blockid generic?

don't think so because this is always passed as input, right?

so any reason why this can't be `EthGetBlock<N::BlockResponse>` instead

crates/provider/src/utils.rs
```diff
@@ -126,6 +127,18 @@ pub(crate) fn convert_u64(r: U64) -> u64 {
...
+    r: Option<N::BlockResponse>,
+) -> Option<N::BlockResponse> {
+    r.map(|mut block| {
+        if block.transactions().is_empty() {
+            block.transactions_mut().convert_to_hashes();
+        }
+
+        block
+    })
+}
```
mattsse, line 140: why can't the EthGetBlock do this?

it knows the blockresponse

mattsse (changes requested):
some additional questions

crates/provider/src/fillers/mod.rs
```diff
@@ -396,28 +396,19 @@ where
         self.inner.get_balance(address)
     }
 
-    async fn get_block(
-        &self,
-        block: BlockId,
-        kind: BlockTransactionsKind,
-    ) -> TransportResult<Option<N::BlockResponse>> {
-        self.inner.get_block(block, kind).await
+    fn get_block(&self, block: BlockId) -> EthGetBlock<N, Option<N::BlockResponse>> {
```
mattsse, line 399: why does this need an N generic and an additional generic for N::BlockResponse, doesn't make much sense

mattsse (changes requested):

crates/provider/src/fillers/mod.rs
```diff
@@ -396,28 +396,16 @@ where
...
     }
 
-    async fn get_block(
-        &self,
-        block: BlockId,
-        kind: BlockTransactionsKind,
-    ) -> TransportResult<Option<N::BlockResponse>> {
-        self.inner.get_block(block, kind).await
+    fn get_block(&self, block: BlockId) -> EthGetBlock<N::BlockResponse> {
+        self.inner.get_block(block)
```
mattsse, line 400: nice, this looks great!

mattsse (approved):
lgtm!

PR #2049, opened 2025-02-13
migrate eip5792 and eip7547 from alloy repo to eips repo

Pana:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Cargo.toml
```diff
@@ -83,6 +81,8 @@ alloy-chains = { version = "0.1.18", default-features = false }
 alloy-eip2124 = { version = "0.1.0", default-features = false }
 alloy-eip2930 = { version = "0.1.0", default-features = false }
 alloy-eip7702 = { version = "0.5.0", default-features = false }
+alloy-eip7547 = { version = "0.11", default-features = false }
+alloy-eip5792 = { version = "0.11", default-features = false }
```
DaniPopes, line 85: should these be re-exported in alloy-eips? @mattsse

mattsse: I'd rather keep alloy-eips leaner and start converting them to standalone crates instead.
however there are a few edge cases for example if an eip depends on rpc types (like engine)
so for now we can't really extract those to alloy-eips and since these are not finalized we shouldn't move them into the eips crate

PR #2050, opened 2025-02-13
feat(`eip4844`): heap allocated blob

### LINKED ISSUE #1907: [Feature] Add Blob type that behaves like `Bytes`
### Component

consensus, eips, genesis

the current blob type is a fixedbytes alias allocated on the stack

https://github.com/alloy-rs/alloy/blob/1ba17f23b9a077f29fd42e1818c3e0edecc95b78/crates/eips/src/eip4844/mod.rs#L90-L92

this is expensive to clone and has issues with StackOverflow in serde.

ideally we want a this to behave like `Bytes` but enforces the length at all times

we can solve this with a wrapper type `Blob(Bytes)` and only allow BYTES_PER_BLOB 

but we need a way to transmute the blob to ckzg type so perhaps this should instead be `Arc<FixedBytes<BYTES_PER_BLOB>>`

so this will be problematic

https://github.com/alloy-rs/alloy/blob/1ba17f23b9a077f29fd42e1818c3e0edecc95b78/crates/eips/src/eip4844/sidecar.rs#L274-L274

so I think we can do this unless we change the ckzg validation interfaces
or we arc the entire sidecar, which would then be incompatible with the blobsbundlev1 type

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #1907 
Supersedes #1912

## Solution

Create a wrapper type `HeapBlob` over `Bytes`.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip4844/mod.rs
+++ b/crates/eips/src/eip4844/mod.rs
@@ -110,6 +110,8 @@ where
     Clone,
     PartialEq,
     Eq,
+    PartialOrd,
+    Ord,
     Hash,
     alloy_rlp::RlpEncodableWrapper,
     alloy_rlp::RlpDecodableWrapper,
@@ -140,7 +142,7 @@ impl HeapBlob {
 
     /// Generate a new heap blob with all bytes set to `byte`.
     pub fn repeat_byte(byte: u8) -> Self {
-        Self(Bytes::from(vec![byte; BYTES_PER_BLOB]))
+        unsafe { Self::new_unchecked(Bytes::from(vec![byte; BYTES_PER_BLOB])) }
     }
 
     /// Get the inner
@@ -339,5 +341,15 @@ mod tests {
             let actual = fake_exponential(factor as u128, numerator as u128, denominator as u128);
             assert_eq!(actual, expected, "test: {t:?}");
         }
+    }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    fn serde_heap_blob() {
+        let blob = HeapBlob::repeat_byte(0x42);
+        let serialized = serde_json::to_string(&blob).unwrap();
+
+        let deserialized: HeapBlob = serde_json::from_str(&serialized).unwrap();
+        assert_eq!(blob, deserialized);
     }
 }
--- a/crates/eips/src/eip4844/mod.rs
+++ b/crates/eips/src/eip4844/mod.rs
@@ -105,19 +105,7 @@ where
 }
 
 /// A heap allocated blob that serializes as 0x-prefixed hex string
-#[derive(
-    Debug,
-    Clone,
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
-    Hash,
-    alloy_rlp::RlpEncodableWrapper,
-    alloy_rlp::RlpDecodableWrapper,
-)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize))]
-#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, alloy_rlp::RlpEncodableWrapper)]
 pub struct HeapBlob(Bytes);
 
 impl HeapBlob {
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,85 @@ where
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    alloy_rlp::RlpEncodableWrapper,
+    alloy_rlp::RlpDecodableWrapper,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
```
mattsse, line 120: this we also need to do manually because we need to ensure fixed size

crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,85 @@ where
...
+    }
+
+    /// Create a new heap blob from a byte slice without checking the length.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure that the length of `blob` is [`BYTES_PER_BLOB`].
+    pub unsafe fn new_unchecked(blob: Bytes) -> Self {
+        Self(blob)
+    }
```
mattsse, line 140: this is safe

but this function shouldn't exist because this is a massive footgun

crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,85 @@ where
...
+    alloy_rlp::RlpEncodableWrapper,
+    alloy_rlp::RlpDecodableWrapper,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+pub struct HeapBlob(Bytes);
+
+impl HeapBlob {
+    /// Create a new heap blob from a byte slice.
+    pub fn new(blob: &[u8]) -> Result<Self, InvalidBlobLength> {
```
mattsse, line 125: we also need a fn that initialized from an array

mattsse (changes requested):

crates/eips/src/eip4844/mod.rs
yash-atreya: [0a13228](https://github.com/alloy-rs/alloy/pull/2050/commits/0a132285b5a69cf27939d4b56498cb9dd3e2a889)

crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,96 @@ where
...
+}
+impl core::error::Error for InvalidBlobLength {}
+
+#[cfg(feature = "serde")]
+impl serde::Serialize for HeapBlob {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        if self.inner().len() != BYTES_PER_BLOB {
```
DaniPopes, line 159: this should be an invariant of the type, so only checked at initialization

crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,96 @@ where
...
+#[cfg(feature = "serde")]
+impl<'de> serde::Deserialize<'de> for HeapBlob {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        // Deserialize the bytes
+        let inner = <Bytes>::deserialize(deserializer)?;
+
+        Self::new(&inner).map_err(serde::de::Error::custom)
```
DaniPopes, line 185: this copies the bytes twice, we want to have a `Bytes` constructor too

crates/eips/src/eip4844/mod.rs
```diff
@@ -104,6 +104,96 @@ where
...
+    }
+}
+
+impl alloy_rlp::Decodable for HeapBlob {
+    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        let bytes = <Bytes>::decode(buf)?;
+
+        Self::new(&bytes).map_err(|_| alloy_rlp::Error::Custom("invalid blob length"))
```
DaniPopes, line 193: same

yash-atreya: [b48cb63](https://github.com/alloy-rs/alloy/pull/2050/commits/b48cb637c1a6d873dd5f6e925b1e4d1f0967847b)

crates/eips/src/eip4844/mod.rs
yash-atreya: [b48cb63](https://github.com/alloy-rs/alloy/pull/2050/commits/b48cb637c1a6d873dd5f6e925b1e4d1f0967847b)

crates/eips/src/eip4844/mod.rs
yash-atreya: [b48cb63](https://github.com/alloy-rs/alloy/pull/2050/commits/b48cb637c1a6d873dd5f6e925b1e4d1f0967847b)

DaniPopes (approved):

mattsse (approved):

PR #2057, opened 2025-02-14
feat(`consensus`): introduce block traits

### LINKED ISSUE #2033: [Feature] Block and Body trait
### Component

consensus, eips, genesis

Introduce Block and Body trait

we have these in reth

https://github.com/paradigmxyz/reth/blob/1a4f5441f38447e5c6e2277eb6849f11f7492cf0/crates/primitives-traits/src/block/body.rs#L21-L21

https://github.com/paradigmxyz/reth/blob/1a4f5441f38447e5c6e2277eb6849f11f7492cf0/crates/primitives-traits/src/block/mod.rs#L51-L51

for header and transaction, reth's specific traits inherit core logic from alloy, e.g:

https://github.com/paradigmxyz/reth/blob/1a4f5441f38447e5c6e2277eb6849f11f7492cf0/crates/primitives-traits/src/block/header.rs#L15-L28

we should start doing the same for block and body, but without the associated types

this will make compat easier, eg: #2032

## TODO
* introduce trait Block which is an ethereum opinionated block trait
* to unblock #2032 it should have fn `withdrawals`

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2033 
Unblocks #2032 

## Solution

Introduce `BlockBodyT` and `BlockT`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/mod.rs
+++ b/crates/consensus/src/block/mod.rs
@@ -6,10 +6,10 @@ pub use header::{BlockHeader, Header};
 #[cfg(all(feature = "serde", feature = "serde-bincode-compat"))]
 pub(crate) use header::serde_bincode_compat;
 
-use crate::Transaction;
+use crate::{transaction::SignerRecoverable, Transaction};
 use alloc::vec::Vec;
-use alloy_eips::{eip4895::Withdrawals, Typed2718};
-use alloy_primitives::B256;
+use alloy_eips::{eip4895::Withdrawals, Encodable2718, Typed2718};
+use alloy_primitives::{Address, Bytes, SignatureError, B256};
 use alloy_rlp::{Decodable, Encodable, RlpDecodable, RlpEncodable};
 
 /// Ethereum full block.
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
can we keep this very minimal for now

mattsse (changes requested):
let's do it even more simple for now, I want to avoid locking in something that we might change later which would be breaking

`trait Block {fn withdrawals}`

crates/consensus/src/block/mod.rs
mattsse, line 1: this file is already quite large, can we move this to traits.rs?

crates/consensus/src/block/mod.rs
```diff
@@ -306,72 +306,15 @@ mod block_rlp {
     }
 }
 
-/// Represents the structure of a block.
+/// Minimal block type.
 pub trait BlockT<T, H> {
```
mattsse, line 310: we can drop the t,h for now

mattsse (changes requested):

mattsse (approved):

PR #2059, opened 2025-02-14
feat: remove preimage hashes from execution witness

meyer9:
## Motivation

Including the hashes gives callers an unnecessary reason to trust the client. They should just hash the returned preimages. Also saves on RPC bandwidth. Motivated by feedback on this geth PR: https://github.com/ethereum/go-ethereum/pull/30613#issuecomment-2531732780

reth PR: https://github.com/paradigmxyz/reth/pull/14503


mattsse (approved):
lgtm,

this is breaking so I will delay merging this

PR #2060, opened 2025-02-14
feat: add try_apply

mattsse:
add fallible helper to builder


DaniPopes (approved):

DaniPopes:
maybe should have a global trait like `tap`? :)

PR #2061, opened 2025-02-14
fix(rpc-types-mev): SimBundleLogs should contain all logs fields.

ryanschneider:
## Motivation

The `mev_simBundle` RPC should return "rpc" log objects that contain `transactionHash`, `blockNumber` etc.  Currently we are using the "primitive" version that just contains the "consensus" fields.

## Solution

Use `alloy_rpc_types_eth::Log` instead of `alloy_primitives::Log`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types-mev/src/mev_calls.rs
+++ b/crates/rpc-types-mev/src/mev_calls.rs
@@ -1,8 +1,8 @@
 use crate::common::{Privacy, ProtocolVersion, Validity};
 
 use alloy_eips::BlockId;
-use alloy_primitives::{Bytes, Log, TxHash, U256};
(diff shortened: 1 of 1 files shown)


mattsse (commented):
yeah I remember struggling to find specs for this

ryanschneider:
> yeah I remember struggling to find specs for this

Ya I don't think we have specs for it anywhere. :/

mattsse (approved):
lgtm, since this is blocking, 
I need to block this until until the next minor bump

DaniPopes (approved):

PR #2064, opened 2025-02-15
feat: add optional builder APIs for AccountOverride

### LINKED ISSUE #2062: [Feature] Add optional builder APIs for `AccountOverride`
### Component

rpc

Sometimes you might only want to set e.g. code for an account in `AccountOverride` in some cases. Since `AccountOverride` is mostly used inside of an `AddressMap`, it's a bit annoying having to reach into the map, get the value, modify it with e.g. `with_code` (which takes `self`) and then overriding the value. Instead, it would be nice to have either builder-style APIs that take `&mut self` or `Option<...>`, e.g. `set_code(self, code: Option<Into<Bytes>>)` or `set_code(&mut self, code: Into<Bytes>)`.

### Additional context

An example real-world use case of account overrides:

```rs
  let overrides = AddressMap::from_iter([
      (
          self.inner.quote_signer.address(),
          AccountOverride::default().with_balance(U256::MAX.div_ceil(2.try_into().unwrap())),
      ),
      (
          request.op.eoa,
          AccountOverride {
              state_diff: Some(key.storage_slots()),
              // we manually etch the 7702 designator since we do not have a signed auth item
              code: request.auth.map(|addr| {
                  Bytes::from([&EIP7702_DELEGATION_DESIGNATOR, addr.as_slice()].concat())
              }),
              ..Default::default()
          },
      ),
  ]);
```

Using the builder API for `request.op.eoa` is cumbersome for setting the code for `request.op.eoa` in *some* cases.

stevencartavia:
## Motivation
closes #2062 

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/state.rs
+++ b/crates/rpc-types-eth/src/state.rs
@@ -83,6 +83,36 @@ impl AccountOverride {
         self.nonce = Some(nonce);
         self
     }
+
+    /// Sets the bytecode override in place.
+    pub fn set_code(&mut self, code: impl Into<Bytes>) {
+        self.code = Some(code.into());
+    }
+
+    /// Sets the state overrides in place.
+    pub fn set_state(&mut self, state: impl IntoIterator<Item = (B256, B256)>) {
+        self.state = Some(state.into_iter().collect());
+    }
+
+    /// Sets the state diffs in place.
+    pub fn set_state_diff(&mut self, state_diff: impl IntoIterator<Item = (B256, B256)>) {
+        self.state_diff = Some(state_diff.into_iter().collect());
+    }
+
+    /// Sets the balance override in place.
+    pub fn set_balance(&mut self, balance: U256) {
+        self.balance = Some(balance);
+    }
+
+    /// Sets the nonce override in place.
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
these are good, we also need additional ones that accept option as input and return self for code balance and once like with_code_opt(Option) -> Self

crates/rpc-types-eth/src/state.rs
```diff
@@ -83,6 +83,87 @@ impl AccountOverride {
...
+    /// Conditionally sets the state diffs and returns self.
+    pub fn with_state_diff_opt(
+        mut self,
+        state_diff: Option<impl IntoIterator<Item = (B256, B256)>>,
+    ) -> Self {
+        if let Some(state_diff) = state_diff {
+            self.state_diff = Some(state_diff.into_iter().collect());
+        }
+        self
+    }
```
mattsse, line 142: these are redundant because Option is also IntoIter so we don't need these

mattsse (changes requested):

mattsse (approved):

PR #2068, opened 2025-02-16
fix: ws transport now checks for missed pongs within two pings

### LINKED ISSUE #2067: [Bug] WebSocket transport implementation should check for Pong responses
### Component

transports

0.11.0

### Operating System

Linux

### Describe the bug

Hello,

I have encountered an interesting issue that, unfortunately, is difficult to reproduce. However, the root problem is observable by simply reading the websocket implementation.

I am currently working from a place with a connection that, for a reason I do not know, sometimes breaks silently. Some services on my laptop recover, some do not (for instance, ssh just hangs). I receive no notification that something went wrong, besides http requests that were running at that exact moment failing (when done by one of my own Rust-built services, hyper returns the "IncompleteMessage" error).

I happen to be currently working on a service that requires a constant websocket connection to an EVM node hosted externally, using alloy. When the connection breaks in any other way I could test (for instance, shutting down my wifi and restarting it), alloy's websocket implementation catches it and restarts the connection as expected.

However, when the connection breaks in that silent way I have been experienced, alloy's websocket implementation does not detect it. THe keepalive timer still ticks, and the socket sends Ping frames without encountering an error. No message is received either from the socket. It's only after about 15 minutes of wait that the `socket.next()` call finally returns an error:

`IO error: peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof`

It's only then that I investigated the websocket implementation and found out that while it does send Ping frames, it does not check that a Pong frame was received in an appropriate time.
I believe that to prevent issues like mine and similar (as I am certain there is other cases where the connection can break without an immediate error showing, causing a too-long void of data), there should be a check that the server replies with a Pong frame in a timely manner, and if it does not, it should mark the connection as erorred. Now all servers might not be replying with Pong as they should, so I believe this additional check should be optional, although perhaps enabled by default.

I am for now, as a workaround, using a modified WsConnect, which in turns uses a modified WsBackend that checks that in-between two "Ping", a "Pong" was received. If it was not, it marks the connection as erorred and breaks the loop. Not the cleanest implementation as the timeout should be configurable separately from the keepalive interval, but it works well enough for my purpose at the moment.

the-solid-box:
## Motivation

fixes #2067 

## Solution

Missed Pong frame marks the connection as errored.

### Notes

Not sure a test is required for this, not sure what test to write either. Let me know if one should be done and I'll think about it when I can.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -116,6 +116,7 @@ impl WsBackend<TungsteniteStream> {
     pub fn spawn(mut self) {
         let fut = async move {
             let mut errored = false;
+            let mut expecting_pong = false;
             let keepalive = sleep(Duration::from_secs(KEEPALIVE));
             tokio::pin!(keepalive);
             loop {
@@ -153,6 +154,13 @@ impl WsBackend<TungsteniteStream> {
                     // Send a ping to the server, if no other messages have been
                     // sent in the last 10 seconds.
                     _ = &mut keepalive => {
+                        // Still expecting a pong from the previous ping,
+                        // meaning connection is errored.
+                        if expecting_pong {
+                            error!("WS server missed a pong");
+                            errored = true;
+                            break
+                        }
                         // Reset the keepalive timer.
                         keepalive.set(sleep(Duration::from_secs(KEEPALIVE)));
                         if let Err(err) = self.socket.send(Message::Ping(Default::default())).await {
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
Thanks!

PR #2070, opened 2025-02-17
chore: add From<Signed<TypedTransaction>> for TxEnvelope

merklefruit:
## Motivation

Useful helper to avoid unnecessary matching in applications that use both TypedTransaction and TxEnvelope types.


mattsse (approved):
lgtm

mattsse:
ci unrelated

PR #2072, opened 2025-02-18
feat(`contract`): decode as `SolError`

yash-atreya:
## Motivation

Currently, we have a method for decoding `ErrorPayload`'s and `contract::Error`'s into one of the `SolError`'s present in the `SolInterface`.

`fn as_decoded_error<E: SolInterface>(data) -> Option<E>`.

We also want a method that attempts decoding to a specific `SolError` and returns a `Result`. 

## Solution

- Add method `fn as_decoded_error<E: SolError>(data) -> Result<E, alloy_sol_types::Error>`
- [Breaking] Rename previous `as_decoded_error<E: SolInterface>` to `as_decoded_interface_error`. Avoidable but I think this naming is better, and breaking pre-1.0 makes sense.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/error.rs
+++ b/crates/contract/src/error.rs
@@ -68,8 +68,7 @@ impl Error {
         self.as_revert_data().and_then(|data| E::abi_decode(&data, false).ok())
     }
 
-    /// Decode the revert data into the specified typed
-    /// [`SolError`].
+    /// Decode the revert data into a custom [`SolError`] type.
     pub fn as_decoded_error<E: SolError>(&self) -> core::result::Result<E, alloy_sol_types::Error> {
         let data =
             self.as_revert_data().ok_or(alloy_sol_types::Error::custom("no revert data found"))?;
--- a/crates/contract/src/error.rs
+++ b/crates/contract/src/error.rs
@@ -64,11 +64,89 @@ impl Error {
     }
 
     /// Attempts to decode the revert data into one of the custom errors in [`SolInterface`].
+    ///
+    /// Returns an enum container type consisting of the custom errors defined in the interface.
+    ///
+    /// None is returned if the revert data is empty or if the data could not be decoded into one of
+    /// the custom errors defined in the interface.
+    ///
+    /// ## Example
+    ///
+    /// ```no_run
+    /// use alloy_provider::ProviderBuilder;
+    /// use alloy_sol_types::sol;
+    ///
+    /// sol! {
+    ///     #[derive(Debug, PartialEq, Eq)]
+    ///     #[sol(rpc, bytecode = "694207")]
+    ///     contract ThrowsError {
+    ///         error SomeCustomError(uint64 a);
+    ///         error AnotherError(uint64 b);
+    ///
+    ///         function error(uint64 a) external {
+    ///             revert SomeCustomError(a);
+    ///         }
+    ///     }
+    /// }
+    ///
+    /// #[tokio::main]
+    /// async fn main() {
+    ///     let provider = ProviderBuilder::new().on_anvil_with_wallet();
+    ///
+    ///     let throws_err = ThrowsError::deploy(provider).await.unwrap();
+    ///
+    ///     let err = throws_err.error(42).call().await.unwrap_err();
+    ///
+    ///     let custom_err =
+    ///         err.as_decoded_interface_error::<ThrowsError::ThrowsErrorErrors>().unwrap();
+    ///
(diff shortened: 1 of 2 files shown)


crates/json-rpc/src/response/error.rs
```diff
@@ -349,10 +349,18 @@ where
...
+    pub fn as_decoded_interface_error<E: SolInterface>(&self, validate: bool) -> Option<E> {
         self.as_revert_data().and_then(|data| E::abi_decode(&data, validate).ok())
     }
+
+    /// Extracts revert data and tries decoding it into custom [`SolError`].
+    pub fn as_decoded_error<E: SolError>(&self) -> Result<E, alloy_sol_types::Error> {
+        let data =
+            self.as_revert_data().ok_or(alloy_sol_types::Error::custom("revert data not found"))?;
+        E::abi_decode(&data, false)
+    }
```
onbjerg, line 363: let's do `Option` for consistency ?

onbjerg (commented):
small suggestion on api, for renaming to `as_decoded_interface_error` unsure if there is a better name, but i don't have a strong opinion/suggestion

yash-atreya:
> small suggestion on api, for renaming to `as_decoded_interface_error` unsure if there is a better name, but i don't have a strong opinion/suggestion

Yeah, couldn't think of a better one, open to ideas

crates/json-rpc/src/response/error.rs
yash-atreya: [71a33a1](https://github.com/alloy-rs/alloy/pull/2072/commits/71a33a1f7eb80b42c46bba88a2ce36d879f821e2)

crates/contract/src/error.rs
```diff
@@ -63,8 +63,13 @@ impl Error {
         None
     }
 
-    /// Decode revert data into a typed [`SolError`](alloy_sol_types::SolError).
-    pub fn as_decoded_revert<E: SolInterface>(&self) -> Option<E> {
```
mattsse, line 67: this is breaking, when did we add this?

yash-atreya: Added here https://github.com/alloy-rs/alloy/pull/2058

Not been released yet.

crates/contract/src/error.rs
```diff
@@ -63,8 +63,13 @@ impl Error {
...
+    /// Attempts to decode the revert data into one of the custom errors in [`SolInterface`].
```
mattsse, line 66: I find the docs for both functions a bit confusing

> one of the custom errors in [`SolInterface`]

> into a custom [`SolError`] type.

what's the difference here exactly?
can we document this and recommend when to use which

crates/json-rpc/src/response/error.rs
```diff
@@ -349,10 +349,16 @@ where
         }
     }
 
-    /// Extracts revert data and tries decoding it into given custom errors set.
-    pub fn as_decoded_error<E: SolInterface>(&self, validate: bool) -> Option<E> {
+    /// Extracts revert data and tries decoding it into given custom errors set present in the
+    /// [`SolInterface`].
+    pub fn as_decoded_interface_error<E: SolInterface>(&self, validate: bool) -> Option<E> {
```
DaniPopes, line 354: let's remove the validate

crates/contract/src/error.rs
DaniPopes: should have an example showing sol!
`SolInterface` is for the generated enum that contains all the contract errors
`SolError` is for an individual error

yash-atreya: I've added an example here: https://github.com/alloy-rs/examples/blob/bc7e4ba593deed094157525419f45b6ed259b241/examples/contracts/examples/revert_decoding.rs#L50-L65, explaining the difference.

Will add a doctest as well

yash-atreya: [71f2bf7](https://github.com/alloy-rs/alloy/pull/2072/commits/71f2bf767f57ba0ee134e56255e678940e8c0a34)

crates/json-rpc/src/response/error.rs
yash-atreya: [87968f1](https://github.com/alloy-rs/alloy/pull/2072/commits/87968f11cc1e99879389d5c63fd6e7871d681219)

mattsse (approved):
lgtm!

PR #2073, opened 2025-02-18
feat(genesis): Deserialize ChainConfig from toml

cakevm:
Allow to read ChainConfig from a toml file. The network configuration is stored as toml in [superchain-registry](https://github.com/ethereum-optimism/superchain-registry). The genesis files do not contain always the config field e.g. `base.json.zst` contains `"config":null`. Step toward resolving: https://github.com/paradigmxyz/reth/issues/14240

Changes:
- Allow parsing from json and toml
  - `serde_alias` introduces unmaintained exception for `proc-macro-error`
- Allow to parse Genesis from missing config field and `"config":null`

@mattsse

### DIFF
--- a/crates/transport/README.md
+++ b/crates/transport/README.md
@@ -23,8 +23,8 @@ specific Ethereum endpoints.
 Alloy maintains the following transports:
 
 - [alloy-transport-http]: JSON-RPC via HTTP.
-- [alloy-transport-ws]: JSON-RPC via Websocket, supports pubsub via
-    [alloy-pubsub].
+- [alloy-transport-ws]: JSON-RPC via Websocket, supports pubsub via.
+- [alloy-pubsub]: JSON-RPC publish-subscribe tower service.
 - [alloy-transport-ipc]: JSON-RPC via IPC, supports pubsub via [alloy-pubsub].
 
 [alloy-transport-http]: https://docs.rs/alloy_transport_http/
--- a/crates/transport/README.md
+++ b/crates/transport/README.md
@@ -23,8 +23,7 @@ specific Ethereum endpoints.
 Alloy maintains the following transports:
 
 - [alloy-transport-http]: JSON-RPC via HTTP.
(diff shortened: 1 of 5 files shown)


cakevm:
Yes, you are right. I removed the `DefaultOnNull` and will patch the json if required. Second as you suggesting I will load the config only when required.

crates/transport/README.md
```diff
@@ -23,8 +23,8 @@ specific Ethereum endpoints.
 Alloy maintains the following transports:
 
 - [alloy-transport-http]: JSON-RPC via HTTP.
-- [alloy-transport-ws]: JSON-RPC via Websocket, supports pubsub via
-    [alloy-pubsub].
+- [alloy-transport-ws]: JSON-RPC via Websocket, supports pubsub via.
+- [alloy-pubsub]: JSON-RPC publish-subscribe tower service.
```
DaniPopes, line 27: this is not right, it's part of the sentence on the line above

cakevm: You are right, cargo check was complaining and I missed the idea. Reverted the change and removed the line break.

cakevm:
I gave the topic a fresh look and mixed up to different types here. Sorry for that, this PR does not makes sense.

There is op-geth the type, used for the json, is defined [here](https://github.com/ethereum-optimism/op-geth/blob/optimism/core/genesis.go#L58):
```go
type Genesis struct {
	Config     *params.ChainConfig `json:"config"`
	Nonce      uint64              `json:"nonce"`
...
type ChainConfig struct {
	ChainID *big.Int `json:"chainId"`
...
```
It looks like config can be null here. I still looking for the reason why there is `"config": null" at all.

Second there is Chain in superchain-registry [here](https://github.com/ethereum-optimism/superchain-registry/blob/main/ops/internal/config/chain.go#L61) used for the toml:
```go
type Chain struct {
	Name                 string              `toml:"name"`
	PublicRPC            string              `toml:"public_rpc"`
...
```

I am considering proposing to allow `None` for the config field and introduce a new struct for the superchain chain config. When I found the reason for the `null`.

PR #2085, opened 2025-02-19
Feat: Declared `BlockBody` trait

### LINKED ISSUE #2083: [Feature] Declare `BlockBody` trait
### Component

consensus, eips, genesis

Declare trait with utility methods
```rust
pub trait BlockBody {
    fn has_eip4844_transactions(&self) -> bool;
    ...
}
```
from
https://github.com/alloy-rs/alloy/blob/9975e8a31ecf3871cee0568e2acf89ef7b271baa/crates/consensus/src/block/mod.rs#L243-L253

### Additional context

_No response_

PoulavBhowmick03:
## Motivation

Fixes #2083

## Solution

- Added a `BlockBodyTrait` in crates/consensus/src/block/traits.rs and it's implementations and then imported it in 
`crates/consensus/src/block/mod.rs` and `crates/consensus/src/lib.rs`

- I have not implemented the transaction associated type, after reading the comment in the issue page
- All the tests pass

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/traits.rs
+++ b/crates/consensus/src/block/traits.rs
@@ -1,6 +1,6 @@
 use super::Block;
+use crate::block::BlockBody;
 use alloy_eips::eip4895::Withdrawals;
-
 /// A trait for ethereum like blocks.
 pub trait EthBlock {
     /// Returns reference to withdrawals in the block if present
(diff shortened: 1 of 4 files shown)


crates/consensus/src/block/traits.rs
```diff
@@ -12,3 +12,21 @@ impl<T, H> EthBlock for Block<T, H> {
         self.body.withdrawals.as_ref()
     }
 }
+
+/// A trait for Ethereum block body utilities.
+pub trait BlockBodyTrait {
```
mattsse, line 17: can we name this `EthBody` perhaps?

mattsse (commented):
this is a good start,

will need to think about this a bit longer before I'll merge this

crates/consensus/src/block/traits.rs
PoulavBhowmick03: Sure!

mattsse (commented):
we decided against this for now, because this would just conflict with what we have in reth

PR #2086, opened 2025-02-20
feat(`rpc-types`): decode log from receipt

yash-atreya:
## Motivation

Currently, we don't have any direct way to decode the receipt logs into a typed `SolEvent` log. See: https://t.me/ethers_rs/43321

## Solution

- Add a `fn decoded_log<E: SolEvent>(&self) -> Option<Log<E>>` to `TransactionReceipt`. 

This enables users to do this:

```rust
let receipt = counter.increment().send().await.unwrap().get_receipt().await.unwrap();
let log = receipt.decoded_log::<Counter::Increment>();
assert!(log.is_some());
```

instead of 

```rust
let receipt = counter.increment().send().await.unwrap().get_receipt().await.unwrap();
let receipt = receipt.into_primitives_receipt();
let logs = receipt.logs();
let increment_log = Counter::Increment::decode_log(&logs[0], false).unwrap();
```

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/receipt.rs
+++ b/crates/rpc-types-eth/src/transaction/receipt.rs
@@ -109,8 +109,8 @@ impl TransactionReceipt {
     ///
     /// Returns None, if none of the logs could be decoded to the provided log type or if there
     /// are no logs.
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense

PR #2087, opened 2025-02-20
chore(engine): Add missing variants for parent beacon block root to `PayloadError`

emhane:
Adds variants for beacon block root presence pre and post cancun to `PayloadError`

Ref https://github.com/paradigmxyz/reth/issues/14604

### DIFF
--- a/crates/contract/src/error.rs
+++ b/crates/contract/src/error.rs
@@ -1,7 +1,7 @@
 use alloy_dyn_abi::Error as AbiError;
 use alloy_primitives::{Bytes, Selector};
 use alloy_provider::PendingTransactionError;
-use alloy_sol_types::SolInterface;
+use alloy_sol_types::{SolError, SolInterface};
 use alloy_transport::TransportError;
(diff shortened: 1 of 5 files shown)


emhane:
these variants are needed to replace `VersionSpecificValidationError` here
https://github.com/paradigmxyz/reth/blob/c5df8fbaaf8a7f6101980beff9d94e2d76b8bcee/crates/payload/primitives/src/lib.rs#L241-L254

we could return less specific variants `PayloadError::PreCancunWithCancunFields`/`PayloadError::PostCancunWithoutCancunFields` here instead because the parent beacon block root is not optional in `CancunPayloadFields`, but think it's better to be more specific because the layout of reth's `PayloadOrAttributes` used here is different, instead of making assumptions about how the two types `CancunPayloadFields` and `PayloadOrAttributes` are (and will remain) integrated with each other in reth.

wdyt @mattsse

mattsse (approved):

PR #2088, opened 2025-02-20
fix(engine): Fix BlockHash display message

hvanz:
Hash values displayed in the message are swapped. We "want" the hash computed in execution but "got" the hash coming from consensus.

## PR Checklist

- [ ] ~Added Tests~
- [ ] ~Added Documentation~
- [ ] ~Breaking changes~

### DIFF
--- a/crates/rpc-types-engine/src/error.rs
+++ b/crates/rpc-types-engine/src/error.rs
@@ -51,7 +51,7 @@ pub enum PayloadError {
     #[display("requests present in pre-prague payload")]
     PrePragueBlockRequests,
     /// Invalid payload block hash.
-    #[display("block hash mismatch: want {consensus}, got {execution}")]
+    #[display("block hash mismatch: want {execution}, got {consensus}")]
     BlockHash {
         /// The block hash computed from the payload.
         execution: B256,


crates/rpc-types-engine/src/error.rs
```diff
@@ -51,7 +51,7 @@ pub enum PayloadError {
     #[display("requests present in pre-prague payload")]
     PrePragueBlockRequests,
     /// Invalid payload block hash.
-    #[display("block hash mismatch: want {consensus}, got {execution}")]
+    #[display("block hash mismatch: want {execution}, got {consensus}")]
```
mattsse, line 54: yeah this isn't super obvious which one is which like what is want

so unsure if this is change matches got vs expected

hvanz: Agree it can be confusing. I was getting this error from Reth when sending a payload from a consensus client and, from that perspective, it was clear it was the other way around.

mattsse (approved):
i see now

https://github.com/paradigmxyz/reth/blob/0611e7ad90f7a0bab72a7d8b79bea7a1b8fa5152/crates/optimism/payload/src/validator.rs#L60-L63

PR #2092, opened 2025-02-21
fix(`provider`): fill txs on `eth_call` ops

### LINKED ISSUE #1942: [Bug] Set `from` via filler for `eth_call` and `eth_estimateGas`
### Component

provider, pubsub

0.9.2

### Operating System

Linux

### Describe the bug

hey there i am having an issue with transaction simulation failing (where if i just skip simulation with the exact same payload and parameters it works)

seems like it has something to do with doing native transfers during my call (simulation) because I get back that I got this error execution reverted, data: "0x9fb7cfe2" which corresponds to this spot in my SC 

https://github.com/kassandraoftroy/stealth-gas-contracts/blob/main/src/StealthGasStation.sol#L63

tx payload:
```
0x589f5407000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000002386f26fc100000000000000000000000000000000000000000000000000000000000000000001000000000000000000000000a4b0855e70013ea969932c7614f34db012cb00d40000000000000000000000000000000000000000000000000000000000000000
```

on contract address: `0x943285f1a29281e59514fF35Dc16E5a14E123a27`

network: holesky testnet, using an alchemy rpc `wss://eth-holesky.g.alchemy.com/v2/xxxxxxx`

block number: ~3196935 (or current)

source code doing simulation: https://github.com/kassandraoftroy/stealth-gas-backend/blob/main/src/http_server.rs#L164-L186

caller/signer of tx being simulated: `0x4B5BaD436CcA8df3bD39A095b84991fAc9A226F1`

here you can see that on tenderly simulation the same tx succeeds! https://www.tdly.co/shared/simulation/1efe08e3-a571-4511-82e1-46df3d3bbf05

my suspicion: it may be a bug in simulation where native transfers fail for some reason (maybe forwarding gas? maybe thinking the EOA can't receive eth? idk)

let me know what you think!

yash-atreya:
## Motivation

Closes #1942

## Solution

- Overrides `call` and `estimate_gas` in `FillProvider`.
- Introduces `prepare_res` builder method on `EthCall` that stores the result of the `filler.prepare_call` in the future and swaps the unfilled tx with the filled tx while polling 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/provider/src/fillers/wallet.rs
```diff
@@ -108,6 +108,16 @@ where
...
         Ok(())
     }
+
+    fn prepare_call_sync(
+        &self,
+        tx: &mut <N as Network>::TransactionRequest,
+    ) -> TransportResult<()> {
+        if tx.from().is_none() {
+            tx.set_from(self.wallet.default_signer_address());
+        }
```
mattsse, line 118: smol nit here, can we call this sync fn in the async variant, then we don't need this logic twice

crates/provider/src/provider/eth_call/caller.rs
```diff
@@ -64,7 +64,8 @@ where
     }
 }
 
-fn provider_rpc_call<N: Network, Resp: RpcRecv>(
+/// Returns a [`ProviderCall::RpcCall`] from the provided method and [`EthCallParams`].
+pub fn provider_rpc_call<N: Network, Resp: RpcRecv>(
```
mattsse, line 68: do we need this pub or can this be pub(crate) ?

crates/provider/src/fillers/mod.rs
```diff
@@ -301,6 +315,30 @@ where
         self.inner.root()
     }
 
+    fn call<'req>(&self, tx: &'req N::TransactionRequest) -> EthCall<'req, N, Bytes> {
+        let call_filler = match self.prepare_call(tx.clone()) {
```
mattsse, line 319: I think we should revisit this borrow here on the API level because this is a bit horrible

crates/provider/src/fillers/mod.rs
```diff
@@ -301,6 +315,30 @@ where
...
+            }
+        };
+        EthCall::call(call_filler, tx).block(BlockNumberOrTag::Pending.into())
+    }
+
+    fn estimate_gas<'req>(&self, tx: &'req N::TransactionRequest) -> EthCall<'req, N, U64, u64> {
+        let call_filler = match self.prepare_call(tx.clone()) {
+            Ok(filled_tx) => EthCallFiller::new(filled_tx, self.weak_client()),
+            Err(e) => {
+                tracing::error!(?e, "preparing call for eth_call, using tx without filling");
```
mattsse, line 333: this silently swallows the error here.

wouldn't it be possible to move this entire logic into the `trait Caller` impl?

like a boxedfuture that does:

https://github.com/alloy-rs/alloy/blob/86f5895c55e9daeab5a19eb422b2233f78952341/crates/provider/src/provider/eth_call/caller.rs#L75-L77

but https://github.com/alloy-rs/alloy/blob/86f5895c55e9daeab5a19eb422b2233f78952341/crates/provider/src/provider/prov_call.rs#L36-L36

mattsse (changes requested):
nits and some questions

crates/provider/src/fillers/wallet.rs
yash-atreya: [fcfe4af](https://github.com/alloy-rs/alloy/pull/2092/commits/fcfe4af5c635f8681a430f5b30932315add68be2)

crates/provider/src/fillers/mod.rs
yash-atreya: [992faf2](https://github.com/alloy-rs/alloy/pull/2092/commits/992faf2a240f20666db4a93b52500662c0c62785)

crates/provider/src/provider/eth_call/caller.rs
yash-atreya: [992faf2](https://github.com/alloy-rs/alloy/pull/2092/commits/992faf2a240f20666db4a93b52500662c0c62785)

crates/provider/src/fillers/mod.rs
yash-atreya: I tried moving the entier logic into the `Caller` impl by storing the `self.filler` from `FillProvider` in `EthCallFiller`.
e.g 
```rust
struct EthCallFiller<F: TxFiller<N>, N> {
filler: F
}
```
But doing this requires adding the `'static` bound to `F: TxFiller<N>` on the `FillProvider` (and up the chain into `JoinFill` as well) as `EthCall` expects `Caller` to be `'static`.

yash-atreya: [a19e906](https://github.com/alloy-rs/alloy/pull/2092/commits/a19e90657dd9ec4369ad1d85c3a2f6f6a90a19d1)

crates/provider/src/fillers/mod.rs
```diff
@@ -329,7 +339,9 @@ where
     }
 
     fn call<'req>(&self, tx: &'req N::TransactionRequest) -> EthCall<'req, N, Bytes> {
-        self.inner.call(tx)
+        let prepare_res = self.prepare_call(tx.clone());
```
mattsse, line 342: this seems okay, here but the alternative is moving this inside the ethcallfiller and use the async call there.

because isn't this now just the same as calling self.inner.call?

actually perhaps it's easier that way for now, so let's simplify this drastically and only add this to relevant calls and not introduce the ethcallfiller type.

but the proper solution here would be to move this into the future I think

mattsse (changes requested):

yash-atreya:
ci failures are unrelated

crates/provider/src/fillers/mod.rs
```diff
@@ -329,7 +338,8 @@ where
     }
 
     fn call<'req>(&self, tx: &'req N::TransactionRequest) -> EthCall<'req, N, Bytes> {
-        self.inner.call(tx)
+        let prepare_res = self.prepare_call(tx.clone()).ok();
```
mattsse, line 341: let's just ignore the error here for now

I think all of this needs some design changes but we can make his work without introducing the `filled_tx` fn and field

mattsse (changes requested):

crates/provider/src/fillers/mod.rs
yash-atreya: Resolved by #2127

crates/provider/src/fillers/mod.rs
yash-atreya: [e16d528](https://github.com/alloy-rs/alloy/pull/2092/commits/e16d5282b6cc7d2ae5b30152d967e99ee3dd6b46)

mattsse (approved):

Issue #2051, opened 2025-02-13
[Feature] Implement Transaction for either


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

there are usecase where we need to handle two different tx types:

https://github.com/paradigmxyz/reth/pull/14472/files#diff-068a87546e97f457bc31013fabf9a116441c1050fb2f84cd6c3c49eaae5cce13R63-R68

we can solve this by implementing it for `Either` from the `either` crate:

https://github.com/alloy-rs/alloy/blob/f717c9e97f238b1376c615ce97632cf243244941/crates/signer/Cargo.toml#L27-L27

## TODO
* impl Transaction for Either
* lift either dep to workspace dep

neithanmo:
I can implement this @mattsse

mattsse:
@neithanmo it could be that it's more useful to roll our own `Either` type for this

so we can impl more logic natively

neithanmo:
@mattsse 
Where would this type live? maybe alloy-primitives?

mattsse:
i think we should put this in `consensus/src/transaction/either.rs`

> There is [an either](https://github.com/alloy-rs/alloy/blob/main/crates/network/src/any/either.rs) module

hmm, this should really be called any

neithanmo:
Ok, I will complete it, this weekend

PR #2097, opened 2025-02-21
Add custom Either type for transactions

neithanmo:
## Motivation
See #2051

## Solution
A new module called either is added, this is useful to handle different transactions types
as it implements `Transaction` trait. The Either type here is a bit similar to [this](https://crates.io/crates/either), however a custom one might be more useful to customize.

## PR Checklist

- [x ] Added Tests
- [x ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/either.rs
+++ b/crates/consensus/src/transaction/either.rs
@@ -1,0 +1,213 @@
+/// A type that represents exactly one of two possible variants: `Left(L)` or `Right(R)`.
+///
+/// `Either` is completely neutral with respect to the meaning of each variants.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Either<L, R> {
+    /// The left variant.
+    Left(L),
+    /// The right variant.
+    Right(R),
+}
+
+impl<L, R> Either<L, R> {
+    /// Returns `true` if this is a `Left` variant.
+    #[inline]
+    pub fn is_left(&self) -> bool {
+        matches!(self, Self::Left(_))
+    }
+
+    /// Returns `true` if this is a `Right` variant.
+    #[inline]
+    pub fn is_right(&self) -> bool {
+        matches!(self, Self::Right(_))
+    }
+
+    /// Converts from `Either<L, R>` to `Option<L>`.
+    ///
+    /// Returns `Some` if this is a `Left` variant, otherwise returns `None`.
+    #[inline]
+    pub fn left(self) -> Option<L> {
+        match self {
+            Self::Left(l) => Some(l),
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
we can simplify this a lot if we reuse the existing `Either` type from the either crate and implement our Tx traits for that type

neithanmo:
> we can simplify this a lot if we reuse the existing `Either` type from the either crate and implement our Tx traits for that typeD

@mattsse 
I think, the wrapper approach sacrifices Rust's pattern matching ergonomics:
```
// With wrapper (cumbersome):
if let Some(tx) = self.as_left() {
    tx.authorization_list()
} else if let Some(tx) = self.as_right() {
    tx.authorization_list()
} else {
    unreachable!()
}
```
With direct Either enum (clean):
```
match tx {
    Either::Left(t) => t.authorization_list(),
    Either::Right(t) => t.authorization_list(),
}
```

While reusing either crate saves implementation code, it significantly impacts usability. Unless we make the inner field public (breaking encapsulation) or implement our own enum directly, we lose the clean pattern matching that makes working with sum types handy and easy to use and clean.

DaniPopes:
What do you mean? https://docs.rs/either is exactly the same as what you defined

neithanmo:
>What do you mean? https://docs.rs/either is exactly the same as what you defined

Yep, it is the same to take advantage of the ergonomics of sum types.
defining it  as a wrapper:

`pub enum Either(pub either::Either);`

would require an indirection layer, assuming inner field is defined as a public:

```
match tx_either.0 {
       either::Either::Left(..)
       either::Either::Right(..)
}
```

if defined as private would be even less ergonomic.

Alright, i will change this to use a wrapper type with the inner field public

neithanmo:
What would be a good name here? 
- Either could cause name collision with the original either crates in users code
- TransactionEither seems good but?

What do you think? @DaniPopes @mattsse

DaniPopes:
You shouldn't need a wrapper type at all, all the traits you want to implement are defined in this repository so you can just implement them on the foreign type

neithanmo:
> You shouldn't need a wrapper type at all, all the traits you want to implement are defined in this repository so you can just implement them on the foreign type

Done !!

crates/signer/Cargo.toml
```diff
@@ -24,7 +24,7 @@ workspace = true
 
 [dependencies]
 alloy-primitives = { workspace = true, features = ["std", "k256"] }
-either = "1.13.0"
+either = { workspace = true }
```
mattsse, line 27:
```suggestion
either.workspace
```

mattsse (approved):

PR #2098, opened 2025-02-22
Move WithEncoded helper type to alloy

programskillforverification:
## Motivation

solve [#14634](https://github.com/paradigmxyz/reth/issues/14634)


mattsse (approved):
great, ty

PR #2099, opened 2025-02-22
chore: delegate provider fns in fill provider

mattsse:
this previously always used the default delegation

@onbjerg

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -7,6 +7,8 @@
 //! [`Provider`]: crate::Provider
 
 mod chain_id;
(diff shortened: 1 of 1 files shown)


DaniPopes:
we can maybe have a macro for this where you just pass in the name of the fn and it delegates

mattsse:
yeah, the problem with that is that we usually need to filter out a few calls, so we with a macro that must handle function names.
now that we have a provider that does full delegation, it's easier to just copy paste this and then remove the calls if any

DaniPopes:
right but then when adding new endpoints that must be overridden this is easily forgotten

PR #2103, opened 2025-02-22
feat: add more transaction conversion helpers

mattsse:
adds various helpers to convert from envelope -> Signed and for the rpc tx variant


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -25,12 +28,9 @@ pub use receipt::TransactionReceipt;
...
 pub use request::{TransactionInput, TransactionRequest};
 
-pub use alloy_consensus::{
-    Receipt, ReceiptEnvelope, ReceiptWithBloom, Transaction as TransactionTrait,
-};
-pub use alloy_consensus_any::AnyReceiptEnvelope;
-
 /// Transaction object used in RPC
+///
+/// This
```
DaniPopes, line 33: ..

DaniPopes (approved):

PR #2104, opened 2025-02-22
fix: drop geth's stderr handle

### LINKED ISSUE #2091: [Bug] `GethInstance` dropping connections with the underlying node
### Component

node-bindings

v0.11.1

### Operating System

Linux

### Describe the bug

# Description

I have a test suite which needs to setup 2 sets of wallets (each set has both EOAs and Smart Contract Accounts) after the generation of their related seeds.

The **1st** set of wallets is created in a certain amount of time `X`, and if the seed is related to the a Smart Contract Account then the deployment through a `GethInstance` is done right after the generation it.

The **2nd** set of wallets requires an amount of time `Y > X`, since the seed generation process needs more computation time, but the procedure then follows the same logic ad the previous case (if SCA, deploy).

These sets of wallets are generated (and eventually deployed) sequentially, first going with set `1` and then with `2`.

While for set `1` everything goes smoothly, for set `2` it always happens that when I reach the point in which I need to deploy a SCA I got the following error message from the underlying `GethInstance`:

```
 Unable to send raw request: server returned an error response: error code -32002: request timed ou
```

This error comes from checking the status of the underlying `GethInstance` by sending a raw request to it for the `net_listening`  RPC method. Basically, I create the connection to the `GethInstance` and I do a first check as mentioned, then I start generating the seeds (which requires ~20s) and then I do further checks with `net_listening` each time I need to deploy a SCA.

I ran the same test flow with both an `AnvilInstance` and a `GethInstance`, but for Anvil it always goes to completion while for Geth doesn't.
More over, I have an implementation of `geth` command binding (let's call it `GethD`) that tears up the geth instance and peforms the operations mentioned before, and also with this one it doesn't fail at all.

To me, it looks like the first connection to the `GethInstance` works good during the init phase while after waiting for the process of generating the seeds the connection drops for some reason and so that error is returned.

# Details
For all the 3 cases mentioned (`AnvilInstance`, `GethInstance` and my own `GethD` binding) I have the following setup:
- `Dev` mode
- `chain_id` = 1337 (ofc)
- `block_time` = 1s

mattsse:
same as #1985 but for stderr because geth logs everything to stderr

closes #2091

@gnapoli23 could you perhaps try this?


gnapoli23:
@mattsse 

Not working. I checked with a simple test:

```rust
#[tokio::test]
async fn test_geth_connection() {
    let geth = Geth::new().dev().spawn();
    let provider = ProviderBuilder::new().on_http(geth.endpoint_url());

    let _ = provider.get_block_number().await.unwrap();
}
```

It's unable to get the connection through the provider:
```rust
thread 'http::test_geth_connection' panicked at crates/provider/tests/it/http.rs:9:47:
called `Result::unwrap()` on an `Err` value: Transport(Custom(reqwest::Error { kind: Request, url: "http://localhost:45635/", source: hyper_util::client::legacy::Error(Connect, ConnectError("tcp connect error", Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })) }))
```

yash-atreya (approved):

Issue #2105, opened 2025-02-22
[Feature] Add builder type for StateOverrides


mattsse:
### Component

rpc

### Describe the feature you would like

this type is an alias making it harder to configure

https://github.com/alloy-rs/alloy/blob/1495d10467cc5d47040cae16b21188e011bbcdc0/crates/rpc-types-eth/src/state.rs#L11-L11

we should add a helper new type here that works like a builder `StateOverridesBuilder(Overrides)` and exposes functions to insert new overrides easily

varun-doshi:
I'd like to take this

PR #2106, opened 2025-02-22
feat: create StateOverridesBuilder

varun-doshi:
## Motivation

Ref #2105 
Create a `Builder` for `StateOverride`

## Solution

Creates a struct called `StateOverridesBuilder` to make using `StateOverride` easier

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/state.rs
+++ b/crates/rpc-types-eth/src/state.rs
@@ -7,9 +7,89 @@ use alloy_primitives::{
     Address, Bytes, B256, U256,
 };
 
+/// A StateOverride builder.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct StateOverridesBuilder {
+    overrides: StateOverride,
+}
+
+impl StateOverridesBuilder {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/state.rs
```diff
@@ -7,6 +7,71 @@ use alloy_primitives::{
...
+    /// Create a new StateOverridesBuilder.
+    pub fn new(map: AddressHashMap<AccountOverride>) -> Self {
+        Self { overrides: map }
+    }
+
+    /// Adds an account override for a specific address.
+    pub fn insert(mut self, address: Address, account_override: AccountOverride) -> Self {
+        self.overrides.insert(address, account_override);
+        self
+    }
```
mattsse, line 26: I'd like to rename this to append, and then we also need `fn extend` that accepts in impl intoiter

mattsse (changes requested):
great,

smol suggestions

mattsse (approved):

PR #2110, opened 2025-02-23
feat(serde): Optimize `WithOtherFields` deserialization

### LINKED ISSUE #2065: [Feature] Optimize WithOtherFields deserialization
### Component

rpc, serde

we currently use this workaround:

https://github.com/alloy-rs/alloy/blob/3866c39fe18681a80f776899555497d4df5328c5/crates/serde/src/other/mod.rs#L235-L257

however we can likely achieve the same thing if the have a manual deserialize impl + visitor and use serde's helper

see also expanded code

### Additional context

_No response_

PanGan21:
## Motivation

Closes #2065 

- Manually deserializes `WithOtherFields`
- Implements `Visitor`


PanGan21:
Closing, my idea was not valid finally

mattsse (changes requested):
I've been thinking about this a bit and this isn't trivial and likely not doable without additional serde+alloc effort

this solution has additional serde overhead, but what we can try is optimize this step

```
serde_json::to_value(&helper.inner).map_err(serde::de::Error::custom)?
```

but writing a custom visitor that serializes the inner value but only remove the fields in `serialize_map` for example

PanGan21:
> I've been thinking about this a bit and this isn't trivial and likely not doable without additional serde+alloc effort
> 
> this solution has additional serde overhead, but what we can try is optimize this step
> 
> ```
> serde_json::to_value(&helper.inner).map_err(serde::de::Error::custom)?
> ```
> 
> but writing a custom visitor that serializes the inner value but only remove the fields in `serialize_map` for example

I might be missunderstanding this since I thought that this was what I implemented. Maybe some more pointers will help me figuring out what is required

DaniPopes:
This PR as it is right now is the same as the derived implementation, just written differently/more verbosely; I also spent some time looking into this and can't see an easy solution, and I think the current implementation is fine, so I would close this PR and issue. @mattsse

PanGan21:
Closing for now, but I am interested on the solution if any new pr will be opened

mattsse:
thanks @PanGan21 

sorry about this...

PR #1086, opened 2024-07-21
feat: Add a TraceRpc helper type

moricho:
## Motivation
Closes: https://github.com/alloy-rs/alloy/issues/820
This allows users to set trace types for trace rpc calls using the builder pattern as with `RpcWithBlock`.

## Solution
Adds `TraceRpcWithBlock`, allowing users to configure TraceType for trace RPC calls using the builder pattern. This includes `RpcWithBlock` as a member and implements Deref, enabling the setting of block kind arguments as with RpcWithBlock.

On the `RpcWithBlock` side, a `trace_types` field is added, which is enabled when the trace-api feature is on. The `trace_types` is passed by `TraceRpcWithBlock` and merged into the request parameters when calling `RpcWithBlockFut::poll_preparing`.

### Proposed API

```rs
#[derive(Debug, Clone)]
pub struct TraceRpcWithBlock<T, Params, Resp, Output = Resp, Map = fn(Resp) -> Output>
where
    T: Transport + Clone,
    Params: RpcParam,
    Resp: RpcReturn,
    Map: Fn(Resp) -> Output,
{
    inner: RpcWithBlock<T, Params, Resp, Output, Map>,
    trace_types: HashSet<TraceType>,
}

impl<T, Params, Resp, Output, Map> TraceRpcWithBlock<T, Params, Resp, Output, Map>
where
    T: Transport + Clone,
    Params: RpcParam,
    Resp: RpcReturn,
    Map: Fn(Resp) -> Output + 'static,
{
    /// Set the trace type.
    pub fn trace_type(mut self, trace_type: TraceType) -> Self {
        self.trace_types.insert(trace_type);
        self
    }

    /// Set the trace types.
    pub fn trace_types<I: IntoIterator<Item = TraceType>>(mut self, trace_types: I) -> Self {
        self.trace_types.extend(trace_types);
        self
    }

    /// Set the trace type to "trace".
    pub fn trace(self) -> Self {
        self.trace_type(TraceType::Trace)
    }

    /// Set the trace type to "vmTrace".
    pub fn vm_trace(self) -> Self {
        self.trace_type(TraceType::VmTrace)
    }

    /// Set the trace type to "stateDiff".
    pub fn state_diff(self) -> Self {
        self.trace_type(TraceType::StateDiff)
    }
}

impl<T, Params, Resp, Output, Map> Deref for TraceRpcWithBlock<T, Params, Resp, Output, Map>
where
    T: Transport + Clone,
    Params: RpcParam,
[...]

PR #2119, opened 2025-02-26
feat(`provider`)!: trace api builder

### LINKED ISSUE #820: Add a TraceRpc helper type for configuring the trace kind argument
cc @mattsse my guess is we could also return a builder that's called `TraceRpcWithBlock` or similar? since we default to `["trace"]` most times?

_Originally posted by @gakonst in https://github.com/alloy-rs/alloy/pull/818#discussion_r1624830808_

similar to 

https://github.com/alloy-rs/alloy/blob/2bff1dd43fb752c15f6179f2bfcebf4c3e0a227d/crates/provider/src/provider/with_block.rs#L164-L164

yash-atreya:
## Motivation

Closes #820 
Supersedes #1086 

## Solution

- Introduce `TraceBuilder` 
- `TraceBuilder` works over `WithBlockInner` to accommodate `ProviderCall`.
- Uses `TraceParams` to handle serialization of params for different types of `trace_*` methods.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -95,6 +95,7 @@ http-body-util.workspace = true
 http.workspace = true
 alloy-rpc-types-engine = { workspace = true, features = ["jwt"] }
 ci_info.workspace = true
+similar-asserts.workspace = true
 
 [features]
 default = ["reqwest", "reqwest-default-tls", "trace-api"]
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -98,7 +98,7 @@ ci_info.workspace = true
 similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls", "trace-api"]
+default = ["reqwest", "reqwest-default-tls"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
 reqwest = [
     "dep:reqwest",
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -98,7 +98,7 @@ ci_info.workspace = true
 similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "trace-api"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
 reqwest = [
     "dep:reqwest",
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -97,7 +97,7 @@ alloy-rpc-types-engine = { workspace = true, features = ["jwt"] }
 ci_info.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "trace-api"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
 reqwest = [
     "dep:reqwest",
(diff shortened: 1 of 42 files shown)


crates/provider/Cargo.toml
```diff
@@ -95,9 +95,10 @@ http-body-util.workspace = true
 http.workspace = true
 alloy-rpc-types-engine = { workspace = true, features = ["jwt"] }
 ci_info.workspace = true
+similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "trace-api"]
```
mattsse, line 101: why is this now default?

crates/provider/src/provider/with_block.rs
```diff
@@ -142,46 +149,65 @@ where
...
+{
+    fn block_id(mut self, block_id: BlockId) -> Self {
+        self.block_id = block_id;
+        self
+    }
+}
+
+/// Trait to set the block id for a call. This should be implemented by any builder types that
+/// require a block id. e.g [`RpcWithBlock`].
+pub trait WithBlock: Sized {
```
mattsse, line 168: where is this useful?

this now forces users to import an extra trait?

mattsse (changes requested):
we can do all of this without introducing a new trait that simplifies 4 functions

crates/provider/Cargo.toml
yash-atreya: oversight, sorry [14c8ad8](https://github.com/alloy-rs/alloy/pull/2119/commits/14c8ad834d6528ba2bc9c7a8dde39a6eac957265)

crates/provider/src/ext/trace/with_block.rs
```diff
@@ -0,0 +1,308 @@
...
+                    trace_params(&method, params, block_id, trace_types.clone())
+                });
+                ProviderCall::RpcCall(inner)
+            }
+            WithBlockInner::ProviderCall(get_call) => get_call(self.block_id),
+        }
+    }
+}
+
+fn trace_params<Params: RpcSend>(
```
mattsse, line 185: this should be a new fn

crates/provider/src/ext/trace/with_block.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        // Get mutable reference to array
+        let arr = ser.as_array_mut().unwrap();
+
+        // Add trace_types if present
+        if let Some(trace_types) = &self.trace_types {
+            let trace_types =
+                serde_json::to_value(trace_types).map_err(serde::ser::Error::custom)?;
+            arr.push(trace_types);
+        }
```
mattsse, line 251: we need to serialize this as array, right?

can we use the serializer type here instead?

I believe this should be possible with serialize_tuple?

mattsse (changes requested):

crates/provider/src/ext/trace/with_block.rs
yash-atreya: [4bec4f3](https://github.com/alloy-rs/alloy/pull/2119/commits/4bec4f33cb072f798b2fdb0387809fce5522dc49)

crates/provider/src/ext/trace/with_block.rs
yash-atreya: [bfc2711](https://github.com/alloy-rs/alloy/pull/2119/commits/bfc27110e75ed5e5fd9c85444363005243b39305)

mattsse (approved):
lgtm!

PR #2120, opened 2025-02-27
feat: introduce `IntoWallet` to pass signer directly to `ProviderBuilder`

### LINKED ISSUE #2094: Make it easier to configure a wallet in a provider
this currently requires

https://github.com/alloy-rs/alloy/blob/38060a65949bbb6b6dcc52f54fdeeeb0b9434943/crates/provider/src/builder.rs#L253-L258

```
 let wallet = EthereumWallet::from(signer);
```

which is usually initialized like this:

https://github.com/alloy-rs/alloy/blob/38060a65949bbb6b6dcc52f54fdeeeb0b9434943/crates/provider/src/provider/trait.rs#L1837-L1839

it would be easier if this could be directly initialized which a Privatekeysigner.

to make this more ergonomic, we can introduce a helper trait that serves as a specialization for PrivateKeysigner so we can accept something like impl IntoWallet or similar.

yash-atreya:
…rovider builder

## Motivation

Closes #2094 

## Solution

WIP

- Introduce the `IntoWallet` trait, to be implemented by signers to return a wallet that implements `NetworkWallet`.
- Changes bounds on `WalletFiller` to `IntoWallet`. 
- Retains the type of signer `.wallet(signer)` was called with.
- Note: Some cloning overhead while calling `into_wallet`.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/ethereum/wallet.rs
+++ b/crates/network/src/ethereum/wallet.rs
@@ -1,7 +1,7 @@
 use crate::{AnyNetwork, AnyTxEnvelope, AnyTypedTransaction, Network, NetworkWallet, TxSigner};
 use alloy_consensus::{SignableTransaction, TxEnvelope, TypedTransaction};
 use alloy_primitives::{map::AddressHashMap, Address, PrimitiveSignature as Signature};
-use std::sync::Arc;
+use std::{fmt::Debug, sync::Arc};
 
 use super::Ethereum;
 
@@ -167,5 +167,29 @@ impl NetworkWallet<AnyNetwork> for EthereumWallet {
             )),
             _ => Err(alloy_signer::Error::other("cannot sign UnknownTypedTransaction")),
         }
+    }
+}
+
+/// A trait for converting a signer into a [`NetworkWallet`].
+pub trait IntoWallet<N: Network>: Send + Sync + Debug {
+    /// The wallet type for the network.
+    type NetworkWallet: NetworkWallet<N>;
+    /// Convert the signer into a wallet.
+    fn into_wallet(&self) -> Self::NetworkWallet;
+}
+
+impl IntoWallet<Ethereum> for EthereumWallet {
+    type NetworkWallet = Self;
+
+    fn into_wallet(&self) -> Self::NetworkWallet {
+        self.clone()
+    }
+}
+
+impl IntoWallet<AnyNetwork> for EthereumWallet {
+    type NetworkWallet = Self;
+
+    fn into_wallet(&self) -> Self::NetworkWallet {
+        self.clone()
     }
 }
--- a/crates/provider/src/fillers/wallet.rs
+++ b/crates/provider/src/fillers/wallet.rs
@@ -1,6 +1,8 @@
+use std::fmt::Debug;
+
 use crate::{provider::SendableTx, Provider};
 use alloy_json_rpc::RpcError;
-use alloy_network::{Network, NetworkWallet, TransactionBuilder};
+use alloy_network::{IntoWallet, Network, NetworkWallet, TransactionBuilder};
 use alloy_transport::TransportResult;
 
 use super::{FillerControlFlow, TxFiller};
@@ -53,7 +55,7 @@ impl<W> WalletFiller<W> {
 impl<W, N> TxFiller<N> for WalletFiller<W>
 where
     N: Network,
-    W: NetworkWallet<N> + Clone,
+    W: IntoWallet<N> + Clone,
 {
     type Fillable = ();
 
@@ -71,7 +73,7 @@ where
     fn fill_sync(&self, tx: &mut SendableTx<N>) {
         if let Some(builder) = tx.as_mut_builder() {
             if builder.from().is_none() {
-                builder.set_from(self.wallet.default_signer_address());
+                builder.set_from(self.wallet.into_wallet().default_signer_address());
             }
         }
     }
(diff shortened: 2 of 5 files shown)


crates/network/src/ethereum/wallet.rs
```diff
@@ -169,3 +169,27 @@ impl NetworkWallet<AnyNetwork> for EthereumWallet {
         }
     }
 }
+
+/// A trait for converting a signer into a [`NetworkWallet`].
+pub trait IntoWallet<N: Network>: Send + Sync + Debug {
+    /// The wallet type for the network.
+    type NetworkWallet: NetworkWallet<N>;
+    /// Convert the signer into a wallet.
+    fn into_wallet(&self) -> Self::NetworkWallet;
```
mattsse, line 178: his should consume the type, the idea here is that we can use this helper to accept multiple `IntoWallet` types on the builder, like private key, ethereumwallet or anything that is networkwallet

crates/network/src/ethereum/wallet.rs
```diff
@@ -169,3 +169,27 @@ impl NetworkWallet<AnyNetwork> for EthereumWallet {
...
+impl IntoWallet<Ethereum> for EthereumWallet {
+    type NetworkWallet = Self;
+
+    fn into_wallet(&self) -> Self::NetworkWallet {
+        self.clone()
+    }
+}
+
+impl IntoWallet<AnyNetwork> for EthereumWallet {
+    type NetworkWallet = Self;
```
mattsse, line 190: we need a blanket impl here otherwise a user must always impl this trait if they use a custom networkwallet which isn't ideal

mattsse (changes requested):

crates/network/src/ethereum/wallet.rs
yash-atreya: [17fbc8c](https://github.com/alloy-rs/alloy/pull/2120/commits/17fbc8c997176585e21acf73427394c73b925d29)

crates/network/src/ethereum/wallet.rs
yash-atreya: [17fbc8c](https://github.com/alloy-rs/alloy/pull/2120/commits/17fbc8c997176585e21acf73427394c73b925d29)

crates/provider/src/builder.rs
```diff
@@ -393,8 +386,22 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
     }
 }
 
+impl<L, F, N: Network> ProviderBuilder<L, F, N> {
+    /// Add a wallet layer to the stack being built.
+    ///
+    /// See [`WalletFiller`].
```
mattsse, line 392: can we add an example here for privatekey?

crates/provider/src/fillers/wallet.rs
```diff
@@ -27,30 +29,43 @@ use super::{FillerControlFlow, TxFiller};
 /// # }
 /// ```
 #[derive(Clone, Debug)]
-pub struct WalletFiller<W> {
+pub struct WalletFiller<W, N: Network> {
```
mattsse, line 32: why do we need this change now?

crates/provider/src/builder.rs
```diff
@@ -393,8 +386,22 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
...
+    #[allow(clippy::type_complexity)]
+    pub fn wallet<W: IntoWallet<N>>(
+        self,
+        wallet: W,
+    ) -> ProviderBuilder<L, JoinFill<F, WalletFiller<W::NetworkWallet, N>>, N> {
+        self.filler(WalletFiller::new(wallet))
```
mattsse, line 398: here we need to convert the wallet into the networkwallet

crates/signer-local/src/lib.rs
```diff
@@ -36,6 +36,22 @@ pub use coins_bip39;
 /// A signer instantiated with a locally stored private key.
 pub type PrivateKeySigner = LocalSigner<k256::ecdsa::SigningKey>;
 
+impl IntoWallet<Ethereum> for PrivateKeySigner {
```
mattsse, line 39: can we move this to the trait def instead?

mattsse, line 39: ah can we not relax this over network? because of foreign trait?

mattsse (changes requested):

crates/provider/src/builder.rs
yash-atreya: [47259e9](https://github.com/alloy-rs/alloy/pull/2120/commits/47259e9f4b4b25e107a56d7150fa576c5ef48645)

crates/provider/src/fillers/wallet.rs
yash-atreya: [94ea00a](https://github.com/alloy-rs/alloy/pull/2120/commits/94ea00ab3658c56fd3efa254a29e59c5934c9c8e)

crates/signer-local/src/lib.rs
yash-atreya: would cause cyclic deps

yash-atreya: Made it cleaner by making Ethereum the default network: [957df23](https://github.com/alloy-rs/alloy/pull/2120/commits/957df2330f143f27f9ba9d8763e7b6eb90d6478c)

crates/provider/src/builder.rs
yash-atreya: [3563b55](https://github.com/alloy-rs/alloy/pull/2120/commits/3563b555a13fa502f3b78c7309a2e63d027bb3f2)

crates/signer-local/src/lib.rs
mattsse: I see, makes sense

crates/provider/src/fillers/wallet.rs
```diff
@@ -45,7 +47,7 @@ impl<W> AsMut<W> for WalletFiller<W> {
 
 impl<W> WalletFiller<W> {
     /// Creates a new wallet layer with the given wallet.
-    pub const fn new(wallet: W) -> Self {
+    pub fn new(wallet: W) -> Self {
```
mattsse, line 50:
```suggestion
    pub const fn new(wallet: W) -> Self {
```

mattsse (approved):
makes sense now

we can't have iml IntoNetworkWallet for where Into EtereumWallet because that would clash with the blanket impl so we have to use the macro, which seems fine

this is also technically breaking, so I'm blocking this

PR #2124, opened 2025-02-28
add `rlp` feature to `full` feature

zemse:
## Motivation

I was writing a basic script to create transaction, sign them and submit to a remote node. However, during the `send_raw_transaction` we need a rlp encoded transaction. I had to do the following:

```
alloy = { version = "0.11", features = ["full", "rlp"] }
```

I assume that `full` feature covers everything that is required for basic ethereum interaction and it appears `rlp` was missing in that. 

## Solution

This PR simply adds `rlp` feature to the `full` feature.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -94,6 +94,7 @@ full = [
     "provider-debug-api",
     "provider-anvil-api",
     "pubsub",
+    "rlp",
 ]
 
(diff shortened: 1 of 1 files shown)


yash-atreya (approved):
makes sense!

Issue #1136, opened 2024-08-08
[Bug] Investigate failing `subscribe_block_ws` and `subscribe_ws_boxed` tests on Windows


zerosnacks:
### Component

provider, pubsub

### Operating System

None

### Describe the bug

Goal of the ticket is to remove the edge case handling added in https://github.com/alloy-rs/alloy/pull/1135

To investigate why `subscribe_block_ws` and `subscribe_ws_boxed` are failing, providing a fix for each

To reproduce run `cargo nextest run --workspace --all-features` on Windows or see failure in CI: https://github.com/alloy-rs/alloy/actions/runs/10299750694/job/28507683340

jenpaff:
fixed by https://github.com/alloy-rs/alloy/pull/1135

PR #2126, opened 2025-02-28
test: enable more tests on windows

DaniPopes:
Closes https://github.com/alloy-rs/alloy/issues/1136

Plus adds some reasons to `#[ignore]`


mattsse (approved):
windows 👎

PR #2127, opened 2025-02-28
feat(`eth-call`)!: rm borrowing from provider api

### LINKED ISSUE #2125: [Feature] Remove borrowing from provider APIs
### Component

provider, pubsub

this is a bit horrible

https://github.com/alloy-rs/alloy/blob/c947f4102fba1b343f34813ed007f80c0cba84d7/crates/provider/src/provider/trait.rs#L255-L255

because this does not work for things like filler where we need to modify the call (setting from field) and also introduces lifetimes in the call type.

we should accept ownership here

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2125 

## Solution

WIP

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/provider/src/provider/eth_call/params.rs
+++ b/crates/provider/src/provider/eth_call/params.rs
@@ -7,9 +7,9 @@ use std::borrow::Cow;
 /// The parameters for an `"eth_call"` RPC request.
 #[derive(Clone, Debug)]
 pub struct EthCallParams<N: Network> {
-    data: Cow<'static, N::TransactionRequest>,
+    data: N::TransactionRequest,
     pub(crate) block: Option<BlockId>,
-    pub(crate) overrides: Option<Cow<'static, StateOverride>>,
+    pub(crate) overrides: Option<StateOverride>,
 }
 
 impl<N> EthCallParams<N>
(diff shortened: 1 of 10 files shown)


crates/provider/src/provider/eth_call/params.rs
```diff
@@ -6,19 +6,19 @@ use std::borrow::Cow;
 
 /// The parameters for an `"eth_call"` RPC request.
 #[derive(Clone, Debug)]
-pub struct EthCallParams<'req, N: Network> {
-    data: Cow<'req, N::TransactionRequest>,
+pub struct EthCallParams<N: Network> {
+    data: Cow<'static, N::TransactionRequest>,
     pub(crate) block: Option<BlockId>,
-    pub(crate) overrides: Option<Cow<'req, StateOverride>>,
+    pub(crate) overrides: Option<Cow<'static, StateOverride>>,
```
DaniPopes, line 12: can drop cows since 'static is not going to happen

mattsse (approved):
imo taking ownership is preferred here, especially now that ethcall is a builder style type

PR #2133, opened 2025-03-01
feat: add conversions for UnknownTxEnvelope

mattsse:
this makes it possible to get an `Either` instance with two different tx types


yash-atreya (approved):
send it!

Issue #1938, opened 2025-01-22
[Feature] Fallback transports


erhant:
### Component

network, json-rpc, rpc, transports

### Describe the feature you would like

I would like to provide a several RPC URLs to provide to the provider so that if one fails, the other shall take its place and so on.

Some implementations in other libraries for reference:
- [Viem `fallback` transport](https://viem.sh/docs/clients/transports/fallback.html)
- [Ethers.js `FallbackProvider`](https://docs.ethers.org/v5/api/providers/other/#FallbackProvider)

yash-atreya:
Closing this as a duplicate of #1760

Issue #1760, opened 2024-12-06
[Feature] `HttpMultiClient<T>` Transport


0xrusowsky:
### Component

rpc, transports

### Describe the feature you would like

I’d like to propose a feature that could enhance the robustness of applications using alloy by implementing a built-in fallback mechanism for the http transport layer. This would help manage scenarios where the primary RPC provider is unavailable, improving reliability without requiring developers to implement custom in-app solutions or gateways.

I see 2 main benefits with such design:
- **"zero-cost" reliability:** by automatically switching to a fallback provider, the app availability increases (while users simply need to inform an extra rpc url when initializing the provider).
- **better dev-ex:** a built-in solution reduces development time for devs that would otherwise implement custom mechanisms + ensures consistency across apps.

Because of that i thought of implementing a new http transport to support multiple clients and potentially multiple request orchestration modes (although we could keep it simple and only use it as a fallback). Here the proposed impl:

```rs
pub struct HttpMultiClient<T> {
   primary: Http<T>,
   fallback: Http<T>,
   mode: MultiClientMode
}

pub enum MultiClientMode {
   Fallback,
   Fastest
}

impl<T> Service<RequestPacket> for HttpMultiClient<T> {
    type Response = ResponsePacket;
    type Error = TransportError;
    type Future = TransportFut<'static>;

    #[inline]
    fn poll_ready(&mut self, _cx: &mut task::Context<'_>) -> task::Poll<Result<(), Self::Error>> {
        task::Poll::Ready(Ok(()))
    }

    #[inline]
    fn call(&mut self, req: RequestPacket) -> Self::Future {
        match self.mode {
          MultiClientMode::Fallback => self.fallback_request(req) // todo
          MultiClientMode::Fastest => self.fastest_request(req)  // todo
       }
    }
}
```

then we could enhance the `ProviderBuilder` so that users would simply do something like:

```rs
//! Example of creating an HTTP provider using the `on_multi_http` method on the `ProviderBuilder`.

use alloy::providers::{Provider, ProviderBuilder};
use eyre::Result;

#[tokio::main]
async fn main() -> eyre::Result<()> {
    // Set up the HTTP transport which is consumed by the RPC client.
    let primary = "https://eth.merkle.io".parse()?;
    let fallback = "https://eth.llamarpc.com".parse()?;

    // Create a provider with the HTTP transport with fallback using the `reqwest` crate.
[...]

PR #2135, opened 2025-03-01
feat: `FallbackLayer` transport

merklefruit:
## Motivation

Having a provider be able to consume a list of transports can increase both availability and performance, both desirable features of many applications that need a connection to Ethereum.

Several issues such as https://github.com/alloy-rs/alloy/issues/1938 and https://github.com/alloy-rs/alloy/issues/1760 outline a feature like this in Alloy. 
Other libraries such as [Viem](https://viem.sh/docs/clients/transports/fallback.html) and [Ethers.js](https://docs.ethers.org/v5/api/providers/other/#FallbackProvider) also have something similar.

## Solution

This PR implements `FallbackLayer` and `FallbackService`, following the usual `tower` abstractions. Most notably, the layer here works with a `Vec<S>` because this service is meant for managing a list of transports.

When sending a request to a provider using this layer, it queries the top ranked `N` transports (configurable with the `active_transport_count` option) concurrently, returning the first successful result. Whenever a result arrives, metrics are updated to keep track of the performance of each transport, using a simple combination of average latency and success rate.

I tried to keep this simple, but there is potential for future work if someone finds it useful: 
- adding a health check task to keep ranks always up to date
- adding a minimum threshold of consistent responses (aka "quorum response")
- adding rate-limiting per transport
- adding circuit-breakers for non-responding transports

## Usage

I've added an example usage binary here: https://github.com/merklefruit/alloy-examples/pull/1, but in short here is how it works:

```rust
// Configure the fallback layer
let fallback_layer = FallbackLayer::default()
    .with_active_transport_count(NonZeroUsize::new(3).unwrap())
    .with_log_transport_rankings(true);

// Define your list of transports to use
let transports = vec![
    Http::new(Url::parse("https://eth.merkle.io/")?),
    Http::new(Url::parse("https://eth.llamarpc.com/")?),
    Http::new(Url::parse("https://ethereum-rpc.publicnode.com/")?),
];

// Apply the FallbackLayer to the transports
let transport = ServiceBuilder::new().layer(fallback_layer).service(transports);
let client = RpcClient::builder().transport(transport, false);
let provider = ProviderBuilder::new().on_client(client);

// use the provider as normal
```

## PR Checklist

Tested on a live [example](https://github.com/merklefruit/alloy-examples/pull/1), as I couldn't add comprehensive unit tests without running into cyclic dependencies in alloy itself.

### DIFF
--- a/crates/transport/src/layers/fallback.rs
+++ b/crates/transport/src/layers/fallback.rs
@@ -21,7 +21,7 @@ const STABILITY_WEIGHT: f64 = 0.7;
 const LATENCY_WEIGHT: f64 = 0.3;
 const DEFAULT_INTERVAL: Duration = Duration::from_secs(10);
 const DEFAULT_SAMPLE_COUNT: usize = 10;
-
+const DEFAULT_ACTIVE_TRANSPORT_COUNT: usize = 3;
 /// Represents performance metrics for a transport.
 #[derive(Debug)]
 struct TransportMetrics {
@@ -456,17 +456,10 @@ pub struct FallbackLayer {
 }
 
 impl FallbackLayer {
-    /// Create a new fallback layer.
-    ///
-    /// The `active_transport_count` parameter controls how many
-    /// transports are used for requests at any one time.
-    pub fn with_active_transport_count(count: usize) -> Self {
-        Self {
-            active_transport_count: count,
-            rank: false,
-            interval: DEFAULT_INTERVAL,
-            sample_count: DEFAULT_SAMPLE_COUNT,
-        }
+    /// Set the number of active transports to use
+    pub fn with_active_transport_count(mut self, count: usize) -> Self {
+        self.active_transport_count = count;
+        self
     }
 
     /// Enable or disable automatic transport ranking
@@ -485,6 +478,17 @@ impl FallbackLayer {
     pub fn with_sample_count(mut self, count: usize) -> Self {
         self.sample_count = count;
         self
+    }
+}
+
+impl Default for FallbackLayer {
+    fn default() -> Self {
+        Self {
+            active_transport_count: DEFAULT_ACTIVE_TRANSPORT_COUNT,
+            rank: false,
+            interval: DEFAULT_INTERVAL,
+            sample_count: DEFAULT_SAMPLE_COUNT,
+        }
     }
 }
 
--- a/crates/transport/src/layers/fallback.rs
+++ b/crates/transport/src/layers/fallback.rs
@@ -21,7 +21,7 @@ const DEFAULT_SAMPLE_COUNT: usize = 10;
 
 /// Represents performance metrics for a transport.
 #[derive(Debug)]
-pub(crate) struct TransportMetrics {
+struct TransportMetrics {
     // Latency history - tracks last N responses
     latencies: VecDeque<Duration>,
     // Success history - tracks last N successes (true) or failures (false)
@@ -132,7 +132,7 @@ impl Default for TransportMetrics {
 
 /// A scored transport that can be ordered in a heap.
 #[derive(Debug, Clone)]
-pub(crate) struct ScoredTransport<S> {
+struct ScoredTransport<S> {
     id: usize, // Unique identifier for the transport
     transport: S,
     metrics: Arc<Mutex<TransportMetrics>>,
@@ -203,10 +203,7 @@ pub struct FallbackService<S> {
     rank: bool,
 }
 
-impl<S> FallbackService<S>
-where
-    S: std::fmt::Debug,
-{
+impl<S> FallbackService<S> {
     /// Create a new fallback provider from a list of transports.
     ///
     /// - The `active_transport_count` parameter controls how many
@@ -261,8 +258,8 @@ where
     S: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
         + Send
         + Clone
-        + 'static
-        + std::fmt::Debug,
+        + std::fmt::Debug
+        + 'static,
 {
     async fn make_request(&self, req: RequestPacket) -> Result<ResponsePacket, TransportError> {
         let mut top_transports = Vec::new();
@@ -400,8 +397,8 @@ where
     S: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
         + Send
         + Clone
-        + 'static
-        + std::fmt::Debug,
+        + std::fmt::Debug
+        + 'static,
 {
     type Response = ResponsePacket;
     type Error = TransportError;
--- a/crates/transport/src/layers/fallback.rs
+++ b/crates/transport/src/layers/fallback.rs
@@ -3,12 +3,15 @@ use std::{
     ops::{Deref, DerefMut},
     sync::{Arc, Mutex},
     task::{Context, Poll},
-    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
+    time::{Duration, Instant},
 };
 
 use alloy_json_rpc::{Request, RequestPacket, ResponsePacket, SerializedRequest};
 use futures::{stream::FuturesUnordered, StreamExt};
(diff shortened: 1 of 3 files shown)


crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,447 @@
+use std::{
+    collections::{BinaryHeap, VecDeque},
+    num::NonZeroUsize,
+    ops::{Deref, DerefMut},
+    sync::{Arc, Mutex},
```
mattsse, line 5: please use parking_lot mutex for this

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,447 @@
...
+
+// Constants for the ranking algorithm
+const STABILITY_WEIGHT: f64 = 0.7;
+const LATENCY_WEIGHT: f64 = 0.3;
+const DEFAULT_SAMPLE_COUNT: usize = 10;
+const DEFAULT_ACTIVE_TRANSPORT_COUNT: usize = 3;
+
+/// Represents performance metrics for a transport.
+#[derive(Debug)]
+struct TransportMetrics {
```
mattsse, line 25: can we move this below the main type

makes it easier to navigate if the main type is top of file

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,447 @@
...
+}
+
+impl<S> FallbackService<S> {
+    /// Create a new fallback provider from a list of transports.
+    ///
+    /// - The `active_transport_count` parameter controls how many transports are used for requests
+    ///   at any one time.
+    /// - The `log_transport_rankings` parameter controls whether the current transport rankings are
+    ///   output as traces. In order to see them, make sure to enable the `alloy_transport` target
+    ///   (i.e. `RUST_LOG=alloy_transport=debug`).
```
mattsse, line 214: unclear what fallback here means if this uses multiple transports

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,447 @@
...
+
+impl<S> FallbackService<S>
+where
+    S: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
+        + Send
+        + Clone
+        + std::fmt::Debug
+        + 'static,
+{
+    async fn make_request(&self, req: RequestPacket) -> Result<ResponsePacket, TransportError> {
```
mattsse, line 263: not obvious what this does, can we get some docs here

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,447 @@
...
+        let mut top_transports = Vec::new();
+        {
+            let mut transports = self.transports.lock().expect("Lock poisoned");
+            for _ in 0..self.active_transport_count.min(transports.len()) {
+                if let Some(transport) = transports.pop() {
+                    top_transports.push(transport);
```
mattsse, line 269: not exactly sure I fully understood this but it seems like this pops the transports and on success it adds them back?

what happens if there are no available transports?

mattsse (commented):
mostly questions, I think overall this makes sense and I was able to follow, but I have some concerns about the make_request impl, because if there are multiple concurrent requests, would this result in empty transports error rn?

crates/transport/src/layers/fallback.rs
merklefruit: Yes, I've updated this to use a `Vec` instead of a `BinaryHeap` and re-sort it whenever a score changes. 
This way I don't need to take the transports from the list on each request.

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+    ///
+    /// - The `active_transport_count` parameter controls how many transports are used for requests
+    ///   at any one time.
+    /// - The `log_transport_rankings` parameter controls whether the current transport rankings are
+    ///   output as traces. In order to see them, make sure to enable the `alloy_transport` target
+    ///   (i.e. `RUST_LOG=alloy_transport=debug`).
+    pub fn new(
+        transports: Vec<S>,
+        active_transport_count: usize,
+        log_transport_rankings: bool,
```
mattsse, line 51: can we make this bool a separate setter instead, but I'm also okay with always emitting this as trace!

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// Log the current ranking of transports
+    fn log_transport_rankings(&self)
+    where
+        S: Debug,
+    {
+        let transports = self.transports.read();
+
+        debug!("Current transport rankings:");
+        for (idx, transport) in transports.iter().enumerate() {
+            debug!("  #{}: Transport[{}] - {}", idx + 1, transport.id, transport.metrics_summary());
```
mattsse, line 75: we can make this trace imo

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+                        transports.sort_by(|a, b| b.cmp(a));
+                    }
+
+                    // Log current rankings if enabled
+                    if self.log_transport_rankings {
+                        self.log_transport_rankings();
+                    }
+
+                    // clear the remaining futures to avoid unnecessary work
+                    futures.clear();
```
mattsse, line 156: this isn't required because this is dropped here anyway

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+            let transports = self.transports.read();
+            transports.iter().take(self.active_transport_count).cloned().collect::<Vec<_>>()
+        };
+
+        // Create a collection of future requests
+        let mut futures = FuturesUnordered::new();
+
+        // Launch requests to all active transports in parallel
+        for transport in top_transports {
+            let start = Instant::now();
```
mattsse, line 113: this instant init should be moved into the async move, otherwise this doesn't accurately track the time this future takes

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+        // Wait for the first successful response or until all fail
+        let mut last_error = None;
+
+        while let Some((result, transport, duration)) = futures.next().await {
+            match result {
+                Ok(response) => {
+                    // Record success
+                    {
+                        let mut metrics = transport.metrics.write();
+                        metrics.track_success(duration);
```
mattsse, line 141: can we make this a function of transport instead

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+                    }
+
+                    // Sort transports after updating metrics
+                    {
+                        let mut transports = self.transports.write();
+                        transports.sort_by(|a, b| b.cmp(a));
+                    }
```
mattsse, line 148: I think we can avoid this write here if we instead sort when we retrieve all the transports then we don't need an additional write lock here

like first clone all transports, sort them, then take only self.active_transport_count

crates/transport/src/layers/fallback.rs
```diff
@@ -0,0 +1,453 @@
...
+                    // Record failure
+                    {
+                        let mut metrics = transport.metrics.write();
+                        metrics.track_failure();
+                    }
+
+                    // Sort transports after updating metrics
+                    {
+                        let mut transports = self.transports.write();
+                        transports.sort_by(|a, b| b.cmp(a));
```
mattsse, line 170: same here

mattsse (changes requested):
a few more suggestions

crates/transport/src/layers/fallback.rs
merklefruit: Got it, then I can drop the RwLock for the transports altogether!

mattsse (approved):
lgtm!

PR #2138, opened 2025-03-03
feat(`consensus`): separate RlpTx trait functionality

### LINKED ISSUE #2134: [Feature] Separate txhashing from RlpEcdsaTx
### Component

consensus, eips, genesis

we want #2025

however this requires that we can obtain the hash on demand via:

https://github.com/alloy-rs/alloy/blob/3872937ebd21e41e867261517bba70d5e4e11d4a/crates/consensus/src/transaction/rlp.rs#L206-L216

but the RlpxEcdsa trait can't be implemented for enums making `Signed<TypedTx>` problematic:

https://github.com/alloy-rs/alloy/blob/3872937ebd21e41e867261517bba70d5e4e11d4a/crates/consensus/src/transaction/rlp.rs#L102-L105

we can do either 
a) split RlpxEcdsa into Encodable and Decodable
b) extract the tx_hash to a standalone trait
c) do a) and b)

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2134 

## Solution

Introduces: 
- `RlpEcdsaEncodableTx` for encoding methods
- `RlpEcdsaDecodableTx` for decoding methods

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes


crates/consensus/src/transaction/rlp.rs
```diff
@@ -4,9 +4,233 @@ use alloy_eips::eip2718::{Eip2718Error, Eip2718Result};
...
+    /// Network encode the transaction with the given signature and the default
+    /// type flag.
+    fn network_encode(&self, signature: &Signature, out: &mut dyn BufMut) {
+        self.network_encode_with_type(signature, Self::DEFAULT_TX_TYPE, out);
+    }
+}
+
+/// Helper trait for calculating transaction hashes.
+#[doc(hidden)]
+pub trait RlpTxHash: RlpEcdsaEncodableTx {
```
mattsse, line 105: why do we need to extend it like this?

this only restricts the RlpEcdsaEncodableTx, unsure how this would be beneficial if all these functions can be default and could be part of the rncodabletx directly

crates/consensus/src/transaction/rlp.rs
```diff
@@ -4,9 +4,233 @@ use alloy_eips::eip2718::{Eip2718Error, Eip2718Result};
...
+    /// Calculate the transaction hash for the given signature.
+    fn tx_hash(&self, signature: &Signature) -> TxHash {
+        self.tx_hash_with_type(signature, Self::DEFAULT_TX_TYPE)
+    }
+}
+
+/// Helper trait for managing RLP decoding of transactions inside 2718 envelopes.
+#[doc(hidden)]
+#[doc(alias = "RlpDecodableTx", alias = "RlpTxDecoding")]
+pub trait RlpEcdsaDecodableTx: RlpEcdsaEncodableTx {
```
mattsse, line 122: do we need to enforce decodable: encodable?

crates/consensus/src/transaction/rlp.rs
```diff
@@ -4,9 +4,233 @@ use alloy_eips::eip2718::{Eip2718Error, Eip2718Result};
...
+    }
+}
+
 /// Helper trait for managing RLP encoding of transactions inside 2718
 /// envelopes.
 #[doc(hidden)]
+#[deprecated(
+    note = "Use RlpEcdsaEncodableTx for encoding, RlpEcdsaDecodableTx for decoding, and RlpTxHash for tx hashing functionality"
+)]
 pub trait RlpEcdsaTx: SignableTransaction<Signature> + Sized {
```
mattsse, line 234: we can change this one so that this acts as `encodable +decodable` and impl for all T

mattsse (changes requested):
mostly questions,

ptal as well @klkvr

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -154,6 +160,8 @@ impl RlpEcdsaTx for TxEip1559 {
     }
 }
 
+// impl RlpEcdsaTx for TxEip1559 {}
+
```
klkvr, line 164:
```suggestion
```

crates/consensus/src/transaction/eip2930.rs
```diff
@@ -80,7 +82,7 @@ impl TxEip2930 {
     }
 }
 
-impl RlpEcdsaTx for TxEip2930 {
+impl RlpEcdsaEncodableTx for TxEip2930 {
     const DEFAULT_TX_TYPE: u8 = { Self::tx_type() as u8 };
```
klkvr, line 86: still unsure about the type AT, it's not relevant for the transaction enums so we'd need to either remove it or set to some random value?

klkvr (changes requested):

crates/consensus/src/transaction/rlp.rs
klkvr: yeah I think hashing should just be part of encodable trait

crates/consensus/src/transaction/eip2930.rs
mattsse: yeah this still doesn't make sense to me either but we can just use 0 and then never use it internally

crates/consensus/src/transaction/rlp.rs
yash-atreya: [f1fb713](https://github.com/alloy-rs/alloy/pull/2138/commits/f1fb713298fa114df572d9a09845db8999e830bd)

crates/consensus/src/transaction/rlp.rs
yash-atreya: We use the AT `DEFAULT_TX_TYPE` from RlpEncodableTx  in `eip2718_decode` and `network_decode`.

Moreover, we're using `eip2718_encoded_length` in `eip2718_decode_with_type` to avoid re-serialization

crates/consensus/src/transaction/eip2930.rs
yash-atreya: Using the AT in decodable as well. Ref: https://github.com/alloy-rs/alloy/pull/2138#discussion_r1979599776

Here: https://github.com/alloy-rs/alloy/blob/f1fb713298fa114df572d9a09845db8999e830bd/crates/consensus/src/transaction/rlp.rs#L181

yash-atreya: Should we remove this check and set `DEFAULT_TX_TYPE` to `0`?

This will also enabled us to do `Decodable: SignableTx` instead of `Decodable: Encodable`. See: https://github.com/alloy-rs/alloy/pull/2138#discussion_r1979599776

crates/consensus/src/transaction/rlp.rs
yash-atreya: Addressing this now

crates/consensus/src/signed.rs
```diff
@@ -106,7 +106,7 @@ impl<T: SignableTransaction<Sig>, Sig> Signed<T, Sig> {
 
 impl<T> Signed<T>
 where
-    T: RlpEcdsaTx,
+    T: RlpEcdsaEncodableTx + RlpEcdsaDecodableTx,
```
mattsse, line 109: this shouldn't require a change if we keep RlpEcdsaTx

crates/consensus/src/transaction/rlp.rs
```diff
@@ -4,9 +4,229 @@ use alloy_eips::eip2718::{Eip2718Error, Eip2718Result};
...
```
mattsse, line 230: can we change this so that this is encodable + decodable and add a blanket impl?

mattsse (changes requested):

crates/consensus/src/signed.rs
klkvr: yeah let's keep this bound as is

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -154,6 +158,8 @@ impl RlpEcdsaTx for TxEip1559 {
...
```
klkvr, line 162:
```suggestion
```

klkvr (changes requested):

crates/consensus/src/signed.rs
yash-atreya: [7e90307](https://github.com/alloy-rs/alloy/pull/2138/commits/7e90307ad1910c617c681964febc0d8cd1a6a043)

klkvr (approved):
lgtm

PR #2140, opened 2025-03-04
refactor: improve and simplify event filters

prestwich:
## Motivation

Filtering is a bit wonky and crufty, presumably because the logic was ported from elsewhere. 

## Solution

- remove FilteredParams, as it was a dead struct providing functionality that should just be on the filter
- memoize blooms for `FilterSet` when `"std"` is enabled
- delete the `BloomFilter` type, use a SINGLE bloom filter per `FilterSet` instead of `Vec<Bloom>`. This appears to have been resource over-use. Blooms are designed to be accrued
- manually `impl Default` as the derived impl has a `T: Default` bound that is unnecessary
- make `FilterSet` inner `HashSet` private, to avoid invalidating memoized blooms
- expose basic mutators for the `FilterSet`

drive-by
- add a bunch of missing [cfg(feature = "serde")] in tests
- fix up some docs

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types-eth/src/filter.rs
```diff
@@ -14,47 +15,50 @@ use itertools::{
...
+pub struct FilterSet<T: Eq + Hash> {
+    set: HashSet<T>,
 
-impl From<Vec<Bloom>> for BloomFilter {
-    fn from(src: Vec<Bloom>) -> Self {
-        Self(src)
-    }
+    #[cfg(feature = "std")]
+    #[cfg_attr(feature = "serde", serde(skip, default))]
+    bloom: std::sync::OnceLock<Bloom>,
```
mattsse, line 26: I think we can use once_cell here and get rid of the std features like this once_cell should be already part of the dep graph anyway

crates/rpc-types-eth/src/filter.rs
```diff
@@ -800,135 +1013,6 @@ where
...
-    }
-
-    /// Returns `true` if the bloom contains one of the address blooms, or the address blooms
-    /// list is empty (thus, no filters)
-    pub fn matches_address(bloom: Bloom, address_filter: &BloomFilter) -> bool {
-        address_filter.matches(bloom)
-    }
-
-    /// Returns true if the filter matches the given block number
-    pub fn filter_block_range(&self, block_number: u64) -> bool {
```
mattsse, line 851: what's the equivalent of this now?

because we need this on the server side

crates/rpc-types-eth/src/filter.rs
```diff
@@ -800,135 +1013,6 @@ where
     }
 }
 
-/// Support for matching [Filter]s
-#[derive(Debug, Default)]
-pub struct FilteredParams {
-    /// The original filter, if any
-    pub filter: Option<Filter>,
-}
```
mattsse, line 808: what's the right way to handle an optional filter now?

mattsse (commented):
some qs, hesitant to touch any of the filter logic -.-

crates/rpc-types-eth/src/filter.rs
prestwich: - `matches_block_range` for numbers 
- `matches_block_hash` for hashes
- `matches_block` for numhashes

crates/rpc-types-eth/src/filter.rs
prestwich: an empty Filter via Filter::default

crates/rpc-types-eth/src/filter.rs
prestwich: `once_cell`'s no_std support is a little wonky. do you use it with race or critical section?

crates/rpc-types-eth/src/filter.rs
prestwich: another easy approach would be `filter_opt.map(|f| f.matches(thing)).unwrap_or(true)`

prestwich:
bump

crates/rpc-types-eth/src/filter.rs
```diff
@@ -534,10 +593,163 @@ impl Filter {
...
+    ///
+    /// This function checks [`crate::Log<LogData>`], the RPC type carrying
+    /// un-parsed [`LogData`].
+    ///
+    /// - For parsed [`Log<T>`]s (e.g. those returned by a contract), see [`Self::matches_parsed`].
+    /// - For parsed [`crate::Log<T>`]s (e.g. those returned by a contract), see
+    ///   [`Self::rpc_matches`].
+    pub fn rpc_matches(&self, log: &crate::Log) -> bool {
+        let Some(number) = log.block_number else { return false };
+        let Some(hash) = log.block_hash else { return false };
```
mattsse, line 700: this will never match for pending logs,
should this behave like an OR

crates/rpc-types-eth/src/filter.rs
```diff
@@ -14,47 +15,50 @@ use itertools::{
...
 };
 
-/// Helper type to represent a bloom filter used for matching logs.
-#[derive(Debug, Default)]
-pub struct BloomFilter(Vec<Bloom>);
+/// FilterSet is a set of values that will be used to filter logs.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Deserialize))]
+pub struct FilterSet<T: Eq + Hash> {
+    set: HashSet<T>,
```
mattsse, line 22: I believe this deserde impl no longer behaves like before, but unsure where this is even used, but we should treat this as serde transparent

mattsse (changes requested):
this is def better, only have 1 serde nit

dealing with filters and logs makes me feel dizzy, and I dread touching any of this https://github.com/paradigmxyz/reth/blob/f24570844545b55f2293a3a3c475133a3f738b68/crates/rpc/rpc/src/eth/filter.rs#L523-L525

so I'd need to prep a reth pr first just to make sure we're not missing anything

mattsse:
@prestwich I haven't started migrating https://github.com/paradigmxyz/reth/pull/16077 but maybe this is actually quite easy

crates/rpc-types-eth/src/filter.rs
prestwich: `serde(from = "HashSet<T>")` should cover this, ya?

crates/rpc-types-eth/src/filter.rs
prestwich: fixed by adding `matches_log_block`

mattsse:
that'd be much appreciated

Issue #2114, opened 2025-02-24
[Feature] Implement operation `sign_hash` on Ledger and Trezor signers


zerosnacks:
### Component

signers

### Describe the feature you would like

Related: https://github.com/foundry-rs/foundry/issues/9944

`cast wallet sign --trezor --no-hash 0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8` 

and 

``cast wallet sign --ledger --no-hash 0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8` 

currently yield the error "Error: operation `sign_hash` is not supported by the signer"

as they are not implemented:

https://github.com/alloy-rs/alloy/blob/88fc40338270756cc1f468319eda642cf01eb153/crates/signer-ledger/src/signer.rs#L82-L89

https://github.com/alloy-rs/alloy/blob/88fc40338270756cc1f468319eda642cf01eb153/crates/signer-trezor/src/signer.rs#L39-L47

### Additional context

For safety, the hash is just a simple `keccak256` of "hello"

```
0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8 == keccak256("hello")
```

neithanmo:
@zerosnacks if possible, can I look at this? I am familiar with ledger signing mechanism, I would like to see how this is handled in alloy

zerosnacks:
Sure, that would be great! Assigned @neithanmo

klkvr:
Not sure this is possible, iiuc it's common for hardware wallets to only allow signing transactions or EIP-712 messages. Allowing to blindly sign random hashes is not great in terms of security

zerosnacks:
I agree we should be hesitant to implement this as we want users to move away from blind signing. Alternative way to resolve this - if we choose not to implement it - is to provide a better error as we choose it to be not implemented rather than we lack support for it. Right now it can appear to users that their device does not support an operation.

neithanmo:
I agree, actually, some ledger applications require enabling blind-signing mode to sign hashes otherwise, this could get rejected. depending on the context here, another alternative is to use EIP-191 or EIP-712 as @klkvr mentioned

neithanmo:
A good documentation and an clear error message seem a good approach, making clear that it is not the device neither us, but a design choice based on security and best [practices](https://www.ledger.com/academy/enable-blind-signing-why-when-and-how-to-stay-safe)

PR #2141, opened 2025-03-04
Improve SignHash error message to clarify intentional security design

neithanmo:
## Motivation
see #2114 

## Solution
Update the `UnsupportedSignerOperation` error display to explicitly mention that blind signing is intentionally disabled as a security best practice

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/signer/src/error.rs
```diff
@@ -79,7 +79,12 @@ pub enum UnsupportedSignerOperation {
 
 impl fmt::Display for UnsupportedSignerOperation {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.as_str().fmt(f)
+        match self {
```
DaniPopes, line 82: you want to change this above in enum Error, but i think it should be another variant

mattsse: do you mean convert UnsupportedSignerOperation into an error via thiserror?

I think we should do this because otherwise this new message will mess with the error message here:

https://github.com/alloy-rs/alloy/blob/750004c1ce56b5044b27d19a980cb3b7d1efc898/crates/signer/src/error.rs#L13-L14

mattsse (changes requested):

crates/signer/src/error.rs
DaniPopes: I think we should keep things as is and just add another variant with a message that it is intentionally unimplemented

neithanmo: right. Your suggestion is better by making a distinction between what is unsupported and what is opted out due to security reasons. maybe a new variant to the Error definition:
```
pub enum Error {
    // ...
    
    /// The operation is deliberately disabled for security reasons
    #[error("Operation {0} is restricted due to security restrictions")]
    SecurityRestriction(String),
    
    // ...
}
```

neithanmo:
I updated and applied the suggested approach of adding a new variant

DaniPopes:
Thanks, however we believe that this functionality is not implemented because the underlying devices do not expose signing raw hashes directly, thus the current error message is correct.

Blind signing is unrelated to this; AFAICT it refers to when the device cannot decode the message or transaction calldata with a scheme known to the device.

We can pick this back up if we ever have an intentionally disabled API for security reasons.

PR #2143, opened 2025-03-05
fix(`rpc-types`): allow missing `effectiveGasPrice` in TxReceipt

### LINKED ISSUE #2019: [Feature] Allow certain missing fields for TransactionReceipt deserialization
### Component

rpc

ref https://github.com/foundry-rs/foundry/issues/7640

this mostly affects the effectiveGasPrice

some networks don't care about the eth spec, but for UX we have those mandatory,

for the eth_getTransactionByHash response we should find a way to deserialize and allowing missing effectiveGasPrice and use `gasPrice` instead if missing, similar to:

https://github.com/alloy-rs/alloy/blob/7643613acf6806b0517d24985a92cc2bbfc15b75/crates/rpc-types-eth/src/transaction/mod.rs#L390-L391

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2019 

## Solution

Manually implement deser logic to accommodate the missing `effectiveGasPrice` field in receipt.

- If missing, it tries to use `gasPrice`.
- Sets `effective_gas_price` to zero if both are missing.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/receipt.rs
+++ b/crates/rpc-types-eth/src/transaction/receipt.rs
@@ -10,7 +10,7 @@ use alloy_sol_types::SolEvent;
 /// consensus data and metadata.
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 #[doc(alias = "TxReceipt")]
 pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/transaction/receipt.rs
```diff
@@ -68,6 +68,74 @@ pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
...
+            #[serde(default, with = "alloy_serde::quantity::opt")]
+            block_number: Option<u64>,
+            #[serde(with = "alloy_serde::quantity")]
+            gas_used: u64,
+            // Optional fields for gas prices
+            // 1. Use effectiveGasPrice if present
+            // 2. Fallback to use gasPrice
+            // 3. Default to 0 if neither is present
+            #[serde(default = "default_gas", alias = "gasPrice", with = "alloy_serde::quantity")]
+            pub effective_gas_price: u128,
```
mattsse, line 99: this can also be private?
```suggestion
            effective_gas_price: u128,
```

mattsse: let's make this an option instead and use `Transaction::effective_gas_price` in case it's missing

mattsse (changes requested):
smol nit

crates/rpc-types-eth/src/transaction/receipt.rs
yash-atreya: Made it an Option

wdym by using `Transaction::effective_gas_price`? `TransactionReceipt` doesn't implement `Transaction` trait

crates/rpc-types-eth/src/transaction/receipt.rs
```diff
@@ -68,6 +68,70 @@ pub struct TransactionReceipt<T = ReceiptEnvelope<Log>> {
...
+
+        let helper = ReceiptDeserHelper::deserialize(deserializer)?;
+        Ok(Self {
+            inner: helper.inner,
+            transaction_hash: helper.transaction_hash,
+            transaction_index: helper.transaction_index,
+            block_hash: helper.block_hash,
+            block_number: helper.block_number,
+            gas_used: helper.gas_used,
+            effective_gas_price: helper.effective_gas_price.unwrap_or(0),
```
mattsse, line 125: ah I see we don't have access to the trait here, this seems fine

this is only affects weird networks so I don't really care here

mattsse (approved):

PR #2149, opened 2025-03-06
feat: add bincode compat for receipt

mattsse:
we also need this because the status type uses flatten which isn't compatible with bincode 👍


yash-atreya (approved):
lgtm!

PR #2150, opened 2025-03-06
feat(`consensus`): impl RlpEncodableTx for TypedTx

### LINKED ISSUE #2147: [Feature] Add RlpEcdsaEncodableTx for typedtransaction
### Component

consensus, eips, genesis

for #2025 we need an impl of RlpEcdsaEncodableTx for TypedTransaction otherwise we can't use `Signed<TypedTransaction>`

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2147 

## Solution

- Implements `RlpEcdsaEncodableTx` for `TypedTransaction`.
- AT: `DEFAULT_TX_TYPE` is set 0.
- Overrides any trait methods that require a tx type as arg i.e `*_with_type`.
- Delegates the tx type to the enum variant in those cases. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/typed.rs
+++ b/crates/consensus/src/transaction/typed.rs
@@ -391,6 +391,16 @@ impl RlpEcdsaEncodableTx for TypedTransaction {
         }
     }
 
+    fn eip2718_encode(&self, signature: &Signature, out: &mut dyn BufMut) {
+        match self {
+            Self::Legacy(tx) => tx.eip2718_encode(signature, out),
+            Self::Eip2930(tx) => tx.eip2718_encode(signature, out),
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this isn't quite right for this enum type because we also need to override the default impl for

https://github.com/alloy-rs/alloy/blob/52a8199bdb1f6488b4821b36740e30a8672a3e1a/crates/consensus/src/transaction/rlp.rs#L70-L74

otherwise this will encode incorrectly

mattsse (approved):
this lgtm now

still think we should remove the associated constant

PR #2154, opened 2025-03-07
feat: Add `ThrottleLayer` to Transport layers

### LINKED ISSUE #1636: [Feature] Alloy `ThrottleLayer`
### Component

rpc, transports

In some instances, it would be useful to have a `ThrottleLayer` to not flood the node with requests all at once. While the `RetryLayer` works perfectly for retrying requests that exceed RPC provider limits, the `ThrottleLayer` could throttle outbound requests before they are sent.

The following snippet shows an example that fetches 100k blocks concurrently. In this example a `ThrottleLayer` would be helpful to limit `n` requests per second.

```rust

#[tokio::main]

   // --snip--

   // Init the provider
    let client = ClientBuilder::default()
        .layer(RetryBackoffLayer::new(10, 300, 330))
        .http(rpc_endpoint.parse()?);
    let provider = ProviderBuilder::new().on_client(client);

    let futures_unordered = FuturesUnordered::new();

    // Get 100_000 blocks
    let block_number = provider.get_block_number().await?;
    for i in block_number - 100_000..block_number {
        let provider = provider.clone();
        futures_unordered.push(async move {
            provider
                .get_block(i.into(), BlockTransactionsKind::Hashes)
                .await
        });
    }

    // Await the results
    let res = futures_unordered.collect::<Vec<_>>().await;

    Ok(())
}
```

[We currently have a working version of a ThrottleLayer built](https://github.com/darkforestry/alloy-throttle/blob/main/src/lib.rs#L45-L97) and I would be happy to open a draft PR with this if you think it would be useful to add to `alloy`. 

### Additional context

_No response_

0xKitsune:
Closes #1636 

This PR adds a `ThrottleLayer` for alloy providers to throttle outbound rpc requests. While the `RetryLayer` works great for for retrying requests that exceed RPC provider limits, the `ThrottleLayer` reduces the frequency that requests need to be retried.

```rust
// --snip--
 let requests_per_second = NonZeroU32::new(50).unwrap();

 let client = ClientBuilder::default()
    // Throttle RPC requests at 40 requests per second 
    .layer(ThrottleLayer::new(requests_per_second))
    // The RetryBackoffLayer can be stacked with the throttle layer to retry failed requests
    .layer(RetryBackoffLayer::new(10, 300, 330))
    .http(rpc_endpoint.parse()?);
 
    // Get 100_000 blocks concurrently
    let futures_unordered = FuturesUnordered::new();
    let block_number = provider.get_block_number().await?;
    for i in block_number - 100_000..block_number {
        let provider = provider.clone();
        futures_unordered.push(async move {
            provider
                .get_block(i.into(), BlockTransactionsKind::Hashes)
                .await
        });
    }

    // Await the results
    let res = futures_unordered.collect::<Vec<_>>().await;

    }
```


yash-atreya (changes requested):
Thanks for this!

Can we put this behind a feature flag since it adds a new dependency?

A test would be great as well, you can put it in`provider/trait.rs`

0xKitsune:
@yash-atreya Ive updated the PR to put the Throttle logic behind a feature flag as well as added tests.

Note that I changed the `ThrottleLayer::new()` function to take a `u32` instead of `NonZeroU32` for simplicity. If `0` is passed to this function it will panic. can revert this change if preferred. 

Lmk if you would like any other updates.

mattsse (approved):
seems ok

PR #2155, opened 2025-03-07
chore(`consensus`)!: rename `Recovered` methods

yash-atreya:
## Motivation

`Recovered` can be used for any signed object not just transactions. 

## Solution

- Renames the methods and changes docs slightly to indicate the above

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/recovered.rs
+++ b/crates/consensus/src/transaction/recovered.rs
@@ -12,7 +12,7 @@ pub struct Recovered<T> {
     /// Signed object
     #[deref]
     #[as_ref]
-    tx: T,
+    inner: T,
 }
 
 impl<T> Recovered<T> {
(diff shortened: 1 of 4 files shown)


crates/consensus/src/transaction/recovered.rs
```diff
@@ -3,96 +3,150 @@ use alloy_primitives::{bytes, Address, B256};
...
     /// Note: This does not check if the signer is the actual signer of the transaction.
     #[inline]
-    pub const fn new_unchecked(tx: T, signer: Address) -> Self {
-        Self { tx, signer }
+    pub const fn new_unchecked(inner: T, signer: Address) -> Self {
+        Self { inner, signer }
+    }
+
+    /// Converts the inner signed object to the given alternative that is `From<T>`
+    pub fn convert_inner<Tx>(self) -> Recovered<Tx>
```
mattsse, line 91:
```suggestion
    pub fn convert<Tx>(self) -> Recovered<Tx>
```

mattsse (changes requested):
lets drop the _inner from some fn names

PR #2162, opened 2025-03-07
feat: add Eip4844 variant generic to TypedTransaction

klkvr:
## Motivation

Renames `TypedTransaction` to `EthereumTypedTransaction` and makes it generic over `Eip4844` variant. `TypedTranaction` is still kept as a type alias making this change non-breaking

We can apply same approch to `TxEnvelope` and and de-duplicate it with `PooledTransaction` https://github.com/alloy-rs/alloy/blob/9442fefece4bd78587f3a5db8f6b871a2a426f7b/crates/consensus/src/transaction/pooled.rs#L32

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -36,8 +36,9 @@ pub mod transaction;
 #[cfg(feature = "kzg")]
 pub use transaction::BlobTransactionValidationError;
 pub use transaction::{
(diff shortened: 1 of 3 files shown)


mattsse (approved):
this is a lot easier than introducing yet another variant

PR #2170, opened 2025-03-08
chore: impl Hash for Signed

mattsse:
use tx hash as hash, same as

https://github.com/alloy-rs/alloy/blob/356dc672d31d44493386a7ff65d30223f18995b5/crates/consensus/src/transaction/pooled.rs#L278-L282


DaniPopes (commented):
if the hash can be invalidated through mutation then this is incorrect

if not then equality should also be implemented like this

mattsse:
max pain

mattsse:
@DaniPopes impled hash for signed, mirroring partialeq and derived it for txenvelope instead

PR #2172, opened 2025-03-10
chore: remove associated constant from RlpEcdsaEncodableTx

### LINKED ISSUE #2171: [Feature] Remove associated constant from RlpEcdsaDecodableTx
### Component

consensus, eips, genesis

the RlpEcdsaDecodableTx trait currently uses associated constants for some decoding functions

https://github.com/alloy-rs/alloy/blob/852f0467074e8c46dde239f09926a93be12dafe1/crates/consensus/src/transaction/rlp.rs#L119-L119

https://github.com/alloy-rs/alloy/blob/852f0467074e8c46dde239f09926a93be12dafe1/crates/consensus/src/transaction/rlp.rs#L198-L200

this does not work for enum types.

while we can replace this const for encoding with `Typed2718`

https://github.com/alloy-rs/alloy/blob/852f0467074e8c46dde239f09926a93be12dafe1/crates/consensus/src/transaction/rlp.rs#L11-L13

this does not work for decoding.

so we need to change the functions so that they don't use the constant.

the problematic functions are:

https://github.com/alloy-rs/alloy/blob/852f0467074e8c46dde239f09926a93be12dafe1/crates/consensus/src/transaction/rlp.rs#L123-L126

because the assume we can decode the fields. but this isn't enough for enums, which require 2718 byte

so we likely need to split this

### Additional context

_No response_

stevencartavia:
## Motivation
closes #2171 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/consensus/src/transaction/rlp.rs
```diff
@@ -195,8 +195,8 @@ pub trait RlpEcdsaDecodableTx: RlpEcdsaEncodableTx {
 
     /// Decodes the transaction from eip2718 bytes, expecting the default type
     /// flag.
-    fn eip2718_decode(buf: &mut &[u8]) -> Eip2718Result<Signed<Self>> {
-        Self::eip2718_decode_with_type(buf, Self::DEFAULT_TX_TYPE)
+    fn eip2718_decode(&self, buf: &mut &[u8]) -> Eip2718Result<Signed<Self>> {
```
mattsse, line 198: we can't give the decoding fns a &self arg because this defeats the purpose of decoding, so we need to fix this separately

crates/consensus/src/transaction/rlp.rs
```diff
@@ -1,17 +1,17 @@
...
 
 /// Helper trait for managing RLP encoding of transactions inside 2718
 #[doc(hidden)]
 #[doc(alias = "RlpEncodableTx", alias = "RlpTxEncoding")]
 #[auto_impl::auto_impl(&, Arc)]
-pub trait RlpEcdsaEncodableTx: Sized {
-    /// The default transaction type for this transaction.
-    const DEFAULT_TX_TYPE: u8;
-
+pub trait RlpEcdsaEncodableTx: Sized + Typed2718 {
```
mattsse, line 14: this is great

crates/consensus/src/transaction/rlp.rs
```diff
@@ -109,14 +109,14 @@ pub trait RlpEcdsaEncodableTx: Sized {
...
-        self.tx_hash_with_type(signature, Self::DEFAULT_TX_TYPE)
+        self.tx_hash_with_type(signature, self.ty())
     }
 }
 
 /// Helper trait for managing RLP decoding of transactions inside 2718 envelopes.
 #[doc(hidden)]
 #[doc(alias = "RlpDecodableTx", alias = "RlpTxDecoding")]
-pub trait RlpEcdsaDecodableTx: RlpEcdsaEncodableTx {
+pub trait RlpEcdsaDecodableTx: RlpEcdsaEncodableTx + Typed2718 {
```
mattsse, line 119: here we now need to introduce the associated constant for now, because we can't remove that in one go, but we want to remove it from the encodable trait right away

mattsse (changes requested):
a good start, but for this pr we can simply move the associated constant from the encodable to the decodable trait until we figure out what to do with the default decode impls

mattsse (approved):
lgtm

Karrq:
Hey, couldn't help but notice this API breaking change was included in the patch release `0.12.5` as described in the [changelog](https://github.com/alloy-rs/alloy/blob/fe19b2bab1de419ac6995e15c7f0007b5c370e6c/CHANGELOG.md?plain=1#L26).
What this intended? Thanks!

Issue #2089, opened 2025-02-20
[Feature] eth_call batching layer


mattsse:
### Component

rpc

### Describe the feature you would like

ref https://github.com/foundry-rs/foundry/issues/5363

to save on compute units we can collect concurrent eth_calls into a single multicall.

this needs to introduce a delay that is used to await potential incoming calls, because otherwise it's unlikely that multiple ethcalls are queued up at the very same time.
so the tradeoff here is latency of single ethcalls vs compute unit consumption or rather outgoing requests.

This can be implemented as a shared buffer and moved into the `EthCall`, where new calls push to the buffer of oneshots and if it's the first entry it waits until a timeout and then drains the buffer and batches all pending calls

PoulavBhowmick03:
Can I take this up?

0xriazaka:
can i work on this?

DaniPopes:
Sorry, we were already working on this, and this shouldn't have been tagged with "good first issue".

PR #2174, opened 2025-03-10
feat(providers): add multicall batch layer

DaniPopes:
Implement a provider layer that aggregates contract calls (`eth_call`) over a time period into a single Multicall3 contract call.

Closes https://github.com/alloy-rs/alloy/issues/2089.


crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,359 @@
...
+            match self.rx.recv().await {
+                Some(msg) => self.process_msg(msg),
+                None => break,
+            }
+
+            // Handle messages within the wait.
+            debug_assert!(!self.pending.is_empty());
+            'inner: loop {
+                sleep(self.wait).await;
+                match self.rx.try_recv() {
```
mattsse, line 186: this only pocesses one more message before sleeping again?

we want to drain the entire channel here

DaniPopes: the sleep should've been outside the loop

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,491 @@
...
+/// can be sent with a multicall.
+///
+/// This layer is useful for reducing the number of network requests made.
+/// However, this only works when requests are made in parallel, for example when using the
+/// [`tokio::join!`] macro or in multiple threads/tasks, as otherwise the requests will be sent one
+/// by one as normal, but with an added delay.
+///
+/// [Multicall3]: https://github.com/mds1/multicall3
+#[derive(Debug)]
+pub struct BatchLayer {
```
DaniPopes, line 42: Name q: CallBatchLayer?

mattsse: yeah let's be explicit and name this CallBatchLayer

crates/provider/src/builder.rs
```diff
@@ -398,6 +395,13 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
         let client = ClientBuilder::default().hyper_http(url);
         self.on_client(client)
     }
+
+    /// Aggregate multiple `eth_call` requests into a single batch request using Multicall3.
+    ///
+    /// See [`BatchLayer`] for more information.
+    pub fn with_call_batching(self) -> ProviderBuilder<Stack<BatchLayer, L>, F, N> {
+        self.layer(BatchLayer::new())
+    }
```
mattsse, line 404: this makes sense, if the user wants to configure the layer they can just call layer with the configured batchlayer manually.

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+use std::{fmt, future::IntoFuture, marker::PhantomData, sync::Arc, time::Duration};
+use tokio::sync::{mpsc, oneshot};
+
+#[cfg(target_arch = "wasm32")]
+use wasmtimer::tokio::sleep;
+
+#[cfg(not(target_arch = "wasm32"))]
+use tokio::time::sleep;
+
+const DEFAULT_WAIT: Duration = Duration::from_millis(1);
```
mattsse, line 20: yeah unsure what the default should be maybe higher than 1ms? but we can keep it at 1ms,

imo very unrealistic that this feature will be useful with 1ms

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+    }
+}
+
+impl BatchLayer {
+    /// Create a new `BatchLayer` with a default wait of 1ms.
+    pub fn new() -> Self {
+        Self { m3a: MULTICALL3_ADDRESS, wait: DEFAULT_WAIT }
+    }
+
+    /// Set the amount of time to wait before sending the batch.
```
mattsse, line 60: can we clarify here that this is the time it waits for additional calls before sending the the call, so this the max delay overhead per call

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+        }
+        let tx = params.data();
+        if tx.to().is_none() {
+            return false;
+        }
+        macro_rules! any_is_some {
+            ($($name:ident),* $(,)?) => { false $(|| tx.$name().is_some())* };
+        }
+        // TODO: Not exhaustive, should we serialize to JSON and check that instead?
+        if any_is_some!(
```
mattsse, line 198: I think this is also missing the blob fee param?

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+            return false;
+        }
+        let tx = params.data();
+        if tx.to().is_none() {
+            return false;
+        }
+        macro_rules! any_is_some {
+            ($($name:ident),* $(,)?) => { false $(|| tx.$name().is_some())* };
+        }
+        // TODO: Not exhaustive, should we serialize to JSON and check that instead?
```
mattsse, line 197: I think the macro is fine

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+            match self.rx.recv().await {
+                Some(msg) => self.process_msg(msg),
+                None => break,
+            }
+
+            // Handle all remaining messages after waiting the duration.
+            debug_assert!(!self.pending.is_empty());
+            sleep(self.wait).await;
+            'inner: loop {
+                match self.rx.try_recv() {
```
mattsse, line 275: is there an upper limit of requests we can process in a single request?

crates/provider/src/layers/batch.rs
```diff
@@ -0,0 +1,515 @@
...
+            // Wait for the first message.
+            debug_assert!(self.pending.is_empty());
+            match self.rx.recv().await {
+                Some(msg) => self.process_msg(msg),
+                None => break,
+            }
+
+            // Handle all remaining messages after waiting the duration.
+            debug_assert!(!self.pending.is_empty());
+            sleep(self.wait).await;
```
mattsse, line 273: this overhead is now always applied per batch but it could be that the channel already has sufficient pending messages (but unlikely) so this is fine imo

mattsse (commented):
this looks pretty good

some nits and comments

crates/provider/src/layers/batch.rs
DaniPopes: currently not but it can be implemented. in viem it's done as a "calldata bytes" limit, which we can also do

crates/provider/src/layers/batch.rs
klkvr: > imo very unrealistic that this feature will be useful with 1ms

I think 1ms should be OK given that this should basically serve as a timeframe for receiving multiple `join`'ed requests?

DaniPopes: yes, this covers the join! usecase and will also catch highly parallel requests like in fork tests while 1ms per request isn't generally a big deal

crates/provider/src/layers/batch.rs
DaniPopes: I copy pasted every method that is available in the trait

crates/provider/src/layers/batch.rs
DaniPopes: we'd still have to wait, if we had another limit then yes

crates/provider/src/layers/batch.rs
DaniPopes: I changed to json for exhaustiveness

crates/provider/src/builder.rs
DaniPopes: Added example to the layer docs

mattsse (approved):
cool, lgtm!

jenpaff:
nice work ! @DaniPopes is this worth adding to the examples repo or book https://alloy.rs/ ? can create a new issue

PR #2176, opened 2025-03-10
Remove constant from rlp ecdsa decodable tx 

### LINKED ISSUE #2171: [Feature] Remove associated constant from RlpEcdsaDecodableTx

Ayushdubey86:
Hi @yash-atreya, could you please review this PR? Also, it would be great if you could assign me a few more issues beyond just 'good first issue,'  in any Alloy or Foundry. Thanks!

## Motivation

Closes #2171 

PR is linked with this [issue](https://github.com/alloy-rs/alloy/issues/2171)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -184,12 +184,12 @@ where
     }
 
     /// EIP-2718 decode the signed transaction.
-    pub fn eip2718_decode(buf: &mut &[u8]) -> Eip2718Result<Self> {
+    pub fn eip2718_decode(self, buf: &mut &[u8]) -> Eip2718Result<Self> {
         T::eip2718_decode(buf)
     }
 
     /// Network decode the signed transaction with a specified type flag.
-    pub fn network_decode_with_type(buf: &mut &[u8], ty: u8) -> Eip2718Result<Self> {
+    pub fn network_decode_with_type(&self.tx, buf: &mut &[u8], ty: u8) -> Eip2718Result<Self> {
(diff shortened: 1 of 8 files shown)


yash-atreya:
Hey @Ayushdubey86 sorry but #2171 was resolved by #2172, closing this.

PR #2177, opened 2025-03-10
fix docs of input field for different tx

gftea:
## Motivation

Fix docs which seems copy/paste from old version

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (approved):
these actually make the docs more correct

PR #2178, opened 2025-03-11
feat(`contract`): build signed and usigned txs from CallBuilder

### LINKED ISSUE #793: [Feature] Add a method `build_raw_transaction` in CallBuilder
### Component

contract

In [CallBuilder](https://github.com/alloy-rs/alloy/blob/4000b7754663d4350022e15c9fab6e75d3da8afe/crates/contract/src/call.rs#L123) it would be nice to have a method:
`build_raw_transaction`

The goal of this method would be to build the raw transaction from the given parameters in Bytes.

The transaction built shouldn't be signed and the caller should handle the signing the transaction.

### Additional context

This feature can be useful if we don't want just to directly `send` the transaction but do some "lower level" stuff.

Additionally, we may want to send a transaction using multiple different wallets and not only the one in Provider

yash-atreya:
## Motivation

Closes #793 

## Solution

Adds the following methods to the `CallBuilder`

1. `build_unsigned_raw_transaction`
2. `build_raw_transactions`. Returns the signed 2718 representation

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -140,7 +140,38 @@ impl<T, P, D, N: Network> CallBuilder<T, P, D, N> {
         self.request
     }
 
-    /// Build a raw unsigned transaction for the call.
+    /// Builds and returns a RLP-encoded unsigned transaction from the call that can be signed.
+    ///
+    /// ## Example
+    ///
+    /// ```no_run
+    /// # use alloy_provider::ProviderBuilder;
+    /// # use alloy_sol_types::sol;
+    ///
+    /// sol! {
+    ///     #[sol(rpc, bytecode = "0x")]
+    ///    contract Counter {
+    ///        uint128 public counter;
+    ///
+    ///        function increment() external {
+    ///            counter += 1;
+    ///        }
+    ///    }
+    /// }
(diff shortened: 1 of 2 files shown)


crates/contract/src/call.rs
```diff
@@ -135,6 +139,27 @@ impl<T, P, D, N: Network> CallBuilder<T, P, D, N> {
...
+    /// Build a raw unsigned transaction for the call.
+    pub fn build_unsigned_raw_transaction(self) -> Result<Vec<u8>, TransactionBuilderError<N>>
+    where
+        N::UnsignedTx: SignableTransaction<Signature>,
+    {
+        let tx = self.request.build_unsigned().map_err(|e| e.error)?;
+        Ok(tx.encoded_for_signing())
+    }
+
+    /// Build a signed raw transaction for the call.
```
mattsse, line 152: these need a few more docs, not obvious what these are, what these return and when to use which

mattsse (changes requested):
doc nits

mattsse (approved):

PR #2192, opened 2025-03-13
feat: impl into_transaction TxEnvelope

stevencartavia:
## Motivation
https://github.com/paradigmxyz/reth/pull/15005

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -48,6 +48,18 @@ impl TxEnvelope {
             Self::Eip7702(tx) => Ok(tx.into()),
         }
     }
+
+    /// Consumes the type and returns the transaction.
+    #[inline]
+    pub fn into_transaction(self) -> EthereumTypedTransaction<TxEip4844Variant> {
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -48,6 +48,18 @@ impl TxEnvelope {
             Self::Eip7702(tx) => Ok(tx.into()),
         }
     }
+
+    /// Consumes the type and returns the transaction.
+    #[inline]
+    pub fn into_transaction(self) -> EthereumTypedTransaction<TxEip4844Variant> {
```
mattsse, line 54:
```suggestion
    pub fn into_typed_transaction(self) -> EthereumTypedTransaction<TxEip4844Variant> {
```

mattsse, line 54: let's also mention that this removes the signature

mattsse (changes requested):
smol nit

Issue #2191, opened 2025-03-13
[Feature] Rlp Encodable, Decodable for Signed<T>


gtsui:
### Component

signers

### Describe the feature you would like

Ive got a Signed<TxEip1559> object - how do i rlp encode it to bytes?

The docs suggests theres a pub fn eip2718_encode(&self, out: &mut dyn BufMut) function but i dont seem to have it in my environment - am i missing something?

mattsse:
this should be fairly simple now and we only need to replicate:

https://github.com/alloy-rs/alloy/blob/a10c8f34b5896d7cbabfcf6d7baf1b0c843d5876/crates/consensus/src/transaction/envelope.rs#L522-L539

for Signed

0xMushow:
Can I take this please?

mattsse:
assigned, ty!

it's possible that the Decodable2718 impl is a bit more complex because we currently only have this impl for

https://github.com/alloy-rs/alloy/blob/a10c8f34b5896d7cbabfcf6d7baf1b0c843d5876/crates/consensus/src/transaction/envelope.rs#L541-L541

so we can likely start with Encodable+Encodable2718 first and figure out what we should do about the decodable impl

mattsse:
@0xMushow Encodable2718 should likely just delegate to the underlying RlpEcdsaEncodableTx

0xMushow:
Working on it in this [PR](https://github.com/alloy-rs/alloy/pull/2193). I have to say I'm very new to this, not so sure if this is the expected change.

yash-atreya:
Resolved by #2193

PR #2193, opened 2025-03-13
Add encodable and decodable for `Signed<T>`

0xMushow:
## Motivation

This is regarding the following [issue](https://github.com/alloy-rs/alloy/issues/2191)

## Solution

Added `Encodable` and `Decodable` for `Signed<T>`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -519,28 +519,6 @@ impl<Eip4844: RlpEcdsaEncodableTx> EthereumTxEnvelope<Eip4844> {
     }
 }
 
-impl<T> Encodable for Signed<T>
-where
-    Self: Encodable2718,
-{
-    fn encode(&self, out: &mut dyn alloy_rlp::BufMut) {
-        self.network_encode(out)
-    }
-
-    fn length(&self) -> usize {
-        self.network_len()
-    }
-}
-
-impl<T> Decodable for Signed<T>
-where
-    T: RlpEcdsaDecodableTx,
-{
-    fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
-        Ok(Self::network_decode(buf)?)
-    }
-}
-
 impl<Eip4844> Encodable for EthereumTxEnvelope<Eip4844>
 where
     Self: Encodable2718,
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -580,6 +580,19 @@ where
     }
 }
 
+impl<T> Decodable2718 for Signed<T>
+where
+    T: RlpEcdsaDecodableTx + Typed2718 + Send + Sync,
+{
+    fn typed_decode(_ty: u8, buf: &mut &[u8]) -> Eip2718Result<Self> {
+        T::rlp_decode_signed(buf).map_err(Into::into)
+    }
+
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -519,6 +519,28 @@ impl<Eip4844: RlpEcdsaEncodableTx> EthereumTxEnvelope<Eip4844> {
     }
 }
 
+impl<T> Encodable for Signed<T>
+where
+    Self: Encodable2718,
```
mattsse, line 524: I think we don't have Encodable2718 for Signed yet,

so we can try adding Encodable2718 by replicating this

https://github.com/alloy-rs/alloy/blob/a10c8f34b5896d7cbabfcf6d7baf1b0c843d5876/crates/consensus/src/transaction/envelope.rs#L557-L560

for Signed

mattsse (changes requested):
cool, good start

crates/consensus/src/transaction/envelope.rs
0xMushow: I tried adding something, could you double-check? Still a beginner here. :)

mattsse (approved):
lgtm, pending @klkvr

crates/consensus/src/transaction/envelope.rs
klkvr: I'm a bit unsure about this impl. The initial motivation for `RlpEcdsa` tx traits was to have a clear API to avoid footguns.

It is not intuitive whether RLP encoding for `Signed<T>` performs the network encoding of a transaction (`network_encode)` or encodes just the transaction with signature (`RlpEcdsaEncodableTx::rlp_encode_signed`)

mattsse: I see, should we limit it to just Encodable2718 then and remove the encodable impl?

this is mostly what you want here anyway

DaniPopes (approved):

crates/consensus/src/transaction/envelope.rs
klkvr: yep, keeping just 2718 impls makes sense

0xMushow: Just to double check, do we also want to remove the `Decodable` impl for `Signed<T>` then?

klkvr: yep, let's remove both `Encodable`/`Decodable` here to not assume any specific RLP encoding here

0xMushow: Should be good now

mattsse (approved):

crates/consensus/src/transaction/envelope.rs
```diff
@@ -554,6 +554,51 @@ impl<Eip4844: RlpEcdsaDecodableTx> Decodable2718 for EthereumTxEnvelope<Eip4844>
...
+        *self.hash()
+    }
+}
+
+impl<T> Decodable2718 for Signed<T>
+where
+    T: RlpEcdsaDecodableTx + Typed2718 + Send + Sync,
+{
+    fn typed_decode(ty: u8, buf: &mut &[u8]) -> Eip2718Result<Self> {
+        let decoded = T::rlp_decode_signed(buf).map_err(Into::into)?;
```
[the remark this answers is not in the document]
0xMushow: Indeed redundant with the `?`

PR #2194, opened 2025-03-13
feat(`provider`): watch_full_blocks

yash-atreya:
## Motivation

Partially addresses #325 

## Solution

- Introduces `WatchBlocks`
- `WatchBlocks` consumes the stream of block hashes from `PollerBuilder` and maps it to `Stream<Item = TransportResult<BlockResponse>`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/provider/trait.rs
```diff
@@ -490,6 +491,45 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
+    /// while let Some(block) = stream.next().await {
+    ///   println!("new block: {block:#?}");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn watch_full_blocks(
+        &self,
+        kind: BlockTransactionsKind,
```
mattsse, line 520: this argument should not exist and instead should also be .full

crates/provider/src/provider/trait.rs
```diff
@@ -490,6 +491,45 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// Get the next 5 full blocks:
+    ///
+    /// ```no_run
+    /// # async fn example(provider: impl alloy_provider::Provider) -> Result<(), Box<dyn std::error::Error>> {
+    /// use futures::StreamExt;
+    /// use alloy_rpc_types_eth::BlockTransactionsKind;
+    ///
+    /// let poller = provider.watch_full_blocks(BlockTransactionsKind::Full).await?;
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
+    /// while let Some(block) = stream.next().await {
```
mattsse, line 512: this should return a stream directly

mattsse (commented):
not immediately obvious why this needs an additional task.

I'm a bit confused by how the PollerBuilder stuff works, or why this appears to be so complex,

isn't this jus feature just mapped stream effectively?

crates/provider/src/provider/trait.rs
yash-atreya: [f324589](https://github.com/alloy-rs/alloy/pull/2194/commits/f32458938fa015c99ba19982e54cf92617e43d05)

crates/provider/src/provider/trait.rs
yash-atreya: [f324589](https://github.com/alloy-rs/alloy/pull/2194/commits/f32458938fa015c99ba19982e54cf92617e43d05)

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
 
+/// A builder type for polling full blocks.
+///
+/// By default, this polls for blocks with [`BlockTransactionsKind::Hashes`].
+///
+/// The polling stream must be consumed by calling [`WatchBlocks::into_stream`].
+#[derive(Debug)]
+#[must_use = "this builder does nothing unless you call `.into_stream`"]
+pub struct WatchBlocks<BlockResp> {
+    poller: PollerBuilder<(U256,), Vec<B256>>,
```
mattsse, line 268: needs docs what the types represent

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
+    kind: BlockTransactionsKind,
```
mattsse, line 269: also needs docs

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
+
+    /// Poll for blocks with just transactions hashes i.e [`BlockTransactionsKind::Hashes`].
+    pub fn hashes(mut self) -> Self {
+        self.kind = BlockTransactionsKind::Hashes;
+        self
+    }
+
+    /// Consumes the stream of block hashes from the inner [`PollerBuilder`] and maps it to a stream
+    /// of [`BlockResponse`].
+    pub fn into_stream(self) -> impl Stream<Item = TransportResult<BlockResp>> + Unpin {
```
mattsse, line 296: if we return anonymous types then this should also enforce Send because this can't be inferred on the callsite

but this is likely problematic for wasm compat

crates/provider/src/provider/trait.rs
```diff
@@ -490,6 +490,36 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// # Examples
+    ///
+    /// Get the next 5 full blocks:
+    ///
+    /// ```no_run
+    /// # async fn example(provider: impl alloy_provider::Provider) -> Result<(), Box<dyn std::error::Error>> {
+    /// use futures::StreamExt;
+    ///
+    /// let poller = provider.watch_full_blocks().await?.full();
+    /// let mut stream = poller.into_stream().flat_map(futures::stream::iter).take(5);
```
mattsse, line 509: why does this need a flaptmap?

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
+        let client = self.poller.client();
+        let kind = self.kind;
+        let stream = self
+            .poller
+            .into_stream()
+            .then(move |hashes| {
```
mattsse, line 302: hmm, why does this yield `hashes` and not individual hashes?

mattsse (changes requested):
a few more questions

crates/provider/src/provider/get_block.rs
DaniPopes: we don't do this anywhere for streams

crates/provider/src/provider/get_block.rs
DaniPopes: eth_getFilterChanges returns a list

crates/provider/src/provider/trait.rs
DaniPopes: See other watch methods, eth_getFilterChanges returns a list

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
+                let client = client.clone();
+                async move { utils::hashes_to_blocks(hashes, client, kind.into()).await }
+            })
+            .flat_map(|res| {
+                futures::stream::iter(match res {
+                    Ok(blocks) => {
+                        // Ignore `None` responses.
+                        blocks.into_iter().filter_map(|opt_block| opt_block.map(Ok)).collect()
+                    }
+                    Err(err) => vec![Err(err)],
```
DaniPopes, line 312: this can use Either Left/Right instead of collect/vec on both sides

crates/provider/src/provider/get_block.rs
```diff
@@ -256,6 +257,74 @@ where
...
+    kind: BlockTransactionsKind,
+    _pd: std::marker::PhantomData<BlockResp>,
+}
+
+impl<BlockResp> WatchBlocks<BlockResp>
+where
+    BlockResp: BlockResponse + RpcRecv,
+{
+    /// Create a new [`WatchBlocks`] instance.
+    pub fn new(poller: PollerBuilder<(U256,), Vec<B256>>) -> Self {
```
DaniPopes, line 278: Don't think this should be public

crates/provider/src/utils.rs
```diff
@@ -138,6 +142,31 @@ pub(crate) fn convert_to_hashes<BlockResp: alloy_network::BlockResponse>(
...
+
+            client
+                .request::<_, Option<BlockResp>>("eth_getBlockByHash", (hash, full))
+                .map_resp(|resp| if !full { convert_to_hashes(resp) } else { resp })
+        };
+
+        futures.push(fut);
+    }
+
+    let blocks = futures::future::try_join_all(futures).await?;
```
DaniPopes, line 166: this takes an iterator so all of this function can be just `try_join_all(hashes.into_iter().map(|hash| ...))`

crates/provider/src/utils.rs
```diff
@@ -138,6 +142,31 @@ pub(crate) fn convert_to_hashes<BlockResp: alloy_network::BlockResponse>(
...
+    hashes: Vec<B256>,
+    client: WeakClient,
+    full: bool,
+) -> TransportResult<Vec<Option<BlockResp>>> {
+    let mut futures = Vec::with_capacity(hashes.len());
+
+    for hash in hashes {
+        let fut = {
+            let client =
+                client.upgrade().ok_or(TransportError::local_usage_str("client dropped"))?;
```
DaniPopes, line 156: can upgrade the client once outside of the loop/iterator

DaniPopes (changes requested):

crates/provider/src/provider/trait.rs
yash-atreya: See https://docs.alchemy.com/reference/eth-getfilterchanges-1

crates/provider/src/provider/get_block.rs
yash-atreya: See: https://docs.alchemy.com/reference/eth-getfilterchanges-1

crates/provider/src/utils.rs
yash-atreya: [00efd83](https://github.com/alloy-rs/alloy/pull/2194/commits/00efd83fa977704936e7ea5d2991367457b8024a)

crates/provider/src/provider/get_block.rs
yash-atreya: [00efd83](https://github.com/alloy-rs/alloy/pull/2194/commits/00efd83fa977704936e7ea5d2991367457b8024a)

crates/provider/src/provider/get_block.rs
yash-atreya: [00efd83](https://github.com/alloy-rs/alloy/pull/2194/commits/00efd83fa977704936e7ea5d2991367457b8024a)

mattsse (approved):

crates/provider/src/provider/get_block.rs
DaniPopes: you can call `into_stream` after sending, but yes ideally these are all named types...

crates/provider/src/provider/trait.rs
```diff
@@ -496,6 +496,36 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
         Ok(PollerBuilder::new(self.weak_client(), "eth_getFilterChanges", (id,)))
     }
 
+    /// Watch for new blocks by polling the provider with
```
DaniPopes, line 499: shouldn't this be just `watch_blocks().full()`?

DaniPopes (changes requested):

crates/provider/src/provider/trait.rs
yash-atreya: I'd like to keep it as a different method because if we want to get block with full txns we'll have to `watch_blocks().await?.full().full_txns()`

Issue #1318, opened 2024-09-20
[Discuss] consider pausing background task if no pending txs


jgalat:
### Component

provider, pubsub

### What version of Alloy are you on?

0.3.6

### Operating System

Linux

### Describe the bug

I noticed my RPC provider getting spammed with `eth_blockNumber` and `eth_getBlockByNumber` calls after sending a tx.
In my use case I am not even watching pending transactions, but still it starts to spam and doesn't stop unless I restart all over.
 
Here is a quick reproduction video and a repository. 
The repo uses 0.3.6, but I tested also with 0.2.1 and 0.1.4 and all behave the same.
In this scenario I am watching and confirming that the transaction went through.

https://github.com/jgalat/alloy-repro-0

https://github.com/user-attachments/assets/7ff3a259-b06a-4175-8e02-543c20abc9b2

nhtyy:
there is a check for is [is_local](https://github.com/alloy-rs/alloy/blob/57dd4c538293421c3d1a793cba79ad6f46d6444f/crates/rpc-client/src/client.rs#L182), maybe there could be a setter at the provider level but i don't think this is a bug

jgalat:
@nhtyy Thanks for answering, but why is it pinging? 
Some RPC providers charge credits for this call, e.g. [Alchemy uses 10 CU](https://docs.alchemy.com/reference/compute-unit-costs#standard-evm-json-rpc-methods-ethereum-polygon-pos-polygon-zkevm-optimism-arbitrum-zksync-base-astar).

Also I see that this doesn't start unless I send a signed tx. `eth_calls` or `eth_subscribe` methods don't trigger the spam.

DaniPopes:
This is the intended behavior. It's the background task that follows the chain for handling pending transactions. It's started when the first transaction is sent and closed when the last provider is dropped. I guess it could also be paused if there are no pending transactions? cc @klkvr @mattsse

jgalat:
Thanks for taking a look @DaniPopes. I've been trying to find an alternative to at least increase the polling interval after reading nhtyy's comment, but seems to be impossible. Here's what I've been trying
```rust
    let ws = WsConnect::new("ws://localhost:8545");
    let transport = ws.into_service().await?;
    let client = ClientBuilder::default()
        .transport(transport, false)
        .with_poll_interval(std::time::Duration::from_secs(3_600));

    client
        .inner()
        .set_poll_interval(std::time::Duration::from_secs(3_600));

[...]

PR #2197, opened 2025-03-14
feat: remove poller task indirection

DaniPopes:
Don't spawn poller in a task by default by using the stream! macro to yield responses instead.

Requirement for https://github.com/alloy-rs/alloy/issues/1318.

~~Unfortunately not spawning it means we lose the ".instrument" span attached to the spawned task future. I have not found a way to "instrument" a stream without writing a Stream utility ourselves.~~

Just decided to claude it.


mattsse (approved):
this is simpler

PR #2202, opened 2025-03-14
feat: add BlobAndProofV2

### LINKED ISSUE #2201: Add `BlobAndProofV2`
Right now we have `BlobAndProofV1`:
https://github.com/alloy-rs/alloy/blob/5ed9660de06adf67db8c5fbe02b20295b04e9524/crates/eips/src/eip4844/engine.rs#L1-L14

A struct exactly like this should be added, called `BlobAndProofV2`, which contains the same fields as `BlobAndProofV1` but with `cell_proof` instead of `proof`. The serde struct cfg will have to be adjusted to use camel case on all fields as well.

This link should be included in the docs as well:
https://github.com/ethereum/execution-apis/pull/630/

varun-doshi:
## Motivation

Add`BlobAndProofV2`

Closes #2201 

## Solution

Added `BlobAndProofV2` type

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip4844/engine.rs
+++ b/crates/eips/src/eip4844/engine.rs
@@ -20,6 +20,6 @@ pub struct BlobAndProofV1 {
 pub struct BlobAndProofV2 {
     /// The blob data.
     pub blob: Box<Blob>,
-    /// The KZG proof for the blob.
+    /// The cell proof for the blob.
     pub cell_proof: Bytes48,
 }
--- a/crates/eips/src/eip4844/engine.rs
+++ b/crates/eips/src/eip4844/engine.rs
@@ -12,3 +12,14 @@ pub struct BlobAndProofV1 {
     /// The KZG proof for the blob.
     pub proof: Bytes48,
 }
+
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip4844/engine.rs
```diff
@@ -12,3 +12,14 @@ pub struct BlobAndProofV1 {
...
 }
+
+/// Blob type returned in responses to `engine_getBlobsV2`: <https://github.com/ethereum/execution-apis/pull/630>
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct BlobAndProofV2 {
+    /// The blob data.
+    pub blob: Box<Blob>,
+    /// The KZG proof for the blob.
```
mattsse, line 23: this doc we need to update that this is the cell proof

mattsse, line 23:
```suggestion
    /// The cell proof for the blob.
```

mattsse (approved):

Issue #2199, opened 2025-03-14
[Feature] Add future type for Subscriptions


mattsse:
### Component

provider, pubsub

### Describe the feature you would like

ref https://github.com/alloy-rs/alloy/issues/2188

This doesn't support providing subscription settings

https://github.com/alloy-rs/alloy/blob/b648a7fdc344d77a7ccdae5746a3f722d4a521d5/crates/provider/src/fillers/mod.rs#L608-L612

the only way to manually configure the channel buffer would be doing this manually:

https://github.com/alloy-rs/alloy/blob/b648a7fdc344d77a7ccdae5746a3f722d4a521d5/crates/provider/src/provider/trait.rs#L938-L943

and setting the channel size:

https://github.com/alloy-rs/alloy/blob/b648a7fdc344d77a7ccdae5746a3f722d4a521d5/crates/pubsub/src/frontend.rs#L96-L103

we replace this asnyc function with a function that returns a future that allows this config so you can do

```rust
subscribe_blocks().buffer(256).await
```

## TODO
* introduce future type for subscription that does https://github.com/alloy-rs/alloy/blob/b648a7fdc344d77a7ccdae5746a3f722d4a521d5/crates/provider/src/provider/trait.rs#L938-L943 internally

this would likely be useful for #2187 as well

cc @naps62

frankudoags:
Hi @mattsse, please can I work on this?

mattsse:
assigned

lmk if you need pointers

PR #2203, opened 2025-03-15
feat: define subscription type

frankudoags:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/2199

Add a custom subscription type that can be configured using the builder pattern

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


frankudoags:
@mattsse please take a look when you can to help confirm I'm on track

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,71 @@
...
+{
+    /// Creates a new [`EthSuscribe`] instance
+    pub fn new(root: RootProvider<N>, params: P) -> Self {
+        Self { root, channel_size: None, params, _marker: std::marker::PhantomData }
+    }
+
+    /// Set the channel_size for the subscription stream.
+    pub fn buffer(mut self, size: usize) -> Self {
+        const MAX_BUFFER_SIZE: usize = 1024; // upper limit
+        self.channel_size = Some(size.min(MAX_BUFFER_SIZE));
```
mattsse, line 35: we can remove this check and leave this to the user

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,71 @@
+use crate::{Provider, RootProvider};
+use alloy_json_rpc::{RpcRecv, RpcSend};
+use alloy_network::{Ethereum, Network};
+use alloy_transport::TransportResult;
+use std::{future::Future, pin::Pin};
+
+/// Future type Subscription struct that wraps client requests to `eth_subscribe`
+/// Allows configuration of channel size
+pub struct EthSuscribe<P, R, N = Ethereum>
```
mattsse, line 9: we can just call this GetSubscription

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,71 @@
+use crate::{Provider, RootProvider};
+use alloy_json_rpc::{RpcRecv, RpcSend};
+use alloy_network::{Ethereum, Network};
+use alloy_transport::TransportResult;
+use std::{future::Future, pin::Pin};
+
+/// Future type Subscription struct that wraps client requests to `eth_subscribe`
+/// Allows configuration of channel size
```
mattsse, line 8: we can change this description slightly so that this a general purpose subscription

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,71 @@
...
+pub struct EthSuscribe<P, R, N = Ethereum>
+where
+    P: RpcSend,
+    R: RpcRecv,
+    N: Network,
+{
+    root: RootProvider<N>,
```
mattsse, line 15: what we also want here is a `method: Cow<'static str>` because then we can reuse this type for other subscription setups as well

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,71 @@
...
+    R: RpcRecv,
+{
+    type Output = TransportResult<alloy_pubsub::Subscription<R>>;
+    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'static>>;
+
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(async move {
+            self.root
+                .pubsub_frontend()?
+                .set_channel_size(self.channel_size.unwrap_or(16)); //default size
```
mattsse, line 65: this now adds another constant, what we should do instead is only set this if a buffer has been set via if let Some

and we need to explicitly mark this request as subscription, like:

https://github.com/alloy-rs/alloy/blob/b648a7fdc344d77a7ccdae5746a3f722d4a521d5/crates/provider/src/ext/admin.rs#L77-L81

mattsse (changes requested):
cool, this is looking pretty good already,

I mostly have nits

crates/provider/src/provider/subscription.rs
frankudoags: done

crates/provider/src/provider/subscription.rs
frankudoags: done

crates/provider/src/provider/subscription.rs
frankudoags: Ah, I see. Very nice, will add. This type can then handle all subscription cases and not just `eth_suscribe` calls, description comment  and set_is_subscription nit are all related to this. 

This will lead to a lot of re-write tho, will that happen in this pr, or fresh pr after this type has been merged? @mattsse 

I'd also love to work on that too :)

crates/provider/src/provider/subscription.rs
frankudoags: done

crates/provider/src/provider/subscription.rs
frankudoags: done

mattsse (approved):
lgtm

pending @yash-atreya for vis

next we can integrate this in existing pubsub fns but for this pr this is okay

yash-atreya (changes requested):
looking good, 
Needs `cargo fmt`

crates/provider/src/provider/subscription.rs
```diff
@@ -0,0 +1,99 @@
...
+            root,
+            method: method.into(),
+            channel_size: None,
+            params,
+            _marker: std::marker::PhantomData,
+        }
+    }
+
+    /// Set the channel_size for the subscription stream.
+    pub fn buffer(mut self, size: usize) -> Self {
```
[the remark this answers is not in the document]
frankudoags: done

mattsse:
@frankudoags do you want to continue with rolling this type out for the subscription functions?

frankudoags:
> @frankudoags do you want to continue with rolling this type out for the subscription functions?

Yes please

PR #2208, opened 2025-03-17
update tx fee comment about Transaction trait

Pana:
## Motivation
Because not long 1559 have max_fee and max_priority_fee, but also 4844 and 7702, so these comment should be updated

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
ty, this makes sense

PR #2212, opened 2025-03-17
fix(pubsub): wrap channel_size with Arc

0xBurbo:
## Motivation

When attempting to increase the PubsubFrontend channel size to accommodate higher throughput, the current handling of channel_size turns out to be inaccessible. This is because it is stored as an AtomicUsize, but is not stored with an Arc making it independent for each PubsubFrontend during cloning.

This poses an issue when initializing a PubsubFrontend using WsConnect:

```rust
let mut ws_config = WebSocketConfig::default();
ws_config.max_frame_size = None;
ws_config.max_message_size = None;
let mut ws = WsConnect::new(self.ws_url.clone()).with_config(ws_config);

let client = ClientBuilder::default().ws(ws).await?;
client.set_channel_size(self.pubsub_channel_size);

let provider: _WsProvider = ProviderBuilder::new().on_client(client);
```

During the ClientBuilder.ws() initialization of PubsubFrontend, it is cloned before being returned to the user to adjust the channel size. This means it is impossible to actually change the channel size, as the PubsubFrontend used to dispatch subscriptions is not the same object as the one returned by ClientBuilder::default().ws(ws).await.

## Solution

The solution is very simple, wrapping an Arc<> around the channel_size AtomicUint inside of PubsubFrontend. This low-impact change means that changes to channel sizes will be reflected across all PubsubFrontends used by the provider. Without this solution, is is difficult/impossible to easily adjust the channel size in this situation.

### DIFF
--- a/crates/pubsub/src/frontend.rs
+++ b/crates/pubsub/src/frontend.rs
@@ -5,7 +5,7 @@ use alloy_transport::{TransportError, TransportErrorKind, TransportFut, Transpor
 use futures::{future::try_join_all, FutureExt, TryFutureExt};
 use std::{
     future::Future,
-    sync::atomic::{AtomicUsize, Ordering},
+    sync::{atomic::{AtomicUsize, Ordering}, Arc},
     task::{Context, Poll},
 };
 use tokio::sync::{mpsc, oneshot};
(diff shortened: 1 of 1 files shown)


mattsse (commented):
I think this was originally intended like this so you can set this per pubsub instance

need to take a closer look if we should do this

0xBurbo:
I figured that was the case - an easy way to see where it goes wrong is by logging when the Pubsub frontend is cloned - you'll see it gets cloned twice during `ClientBuilder::default().ws(ws).await`. All subscription requests seem to use the old (cloned) pubsub frontends to dispatch subscription requests rather than the new ones regardless of setting (will always yield a channel size of 16). If this is not intended I suppose it would be a larger issue.

If you decided to continue to allow this per-frontend behavior then it would make sense to discard this and open an issue to find another solution, my thinking was that if really needed multiple different buffer sizes they would create additional providers, but on further reflection I see how that would require more connection overhead than separating channel sizes per-pubsub.

mattsse (approved):
I think this makes sense

blocking this until #2203 is rolled out because this change her would have some side-effects

DaniPopes (approved):

mattsse (approved):
we can do this now

especially with #2219

PR #2215, opened 2025-03-18
feat(`provider`): subscribe_full_blocks

yash-atreya:
## Motivation

2nd part of #325 

## Solution

- Introduce type `SubFullBlock` 
- Consumes the `"newHeads"` subscriptions and maps the stream to return full `BlockResponse`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/provider/trait.rs
```diff
@@ -943,6 +945,38 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// while let Some(block) = stream.next().await {
+    ///   println!("{block:#?}");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "pubsub")]
+    async fn subscribe_full_blocks(&self) -> TransportResult<SubFullBlocks<N>> {
+        self.root().pubsub_frontend()?;
+        let id = self.client().request("eth_subscribe", ("newHeads", true)).await?;
```
mattsse, line 974: hmm, why does his take a bool value?

crates/provider/src/provider/trait.rs
```diff
@@ -943,6 +945,38 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    ///
+    /// while let Some(block) = stream.next().await {
+    ///   println!("{block:#?}");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "pubsub")]
+    async fn subscribe_full_blocks(&self) -> TransportResult<SubFullBlocks<N>> {
+        self.root().pubsub_frontend()?;
```
mattsse, line 973: I think this should return GetSubscription because now we no way to configure the channel size

mattsse, line 973: I think this should return GetSubscription because now we no way to configure the channel size

mattsse (changes requested):

crates/provider/src/provider/trait.rs
yash-atreya: We'll need to write mapping logic, it is easier to embed `GetSubscription` in `SubFullBlocks` and expose `channel_size`

yash-atreya: [75237b2](https://github.com/alloy-rs/alloy/pull/2215/commits/75237b2b7815cacf17bde98cc402ebfd4030a694)

mattsse (approved):
lgtm

PR #2219, opened 2025-03-19
chore(`provider`): use `WeakClient` in `GetSubscription`

yash-atreya:
## Motivation

The `GetSubscription` future type was introduced in https://github.com/alloy-rs/alloy/pull/2203
Remove it's reliance on `N: Network` by using `WeakClient`

## Solution

- Remove `N: Network` generic from `GetSubscription` by using `WeakClient`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -98,7 +98,7 @@ ci_info.workspace = true
 similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "pubsub", "ws"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
 reqwest = [
     "dep:reqwest",
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -98,7 +98,7 @@ ci_info.workspace = true
 similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls", "pubsub", "ws"]
+default = ["reqwest", "reqwest-default-tls"]
 pubsub = ["alloy-rpc-client/pubsub", "dep:alloy-pubsub"]
 reqwest = [
     "dep:reqwest",
--- a/crates/provider/src/provider/subscription.rs
+++ b/crates/provider/src/provider/subscription.rs
@@ -59,8 +59,10 @@ where
 
     fn into_future(self) -> Self::IntoFuture {
         Box::pin(async move {
-            let client =
-                self.client.upgrade().ok_or(TransportErrorKind::custom_str("client dropped"))?;
+            let client = self
+                .client
+                .upgrade()
+                .ok_or_else(|| TransportErrorKind::custom_str("client dropped"))?;
             let pubsub = client.pubsub_frontend().ok_or(TransportErrorKind::PubsubUnavailable)?;
 
             // Set config channel size if any
--- a/crates/provider/src/provider/subscription.rs
+++ b/crates/provider/src/provider/subscription.rs
@@ -1,79 +1,32 @@
-use alloy_json_rpc::RpcRecv;
+use alloy_json_rpc::{RpcRecv, RpcSend};
+use alloy_primitives::B256;
 use alloy_pubsub::Subscription;
-use alloy_rpc_client::WeakClient;
-use alloy_rpc_types_eth::{
-    pubsub::{Params, SubscriptionKind},
-    Filter,
-};
+use alloy_rpc_client::{RpcCall, WeakClient};
 use alloy_transport::{TransportErrorKind, TransportResult};
 
-/// A builder for `"eth_subscribe"`  requests.
+/// A general-purpose subscription request builder
 ///
 /// This struct allows configuring subscription parameters and channel size
 /// before initiating a request to subscribe to Ethereum events.
-///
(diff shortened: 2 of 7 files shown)


crates/provider/Cargo.toml
```diff
@@ -98,7 +98,7 @@ ci_info.workspace = true
 similar-asserts.workspace = true
 
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "pubsub", "ws"]
```
mattsse, line 101: this is breaking and pulls in more deps by default, we should avoid this

crates/provider/src/provider/trait.rs
```diff
@@ -935,12 +937,9 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
     /// # }
     /// ```
     #[cfg(feature = "pubsub")]
-    async fn subscribe_blocks(
-        &self,
-    ) -> TransportResult<alloy_pubsub::Subscription<N::HeaderResponse>> {
-        self.root().pubsub_frontend()?;
-        let id = self.client().request("eth_subscribe", ("newHeads",)).await?;
-        self.root().get_subscription(id).await
+    fn subscribe_blocks(&self) -> GetSubscription<(String,), N::HeaderResponse> {
```
mattsse, line 940: looks like we can use:

https://github.com/alloy-rs/alloy/blob/bc190c61f1bee74271a29ea6275ab674fcfe73d5/crates/rpc-types-eth/src/pubsub.rs#L76-L83

here?

mattsse (changes requested):
nice,

some nits

crates/provider/src/provider/trait.rs
yash-atreya: [daae6e5](https://github.com/alloy-rs/alloy/pull/2219/commits/daae6e590c6bb2d743b381bccbb009d34b5ec310)

crates/provider/src/provider/subscription.rs
```diff
@@ -1,99 +1,140 @@
...
 where
-    P: RpcSend,
     R: RpcRecv,
-    N: Network,
 {
-    root: RootProvider<N>,
-    method: Cow<'static, str>,
-    params: Option<P>,
+    client: WeakClient,
+    kind: SubscriptionKind,
```
mattsse, line 23: this we can't do because then this type no longer works for non eth_subscribe subscription endpoints, so we need to keep a generic for this

crates/provider/src/provider/subscription.rs
```diff
@@ -1,99 +1,140 @@
...
+    params: Params,
```
mattsse, line 24: isn't this the same as params?

not obvious why we need both kind + params here

mattsse (changes requested):

crates/provider/src/provider/subscription.rs
yash-atreya: SubscriptionKind can be `Syncing` or `NewHeads`, params for both are `Params::None`

crates/provider/src/provider/subscription.rs
yash-atreya: [2861b57](https://github.com/alloy-rs/alloy/pull/2219/commits/2861b573007a9dad3c3269c4815ec1f960187000)

crates/provider/src/provider/subscription.rs
```diff
@@ -49,51 +36,41 @@ where
...
 {
     type Output = TransportResult<alloy_pubsub::Subscription<R>>;
-    type IntoFuture = Pin<Box<dyn Future<Output = Self::Output> + Send + 'static>>;
+    type IntoFuture = futures_utils_wasm::BoxFuture<'static, Self::Output>;
 
     fn into_future(self) -> Self::IntoFuture {
         Box::pin(async move {
-            let pubsub = self.root.pubsub_frontend()?;
+            let client =
+                self.client.upgrade().ok_or(TransportErrorKind::custom_str("client dropped"))?;
```
DaniPopes, line 63: ok_or_else

crates/provider/src/provider/subscription.rs
```diff
@@ -1,45 +1,32 @@
...
+use alloy_pubsub::Subscription;
+use alloy_rpc_client::{RpcCall, WeakClient};
+use alloy_transport::{TransportErrorKind, TransportResult};
 
 /// A general-purpose subscription request builder
 ///
 /// This struct allows configuring subscription parameters and channel size
 /// before initiating a request to subscribe to Ethereum events.
-pub struct GetSubscription<P, R, N = Ethereum>
+pub struct GetSubscription<P, R>
```
DaniPopes, line 11: please also gate the entire module to pubsub feature as it's otherwise useless and unused

DaniPopes (changes requested):

crates/provider/src/provider/subscription.rs
yash-atreya: https://github.com/alloy-rs/alloy/blob/210562f8ed4cfaf5c1cd7063e4c94ebafdaceba7/crates/provider/src/provider/mod.rs#L33

crates/provider/src/provider/subscription.rs
yash-atreya: Isn't ok_or cleaner? 
Regarless [5a8d29f](https://github.com/alloy-rs/alloy/pull/2219/commits/5a8d29fe23783971444b6f289b089c3079a60f39)

DaniPopes (approved):

PR #2222, opened 2025-03-19
fix(`pubsub`): fix race condition in ActiveSub

### LINKED ISSUE #1601: [Bug] PubSubService does not subscribe fast enough to the ActiveSubscription Sender and drops first message
### Component

provider, pubsub

alloy-provider v0.5.4

### Operating System

macOS (Apple Silicon)

### Describe the bug

When registering a JSON RPC subscription with WebSocket, the pubsub frontend sends an async request to the service loop to subscribe to the broadcast channel, by the time the receiver is returned by the frontend, any message that was sent by the server in the meantime is missed.

To reproduce, simply run this test:
```rust
  async fn run_server() -> eyre::Result<std::net::SocketAddr> {
        use jsonrpsee::server::{RpcModule, Server, SubscriptionMessage};

        let server = Server::builder().build("127.0.0.1:0").await?;
        let mut module = RpcModule::new(());
        module
            .register_subscription(
                "subscribe_hello",
                "s_hello",
                "unsubscribe_hello",
                |_, pending, _, _| async move {
                    let sub = pending.accept().await.unwrap();

                    for i in 0..usize::MAX {
                        let msg = SubscriptionMessage::from_json(&i).unwrap();
                        sub.send(msg).await.unwrap();
                        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
                    }

                    Ok(())
                },
            )
            .unwrap();
        let addr = server.local_addr()?;

        let handle = server.start(module);

        tokio::spawn(handle.stopped());

        Ok(addr)
    }

    #[tokio::test]
    async fn test_subscription() -> eyre::Result<()> {
        use alloy_provider::{Provider, ProviderBuilder};

        let addr = run_server().await?;

        let ws_provider = ProviderBuilder::new()
            .with_recommended_fillers()
            .on_builtin(&format!("ws://{}", addr).as_str())
            .await?;
        let mut request = ws_provider.client().request("subscribe_hello", ());
        // required if not eth_subscribe
        request.set_is_subscription();
        let sub_id = request.await?;
        // call the pubsub service to get a broadcast receiver.
        let mut sub = ws_provider.root().get_subscription(sub_id).await?;

        let num: usize = sub.recv().await.unwrap();
        assert_eq!(num, 0);

        Ok(())
    }
```
You will find the first message to be `1` which means the client missed the `0`.

### LINKED ISSUE #2187: [Feature] Fix race condition on subscription setup
### Component

provider, pubsub

fix for #1601 which outlined a race condition when we retrieve the subscription for the first time, this is because we drop the the rx here:

https://github.com/alloy-rs/alloy/blob/b78ea8b06fe69b23fbafecc2c5ea3827b8328570/crates/pubsub/src/managers/active_sub.rs#L61-L61

and by the time

```
let mut sub = ws_provider.root().get_subscription(sub_id).await?;
```

we only subscribe to new values.

https://github.com/alloy-rs/alloy/blob/b78ea8b06fe69b23fbafecc2c5ea3827b8328570/crates/pubsub/src/managers/active_sub.rs#L74-L74

we can fix this race in two ways

1. either change the set up in such a way that

```
request.set_is_subscription();
let sub_id = request.await?;
```

this directly awaits into a subscription

2. don't drop the receiver until the first subscription is retrieved

we can do both 1 and 2 imo, 2 should fix this issue right away

@yash-atreya 

### Additional context

_No response_

yash-atreya:
## Motivation

Closes #2187 
Closes #1601 

## Solution

- Retain the initial `Receiver` channel in `ActiveSubscription`
- Return the initial `Receiver` instead of subscribing for a new one if there pending notifs over that channel.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/pubsub/src/managers/active_sub.rs
```diff
@@ -71,7 +72,33 @@ impl ActiveSubscription {
...
+        let lock = self.rx.lock();
+
+        if lock.is_err() {
+            return RawSubscription { rx: new_rx, local_id: self.local_id };
+        }
+        if let Some(mut rx) = lock.unwrap().take() {
+            // Attempt to drain the queue by forwarding all pending notifications to the new
+            // receiver.
+
+            while let Ok(notification) = rx.try_recv() {
```
mattsse, line 92: this looks a bit weird, don't think we need any of this

crates/pubsub/src/managers/active_sub.rs
```diff
@@ -2,18 +2,19 @@ use crate::RawSubscription;
 use alloy_json_rpc::SerializedRequest;
 use alloy_primitives::B256;
 use serde_json::value::RawValue;
-use std::{fmt, hash::Hash};
+use std::{fmt, hash::Hash, sync::Mutex};
```
mattsse, line 5: no std mutex

crates/pubsub/src/managers/active_sub.rs
yash-atreya: Just retaining the initial receiver in ActiveSub doesn't work because the new receiver always starts from the tail-end of the previous, hence the need for forwarding the messages.

crates/pubsub/src/managers/active_sub.rs
yash-atreya: [6bb0177](https://github.com/alloy-rs/alloy/pull/2222/commits/6bb017778222067583e688f4b942a2c8f9b803e8)

crates/pubsub/src/managers/active_sub.rs
```diff
@@ -71,7 +73,29 @@ impl ActiveSubscription {
...
+            // receiver.
+
+            while let Ok(notification) = rx.try_recv() {
+                self.notify(notification);
+                pending -= 1;
+                if pending == 0 {
+                    break;
+                }
+            }
+        }
```
mattsse, line 96: unclear why we're doing this,

why don't we use this buffered subscription if this is the first one?

mattsse (changes requested):

crates/pubsub/src/managers/active_sub.rs
yash-atreya: [eef1b27](https://github.com/alloy-rs/alloy/pull/2222/commits/eef1b2710bb853f9caf68ce6b3f76a6334024db2)

crates/pubsub/src/managers/active_sub.rs
```diff
@@ -1,19 +1,21 @@
...
-#[derive(Clone)]
 pub(crate) struct ActiveSubscription {
     /// Cached hash of the request, used for sorting and equality.
     pub(crate) local_id: B256,
     /// The serialized subscription request.
     pub(crate) request: SerializedRequest,
     /// The channel via which notifications are broadcast.
     pub(crate) tx: broadcast::Sender<Box<RawValue>>,
+    /// The channel via which notifications are received.
+    pub(crate) rx: Mutex<Option<broadcast::Receiver<Box<RawValue>>>>,
```
mattsse, line 18: we need to properly document why we are even doing this

crates/pubsub/src/managers/active_sub.rs
```diff
@@ -71,7 +73,13 @@ impl ActiveSubscription {
 
     /// Get a subscription.
     pub(crate) fn subscribe(&self) -> RawSubscription {
-        RawSubscription { rx: self.tx.subscribe(), local_id: self.local_id }
+        if self.tx.is_empty() {
+            return RawSubscription { rx: self.tx.subscribe(), local_id: self.local_id };
+        }
+        RawSubscription {
+            rx: self.rx.lock().deref_mut().take().unwrap_or(self.tx.subscribe()),
```
mattsse, line 80: this also needs docs

mattsse (changes requested):
more pedantic doc nits

crates/pubsub/src/managers/active_sub.rs
yash-atreya: [78bd472](https://github.com/alloy-rs/alloy/pull/2222/commits/78bd472caea78aff13c862787c8ecff16cc8e68b)

crates/pubsub/src/managers/active_sub.rs
yash-atreya: [78bd472](https://github.com/alloy-rs/alloy/pull/2222/commits/78bd472caea78aff13c862787c8ecff16cc8e68b)

mattsse (approved):
this should do it as a hacky workaround

crates/pubsub/src/managers/active_sub.rs
```diff
@@ -71,7 +79,20 @@ impl ActiveSubscription {
...
+            // subscriber.
+            return RawSubscription { rx: self.tx.subscribe(), local_id: self.local_id };
+        }
+
+        // If there are pending notifications, we need to ensure that they are not dropped.
+        // Hence, we first try to return the initial receiver (if it exists), which will receive
+        // those pending notifications.
+        // Ref: <https://github.com/alloy-rs/alloy/issues/2187>
+        RawSubscription {
+            rx: self.rx.lock().deref_mut().take().unwrap_or(self.tx.subscribe()),
```
DaniPopes, line 93: unwrap_or_else, subscribe has side effects

Issue #2214, opened 2025-03-18
auto_impl(&) for Encodable2718


kpp:
https://github.com/alloy-rs/alloy/blob/d964f082a745c69e084749dce3d384201138bc25/crates/eips/src/eip2718.rs#L181

Because `proofs::calculate_transaction_root` and `proofs::calculate_receipt_root` accept an arg `&[T] where T: Encodable2718`, so `&[&TransactionSigned]` is not accepted which is odd. It adds extra copy/clone for every elem in the list and alloc for the list to create `&[&T]` -> `Vec<T>` -> `&[T] where T: Encodable2718` or `T=TransactionSigned`.

mattsse:
I see,

I think we could do auto_impl for these, don't think this will would cause issues with derefs etc., wdyt @DaniPopes

gitToki:
Could I take this issue? I should be able to optimize it

PR #2230, opened 2025-03-20
auto_impl(&) for Encodable2718

gitToki:
Add auto_impl(&) to Encodable2718 to avoid unnecessary clones. Link to [issues](https://github.com/alloy-rs/alloy/issues/2214).

### DIFF
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -179,7 +179,7 @@ pub trait Decodable2718: Sized {
 /// over the accepted transaction types.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
-#[auto_impl(keep_default_for(&))]
+#[auto_impl(&)]
 pub trait Encodable2718: Typed2718 + Sized + Send + Sync {
     /// Return the type flag (if any).
     ///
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip2718.rs
```diff
@@ -224,9 +226,9 @@ pub trait Encodable2718: Typed2718 + Sized + Send + Sync {
     }
 
     /// Seal the encodable, by encoding and hashing it.
-    fn seal(self) -> Sealed<Self> {
+    fn seal(&self) -> Sealed<Self> where Self: Clone {
```
mattsse, line 229: I think we can keep this as it was if we add

`#[auto_impl(keep_default_for(&))]`

mattsse (changes requested):

crates/eips/src/eip2718.rs
gitToki: Implemented in the latest change, ty for suggestion

crates/eips/src/eip2718.rs
```diff
@@ -178,6 +179,7 @@ pub trait Decodable2718: Sized {
 /// over the accepted transaction types.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
+#[auto_impl(keep_default_for(&))]
```
mattsse, line 182: this must go on the function that returns `Self`

and here we need to add just auto_impl(&)

mattsse (changes requested):

mattsse (approved):
gg

DaniPopes (approved):

PR #2236, opened 2025-03-24
feat(`provider`): eth_signTransaction

yash-atreya:
## Motivation

Closes #2235 

## Solution

- Introduces `eth_signTransaction` to the provider trait
- Add to `FillProvider` and `DynProvider` impls

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #2238, opened 2025-03-24
feat(`signers`): `Web3Signer`

### LINKED ISSUE #2237: [Feature] introduce `Web3Signer`
### Component

signers

Ref: https://t.me/ethers_rs/43582

Introduce [`Web3Signer`](https://docs.web3signer.consensys.io/reference/api/json-rpc) remote signer type in a separate crate.

- This should wrap the `Provider` in order to use `eth_signTransaction` for signing requests.

- Impl `TxSigner` for `Web3Signer`

Needs #2235 

### Additional context

_No response_

yash-atreya:
## Motivation

Builds on #2236 

Closes #2237 

## Solution

- Introduce `Web3Signer` in a separate crate `alloy-signer-web3`
- `Web3Signer` wraps the provider and uses `eth_signTransaction` from #2236 
- impl `TxSigner` for `Web3Signer`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer-web3/Cargo.toml
+++ b/crates/signer-web3/Cargo.toml
@@ -1,0 +1,38 @@
+[package]
+name = "alloy-signer-web3"
+description = "Ethereum Web3 signer"
+
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+authors.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+exclude.workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = [
+    "-Zunstable-options",
+    "--generate-link-to-definition",
+    "--show-type-layout",
+]
+
+[lints]
+workspace = true
+
+[dependencies]
+alloy-consensus = { workspace = true, features = ["std"] }
+alloy-network.workspace = true
+alloy-primitives.workspace = true
+alloy-signer.workspace = true
+alloy-provider.workspace = true
+
+async-trait.workspace = true
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
+[features]
+eip712 = ["alloy-signer/eip712"]
--- a/crates/provider/Cargo.toml
+++ b/crates/provider/Cargo.toml
@@ -71,7 +71,7 @@ parking_lot.workspace = true
 wasmtimer.workspace = true
 
 [dev-dependencies]
-alloy-consensus = { workspace = true, features = ["kzg"] }
+alloy-consensus = { workspace = true, features = ["kzg", "k256"] }
 alloy-primitives = { workspace = true, features = ["rand"] }
 alloy-node-bindings.workspace = true
 alloy-rpc-client = { workspace = true, features = ["reqwest"] }
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -47,7 +47,7 @@ use alloy_rpc_types_eth::{
     AccessListResult, EIP1186AccountProofResponse, EthCallResponse, FeeHistory, Filter,
     FilterChanges, Log,
 };
-use alloy_transport::TransportResult;
+use alloy_transport::{TransportError, TransportResult};
 use async_trait::async_trait;
 use futures_utils_wasm::impl_future;
 use serde_json::value::RawValue;
@@ -607,6 +607,12 @@ where
 
         // Errors in tx building happen further down the stack.
         self.inner.send_transaction_internal(tx).await
+    }
+
+    async fn sign_transaction(&self, tx: N::TransactionRequest) -> TransportResult<Bytes> {
+        let tx = self.fill(tx).await?;
+        let tx = tx.try_into_request().map_err(TransportError::local_usage)?;
+        self.inner.sign_transaction(tx).await
     }
 
     #[cfg(feature = "pubsub")]
--- a/crates/provider/src/provider/erased.rs
+++ b/crates/provider/src/provider/erased.rs
@@ -353,6 +353,10 @@ impl<N: Network> Provider<N> for DynProvider<N> {
         self.0.send_transaction_internal(tx).await
     }
 
+    async fn sign_transaction(&self, tx: N::TransactionRequest) -> TransportResult<Bytes> {
+        self.0.sign_transaction(tx).await
+    }
+
     #[cfg(feature = "pubsub")]
     fn subscribe_blocks(&self) -> GetSubscription<(SubscriptionKind,), N::HeaderResponse> {
         self.0.subscribe_blocks()
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -913,6 +913,12 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
         }
     }
 
+    /// Signs a transaction that can be submitted to the network later using
+    /// [`Provider::send_raw_transaction`].
+    async fn sign_transaction(&self, tx: N::TransactionRequest) -> TransportResult<Bytes> {
+        self.client().request("eth_signTransaction", (tx,)).await
+    }
+
     /// Subscribe to a stream of new block headers.
     ///
     /// # Errors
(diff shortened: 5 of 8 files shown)


crates/signer-web3/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+workspace = true
+
+[dependencies]
+alloy-consensus = { workspace = true, features = ["std"] }
+alloy-network.workspace = true
+alloy-primitives.workspace = true
+alloy-signer.workspace = true
+alloy-provider.workspace = true
+
+async-trait.workspace = true
```
mattsse, line 32: this doesn't need any additional deps, do we really need another crate for this?

mattsse (commented):
can we instead moved this into the provider crate?

don't think it's reasonable to introduce another crate just for this type

crates/consensus/src/transaction/mod.rs
```diff
@@ -184,6 +184,11 @@ pub trait Transaction: Typed2718 + fmt::Debug + any::Any + Send + Sync + 'static
     /// `None`.
     fn blob_versioned_hashes(&self) -> Option<&[B256]>;
 
+    /// Blob sidecar for eip4844 transaction.
+    fn blob_sidecar(&self) -> Option<&BlobTransactionSidecar> {
+        None
+    }
```
mattsse, line 190: dont think we want this because this isn't really part of the tx.

not fully understanding why this is required here

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,173 @@
...
+            request.set_gas_price(max_fee_or_gas_price);
+        }
+
+        if let Some(access_list) = tx.access_list() {
+            request.set_access_list(access_list.clone());
+        }
+
+        if let Some(sidecar) = tx.blob_sidecar() {
+            request.set_blob_sidecar(sidecar.clone());
+        }
```
mattsse, line 100: we can exclude support for 4844 in this impl because this is also not supported by web3signer

https://github.com/Consensys/web3signer/blob/bdf1e72169bf7f6800979b0c765647feea172dc5/core/src/main/java/tech/pegasys/web3signer/core/service/jsonrpc/handlers/signing/TransactionSerializer.java#L43

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,173 @@
...
+        if let Some(auth) = tx.authorization_list() {
+            request.set_authorization_list(auth.to_vec());
+        }
+
+        let raw = self.sign_transaction(request).await?;
+
+        let envelope =
+            TxEnvelope::decode_2718(&mut raw.as_ref()).map_err(alloy_signer::Error::other)?;
+
+        Ok(*envelope.signature())
```
mattsse, line 111: we should add a check here that the signed tx actually matches what we received as signable tx because otherwise we return a signature for a different tx object.

for this we need to ensure that encode_for_signing is the same for both

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,173 @@
...
+    N::TransactionRequest: TransactionBuilder7702 + TransactionBuilder4844,
+{
+    fn address(&self) -> Address {
+        self.address
+    }
+
+    async fn sign_transaction(
+        &self,
+        tx: &mut dyn SignableTransaction<Signature>,
+    ) -> alloy_signer::Result<Signature> {
```
mattsse, line 69: I think this is impl is potentially problematic because we blindly return a signature here but the signed object isn't necessarily the same, so we can only really do this if we ensure the signed object is the same.

because the server can set fields arbitrarily 

the way this will be used in internally in the provider stack therefor problematic because this only returns the signature

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,173 @@
...
+        mut tx: N::TransactionRequest,
+    ) -> alloy_signer::Result<Bytes> {
+        tx.set_from(self.address);
+        self.provider.sign_transaction(tx).await.map_err(alloy_signer::Error::other)
+    }
+}
+
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl<P, N> TxSigner<Signature> for Web3Signer<P, N>
```
mattsse, line 56: do we even need this impl?

mattsse (changes requested):
some more nits.

imo support for 4844 is out of scope for now because also not supported by via the eth_signTx endpoints

https://github.com/Consensys/web3signer/tree/master


the sing_tx impl is problematic because we return a signature for something that doesn't necessarily match the input tx.

I think this should rather be implemented like:

https://github.com/alloy-rs/alloy/blob/70ef51888e4a3411756030e0083408665d75ad50/crates/provider/src/fillers/wallet.rs#L92-L105

crates/provider/src/provider/web3_signer.rs
yash-atreya: [7577dfe](https://github.com/alloy-rs/alloy/pull/2238/commits/7577dfe631524f0e49f248a16c7a2704928ece38)

crates/provider/src/provider/web3_signer.rs
yash-atreya: [7577dfe](https://github.com/alloy-rs/alloy/pull/2238/commits/7577dfe631524f0e49f248a16c7a2704928ece38)

crates/provider/src/provider/web3_signer.rs
yash-atreya: [7577dfe](https://github.com/alloy-rs/alloy/pull/2238/commits/7577dfe631524f0e49f248a16c7a2704928ece38)

crates/consensus/src/transaction/mod.rs
yash-atreya: [7577dfe](https://github.com/alloy-rs/alloy/pull/2238/commits/7577dfe631524f0e49f248a16c7a2704928ece38)

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub async fn sign_and_decode(
+        &self,
+        tx: N::TransactionRequest,
+    ) -> alloy_signer::Result<N::TxEnvelope> {
+        let raw = self.sign_transaction(tx).await?;
+        N::TxEnvelope::decode_2718(&mut raw.as_ref()).map_err(alloy_signer::Error::other)
+    }
+}
+
+impl<N: Network> TxFiller<N> for Web3Signer<N> {
```
mattsse, line 75: is this then even useful?

the setup for this could be a bit complicated,

like in order to configure a web3signer filler you first need to create a provider

the example only shows how this is used as like a provider type, which seems more useful

crates/provider/src/provider/web3_signer.rs
```diff
@@ -0,0 +1,176 @@
...
+///
+/// Note:
+///
+/// `"eth_signTransaction"` is not supported by regular nodes.
+///
+/// [`ProviderBuilder`]: crate::ProviderBuilder
+#[derive(Debug, Clone)]
+pub struct Web3Signer<N: Network = Ethereum> {
+    /// The provider used to make `"eth_signTransaction"` requests.
+    provider: DynProvider<N>,
```
mattsse, line 25: this can now also be used as regular provider, but we don't expose the underlying provider so all this type can do sign stuff and not send stuff

mattsse (changes requested):
this entire feature is a bit odd.

not obvious how we invison this to be used.

to me it feels like we're trying to use the same type for two different use cases, one as filler and one as regular provider, both kinda conflict with each other.

but looks like the fillable API doesn't even allow this and doesn't provide access to the transport so we're kinda forced to first create a basic provider, create a web3singer instance and then use that as the filler for a new provider instance.

a bit convoluted but this kinda works, so this seems fine.
I also don't expect this to be used that much, but we need another example that shows how to use this as filler

crates/provider/src/provider/web3_signer.rs
yash-atreya: Removed [f7b7c5a](https://github.com/alloy-rs/alloy/pull/2238/commits/f7b7c5aaf2f87bafd5dca894e1cc45279c91ac1c)

mattsse (approved):

PR #2240, opened 2025-03-24
chore(deps): c-kzg 2.0

### LINKED ISSUE #1745: [Bug] Upgrade c-kzg dependency to 2.0
### Component

consensus, eips, genesis

_No response_

### Operating System

None

### Describe the bug

I can't use latest c-kzg library with alloy library.
```
error: failed to select a version for `c-kzg`.
    ... required by package `alloy-consensus v0.7.0`
    ... which satisfies dependency `alloy-consensus = "^0.7"` of package `alloy-network v0.7.0`
    ... which satisfies dependency `alloy-network = "^0.7"` of package `alloy-contract v0.7.0`
    ... which satisfies dependency `alloy-contract = "^0.7"` of package `eventlogger v0.1.0 (/Users/redacted/eventlogger)`
versions that meet the requirements `^1.0` are: 1.0.3, 1.0.2, 1.0.0

the package `c-kzg` links to the native library `ckzg`, but it conflicts with a previous package which links to `ckzg` as well:
package `c-kzg v2.0.0`
    ... which satisfies dependency `c-kzg = "^2.0.0"` (locked to 2.0.0) of package `apilsevice v0.1.0 (/Users/redacted/apilsevice)`
Only one package in the dependency graph may specify the same links value. This helps ensure that only one copy of a native library is linked in the final binary. Try to adjust your dependencies so that only one package uses the `links = "ckzg"` value. For more information, see https://doc.rust-lang.org/cargo/reference/resolver.html#links.

failed to select a version for `c-kzg` which could resolve this conflict
```

yash-atreya:
## Motivation

Closes #1745 

## Solution

- Bump c-kzg to 2.0 
- Address breaking changes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip4844/env_settings.rs
```diff
@@ -45,8 +45,10 @@ impl EnvKzgSettings {
         match self {
             Self::Default => {
                 let load = || {
-                    KzgSettings::load_trusted_setup(&G1_POINTS.0, &G2_POINTS.0)
-                        .expect("failed to load default trusted setup")
+                    // FIXME
+                    // KzgSettings::load_trusted_setup(&G1_POINTS.0, &G2_POINTS.0)
+                    //     .expect("failed to load default trusted setup")
+                    todo!()
```
yash-atreya, line 51: @mattsse @DaniPopes 

need some help initializing the default KzgSettings using https://docs.rs/c-kzg/latest/c_kzg/struct.KzgSettings.html#method.load_trusted_setup

crates/eips/src/eip4844/env_settings.rs
```diff
@@ -71,7 +73,9 @@ impl EnvKzgSettings {
     pub fn load_from_trusted_setup_file(
         trusted_setup_file: &std::path::Path,
     ) -> Result<Self, c_kzg::Error> {
-        let settings = KzgSettings::load_trusted_setup_file(trusted_setup_file)?;
+        // FIXME: Precompute value??
+        let precompute = 0;
+        let settings = KzgSettings::load_trusted_setup_file(trusted_setup_file, precompute)?;
```
yash-atreya, line 78: same here

@mattsse @DaniPopes what should the precompute value be?

crates/eips/src/eip4844/env_settings.rs
```diff
@@ -43,25 +42,7 @@ impl EnvKzgSettings {
...
-                {
-                    use once_cell as _;
-                    use std::sync::OnceLock;
-                    static DEFAULT: OnceLock<KzgSettings> = OnceLock::new();
-                    DEFAULT.get_or_init(load)
-                }
-                #[cfg(not(feature = "std"))]
-                {
-                    use once_cell::race::OnceBox;
-                    static DEFAULT: OnceBox<KzgSettings> = OnceBox::new();
```
yash-atreya, line 61: We don't need this anymore as the helper `c_kzg::ethereum_kzg_settings` does this under the hood

crates/eips/src/eip4844/env_settings.rs
```diff
@@ -43,25 +42,7 @@ impl EnvKzgSettings {
...
-                    DEFAULT.get_or_init(|| alloc::boxed::Box::new(load()))
-                }
-            }
+            Self::Default => c_kzg::ethereum_kzg_settings(0),
```
mattsse, line 45: this ckzg api is incredibly weird, because this allows providing an argument which is then ignored the second time this function is called because the kzgsettings are cached.
however, for this crate this is perfectly fine

mattsse (approved):
lgtm

crates/eips/src/eip4844/env_settings.rs
rakita: It is optimization for `compute_cells_and_kzg_proofs` and `recover_cells_and_kzg_proofs`, ref https://github.com/ethereum/c-kzg-4844?tab=readme-ov-file#precompute

It seems fine to leave at zero.

crates/eips/src/eip4844/env_settings.rs
DaniPopes: Ref https://github.com/ethereum/c-kzg-4844/blob/dffa18ee350aeef38f749ffad24a27c1645fb4f8/src/setup/setup.c#L401-L407

crates/eips/src/eip4844/env_settings.rs
```diff
@@ -71,7 +52,8 @@ impl EnvKzgSettings {
     pub fn load_from_trusted_setup_file(
         trusted_setup_file: &std::path::Path,
     ) -> Result<Self, c_kzg::Error> {
-        let settings = KzgSettings::load_trusted_setup_file(trusted_setup_file)?;
+        let precompute = 0; // See: <https://github.com/ethereum/c-kzg-4844/blob/dffa18ee350aeef38f749ffad24a27c1645fb4f8/README.md?plain=1#L112>
```
DaniPopes, line 55: can we extract this to a constant and also pass it in `fn get`?
0 looks fine per the README explanation, I don't think we use those functions so we don't want a value greater than 0

kevaundray: Yeah 0 should be fine -- the one minor caveat is if rollups are importing c-kzg via one of the reth crates when creating blob transactions.

When a rollup creates a blob, they will now call `compute_cells_and_kzg_proofs` on the blob and then send the blob plus the proofs to the EL.

For a rollup, they would likely want somewhere around 8 as that gives the best performance/memory trade-off for `compute_cells_and_kzg_proofs`

Saying that, its unclear if rollups need this method to be very fast, so perhaps 0 is okay for them

kevaundray:
@yash-atreya why is this marked as blocked?

mattsse:
no issue with the integration, but this is a breaking change and would further split crate compatibility until everything is migrated properly.

but I think we can now send 0.13

crates/eips/src/eip4844/env_settings.rs
yash-atreya: [1322198](https://github.com/alloy-rs/alloy/pull/2240/commits/132219872b9b3610a3e48beffb95d29828100a5e)

DaniPopes (approved):

PR #2244, opened 2025-03-24
added input-mut for TxEnvelope

Rimeeeeee:
fixes [#15245(reth)](https://github.com/paradigmxyz/reth/issues/15245)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -49,6 +49,19 @@ impl TxEnvelope {
...
     }
 
+    /// Returns a mutable reference to the transaction's input.
+    pub fn input_mut(&mut self) -> &mut Bytes {
+        match self {
+            Self::Eip1559(tx) => &mut tx.tx_mut().input,
+            Self::Eip2930(tx) => &mut tx.tx_mut().input,
+            Self::Legacy(tx) => &mut tx.tx_mut().input,
+            Self::Eip7702(tx) => &mut tx.tx_mut().input,
+            Self::Eip4844(tx) => match tx.tx_mut() {
```
mattsse, line 59: this we can relax a bit but for that we need to more changes:

`Asref<TxEip4884>` and `AsMut<TxEip4844>` impls for `TxEip4844Variant`

then we can make this function generic over EthereumTxEnvelope<T: AsMut<TxEip4844>>

mattsse (changes requested):
cool, one idea for relaxing the 4844 variant

crates/consensus/src/transaction/envelope.rs
Rimeeeeee: Will do as soon as I reach home :)

Rimeeeeee:
@mattsse , is this ok?

mattsse (approved):
clean, ty

Rimeeeeee:
@mattsse do i need to do something else or should I raise an pr in reth once this and the op-alloy changes are merged?

PR #2246, opened 2025-03-25
feat: add bincode compat for receipt envelope

### LINKED ISSUE #2242: [Bug] `ReceiptEnvelope` not bincode serializable because of use of `serde(flatten)` in `Receipt`
### Component

consensus, eips, genesis

alloy-consensus v0.11.0

### Operating System

macOS (Apple Silicon)

### Describe the bug

The following code gives me an error for a fairly normal receipt:

```rust
        let mut out_buf: Vec<u8> = Vec::new();
        bincode::serialize_into(&mut out_buf, &receipt).expect("serialization failed");
```

Researching a bit further this seems to have been an issue acknowledged as out-of-scope in the bincode repo (https://github.com/bincode-org/bincode/issues/245 & https://github.com/bincode-org/bincode/issues/167)

mattsse:
closes #2242

one more bincode helper type -.-

### DIFF
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -281,10 +281,11 @@ where
 }
 
 /// Bincode-compatible [`ReceiptEnvelope`] serde implementation.
-// #[cfg(all(feature = "serde", feature = "serde-bincode-compat"))]
+#[cfg(all(feature = "serde", feature = "serde-bincode-compat"))]
 pub(crate) mod serde_bincode_compat {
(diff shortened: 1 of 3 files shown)


shekhirin (approved):

Philogy:
@mattsse curious what the philosophy is behind making a new compatible type vs. just making the original type compatible?

Also thanks for the fast turn-around!

mattsse:
> just making the original type compatible?

this does not work if we want this type to also be embedded in the rpc types which may require things like flatten

PR #2249, opened 2025-03-27
chore: added error handling for command execution

defiberrys:
## Motivation  
The original script didn’t have explicit error handling, which could make troubleshooting more difficult in case of a failure. Adding error handling improves robustness and helps identify issues more clearly.

## Solution  
I’ve added an error handling block for the command execution (`"${cmd[@]}"`). This change ensures that any errors during the execution are caught and more explicit feedback is provided in case of failure.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/scripts/check_no_std.sh
+++ b/scripts/check_no_std.sh
@@ -18,4 +18,9 @@ for crate in "${crates[@]}"; do
 done
 
 echo "Running: ${cmd[*]}"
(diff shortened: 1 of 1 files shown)


mattsse (commented):
this is fine, this is only used in ci

PR #2257, opened 2025-03-29
feat: make it easier to configure non u256 topics in filterset

### LINKED ISSUE #2256: [Feature] Make it easier to configure non u256 topics in filterset
### Component

rpc

Value types (uint, int, address, bool, bytes32) are padded to 32 bytes in topics
Addresses are left-padded with zeros to 32 bytes
Boolean values are encoded as 32 bytes representing 0 or 1

we currently only accept:

https://github.com/alloy-rs/alloy/blob/d9ba0b990066a78d324ea71e57680e30acf8ab93/crates/contract/src/event.rs#L148-L148

https://github.com/alloy-rs/alloy/blob/d9ba0b990066a78d324ea71e57680e30acf8ab93/crates/rpc-types-eth/src/filter.rs#L146-L153

we need some helpers to add address,bool into Topic, left padding addresses is just

```rust
        pub fn address_to_topic(address: Address) -> Topic {
            let mut address_bytes = [0u8; 32];
            address_bytes[12..].copy_from_slice(address.as_slice());
            FilterSet::from(FixedBytes::<32>::from_slice(&address_bytes))
        }
```

### Additional context

_No response_

stevencartavia:
## Motivation
closes #2256

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/filter.rs
+++ b/crates/rpc-types-eth/src/filter.rs
@@ -152,6 +152,28 @@ impl From<U256> for Topic {
     }
 }
 
+impl From<Address> for Topic {
+    fn from(address: Address) -> Self {
+        let mut bytes = [0u8; 32];
+        bytes[12..].copy_from_slice(address.as_slice());
+        B256::from(bytes).into()
+    }
+}
+
+impl From<bool> for Topic {
+    fn from(value: bool) -> Self {
+        let mut bytes = [0u8; 32];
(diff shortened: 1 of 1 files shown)


mattsse (approved):
I think we also want these helpers for impl FilterSet<B256> so that we can easily insert new address,u256,bool

for convenience this can be like `fn extend<T: Into<Topic>>`

PR #2258, opened 2025-03-29
feat: filterset topics extend

stevencartavia:
## Motivation
helpers for impl FilterSet to insert new address, u256, bool

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/filter.rs
+++ b/crates/rpc-types-eth/src/filter.rs
@@ -38,6 +38,14 @@ impl BloomFilter {
 #[cfg_attr(feature = "serde", derive(serde::Deserialize))]
 pub struct FilterSet<T: Eq + Hash>(HashSet<T>);
 
+impl FilterSet<Topic> {
+    /// Extends the filter set with a value that can be converted into a Topic
+    pub fn extend<T: Into<Topic>>(mut self, value: T) -> Self {
+        self.0.insert(value.into());
+        self
+    }
+}
+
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/filter.rs
```diff
@@ -38,6 +38,14 @@ impl BloomFilter {
 #[cfg_attr(feature = "serde", derive(serde::Deserialize))]
 pub struct FilterSet<T: Eq + Hash>(HashSet<T>);
 
+impl FilterSet<Topic> {
```
mattsse, line 41: this isn't quite right:

https://github.com/alloy-rs/alloy/blob/72a12eab59df688abce53da845ee66dcc1060c86/crates/rpc-types-eth/src/filter.rs#L147-L147

we basically want `impl Topic` here

`FilterSet<Topic>` is `FilterSet<FilterSet<B256>>`

crates/rpc-types-eth/src/filter.rs
```diff
@@ -38,6 +38,14 @@ impl BloomFilter {
...
+    /// Extends the filter set with a value that can be converted into a Topic
+    pub fn extend<T: Into<Topic>>(mut self, value: T) -> Self {
+        self.0.insert(value.into());
```
mattsse, line 44: here we'd need to actually extend, basically self.0.extend(value.into().0)

mattsse (changes requested):

mattsse (approved):

PR #2263, opened 2025-04-01
chore: port transaction envelope bincode compat function

shane-moore:
This PR intends to close https://github.com/paradigmxyz/reth/issues/15377. At the moment, I could use some tips to see if the implementation is per intent.

Issue regarding why the test fails is noted in a comment below

## Todo:
- implement similar for `OpTxEnvelope`
- write additional test for `test_transaction_signed_bincode_roundtrip`

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1038,6 +1038,257 @@ mod serde_from {
     }
 }
 
+/// Bincode-compatible serde implementations for transaction types
+///
+/// This module defines borrowed representations for signed transactions and
+/// a trait for converting to/from a bincode-friendly representation
+#[cfg(all(feature = "serde", feature = "serde-bincode-compat"))]
+pub mod serde_bincode_compat {
+    use crate::{
+        transaction::{
+            serde_bincode_compat::{TxEip1559, TxEip2930, TxEip7702, TxLegacy},
+            TxEnvelope,
+        },
+        Signed, TxEip4844Variant,
+    };
+    use alloc::borrow::Cow;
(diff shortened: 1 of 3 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -1038,6 +1038,257 @@ mod serde_from {
...
+            let tx: TxEnvelope =
+                TxEnvelope::arbitrary(&mut arbitrary::Unstructured::new(&bytes)).unwrap();
+
+            // Convert to bincode representation
+            let data = Data { transaction: tx.as_repr() };
+
+            // Serialize and then deserialize
+            let encoded = bincode::serialize(&data).expect("Serialization should succeed");
+            let decoded: Data<'_> =
+                bincode::deserialize(&encoded).expect("Deserialization should succeed");
```
shane-moore, line 1283: @mattsse  Error on this line when running test `Deserialization should succeed: Custom("Bincode does not support Deserializer::deserialize_identifier")`

It seems to be related to how the `impl SerdeBincodeCompat for TxEnvelope` is now for `TxEnvelope`, which is an untagged enum, and it seems bincode requires tagged enums to know which variant to choose. So bincode calls `deserialize_identifier` and fails at that point. Could use some thoughts on how to refactor 🙏

shane-moore:
nice! will check out the refactors after implemented and then do similar for `OpTxEnvelope`

mattsse (approved):
I realized we can simplify this if we als add support for this to the typed variant then this simplifies the helper type quite a bit, we also don't need the hash actually because this one can be recovered now

PR #2273, opened 2025-04-02
feat(eips): EIP-7594 sidecar

rkrasiuk:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,289 @@
...
+use alloy_primitives::B256;
+use alloy_rlp::{Buf, BufMut, Decodable, Encodable, Header};
+
+/// This represents a set of blobs, and its corresponding commitments and proofs.
+/// Proof type depends on the sidecar variant.
+///
+/// This type encodes and decodes the fields without an rlp header.
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum BlobTransactionSidecar {
```
mattsse, line 17: I suggest we rename this `BlobTransactionSidecarVariant` following what we did for 4844 tx type

mattsse (changes requested):
can we undo the rename here, this just introduces friction and we can avoid this because this sidecar will still remain the main user facing type

mattsse (changes requested):
we can likely make all of this work without introducing breaking changes (at least twice, introducing the variant format, transitioning to 7594 format)

rkrasiuk:
changes were included in other PRs

PR #2276, opened 2025-04-03
feat!: relax ProviderBuilder bounds

DaniPopes:
Allows:
- calling the filler helper methods at any point of the builder
- using on_anvil* with any network

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/anvil.rs
+++ b/crates/provider/src/layers/anvil.rs
@@ -1,9 +1,11 @@
-use alloy_network::Ethereum;
+use crate::{Provider, ProviderLayer, RootProvider};
+use alloy_network::{Ethereum, Network};
 use alloy_node_bindings::{Anvil, AnvilInstance};
 use reqwest::Url;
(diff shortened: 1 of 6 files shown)


crates/provider/src/layers/anvil.rs
```diff
@@ -55,20 +54,18 @@ where
 /// A provider that wraps an [`AnvilInstance`], preventing the instance from
 /// being dropped while the provider is in use.
 #[derive(Clone, Debug)]
-pub struct AnvilProvider<P> {
+pub struct AnvilProvider<P, N = Ethereum> {
```
mattsse, line 57: this kinda makes sense because we also support op and any on anvil

mattsse (approved):
this is better but will conflict with #2261

but since this isn't as breaking as #2261 this one here should go in first,

delaying this until next minor though

onbjerg (approved):

PR #2278, opened 2025-04-03
feat: add support for Eth69 receipts representation

stevencartavia:
## Motivation
https://github.com/paradigmxyz/reth/issues/15509

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -2,7 +2,10 @@ use crate::receipt::{
     Eip2718EncodableReceipt, Eip658Value, RlpDecodableReceipt, RlpEncodableReceipt, TxReceipt,
 };
 use alloc::{vec, vec::Vec};
-use alloy_eips::{eip2718::Encodable2718, Typed2718};
+use alloy_eips::{
+    eip2718::{Decodable2718, Eip2718Error, Eip2718Result, Encodable2718},
+    Typed2718,
+};
 use alloy_primitives::{Bloom, Log};
 use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use core::fmt;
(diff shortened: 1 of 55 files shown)


mattsse (changes requested):
great, 

we also need this for the Decodable trait and then a few basic roundtrip tests

crates/consensus/src/receipt/receipts.rs
```diff
@@ -553,4 +596,101 @@ mod test {
...
+            ],
+        };
+
+        let mut out = vec![];
+        receipt.rlp_encode(&mut out);
+
+        assert_eq!(out.len(), receipt.rlp_encoded_length());
+
+        let header = alloy_rlp::Header::decode(&mut out.as_slice()).unwrap();
+        assert!(header.list);
```
mattsse, line 623: this isn't a full roundtrip test

this only decodes the header

crates/consensus/src/receipt/receipts.rs
```diff
@@ -126,6 +141,14 @@ impl<T: Encodable> RlpEncodableReceipt for Receipt<T> {
...
         self.rlp_encode_fields_with_bloom(bloom, out);
     }
+
+    fn rlp_encoded_length(&self) -> usize {
+        self.rlp_header().length_with_payload()
+    }
+
+    fn rlp_encode(&self, out: &mut dyn BufMut) {
+        self.rlp_header().encode(out);
+    }
```
mattsse, line 151: why does this only encode the header?

mattsse (changes requested):
I realized the receipts are now 2718 encoded, so instead of introducing those functions, we should require that the rlp trait is also `Encodable2718` and `Decodable2718`.

so we'd need to implement those traits instead

PR #2289, opened 2025-04-07
feat(`fillers`): make CachedNonceManager default

yash-atreya:
## Motivation

Currently, the default nonce manager in the `NonceFiller` is `SimpleNonceManager` which always hits the RPC get next nonce. 

## Solution

Default to `CachedNonceManager` instead, which storesthe  tx count locally after first RPC request.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/anvil.rs
+++ b/crates/provider/src/ext/anvil.rs
@@ -328,7 +328,10 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::ProviderBuilder;
+    use crate::{
+        fillers::{ChainIdFiller, GasFiller},
+        ProviderBuilder,
+    };
     use alloy_eips::BlockNumberOrTag;
     use alloy_network::TransactionBuilder;
     use alloy_primitives::B256;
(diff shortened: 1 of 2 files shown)


crates/provider/src/builder.rs
```diff
@@ -188,7 +188,7 @@ impl<L, N: Network> ProviderBuilder<L, Identity, N> {
     /// See [`SimpleNonceManager`] for more information.
     pub fn with_simple_nonce_management(
         self,
-    ) -> ProviderBuilder<L, JoinFill<Identity, NonceFiller>, N> {
+    ) -> ProviderBuilder<L, JoinFill<Identity, NonceFiller<SimpleNonceManager>>, N> {
```
mattsse, line 191: being explicit here makes sense

mattsse (approved):
lgtm

StackOverflowExcept1on:
@mattsse Why did you decide to make this change? We found an issue in gear-tech/gear#4596 where our tests that use anvil and reorgs don't work correctly. I think it's more of priority to have stable backend, rather than x2 compared to ethers-js as paradigm showed in recent benchmarks.

PR #2294, opened 2025-04-08
chore!: Add `ancestor_headers` to `ExecutionWitness`

kevaundray:
This PR adds an `ancestor_headers` field to the ExecutionWitness. This allows one to use the execution witness to statelessly execute a block, by adding the headers needed to verify state reads and BLOCKHASH opcode executions.

### DIFF
--- a/crates/rpc-types-debug/src/debug.rs
+++ b/crates/rpc-types-debug/src/debug.rs
@@ -25,7 +25,7 @@ pub struct ExecutionWitness {
     /// ## Why this field will be empty in the future
     ///
     /// This field is expected to be empty in the future because:
-    /// - EIP-2935 (Parague) will include block hashes directly in the state
+    /// - EIP-2935 (Prague) will include block hashes directly in the state
     /// - Verkle/Delayed execution will change the block structure to contain the
     /// pre-state root instead of the post-state root.
     ///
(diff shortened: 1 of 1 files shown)


kevaundray:
Initially this field was Option<VecBytes>> to distinguish the case where we have a pre-state root and 2935, but given that this field will never be empty due to needing the pre-state header, Vec::is_empty uniquely identifies the former case

crates/rpc-types-debug/src/debug.rs
```diff
@@ -16,4 +16,49 @@ pub struct ExecutionWitness {
...
+    ///
+    /// Consider a blockchain at block 200, and inside of block 200, a transaction
+    /// calls BLOCKHASH(100):
+    /// - This is valid because block 100 is within the 256-block lookback window
+    /// - To verify this, we need all of the headers from block 100 through block 200
+    /// - These headers form a chain proving the correctness of block 100's hash.
+    ///
+    /// The naive way to construct the headers would be to unconditionally include the last
+    /// 256 block headers. However note, we may not need all 256, like in the example above.
+    pub ancestor_headers: Vec<Bytes>,
```
rkrasiuk, line 63: what are the considerations between returning encoded vs decoded headers?

rkrasiuk (approved):
lgtm

crates/rpc-types-debug/src/debug.rs
mattsse: I'd prefer encoded because this will be easier to handle client side

mattsse (approved):

kevaundray:
Francis from Base noted that geth has a field like this, but it's called headers and not ancestor_headers -- what are your thoughts on changing it to match?

I can also make a PR in geth to change theirs from headers to ancestorHeaders

mattsse:
let's change it to headers then

Issue #2291, opened 2025-04-08
[Feature] Add is_dynamic_fee to TxType


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

having this on TxType would be nice to have:

https://github.com/alloy-rs/alloy/blob/5c3f104c230230983e9ed74b47ef9fd22be96d62/crates/consensus/src/transaction/mod.rs#L147-L148

Ayushdubey86:
can i look into this?

yash-atreya:
Closed by #2296

PR #2296, opened 2025-04-09
 Added Is_dyanamic_fee to TxType

Ayushdubey86:
Hey @mattsse , do check this out once free!
## Motivation

This pr will close issue #2291

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -166,7 +166,7 @@ impl TxType {
     /// Returns true if the transaction type has dynamic fee.
     #[inline]
     pub const fn is_dynamic_fee(&self) -> bool {
-        matches!(self,Self::Legacy | Self::Eip2930 | Self::Eip1559 | Self::Eip4844 | Self::Eip7702)
+        matches!(self, Self::Eip1559 | Self::Eip4844 | Self::Eip7702)
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/consensus/src/transaction/envelope.rs
```diff
@@ -162,6 +162,12 @@ impl TxType {
     pub const fn is_eip7702(&self) -> bool {
         matches!(self, Self::Eip7702)
     }
+    
+    /// Returns true if the transaction type has dynamic fee.
+    #[inline]
+    pub const fn is_dynamic_fee(&self) -> bool {
+        matches!(self,Self::Legacy | Self::Eip2930 | Self::Eip1559 | Self::Eip4844 | Self::Eip7702)
+    }
```
yash-atreya, line 170: This is incorrect.  It will always return true

This method should only check the tx types which has dynamic fee fields i.e 1559, 4844 and 7702

yash-atreya (changes requested):

crates/consensus/src/transaction/envelope.rs
```diff
@@ -408,6 +414,12 @@ impl<Eip4844: RlpEcdsaEncodableTx> EthereumTxEnvelope<Eip4844> {
         matches!(self, Self::Eip7702(_))
     }
 
+    /// Returns true if the transaction supports dynamic fees.
+    #[inline]
+    pub const fn is_dynamic_fee(&self) -> bool {
+        matches!(self, Self::Legacy(_) | Self::Eip2930(_) | Self::Eip1559(_) | Self::Eip4844(_) | Self::Eip7702(_))
+    }
```
mattsse, line 421: please remove this

mattsse (changes requested):

crates/consensus/src/transaction/envelope.rs
Ayushdubey86: Have made changes, do check them out once you're free!

mattsse (approved):

Issue #2252, opened 2025-03-28
[Bug] No custom policy and no retry is done for disconnected web socket


mskrzypkows:
### Component

provider, pubsub

### What version of Alloy are you on?

```
[...]

PR #2301, opened 2025-04-10
feat(transport-ws): expose Ws url

yash-atreya:
## Motivation

towards #2252 

## Solution

- Make the URL field private
- Expose the URL string of the Websocket connection using a getter

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -44,6 +44,11 @@ impl WsConnect {
         self.config = Some(config);
         self
     }
+
+    /// Get the URL string of the connection.
+    pub fn url(&self) -> String {
+        self.url.clone()
+    }
 }
 
 impl IntoClientRequest for WsConnect {
--- a/crates/transport-ws/src/native.rs
+++ b/crates/transport-ws/src/native.rs
@@ -20,7 +20,7 @@ const KEEPALIVE: u64 = 10;
 #[derive(Clone, Debug)]
 pub struct WsConnect {
     /// The URL to connect to.
-    pub url: String,
+    url: String,
     /// The authorization header to use.
     pub auth: Option<Authorization>,
     /// The websocket config.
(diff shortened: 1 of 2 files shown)


crates/transport-ws/src/native.rs
```diff
@@ -20,7 +20,7 @@ const KEEPALIVE: u64 = 10;
 #[derive(Clone, Debug)]
 pub struct WsConnect {
     /// The URL to connect to.
-    pub url: String,
+    url: String,
     /// The authorization header to use.
     pub auth: Option<Authorization>,
```
mattsse, line 25: all of the other fields should also be private

crates/transport-ws/src/native.rs
```diff
@@ -44,6 +44,11 @@ impl WsConnect {
         self.config = Some(config);
         self
     }
+
+    /// Get the URL string of the connection.
+    pub fn url(&self) -> String {
```
mattsse, line 49: let's return &str here, the caller should decide

mattsse (changes requested):

crates/transport-ws/src/native.rs
yash-atreya: [40ec657](https://github.com/alloy-rs/alloy/pull/2301/commits/40ec65786d9a342c76eb9b7b274b1a254292a98f)

crates/transport-ws/src/native.rs
yash-atreya: [40ec657](https://github.com/alloy-rs/alloy/pull/2301/commits/40ec65786d9a342c76eb9b7b274b1a254292a98f)

mattsse (approved):
lgtm

blocking this because breaking

mattsse (approved):

Issue #2311, opened 2025-04-11
[Feature] Allow to implement a custom pubsub backend


Jinwei1987:
### Component

provider, pubsub

### Describe the feature you would like

PubSubFrontend has a crate protected new function, which prevent a customiz pubsub backend. Just making it public will allow user to create a customized backend without hacking.

DaniPopes:
Why can you not use `PubSubConnect`?

Jinwei1987:
Looks like PubSubConnect can work. I will close this issue for now.

Jinwei1987:
Actually it can be a good improvement for customised backend. We have our own subscription management and retry connection logic. Using PubSubConnect can work but it means reuse all subscription manager in alloy which is actually not necessary and overhead for us. The ability to construct PubSubFrontend is more ideal.

Jinwei1987:
what's more, ConnectionInterface hide receivers within crate. To use PubSubConnect to implement a complex backend will create a lot of overhead and is inconvenient.

mattsse:
imo seems fine to expose these

yash-atreya:
@Jinwei1987 mind submitting a PR for this?

Jinwei1987:
sure.
https://github.com/alloy-rs/alloy/pull/2326

@yash-atreya

PR #2312, opened 2025-04-11
make PubSubFrontend new public

Jinwei1987:
## Motivation
We want to implement a customized pub sub backend. i.e. make subscription to multiple nodes and merge data from different nodes. We don't think it is good idea to implement such a feature in alloy itself. So we looking for a way to implement customized pub sub backend. But the crate protected new function in PubSubFrontend prevent us to do so without using unsafe hacking.

## Solution
Just make PubSubFrontend new function public, so that we can channel the pub sub request into our customized backend.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/pubsub/src/frontend.rs
+++ b/crates/pubsub/src/frontend.rs
@@ -27,7 +27,7 @@ pub struct PubSubFrontend {
 
 impl PubSubFrontend {
     /// Create a new frontend.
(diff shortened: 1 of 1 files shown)


mattsse (commented):
hmm, can't think of a valid reason not to do this,

need to think about this a bit more

DaniPopes:
Link https://github.com/alloy-rs/alloy/issues/2311

PR #2320, opened 2025-04-16
fix(`transport`): enable hyper-tls via hyper feature

yash-atreya:
## Motivation

https://github.com/alloy-rs/alloy/pull/1899 added the `HttpsConnector` for hyper behind the `hyper-tls` feature. However, this feature isn't being enabled via the `hyper` feat. See: https://github.com/alloy-rs/examples/actions/runs/14471175068/job/40585218348?pr=212

## Solution

- Enable the `hyper-tls` feature via `hyper`
- This makes the HttpsConnector the default which can be used with both http and https schemes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/Cargo.toml
+++ b/crates/transport-http/Cargo.toml
@@ -60,6 +60,7 @@ hyper = [
     "dep:serde_json",
     "dep:tower",
     "dep:tracing",
+    "hyper-tls",
 ]
 hyper-tls = ["hyper", "dep:hyper-tls"]
 jwt-auth = [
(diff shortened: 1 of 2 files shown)


crates/transport-http/Cargo.toml
```diff
@@ -60,6 +60,7 @@ hyper = [
     "dep:serde_json",
     "dep:tower",
     "dep:tracing",
+    "hyper-tls",
```
mattsse, line 63: we could add this as separate tls feature but imo not useful

mattsse (approved):

yash-atreya:
@mattsse, should we remove the `HttpConnector` altogether? 
Downside would be native-tls is always added via `hyper_tls` lib

mattsse:
no I think this is fine

PR #2331, opened 2025-04-20
fix: change value field in TraceEntry to Option<U256>

ecangncr:


mattsse (approved):
thanks, this looks correct,

but not sure when this would be null

https://github.com/ledgerwatch/erigon/blob/a64008640bdbf5bc0b52f982295c9238840d030a/rpc/jsonrpc/otterscan_trace_transaction.go#L55-L55

ecangncr:
> thanks, this looks correct,
> 
> but not sure when this would be null
> 
> https://github.com/ledgerwatch/erigon/blob/a64008640bdbf5bc0b52f982295c9238840d030a/rpc/jsonrpc/otterscan_trace_transaction.go#L55-L55

Thank you for approving, as you can see here, if there is a DELEGATECALL or STATICCALL during the trace, it is assigned to nil

https://github.com/erigontech/erigon/blob/a64008640bdbf5bc0b52f982295c9238840d030a/rpc/jsonrpc/otterscan_trace_transaction.go#L108

PR #2346, opened 2025-04-24
fix: use correct type in conversion

mattsse:
smol breaking change that slipped in in #2345 via copy paste


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -500,6 +508,11 @@ mod tx_serde {
 mod tests {
     use super::*;
 
+    #[allow(unused)]
+    fn assert_convert_into_envelope(tx: Transaction) -> TxEnvelope {
+        tx.into()
+    }
```
refcell, line 514: Nice!

PR #2357, opened 2025-04-26
feat: added DualTransport implementation that wraps two transport

### LINKED ISSUE #2356: [Feature] Add DualTransport implementation that wraps two transport
### Component

transports

it's often useful to wrap two `Transports` and depending on the request we want to dispatch the request.

for this we need something like

```
pub struct DualTransport<L, R, H> {
    left: L,
    right: R,
    handler: H
}
```

where handler is capable of dispatching the request given the two transports as input

```
 fn call<L, R>(&mut self, req: RequestPacket, left : L, right: R ) -> TransportFut<'static>

```

This will require some service based impl, something like this

```

trait DualTransportHandler<L, R> {

  fn call(&self, request: RequestPacket, left: L, right: R) -> TransportFut<'static>
}


impl<L, R, H> Service<RequestPacket> for DualTransport<L, R, H>
where
    L: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
        + Send
        + Sync
        + 'static,
    R: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
        + Send
        + Sync
        + 'static,
    H: DualTransportHandler<L,R>
        + 'static,
{
    type Response = ResponsePacket;
    type Error = TransportError;
    type Future = TransportFut<'static>;

...
```

## TODO
* create this type

maybe @Rimeeeeee if you're interested in something more advanced

### Additional context

_No response_

Rimeeeeee:
fixes #2356

### DIFF
--- a/crates/transport/src/dual.rs
+++ b/crates/transport/src/dual.rs
@@ -1,0 +1,77 @@
+use crate::{TransportError, TransportFut};
+use alloy_json_rpc::{RequestPacket, ResponsePacket};
+use tower::Service;
+
+/// Trait for converting two transports into a dual transport.
+pub trait DualTransportHandler<L, R> {
+    /// The type of the future returned by the transport.
+    fn call(&self, request: RequestPacket, left: L, right: R) -> TransportFut<'static>;
+}
+
+/// A transport that dispatches requests to one of two inner transports based on a handler.
+///
+/// This type allows RPC clients to dynamically select between two different transports
+/// at runtime depending on the request. It is [Send] + [`Sync` ] and implements [`Transport`]
+/// via the [`Service`] trait.
+///
+/// This is useful for building clients that abstract over multiple backends or protocols,
+/// routing requests flexibly without having to commit to a single transport implementation.
+///
+/// All higher-level types can use  [`DualTransport`] internally to support multiple transport
+/// strategies.
+
+#[derive(Debug)]
+pub struct DualTransport<L, R, H> {
+    /// The left transport.
+    left: L,
+    /// The right transport.
+    right: R,
+    /// The handler that decides which transport to use.
+    handler: H,
+}
+
+impl<L, R, H> DualTransport<L, R, H> {
+    /// Instantiate a new dual transport from a suitable transport.
+    pub fn new(left: L, right: R, handler: H) -> Self {
+        Self { left, right, handler }
+    }
+}
+
+impl<L, R, H> Service<RequestPacket> for DualTransport<L, R, H>
+where
+    L: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
+        + Send
+        + Sync
+        + Clone
+        + 'static,
+    R: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError>
+        + Send
+        + Sync
+        + Clone
+        + 'static,
+    H: DualTransportHandler<L, R> + 'static,
+{
+    type Response = ResponsePacket;
+    type Error = TransportError;
+    type Future = TransportFut<'static>;
+
+    #[inline]
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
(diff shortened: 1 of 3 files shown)


Rimeeeeee:
@mattsse ,lmk if i need to do something :)

crates/transport/src/dual.rs
```diff
@@ -0,0 +1,77 @@
+use crate::{TransportError, TransportFut};
+use alloy_json_rpc::{RequestPacket, ResponsePacket};
+use tower::Service;
+
+/// Trait for converting two transports into a dual transport.
+pub trait DualTransportHandler<L, R> {
```
mattsse, line 6: cool, we probably want an impl of this for `Fn<L,R>(RequestPacket, L, R)`

mattsse (changes requested):
this is great,

we also want a test for this

here is some boilerplate that should be useful


```rust
    /// Helper function that transforms a closure to a alloy transport service
    fn request_fn<T>(f: T) -> RequestFn<T>
    where
        T: FnMut(RequestPacket) -> TransportFut<'static>,
    {
        RequestFn { f }
    }

    #[derive(Copy, Clone)]
    struct RequestFn<T> {
        f: T,
    }

    impl<T> Service<RequestPacket> for RequestFn<T>
    where
        T: FnMut(RequestPacket) -> TransportFut<'static>,
    {
        type Response = ResponsePacket;
        type Error = TransportError;
        type Future = TransportFut<'static>;

        fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), TransportError>> {
            Ok(()).into()
        }

        fn call(&mut self, req: RequestPacket) -> Self::Future {
            (self.f)(req)
        }
    }

    fn make_hash_response() -> ResponsePacket {
        ResponsePacket::Single(Response {
            id: Id::Number(0),
            payload: ResponsePayload::Success(
                RawValue::from_string(serde_json::to_string(&B256::ZERO).unwrap()).unwrap(),
            ),
        })
    }

    #[tokio::test]
    async fn test_dual_transport() {
        let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
        let sequencer_transport = request_fn(move |request: RequestPacket| {
            let tx = tx.clone();
            Box::pin(async move {
                tx.send(request).unwrap();
                Ok::<_, TransportError>(make_hash_response())
            })
        });

        let left = request_fn(|_request: RequestPacket| {
            Box::pin(async move { Ok::<_, TransportError>(make_hash_response()) })
        });

```

this can be extended with a dualtransporthandler and another instance of the service and then select the first one in the handler for example

crates/transport/src/dual.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// The type of the future returned by the transport.
+    fn call(&self, request: RequestPacket, left: L, right: R) -> TransportFut<'static>;
+}
+
+impl<F, L, R> DualTransportHandler<L, R> for F
+where
+    F: Fn(RequestPacket, L, R) -> TransportFut<'static> + Send + Sync,
+    L: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError> + Clone,
+    R: Service<RequestPacket, Future = TransportFut<'static>, Error = TransportError> + Clone,
```
mattsse, line 15: I think we likely don't need these bounds here, because those will be enforced by the the transport

mattsse (changes requested):
this is pretty cool, one more nit

crates/transport/src/dual.rs
```diff
@@ -0,0 +1,197 @@
...
+                RequestPacket::Batch(reqs) => {
+                    reqs.first().map(|r| r.id().as_number().unwrap_or(0)).unwrap_or(0)
+                }
+            };
+
+            if id % 2 == 0 {
+                left.call(req)
+            } else {
+                right.call(req)
+            }
```
mattsse, line 170: ah this is very nice

mattsse (approved):
nice work, ty

PR #2358, opened 2025-04-26
feat: Support deserializing system signatures in legacy transactions

### LINKED ISSUE #2353: [Feature] Support deserializing system signatures in legacy transactions
### Component

rpc

https://github.com/alloy-rs/alloy/issues/2348 has a pre bedrock legacy transaction with a fake signature, that is technically invalid and fails deserialzation:

>  Error("invalid EIP-155 signature parity value"

we can probably add a check for this when we use the legacy fallback and allow deserialzing 0,0,0 because I think this is quite common for off spec L2s

for this we'd need to modify:

https://github.com/alloy-rs/alloy/blob/bf4457798ce6afd3ef799f9f5c93cee9881df819/crates/consensus/src/transaction/envelope.rs#L869-L873

and check if the values are 0,0,0 and if so, create the signature manually.

https://github.com/alloy-rs/alloy/blob/bf4457798ce6afd3ef799f9f5c93cee9881df819/crates/consensus/src/transaction/legacy.rs#L419-L419

### Additional context

_No response_

Haxry:
## Solution
closes #2353 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Haxry:
@mattsse plz review

crates/consensus/src/transaction/legacy.rs
```diff
@@ -510,19 +510,33 @@ pub mod signed_legacy_serde {
...
-        let (parity, chain_id) = from_eip155_value(signature.v.to())
-            .ok_or_else(|| serde::de::Error::custom("invalid EIP-155 signature parity value"))?;
-
-        // Note: some implementations always set the chain id in the response, so we only check if
-        // they differ if both are set.
-        if let Some((tx_chain_id, chain_id)) = tx.chain_id().zip(chain_id) {
-            if tx_chain_id != chain_id {
-                return Err(serde::de::Error::custom("chain id mismatch"));
+        let is_fake_signature =
+            signature.r.is_zero() && signature.s.is_zero() && signature.v.is_zero();
```
mattsse, line 514: could we also reference the reported bug in a comment here and document that this happens for pre bedrock transactions for example

mattsse (changes requested):
cool, let's also add the example from the issue as a test case

Haxry:
@mattsse review plz

mattsse (approved):
ty!

Issue #2372, opened 2025-04-30
Add Ens util functions


abhijeetbhagat:

mattsse:
I think we can actually upstream the entire ens.rs file

abhijeetbhagat:
yes i can do it.

PR #2376, opened 2025-05-01
feat: add ens crate from foundry

abhijeetbhagat:
## Motivation

while working with ENS related stuff, you need utilities like namehash calculation, forward & reverse lookups. these utilities are part of `foundry-common` crate. developers using alloy-rs as a result require importing `foundry-common` crate alongside alloy-rs which doesn't make sense since foundry in turn uses alloy-rs as a dependency. (#2372)

## Solution

we need to provide ENS utilities to developers using alloy-rs by copying the ens crate from foundry with little modifications:
- fix the `EnsResolverInstance` type to reflect the latest version (foundry uses old version of `alloy-sol-types` crate) 
- remove the `inspect_err` call because it in turn calls `sh_eprintln` which belongs to the foundry-common crate.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (changes requested):
can we include this in the provider crate instead, I dont think we need an additional crate for this feature

abhijeetbhagat:
> can we include this in the provider crate instead, I dont think we need an additional crate for this feature

problem is that the ens file needs the `alloy-contract` crate (because of the contracts defined using `sol!`) which in turn references the alloy-provider crate. so adding the ens file to the provider crate results in cyclic dependency.

abhijeetbhagat:
@mattsse hey! what do you think?

crates/ens/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+[lints]
+workspace = true
+
+[dependencies]
+alloy-contract.workspace = true
+alloy-primitives.workspace = true
+alloy-provider.workspace = true
+alloy-sol-types.workspace = true
+async-trait.workspace = true
+thiserror.workspace = true
```
DaniPopes, line 31: we can feature gate the async/provider/contract functionality to allow importing just for the utility functions like namehash

the feature should also be re-exported in the meta crate

crates/ens/README.md
DaniPopes, line 1: let's add this to the meta crate crates/alloy

crates/ens/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![doc = include_str!("../README.md")]
+#![doc(
+    html_logo_url = "https://raw.githubusercontent.com/alloy-rs/core/main/assets/alloy.jpg",
+    html_favicon_url = "https://raw.githubusercontent.com/alloy-rs/core/main/assets/favicon.ico"
+)]
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
+pub mod ens;
```
DaniPopes, line 9: this is unnecessary, we can move everything into lib.rs

DaniPopes (changes requested):

mattsse (approved):

PR #2384, opened 2025-05-02
Rename builder to `with_state_diff`

rnkrtt:
Changes method name from `with_statediff` to `with_state_diff` to match Rust snake_case and other builder names.


mattsse (commented):
this would be breaking,

this is fine

PR #2388, opened 2025-05-02
Adding input helper methods to TransactionRequest

### LINKED ISSUE #2381: [Feature] Add helper function to force either data or input or both for TransactionRequest.input
### Component

rpc

Transactionrequest has:

https://github.com/alloy-rs/alloy/blob/79e47906f8df58f2d0f69516374ea841c485bc39/crates/rpc-types-eth/src/transaction/request.rs#L86-L88

https://github.com/alloy-rs/alloy/blob/79e47906f8df58f2d0f69516374ea841c485bc39/crates/rpc-types-eth/src/transaction/request.rs#L258-L262

we are missing a few helper functions on both `TransactionInput` and `TransactionRequest` that make it easier to force both input and data field or swap data for input

## TODO
* add helpers for this

maybe @ratankaliani you want to take this 

ref #2378

### Additional context

_No response_

Ayushdubey86:
Hello @mattsse  , this should close #2381 . 

I’ve added three helper methods as described in the issue, If there are other helper methods or additional use cases you’d like me to cover, I’m happy to extend this PR or follow up with another one.

### DIFF
--- a/crates/rpc-types-eth/src/transaction/request.rs
+++ b/crates/rpc-types-eth/src/transaction/request.rs
@@ -130,7 +130,7 @@ pub struct TransactionRequest {
         serde(default, flatten, skip_serializing_if = "Option::is_none")
     )]
     pub sidecar: Option<BlobTransactionSidecar>,
-    /// Authorization list for for EIP-7702 transactions.
+    /// Authorization list for EIP-7702 transactions.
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
this isn't quite what we want here, we want a way to set 
* force both fields if just one is set
* swap input/data

on both txrequest and txinput

Ayushdubey86:
closing this as I see #2393 up

PR #2391, opened 2025-05-03
chore: SubmitBlockRequest enum

mattsse:
add enum type for all available versions


crates/rpc-types-beacon/src/relay.rs
```diff
@@ -165,6 +172,74 @@ pub struct SignedBidSubmissionV4 {
     pub signature: BlsSignature,
 }
 
+/// Represents all versions of signed bid submissions (submit block requests).
+///
+/// Note: The fields are ordered starting with the most recent version so that the
+/// untagged,transparent decoding prioritises the newest version.
```
DaniPopes, line 178:
```suggestion
/// untagged, transparent decoding prioritises the newest version.
```

DaniPopes (approved):

PR #2392, opened 2025-05-03
feat: Add Arbitrary Support for payload types

### LINKED ISSUE #2390: [Feature] Add arbitrary support for payload related types
### Component

rpc

all of those types currently don't have `arbitrary` support

https://github.com/alloy-rs/alloy/blob/93d1c98cc841720b5fcb4c577d6881cb51275afe/crates/rpc-types-engine/src/payload.rs#L1-L1

## TODO
* introduce arbitrary feature and derive/implement arbitrary for those types

### Additional context

for example

https://github.com/alloy-rs/alloy/blob/93d1c98cc841720b5fcb4c577d6881cb51275afe/crates/consensus/src/receipt/receipts.rs#L13-L13

or 

https://github.com/alloy-rs/alloy/blob/93d1c98cc841720b5fcb4c577d6881cb51275afe/crates/consensus/src/block/header.rs#L522-L523

Ayushdubey86:
Hello @mattsse , do check this out!   It closes #2390 :)


crates/rpc-types-engine/Cargo.toml
```diff
@@ -44,6 +44,9 @@ ethereum_ssz = { workspace = true, optional = true }
 # jsonrpsee
 jsonrpsee-types = { version = "0.24", optional = true }
 
+# arbitrary
+arbitrary = { workspace = true, features = ["derive"], optional = true }
```
mattsse, line 48: we also want an arbitrary feature here that has "dep:arbitrary"

mattsse (changes requested):
cool

mattsse (approved):
ty

PR #2394, opened 2025-05-03
Move IsTyped2718 to alloy

18aaddy:
## Motivation
Resolves reth issue: https://github.com/paradigmxyz/reth/issues/16026

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/eip1559.rs
+++ b/crates/consensus/src/transaction/eip1559.rs
@@ -1,5 +1,5 @@
 use crate::{SignableTransaction, Transaction, TxType};
-use alloy_eips::{eip2930::AccessList, eip7702::SignedAuthorization, Typed2718};
+use alloy_eips::{eip2718::IsTyped2718, eip2930::AccessList, eip7702::SignedAuthorization, Typed2718};
 use alloy_primitives::{Bytes, ChainId, Signature, TxKind, B256, U256};
 use alloy_rlp::{BufMut, Decodable, Encodable};
 use core::mem;
@@ -288,6 +288,12 @@ impl Encodable for TxEip1559 {
 impl Decodable for TxEip1559 {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         Self::rlp_decode(buf)
+    }
+}
+
+impl IsTyped2718 for TxEip1559 {
+    fn is_type(type_id: u8) -> bool {
+        matches!(type_id, 0x00)
     }
 }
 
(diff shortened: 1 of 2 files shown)


crates/eips/src/eip2718.rs
```diff
@@ -403,3 +403,9 @@ impl<L: Typed2718, R: Typed2718> Typed2718 for either::Either<L, R> {
         }
     }
 }
+
+/// Trait for checking if a transaction envelope supports a given EIP-2718 type ID.
```
mattsse, line 407:
```suggestion
/// Trait for checking if a type supports a given EIP-2718 type ID.
```

crates/consensus/src/transaction/eip1559.rs
```diff
@@ -291,6 +291,12 @@ impl Decodable for TxEip1559 {
     }
 }
 
+impl IsTyped2718 for TxEip1559 {
+    fn is_type(type_id: u8) -> bool {
+        matches!(type_id, 0x00)
+    }
+}
```
mattsse, line 298: the need this impl but this should use the correct type id, see:

https://github.com/alloy-rs/alloy/blob/93d1c98cc841720b5fcb4c577d6881cb51275afe/crates/eips/src/eip2718.rs#L18-L28

mattsse (changes requested):
cool,

now we also need this for 

https://github.com/alloy-rs/alloy/blob/93d1c98cc841720b5fcb4c577d6881cb51275afe/crates/consensus/src/transaction/envelope.rs#L185-L185

crates/consensus/src/transaction/eip1559.rs
18aaddy: Yeah I just opened this draft PR so that no one else would, I'm working on all the changes required.

18aaddy:
@mattsse Can you check now?

crates/consensus/src/transaction/envelope.rs
```diff
@@ -153,6 +153,12 @@ impl<T> EthereumTxEnvelope<T> {
     }
 }
 
+impl<T: IsTyped2718> IsTyped2718 for EthereumTxEnvelope<T> {
+    fn is_type(type_id: u8) -> bool {
+        T::is_type(type_id)
```
mattsse: this needs to check that id is in 0..=4

mattsse (approved):
great, ty!

next step will be unlocked once I send a patch release

PR #2397, opened 2025-05-05
fix(`CallBatchLayer`): don't batch if single request

### LINKED ISSUE #2386: CallBatchLayer: don't batch if only 1 request is received
### Component

provider, pubsub

Currently CallBatchLayer will send a Multicall3 call even if only 1 request is received: e.g. a single eth_getBalance will still get sent as a Multicall3 call

This is not efficient as RPC providers charge more for eth_call than the other requests that CallBatchLayer handles

If there's only 1 request in the queue when we want to fire off the Multicall3 call, we should instead fall back to normal RPC

### Additional context

_No response_

Ayushdubey86:
Hello @mattsse @DaniPopes,

Raising a draft PR for #2386 — please let me know if the approach looks good.

I’m not converting to Multicall3 immediately upon receiving a request. Instead, I defer conversion until I’ve intercepted and counted the total number of calls. Based on that count, I either perform a normal RPC call (for a single request) or batch them using Multicall3.

I have numbed testcases as of now, if current approach is correct, I'll take on tc

Closes #2386.


crates/provider/src/layers/batch.rs
```diff
@@ -144,13 +165,13 @@ enum CallBatchMsgKind<N: Network = Ethereum> {
     Balance(Address),
 }
 
-impl CallBatchMsg {
-    fn new<N: Network>(
+impl<N: Network> CallBatchMsg<N> {
+    fn new(
         kind: CallBatchMsgKind<N>,
-        m3a: Address,
+        //m3a: Address,
```
DaniPopes, line 171: can remove

crates/provider/src/layers/batch.rs
```diff
@@ -81,6 +81,13 @@ impl Default for CallBatchLayer {
     }
 }
 
+#[allow(dead_code)]
+#[derive(Debug)]
+enum CallResult {
```
DaniPopes, line 86: I think we can keep using `IMulticall3::Result` and just set the `success` to true always when we're not calling the contract

crates/provider/src/layers/batch.rs
```diff
@@ -311,18 +342,45 @@ impl<P: Provider<N> + 'static, N: Network> CallBatchBackend<P, N> {
...
+                    let result = self.inner.get_block_number().into_future().await.unwrap();
+                    Ok(CallResult::Single(Bytes::from(result.to_be_bytes())))
+                }
+                CallBatchMsgKind::ChainId => {
+                    let result = self.inner.get_chain_id().into_future().await.unwrap();
+                    Ok(CallResult::Single(Bytes::from(result.to_be_bytes())))
+                }
+                CallBatchMsgKind::Balance(addr) => {
+                    let result = self.inner.get_balance(addr).into_future().await.unwrap();
+                    Ok(CallResult::Single(Bytes::from(result.to_be_bytes::<32>())))
```
DaniPopes, line 370: these `to_be_bytes` happen to work fine but we want to make sure they're ABI encoded with `.abi_encode()` instead

crates/provider/src/layers/batch.rs
```diff
@@ -311,18 +342,45 @@ impl<P: Provider<N> + 'static, N: Network> CallBatchBackend<P, N> {
...
+                CallBatchMsgKind::BlockNumber => {
+                    let result = self.inner.get_block_number().into_future().await.unwrap();
+                    Ok(CallResult::Single(Bytes::from(result.to_be_bytes())))
+                }
+                CallBatchMsgKind::ChainId => {
+                    let result = self.inner.get_chain_id().into_future().await.unwrap();
+                    Ok(CallResult::Single(Bytes::from(result.to_be_bytes())))
+                }
+                CallBatchMsgKind::Balance(addr) => {
+                    let result = self.inner.get_balance(addr).into_future().await.unwrap();
```
DaniPopes, line 369: shouldn't unwrap but send `Err` for all branches here

crates/provider/src/layers/batch.rs
```diff
@@ -445,101 +504,3 @@ impl<N: Network> Caller<N, Bytes> for CallBatchCaller {
         Caller::<N, Bytes>::call_many(&self.weak, params)
     }
 }
-
-#[cfg(test)]
-mod tests {
```
DaniPopes, line 450: please keep the test

DaniPopes (changes requested):

Ayushdubey86:
Hello @DaniPopes pushed those changes, do check once free!

DaniPopes (changes requested):

Ayushdubey86:
Made those changes and pushed, do check @DaniPopes !

crates/provider/src/layers/batch.rs
```diff
@@ -311,13 +326,45 @@ impl<P: Provider<N> + 'static, N: Network> CallBatchBackend<P, N> {
...
-    fn process_msg(&mut self, msg: CallBatchMsg) {
+    fn process_msg(&mut self, msg: CallBatchMsg<N>) {
         self.pending.push(msg);
     }
 
     async fn send_batch(&mut self) {
-        let result = self.send_batch_inner().await;
         let pending = std::mem::take(&mut self.pending);
+
+        if pending.len() == 1 {
```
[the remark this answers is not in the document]
Ayushdubey86: added

crates/provider/src/layers/batch.rs
```diff
@@ -333,28 +351,45 @@ impl<P: Provider<N> + 'static, N: Network> CallBatchBackend<P, N> {
...
+            CallBatchMsgKind::BlockNumber => {
+                self.inner.get_block_number().await.map(|res| m3_res(true, res.abi_encode().into()))
+            }
+            CallBatchMsgKind::ChainId => {
+                self.inner.get_chain_id().await.map(|res| m3_res(true, res.abi_encode().into()))
+            }
+            CallBatchMsgKind::Balance(addr) => {
+                self.inner.get_balance(addr).await.map(|res| m3_res(true, res.abi_encode().into()))
+            }
+        }
```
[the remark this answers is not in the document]
DaniPopes: it doesn't matter

PR #2400, opened 2025-05-06
chore: Make `alloy-rpc-types-debug` `no_std` compatible

### LINKED ISSUE #2399: [Feature] Make alloy-rpc-types-debug no_std
### Component

rpc

## TODO
* relax deps and make sure alloy-rpc-types-debug compiles for no-std

### Additional context

_No response_

kevaundray:
closes #2399

### DIFF
--- a/crates/rpc-types-debug/src/lib.rs
+++ b/crates/rpc-types-debug/src/lib.rs
@@ -5,6 +5,9 @@
 )]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![no_std]
+
+extern crate alloc;
 
 mod debug;
 pub use debug::*;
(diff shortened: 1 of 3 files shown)


crates/rpc-types-debug/src/lib.rs
```diff
@@ -5,6 +5,9 @@
 )]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+#![no_std]
```
mattsse, line 8: not sure if we can make this entirely no-std

kevaundray: Oh can you say why? I think right now it can be, but maybe you mean in the future?

mattsse: oh actually we only have one type in here, so this seems fine

mattsse (approved):
lgtm

DaniPopes:
please also add it to the script that runs in ci

PR #2410, opened 2025-05-08
feat: some covenience signer impls

mattsse:
@yash-atreya we can use this to simplify some examples:

```
let alice: PrivateKeySigner = anvil.keys()[0].clone().into();
    let bob: PrivateKeySigner = anvil.keys()[1].clone().into();
```


yash-atreya (approved):
Nice!

PR #2411, opened 2025-05-08
feat: Add fn `fill_envelope`

### LINKED ISSUE #2406: [Feature] Add Filler::fill_envelope
### Component

rpc

ref #2404 

https://github.com/alloy-rs/alloy/blob/570ff02256fd96abf04c7304219005ce7d05f501/crates/provider/src/fillers/mod.rs#L196-L201

ideally we also have `fn fill_envelope` as a default impl that does `.fill` and:

https://github.com/alloy-rs/alloy/blob/570ff02256fd96abf04c7304219005ce7d05f501/crates/provider/src/provider/sendable.rs#L54-L62

for this we need another error type that can represent both error cases

## TODO
* introduce function and error

### Additional context

_No response_

0xMushow:
Is this a good start @mattsse ?

Closes #2406

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -218,7 +218,8 @@ pub trait TxFiller<N: Network = Ethereum>: Clone + Send + Sync + std::fmt::Debug
         &self,
         fillable: Self::Fillable,
         tx: SendableTx<N>,
-    ) -> impl_future!(<Output = Result<N::TxEnvelope, FillEnvelopeError<N::TransactionRequest>>>) {
+    ) -> impl_future!(<Output = Result<N::TxEnvelope, FillEnvelopeError<N::TransactionRequest>>>)
+    {
         async move {
             let tx = self.fill(fillable, tx).await.map_err(FillEnvelopeError::Transport)?;
             let envelope = tx.try_into_envelope().map_err(FillEnvelopeError::NotReady)?;
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -47,7 +47,7 @@ use alloy_rpc_types_eth::{
     AccessListResult, EIP1186AccountProofResponse, EthCallResponse, FeeHistory, Filter,
     FilterChanges, Log,
 };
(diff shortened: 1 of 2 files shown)


crates/provider/src/fillers/mod.rs
```diff
@@ -200,6 +200,19 @@ pub trait TxFiller<N: Network = Ethereum>: Clone + Send + Sync + std::fmt::Debug
...
 
+    /// Fills in the transaction request and try to convert it to an envelope.
+    fn fill_envelope(
+        &self,
+        fillable: Self::Fillable,
+        tx: SendableTx<N>,
+    ) -> impl_future!(<Output = TransportResult<N::TxEnvelope>>) {
+        async move {
+            let tx = self.fill(fillable, tx).await?;
+            let envelope = tx.try_into_envelope().map_err(/* Not sure where to create the new error type */)?;
```
mattsse, line 211: yeah we can either introduce a new dedicated error type (whcih I think would be fine here)

or we simply type erase this error to https://github.com/alloy-rs/alloy/blob/833f9a22ca7c30079d64554d435561877c15267a/crates/transport/src/error.rs#L57-L57

both approaches are fine here

I'd prefer the additional error enum with two variants

mattsse (changes requested):

crates/transport/src/error.rs
```diff
@@ -39,6 +39,14 @@ pub enum TransportErrorKind {
...
     #[error("{0}")]
     Custom(#[source] Box<dyn StdError + Send + Sync + 'static>),
+
+    /// Transaction filling error.
+    #[error("transaction filling error: {0}")]
+    FillError(String),
+
+    /// Transaction not ready: still a builder.
+    #[error("transaction not ready: still a builder")]
+    NotReady,
```
mattsse, line 49: these we don't need

we can instead introduce a new dedicated error enum that has two variants for transporterr and SendableTxErr

mattsse (changes requested):

crates/transport/src/error.rs
0xMushow: just made changes, let me know if that's better

mattsse (approved):
great, ty!

PR #2419, opened 2025-05-10
docs (README.md): integrating crates.io badges

gap-editor:
By integrating these badges, users can immediately see which versions are published without ever leaving GitHub. Because Shields.io updates badges automatically whenever a new version is released, we no longer need to maintain these by hand. In addition, including standardized version badges helps our project adhere to the conventions commonly used in the Rust ecosystem, giving our documentation a more professional and polished appearance.


mattsse (approved):
sure, why not

PR #2420, opened 2025-05-10
feat(signer): adds signer-stronghold

tuddman:
## Solution

Adds a signer using a Stronghold as the store

## PR Checklist

- [x] Added Tests
- [x] Added Documentation


mattsse (approved):
seems okay, simple enough, not much work to maintain, ty

mattsse:
@tuddman most of these tests are timing out, mind taking a look?

gakonst:
@mattsse @tuddman could maintain this in their own repo? I'm not a fan of having this upstream and it might be hard to maintain

mattsse:
right, now that we have 1.0 which we want to keep stable, I think as a standalone repo this would be more appropriate because iota specific.

tuddman:
Stronghold is an interesting piece of software.  I used it successfully for quite awhile to securely manage keys without any incident.    But as it is unlikely this PR will land upstream, and given that I am also using other more secure key management solutions (hardware based), I am electing to close this PR and am deprecating  `signer-stronghold` into maintenance-mode.   The original authors of `stronghold` have also appeared to have abandoned it.  

Others are free to use this library for their purposes but be advised it comes with no promises or claims of any kind.

tensor-programming:
Technically we haven't abandoned it, we just can't maintain the IOTA version because I and the other main creator are no longer working there. I also heard that IOTA isn't maintaining it any longer since they moved onto a hardware solution. The original purpose of stronghold was to make as secure of a software signer as possible due to most users not having hardware enclaves.  If you can use a hardware solution then its almost always going to be the better more secure choice.  With the current state of the IOTA branch of stronghold, I can't make any guarantees though as the other individual pointed out, it hasn't ever been compromised to my knowledge.

This looks like an interesting project; and I wish you guys luck.

PR #2423, opened 2025-05-12
feat: impl Signerrecoverable trait

mattsse:
we had this but never used it,

since we moved the crypto impls, we can implement this easily, mirroring what we have for reth.
this can then be made a supertrait of `reth::SignedTransaction` getting rid of the shadowed recovery fn footgun

### DIFF
--- a/crates/consensus/src/transaction/recovered.rs
+++ b/crates/consensus/src/transaction/recovered.rs
@@ -1,3 +1,4 @@
+use crate::crypto::RecoveryError;
 use alloy_eips::{
     eip2718::{Encodable2718, WithEncoded},
     Typed2718,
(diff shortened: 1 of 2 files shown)


crates/consensus/src/transaction/recovered.rs
```diff
@@ -239,11 +240,45 @@ pub trait SignerRecoverable {
...
+    where
+        Self: Sized,
+    {
+        let signer = self.recover_signer_unchecked()?;
+        Ok(Recovered::new_unchecked(self, signer))
+    }
+
+    /// Recover the signer via [`SignerRecoverable::recover_signer`] and returns a
+    /// `Recovered<&Self>`
+    fn try_to_recovered_ref(&self) -> Result<Recovered<&Self>, RecoveryError> {
```
klkvr, line 273:
```suggestion
    fn try_into_recovered_ref(&self) -> Result<Recovered<&Self>, RecoveryError> {
```

crates/consensus/src/transaction/recovered.rs
```diff
@@ -239,11 +240,45 @@ pub trait SignerRecoverable {
...
+        let signer = self.recover_signer()?;
+        Ok(Recovered::new_unchecked(self, signer))
+    }
+
+    /// Recover the signer via [`SignerRecoverable::recover_signer_unchecked`] and returns a
+    /// `Recovered<&Self>`
+    fn try_to_recovered_ref_unchecked(&self) -> Result<Recovered<&Self>, RecoveryError> {
```
klkvr, line 280:
```suggestion
    fn try_into_recovered_ref_unchecked(&self) -> Result<Recovered<&Self>, RecoveryError> {
```

klkvr (approved):
lgtm

crates/consensus/src/transaction/recovered.rs
mattsse: this doesn't consume self so I'd rather keep this as `to`

PR #2426, opened 2025-05-12
fix(types-engine): `proofs` field name in `BlobsBundleV2`

rkrasiuk:
## Description

Just `proofs`, not `cell_proofs`. Ref https://github.com/ethereum/execution-apis/pull/630/files


mattsse (approved):
this is now structually the same as v1, but I guess all the sidecar operations will be different

PR #2428, opened 2025-05-12
feat(eips): `BlobTransactionSidecarEip7594`

rkrasiuk:
## Description

Add `BlobTransactionSidecarEip7594`. Ref https://eips.ethereum.org/EIPS/eip-7594

### DIFF
--- a/crates/eips/src/eip7594/sidecar.rs
+++ b/crates/eips/src/eip7594/sidecar.rs
@@ -127,7 +127,7 @@ impl BlobTransactionSidecarEip7594 {
         // Repeat commitments for each cell.
         let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
         for commitment in &self.commitments {
-            commitments.extend((0..CELLS_PER_EXT_BLOB).map(|_| *commitment));
+            commitments.extend(std::iter::repeat_n(*commitment, CELLS_PER_EXT_BLOB));
         }
 
         // SAFETY: ALL types have the same size
--- a/crates/eips/src/eip7594/sidecar.rs
+++ b/crates/eips/src/eip7594/sidecar.rs
@@ -127,7 +127,7 @@ impl BlobTransactionSidecarEip7594 {
         // Repeat commitments for each cell.
         let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
         for commitment in &self.commitments {
-            commitments.extend(std::iter::repeat_n(*commitment, CELLS_PER_EXT_BLOB));
+            commitments.extend(core::iter::repeat_n(*commitment, CELLS_PER_EXT_BLOB));
         }
 
         // SAFETY: ALL types have the same size
--- a/crates/eips/src/eip7594/sidecar.rs
+++ b/crates/eips/src/eip7594/sidecar.rs
@@ -120,18 +120,18 @@ impl BlobTransactionSidecarEip7594 {
             }
         }
 
+        // Repeat cell ranges for each blob.
+        let cell_indices =
+            Vec::from_iter((0..blobs_len).flat_map(|_| 0..CELLS_PER_EXT_BLOB as u64));
+
+        // Repeat commitments for each cell.
+        let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
+        for commitment in &self.commitments {
+            commitments.extend((0..CELLS_PER_EXT_BLOB).map(|_| *commitment));
+        }
+
         // SAFETY: ALL types have the same size
         let res = unsafe {
-            // Repeat commitments for each cell.
-            let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
-            for commitment in &self.commitments {
-                commitments.extend((0..CELLS_PER_EXT_BLOB).map(|_| *commitment));
-            }
-
-            // Repeat cell ranges for each blob.
-            let cell_indices =
-                Vec::from_iter((0..blobs_len).flat_map(|_| 0..CELLS_PER_EXT_BLOB as u64));
-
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+        serde(deserialize_with = "crate::eip4844::deserialize_blobs")
+    )]
+    pub blobs: Vec<Blob>,
+    /// The blob commitments.
+    pub commitments: Vec<Bytes48>,
+    /// List of cell proofs for all blobs in the sidecar, including the proofs for the extension
+    /// indices, for a total of `CELLS_PER_EXT_BLOB` proofs per blob (`CELLS_PER_EXT_BLOB` is the
+    /// number of cells for an extended blob, defined in
+    /// [the consensus specs](https://github.com/ethereum/consensus-specs/tree/9d377fd53d029536e57cfda1a4d2c700c59f86bf/specs/fulu/polynomial-commitments-sampling.md#cells))
+    pub cell_proofs: Vec<Bytes48>,
```
mattsse, line 34: we renamed cell_proofs to proofs in blobsbundlev2

I personally prefer cell_proofs here, because this makes it easier to distinguish the two sidecar types, but I wonder if we should rename this here as well, if this type will be used in jsonrpc?
though we could rename via serde name later

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+                });
+            }
+        }
+
+        // SAFETY: ALL types have the same size
+        let res = unsafe {
+            // Repeat commitments for each cell.
+            let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
+            for commitment in &self.commitments {
+                commitments.extend((0..CELLS_PER_EXT_BLOB).map(|_| *commitment));
```
mattsse, line 128: I think this is just

```
 std::iter::repeat_n(commitment, CELLS_PER_EXT_BLOB)
```

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+            }
+
+            // Repeat cell ranges for each blob.
+            let cell_indices =
+                Vec::from_iter((0..blobs_len).flat_map(|_| 0..CELLS_PER_EXT_BLOB as u64));
+
+            let mut cells = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
+            for blob in &self.blobs {
+                let blob = core::mem::transmute::<&Blob, &c_kzg::Blob>(blob);
+                cells.extend(proof_settings.compute_cells(blob)?.into_iter());
```
mattsse, line 138: let's do this step first because fallible

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+        let res = unsafe {
+            // Repeat commitments for each cell.
+            let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
+            for commitment in &self.commitments {
+                commitments.extend((0..CELLS_PER_EXT_BLOB).map(|_| *commitment));
+            }
+
+            // Repeat cell ranges for each blob.
+            let cell_indices =
+                Vec::from_iter((0..blobs_len).flat_map(|_| 0..CELLS_PER_EXT_BLOB as u64));
```
mattsse, line 133: isn't this just `0..(blobs_len * CELLS_PER_EXT_BLOB)`?

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+            if *versioned_hash != calculated_versioned_hash {
+                return Err(BlobTransactionValidationError::WrongVersionedHash {
+                    have: *versioned_hash,
+                    expected: calculated_versioned_hash,
+                });
+            }
+        }
+
+        // SAFETY: ALL types have the same size
+        let res = unsafe {
```
mattsse, line 124: can we reduce the unsafe scope

mattsse (changes requested):
nits

crates/eips/src/eip7594/sidecar.rs
rkrasiuk: would like to keep `cell_proofs` too. mb the other PR should've been a simple serde rename as well

crates/eips/src/eip7594/sidecar.rs
rkrasiuk: nope, it's  `[0, 1, 2, 3...., 0, 1, 2, 3.....]` repeated :/

crates/eips/src/eip7594/sidecar.rs
rkrasiuk: i restructured to reduce unsafe scope, no longer applicable

crates/eips/src/eip7594/sidecar.rs
mattsse: disusting :D

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        // Repeat cell ranges for each blob.
+        let cell_indices =
+            Vec::from_iter((0..blobs_len).flat_map(|_| 0..CELLS_PER_EXT_BLOB as u64));
+
+        // Repeat commitments for each cell.
+        let mut commitments = Vec::with_capacity(blobs_len * CELLS_PER_EXT_BLOB);
+        for commitment in &self.commitments {
+            commitments.extend(std::iter::repeat_n(*commitment, CELLS_PER_EXT_BLOB));
```
mattsse, line 130:
```suggestion
            commitments.extend(core::iter::repeat_n(*commitment, CELLS_PER_EXT_BLOB));
```

mattsse (approved):

PR #2429, opened 2025-05-12
fix: resolve dead code warnings in crypto.rs

GarmashAlex:
Fix visibility modifiers in crypto.rs to resolve dead code warnings:
- Change public functions to pub(crate) in implementation modules
- Restructure test code to use the public API
- Use conditional compilation for test-specific visibility requirements
- Maintain API compatibility while eliminating unreachable_pub warnings


mattsse (commented):
this should stay

PR #2430, opened 2025-05-12
feat(eips): add `BlobTransactionSidecarVariant`

rkrasiuk:
## Description

Add `BlobTransactionSidecarVariant`, a type that wraps both EIP-4844 and EIP-7594 style sidecars.

### DIFF
--- a/crates/eips/src/eip7594/sidecar.rs
+++ b/crates/eips/src/eip7594/sidecar.rs
@@ -1,16 +1,131 @@
 use crate::{
     eip4844::{
-        kzg_to_versioned_hash, Blob, BlobAndProofV2, Bytes48, BYTES_PER_BLOB, BYTES_PER_COMMITMENT,
-        BYTES_PER_PROOF,
+        kzg_to_versioned_hash, Blob, BlobAndProofV2, BlobTransactionSidecar, Bytes48,
+        BYTES_PER_BLOB, BYTES_PER_COMMITMENT, BYTES_PER_PROOF,
     },
     eip7594::{CELLS_PER_EXT_BLOB, EIP_7594_WRAPPER_VERSION},
 };
 use alloc::{boxed::Box, vec::Vec};
 use alloy_primitives::B256;
-use alloy_rlp::{BufMut, Decodable, Encodable, Header};
+use alloy_rlp::{Buf, BufMut, Decodable, Encodable, Header};
 
 #[cfg(feature = "kzg")]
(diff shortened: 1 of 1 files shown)


crates/eips/src/eip7594/sidecar.rs
```diff
@@ -1,17 +1,132 @@
...
+            Self::Eip7594(sidecar) => sidecar.rlp_encode_fields(out),
+        }
+    }
+
+    /// RLP decode the fields of a [BlobTransactionSidecarVariant] based on the wrapper version.
+    #[doc(hidden)]
+    pub fn rlp_decode_fields(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
+        if buf.first() == Some(&EIP_7594_WRAPPER_VERSION) {
+            buf.advance(1);
+            Ok(Self::Eip7594(BlobTransactionSidecarEip7594::rlp_decode_fields(buf)?))
```
mattsse, line 86: this feels a bit strange, because this kinda behaves like a type tag

that probably shouldn't be encoded like this:

https://github.com/alloy-rs/alloy/blob/cce24039fa259559453d217eb55d7d7eae28d328/crates/eips/src/eip7594/sidecar.rs#L207-L208

because this isn't symmetrical

so one way is wrong I believe

could we add a simple roundtrip test for this as well?

mattsse (changes requested):
now that we have this, we also need another tx type 

or ideally make this one here generic over the sidecar:

https://github.com/alloy-rs/alloy/blob/cce24039fa259559453d217eb55d7d7eae28d328/crates/consensus/src/transaction/eip4844.rs#L782-L782

crates/eips/src/eip7594/sidecar.rs
```diff
@@ -1,17 +1,132 @@
...
 #[cfg(feature = "kzg")]
 use crate::eip4844::BlobTransactionValidationError;
 
+/// This represents a set of blobs, and its corresponding commitments and proofs.
+/// Proof type depends on the sidecar variant.
+///
+/// This type encodes and decodes the fields without an rlp header.
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(untagged))]
```
mattsse, line 21: this untagged impl is quite expensive during deserde, but we can do this as a followup

mattsse (approved):

PR #2434, opened 2025-05-13
feat(consensus): generic sidecar for 4844

rkrasiuk:
## Description

Make `TxEip4844` generic over sidecar. Ref https://github.com/alloy-rs/alloy/pull/2430#pullrequestreview-2834455326


mattsse (approved):
lgtm, there's potential more followup work to do because we can then relax some PooledTransaction impls and need additional conversions for reth, but we can do these seperately

PR #2436, opened 2025-05-13
Stabilize ChainConfig serde for Human-Readable & Binary Formats

### LINKED ISSUE #2435: [Bug] serde for ChainConfig: Roundtrip failures & missing format distinction
### Component

consensus, eips, genesis

v0.15.11

### Operating System

None

### Describe the bug

The current serde logic for Genesis configurations in `alloy-genesis` and `alloy-serde` has the following problems:

1. **Roundtrip errors:** The `ChainConfig` uses custom deserializers (e.g. for TTD or for numbers) but default serializers. However, the default serialization for these types can be different (e.g. hex for `U256`), leading to `deserialize(serialize(data))` failures with serializers other than `serde_json`.
2. **Format ambiguity:** The existing custom serde logic does not consistently distinguish between human-readable (e.g., JSON) and binary data formats, which can lead to parsing errors or incorrect data interpretation.

These issues affect the reliability and predictability of (de)serializing Genesis data across different serde implementations and formats.

Wollac:
## Motivation

This PR refactors Genesis serde logic to fix roundtrip errors and ensure a clear distinction between human-readable (JSON) and binary formats.

Fixes #2435 

## Solution

This PR implements the following changes:

- Updated `deserialize_private_key`, `deserialize_storage_map`, and introduced the `ttd` module. All now use `is_human_readable()` to ensure symmetric (de)serialization:
  - TTD is now (de)serialized as a raw JSON number (using `u128`) in human-readable contexts and via standard U256 serde for binary formats.
  - This replaces the previous TTD deserialization, which included a complex workaround using f64 for very large JSON numbers. That workaround was reportedly to aid with `serde_json`'s limitations regarding untagged enums, but it does not seem to work in this situation either. The u128 approach is simpler and sufficient. (A general `serde_json` issue with untagged enums still exists, as noted by an ignored test in ttd_opt.rs).
- Other optional numeric fields in `ChainConfig` now rely on standard serde behavior, ensuring consistency for both JSON (matching Geth's typical output) and binary formats.
- Added the `hoodi.json` genesis file and a corresponding parsing test.
- `deserialize_json_ttd_opt` is kept for backward compatibility.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (changes requested):

crates/genesis/src/lib.rs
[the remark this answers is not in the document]
Wollac: Only for JSON files that contain hex, which neither alloy nor geth ever produce. Do you still think we should support hex deserialization in human-readable input?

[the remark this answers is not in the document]
Wollac: Arguably, using `alloy-serde::opt` also for serialization, would be an even more breaking change.
So the only alternative would be to add a alloy-genesis specific deserializer, which uses default when non-human-readable and  alloy_serde::quantity::opt::deserialize when human-readable.

crates/serde/src/ttd.rs
```diff
@@ -1,16 +1,58 @@
...
+pub fn serialize<S>(value: &Option<U256>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    if serializer.is_human_readable() {
+        match value {
+            Some(value) => {
+                // convert into an u128 when possible
+                let number = value.try_into().map_err(ser::Error::custom)?;
+                serializer.serialize_u128(number)
```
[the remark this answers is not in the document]
Wollac: It does work with serde_json, and I've even added a test for that.
Serde has (limited) support for u128, even when the arbitrary precision feature is not enabled. However, there is no support for deserializing u128 inside and untagged enum: https://github.com/serde-rs/serde/issues/2230
And as I understand the comments, the workaround has been implemented to support exactly untagged enums. (I even added a test for this particular case).

If we don't need to consider untagged enums (which are irrelevant for ChainConfig), we could even get rid of the workaround altogether and use u128.

crates/genesis/src/lib.rs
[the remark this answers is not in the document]
Wollac: Addressed in 823a16b96f5a4c53456aa5e0fb9dc696267b890f

Wollac:
@mattsse I changed it so that the deserialization is now 100% identical to and compatible with the existing version.

crates/genesis/src/lib.rs
```diff
@@ -293,15 +293,33 @@ fn deserialize_private_key<'de, D>(deserializer: D) -> Result<Option<B256>, D::E
...
+                }
+                B256::from_str(s).map(Some).map_err(D::Error::custom)
+            }
+            None => Ok(None),
         }
-        B256::from_str(s).map(Some).map_err(D::Error::custom)
     } else {
-        Ok(None)
+        Option::<B256>::deserialize(deserializer)
+    }
```
mattsse, line 308: hmm, why we need to change this here, is this because the non human readable impl for b256 is different?

mattsse (commented):
I'm a bit confused why we need to change this via the deserde impls

because this still keeps only `deserialize_with = "deserialize_u64_opt"`

which doesn't seem symmetrical to me

wouldn't this still deserialize hex and serialize u64?

mattsse:
oh I think I get it now

crates/genesis/src/lib.rs
Wollac: In binary, a B256 will not be serialized as a string, but as 32 bytes. Thus in a roundtrip, the Option<String>::deserialize would fail. It only works for human-readable where it is serialized as a hex string.

Wollac:
I think it would conceptually be cleaner (symmetry and Geth compatibility) to only support parsing JSON numbers and not also hex strings.
However, technically this would be a breaking change for config files in a strange format that used to work before. Thus, I've proposed to serialize to JSON numbers, and support the deserialization of both.

crates/serde/src/ttd.rs
```diff
@@ -67,38 +118,13 @@ mod tests {
...
-        );
-    }
-
-    #[test]
-    fn jsonu256_serialize() {
-        let data = U256::from(16);
-        let serialized = serde_json::to_string(&data).unwrap();
-
-        assert_eq!(serialized, r#""0x10""#);
-    }
```
mattsse, line 93: were those test cases moved?
can we add them back or don't the work anymore?

crates/genesis/src/lib.rs
```diff
@@ -293,15 +293,33 @@ fn deserialize_private_key<'de, D>(deserializer: D) -> Result<Option<B256>, D::E
...
+fn deserialize_u64_opt<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    if deserializer.is_human_readable() {
+        alloy_serde::quantity::opt::deserialize(deserializer)
+    } else {
+        Option::<u64>::deserialize(deserializer)
     }
 }
```
mattsse, line 324: this function is a bit out of place here, same for the existing `deserialize_private_key` can we move both to end of file?

mattsse (changes requested):
okay, I think this makes sense, I only have one last question about removed test cases and one nit

crates/genesis/src/lib.rs
Wollac: done 08a6d86173f7c3bac25886fc0ca9a0612e195e72

mattsse:
cool, I think this is good now
I'm delaying merging this a bit because we have fusaka genesis changes wip and it'd be easier to rebase this pr

mattsse:
hey @Wollac sorry for the delay, we just merged https://github.com/alloy-rs/alloy/pull/2542

mind rebasing this pr and update the new fields as well? then we can include this one

mattsse (approved):

PR #2442, opened 2025-05-13
feat(mux-tracer): Support JS tracer in MUX tracer.

kustrun:
Ref: https://github.com/paradigmxyz/revm-inspectors/issues/58

## Motivation

Currently `muxTracer` supports only built-in tracers. It should also support `JS tracers`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [] Breaking changes


mattsse (approved):
cool, that should work.

this is technically breaking, but I think we can get away with this,

just need to time the inspector pr accordingly

PR #2444, opened 2025-05-13
fix(`provider`): FillerNetwork trait should return associated type

yash-atreya:
## Motivation

In the current implementation of the `FillerNetwork` trait the `.network` method returns `Fillers<T, Net>`. 

This is incorrect in the case of `impl FillerNetwork for Identity` as we can end up with `Fillers<Identity, Net>` in the ProviderBuilder stack which should never happen as this is a noop filler. 

## Solution

Change the return type of `FillerNetwork.network` to be the associated type `NewFillers` which corresponds to the `Fillers` stack with the new network. 

In the case of `impl FillerNetwork for Identity` the returned type is `Identity` instead of `Fillers<Identity, Net>`.

Commit to review: [84f6d9b](https://github.com/alloy-rs/alloy/pull/2444/commits/84f6d9be4041fa539f0e83c5ed04497ff51802ea)

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (commented):
we descoped this from v1, and this would be breaking

PR #2451, opened 2025-05-14
chore: add a new fn for TxType derivation

fantasyup:
## Motivation
This is related to https://github.com/paradigmxyz/reth/issues/16231, to upstream `TxType` derive to `TransactionRequest`

## Solution
This PR adds a new function in `TransactionRequest` with logic in https://github.com/paradigmxyz/reth/blob/3ac3e6ff11bac485272d023969bfd0377b820165/crates/optimism/rpc/src/eth/call.rs#L70-L80

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/request.rs
+++ b/crates/rpc-types-eth/src/transaction/request.rs
@@ -332,28 +332,6 @@ impl TransactionRequest {
         self.max_fee_per_gas.is_some() || self.max_priority_fee_per_gas.is_some()
     }
 
-    /// Returns the transaction type based on the fields that are set.
-    ///
-    /// The type is determined in the following order:
-    /// - EIP-7702 if authorization_list is set
-    /// - EIP-4844 if any EIP-4844 fields are set (sidecar, blob hashes, max blob fee)
-    /// - EIP-1559 if any EIP-1559 fee fields are set (max fee per gas, max priority fee)
-    /// - EIP-2930 if access_list is set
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -332,6 +332,28 @@ impl TransactionRequest {
...
 
+    /// Returns the transaction type based on the fields that are set.
+    ///
+    /// The type is determined in the following order:
+    /// - EIP-7702 if authorization_list is set
+    /// - EIP-4844 if any EIP-4844 fields are set (sidecar, blob hashes, max blob fee)
+    /// - EIP-1559 if any EIP-1559 fee fields are set (max fee per gas, max priority fee)
+    /// - EIP-2930 if access_list is set
+    /// - Legacy otherwise
+    pub const fn tx_type(&self) -> TxType {
```
mattsse, line 343: @klkvr this is now copy pasted from reth,

this isn't quite the same as `preferred_type` because for revm iirc we should always fallback to legacy, but cant remember the reason?

should we name this, `derive_tx_type`?

klkvr: I think we could name this e.g `minimal_tx_type`? as the idea here is to have the most primitive txtype that fits all of the specified fields

mattsse (approved):

PR #2463, opened 2025-05-16
feat(consensus): alias sidecar generics

rkrasiuk:
## Description

Make various transaction aliases generic over sidecar.

### DIFF
--- a/crates/consensus/src/transaction/eip2930.rs
+++ b/crates/consensus/src/transaction/eip2930.rs
@@ -259,6 +259,7 @@ impl Decodable for TxEip2930 {
 mod tests {
     use super::*;
     use crate::{SignableTransaction, TxEnvelope};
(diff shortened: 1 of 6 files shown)


crates/consensus/src/transaction/eip2930.rs
```diff
@@ -301,7 +302,7 @@ mod tests {
 
         let tx = request.into_signed(signature);
 
-        let envelope = TxEnvelope::Eip2930(tx);
+        let envelope = TxEnvelope::<BlobTransactionSidecar>::Eip2930(tx);
```
mattsse, line 305: this looks breaking, why is this generic necessary here?

mattsse (commented):
don't think we can do this easily without causing issues,

can we patch this against reth and see what happens?

PR #2480, opened 2025-05-19
fix!: check each bloom

mattsse:
~~this would be an expensive workaround~~

instead of a lazy single bloom, this computes the original BloomFilter again, that was incorrectly changed in #2140 and stores that in a oncelock instead of the single bloom

ref https://github.com/paradigmxyz/reth/issues/16354

### DIFF
--- a/crates/rpc-types-eth/src/filter.rs
+++ b/crates/rpc-types-eth/src/filter.rs
@@ -657,8 +657,10 @@ impl Filter {
     /// Check whether the provided bloom contains all topics and the address we
     /// wish to filter on.
     pub fn matches_bloom(&self, bloom: Bloom) -> bool {
-        bloom.contains(&self.address_bloom())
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/filter.rs
```diff
@@ -644,21 +636,21 @@ impl Filter {
...
 
-    /// Create the [`Bloom`] for the addresses.
-    pub fn address_bloom(&self) -> Cow<'_, Bloom> {
-        self.address.bloom()
+    /// Create the [`BloomFilter`] for the addresses.
+    pub fn address_bloom_filter(&self) -> Cow<'_, BloomFilter> {
+        self.address.bloom_filter()
     }
 
     /// Create a [`Bloom`] for each topic filter.
```
fgimenez, line 644:
```suggestion
    /// Create a [`BloomFilter`] for each topic filter.
```

fgimenez (approved):
lgtm, smol doc nit

Issue #2481, opened 2025-05-20
[Feature] Remove collect from `Variant::versioned_hashes`


mattsse:
### Component

consensus, eips, genesis

### Describe the feature you would like

this currently collects:

https://github.com/alloy-rs/alloy/blob/cb1401f29e106f7ed933e521699c36c1f99ae79f/crates/eips/src/eip7594/sidecar.rs#L97-L103

because we have two different iterator types `impl Iterator`

however both are doing the same:

https://github.com/alloy-rs/alloy/blob/cb1401f29e106f7ed933e521699c36c1f99ae79f/crates/eips/src/eip4844/sidecar.rs#L292-L292

https://github.com/alloy-rs/alloy/blob/cb1401f29e106f7ed933e521699c36c1f99ae79f/crates/eips/src/eip7594/sidecar.rs#L410-L410

which an be replaced with a 

```rust
pub struct VersionedHashIter<'a> {
 commitments: core::slice::Iter<>
}

impl Iterator for VersionedHashIter
fn next() -> {self.next().map(kzg_to_versioned_hash)}
```

cc @rkrasiuk 

then all three functions can return this type

RohitRathore1:
I would like to work on this one. Thanks! @mattsse

PR #2483, opened 2025-05-21
refactor: create VersionedHashIter to remove unnecessary collect()

RohitRathore1:
## Motivation

Fix issue #2481: Remove unnecessary collection into Vec in `versioned_hashes` method.

The `versioned_hashes` method in `BlobTransactionSidecarVariant` was unnecessarily collecting the iterator results into a Vec, which could impact performance. Both implementations of the underlying method were doing the same operation on commitments.

## Solution

Created a common `VersionedHashIter` struct that implements the Iterator trait to handle versioned hash generation from commitments. This allows:

1. All related methods to return the same concrete type instead of `impl Iterator`
2. Removes the need for `.collect()` in `BlobTransactionSidecarVariant::versioned_hashes`
3. Maintains the same functionality while being more memory efficient

## PR Checklist

- [x] Added Tests (existing tests still pass)
- [x] Added Documentation (documented the new iterator type)
- [ ] Breaking changes (no breaking changes introduced)

### DIFF
--- a/crates/eips/src/eip4844/sidecar.rs
+++ b/crates/eips/src/eip4844/sidecar.rs
@@ -5,7 +5,7 @@ use crate::{
         kzg_to_versioned_hash, Blob, BlobAndProofV1, Bytes48, BYTES_PER_BLOB, BYTES_PER_COMMITMENT,
         BYTES_PER_PROOF,
     },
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm

I only moved the type

PR #2488, opened 2025-05-21
feat: Add lenient_block_number_or_tag to support raw integers

louisbrown0212:
## Motivation

BlockNumberOrTag supports only:
- HEX String - an integer block number
- String "latest" - for the latest mined block
- String "finalized" - for the finalized block
- String "safe" - for the safe head block
- String "earliest" for the earliest/genesis block
- String "pending" - for the pending state/transactions

And EIP-1898 does not all calls that use `BlockNumber` like `eth_getBlockByNumber` and doesn't list raw integers as supported.

## Solution

Added a module that deserializes either a BlockNumberOrTag, or a simple number

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip1898.rs
+++ b/crates/eips/src/eip1898.rs
@@ -232,9 +232,13 @@ impl fmt::Debug for BlockNumberOrTag {
     }
 }
 
-/// This is a helper
-/// Various block number representations
-#[derive(Clone, Copy, Debug, Default)]
+/// This is a helper type to allow for lenient parsing of block numbers.
+///
+/// The eth json-rpc spec requires quantities to be hex encoded, which [`BlockNumberOrTag`] strictly
+/// enforces.
+///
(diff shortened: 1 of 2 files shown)


crates/eips/src/eip1898.rs
```diff
@@ -232,6 +232,85 @@ impl fmt::Debug for BlockNumberOrTag {
     }
 }
 
+/// A module that deserializes either a BlockNumberOrTag, or a simple number.
+pub mod lenient_block_number_or_tag {
```
mattsse, line 236: can we not copy paste the existing module and then add the 

```
pub struct LenientBlockNumberOrTag(BlockNumberOrTag);
```

as a new public enum?
this seems easier than doing this

we also want some additional test cases for this

mattsse (changes requested):

mattsse (approved):
lgtm

PR #2491, opened 2025-05-22
feat: add option to always set input+data in MulticallBuilder

### LINKED ISSUE #2371: Using multicall with Tron chain
Tron doesn't accept the format generated by alloy 0.15, and I haven't found a way to override it before calling the `aggragate3()` function. 

Also, on Tron the multicall contract has a different address,  MulticallBuilder allow to change default one so this is fine.

Request generated with alloy 0.15 help:
```sh
curl -X POST https://api.trongrid.io/jsonrpc \
        -d '{"method":"eth_call","params":[{"to":"0x32a4f47a74a6810bd0bf861cabab99656a75de9e","input":"0x82ad56cb0000000000000000000000000000000000000000000000000000000000000
02000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000010000000000000000000000000032a4f47a74a6810bd0bf861cabab99656a75de9e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000244d2301cc000000000000000000000000c45544566aab69ec66720c0ed3ae6b9014a20c0a00000000000000000000000000000000000000000000000000000000000000000000000000000000a614f803b6fd780986a42c78ec9c7f77e6ded13c00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000002470a08231000000000000000000000000c45544566aab69ec66720c0ed3ae6b9014a20c0a00000000000000000000000000000000000000000000000000000000"},"pending"],"id":1,"jsonrpc":"2.0"}'
```

if we change the `input` field to `data` and `"pending"` to `"latest"` it works. A request with just these two changes is shown below:
```sh
curl -X POST https://api.trongrid.io/jsonrpc \
        -d '{"method":"eth_call","params":[{"to":"0x32a4f47a74a6810bd0bf861cabab99656a75de9e","data":"0x82ad56cb000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000010000000000000000000000000032a4f47a74a6810bd0bf861cabab99656a75de9e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000244d2301cc000000000000000000000000c45544566aab69ec66720c0ed3ae6b9014a20c0a00000000000000000000000000000000000000000000000000000000000000000000000000000000a614f803b6fd780986a42c78ec9c7f77e6ded13c00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000060000000000000000000000000000000000000000000000000000000000000002470a08231000000000000000000000000c45544566aab69ec66720c0ed3ae6b9014a20c0a00000000000000000000000000000000000000000000000000000000"}, "latest"],"id":1,"jsonrpc":"2.0"}'
```

In the past, with version 0.12 and the alloy-multicall crate, it worked. Here's an example of a workaround that changed input to data:
https://github.com/Corn3333/alloy-multicall/blob/acfbe372c204970cc94b44056dea890a7d7c75dd/examples/simple_trc20.rs#L56-L72

I’d like to know the recommended way to make this work with Alloy 0.15 on Tron.

### LINKED ISSUE #2489: [Feature] Add option to always set input+data in MulticallBuilder
### Component

rpc

ref #2371 some chains struggle with input/data values, some only support one and fail if both are set.
To address this we need to do a few things

1. Introduce an enum `TransactionInputKind {Input,Data,Both}` to represent these cases
2. Introduce builder functions using this in the TransactionBuilder trait: https://github.com/alloy-rs/alloy/blob/e6dcaaf9ff303fb25ecc2416312ad2f52733b3dd/crates/network/src/transaction/builder.rs#L68-L68
3. add an instance of `TransactionInputKind` to `MultiCallBuilder` https://github.com/alloy-rs/alloy/blob/e6dcaaf9ff303fb25ecc2416312ad2f52733b3dd/crates/provider/src/provider/multicall/mod.rs#L74-L74 and respect here: https://github.com/alloy-rs/alloy/blob/e6dcaaf9ff303fb25ecc2416312ad2f52733b3dd/crates/provider/src/provider/multicall/mod.rs#L530-L530

### Additional context

_No response_

fantasyup:
## Motivation

Closes #2371 and #2489

## Solution
closes #2489

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/transaction/builder.rs
+++ b/crates/network/src/transaction/builder.rs
@@ -56,6 +56,17 @@ impl<N: Network> TransactionBuilderError<N> {
     }
 }
 
+/// Represents how a transaction handles input/data values
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TransactionInputKind {
+    /// Only supports 'input' field
+    Input,
+    /// Only supports 'data' field
+    Data,
+    /// Supports both 'input' and 'data' fields
+    Both,
+}
+
 /// A Transaction builder for a network.
 ///
 /// Transaction builders are primarily used to construct typed transactions that can be signed with
@@ -350,4 +361,39 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
         self,
         wallet: &W,
     ) -> impl_future!(<Output = Result<N::TxEnvelope, TransactionBuilderError<N>>>);
+
+    /// Set the input data for the transaction, respecting the input kind
+    fn set_input_respecting_kind<T: Into<Bytes>>(&mut self, input: T, kind: TransactionInputKind) {
+        match kind {
+            TransactionInputKind::Input => self.set_input(input),
+            TransactionInputKind::Data => self.set_data(input),
+            TransactionInputKind::Both => {
+                let bytes = input.into();
+                self.set_input(bytes.clone());
+                self.set_data(bytes);
+            }
+        }
+    }
+
+    /// Builder-pattern method for setting the input data, respecting the input kind
+    fn with_input_respecting_kind<T: Into<Bytes>>(
+        mut self,
+        input: T,
+        kind: TransactionInputKind,
+    ) -> Self {
+        self.set_input_respecting_kind(input, kind);
+        self
+    }
+
+    /// Get the data for the transaction.
+    fn data(&self) -> Option<&Bytes>;
+
+    /// Set the data for the transaction.
+    fn set_data<T: Into<Bytes>>(&mut self, data: T);
+
+    /// Builder-pattern method for setting the data.
+    fn with_data<T: Into<Bytes>>(mut self, data: T) -> Self {
+        self.set_data(data);
+        self
+    }
 }
--- a/crates/network/src/transaction/builder.rs
+++ b/crates/network/src/transaction/builder.rs
@@ -1,7 +1,7 @@
 use super::signer::NetworkWallet;
 use crate::Network;
 use alloy_primitives::{Address, Bytes, ChainId, TxKind, U256};
-use alloy_rpc_types_eth::AccessList;
+use alloy_rpc_types_eth::{AccessList, TransactionInputKind};
 use alloy_sol_types::SolCall;
 use futures_utils_wasm::impl_future;
 
@@ -54,17 +54,6 @@ impl<N: Network> TransactionBuilderError<N> {
     pub const fn into_unbuilt(self, request: N::TransactionRequest) -> UnbuiltTransactionError<N> {
         UnbuiltTransactionError { request, error: self }
     }
-}
-
-/// Represents how a transaction handles input/data values
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum TransactionInputKind {
-    /// Only supports 'input' field
-    Input,
-    /// Only supports 'data' field
-    Data,
-    /// Supports both 'input' and 'data' fields
-    Both,
 }
 
 /// A Transaction builder for a network.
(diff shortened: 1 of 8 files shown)


crates/network/src/transaction/builder.rs
```diff
@@ -56,6 +56,17 @@ impl<N: Network> TransactionBuilderError<N> {
     }
 }
 
+/// Represents how a transaction handles input/data values
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum TransactionInputKind {
+    /// Only supports 'input' field
+    Input,
```
mattsse, line 63: let's make this the default

and also move this type to where we define:

https://github.com/alloy-rs/alloy/blob/54eb6743c2a281cdfc7b989168d83a492aba69da/crates/rpc-types-eth/src/transaction/request.rs#L1135-L1135

crates/network/src/transaction/builder.rs
```diff
@@ -350,4 +361,39 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
         self,
         wallet: &W,
     ) -> impl_future!(<Output = Result<N::TxEnvelope, TransactionBuilderError<N>>>);
+
+    /// Set the input data for the transaction, respecting the input kind
+    fn set_input_respecting_kind<T: Into<Bytes>>(&mut self, input: T, kind: TransactionInputKind) {
```
mattsse, line 366: I think we can shorten this
```suggestion
    fn set_input_kind<T: Into<Bytes>>(&mut self, input: T, kind: TransactionInputKind) {
```

crates/network/src/transaction/builder.rs
```diff
@@ -350,4 +361,39 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
...
+    fn data(&self) -> Option<&Bytes>;
+
+    /// Set the data for the transaction.
+    fn set_data<T: Into<Bytes>>(&mut self, data: T);
+
+    /// Builder-pattern method for setting the data.
+    fn with_data<T: Into<Bytes>>(mut self, data: T) -> Self {
+        self.set_data(data);
+        self
+    }
```
mattsse, line 398: let's remove these because not relevant

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -102,6 +104,7 @@ where
             block: None,
             state_override: None,
             address: MULTICALL3_ADDRESS,
+            input_kind: TransactionInputKind::Input,
```
mattsse, line 107: these should all be `default`

mattsse (changes requested):
cool, this is already functionally complete

left some nits

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -1121,6 +1121,23 @@ impl From<TxEnvelope> for TransactionRequest {
...
+    Data,
+    /// Supports both 'input' and 'data' fields
+    Both,
+}
+
+impl Default for TransactionInputKind {
+    fn default() -> Self {
+        Self::Input
+    }
+}
```
yash-atreya, line 1139: You can derive Default and mark the Input variant with #[default]

mattsse (approved):
nice

yash-atreya:
> @yash-atreya I think we can add an example to the docs, since this apparently affects some chains that can't handle the input field

Yeah makes sense!

PR #2493, opened 2025-05-22
feat: Introducing BlockOverrides support to EthCallParams

### LINKED ISSUE #2490: [Feature] Add BlockOverrides support to EthCallParams
### Component

rpc

Some ethcall endpoints support blockoverrides in addition to stateoverrides.

we should add this to:
https://github.com/alloy-rs/alloy/blob/54eb6743c2a281cdfc7b989168d83a492aba69da/crates/provider/src/provider/eth_call/params.rs#L12-L12

## TODO
* add block_overrides to `EthCallParams`
* add setters/getts like: https://github.com/alloy-rs/alloy/blob/54eb6743c2a281cdfc7b989168d83a492aba69da/crates/provider/src/provider/eth_call/params.rs#L31-L39
and https://github.com/alloy-rs/alloy/blob/54eb6743c2a281cdfc7b989168d83a492aba69da/crates/provider/src/provider/eth_call/mod.rs#L234-L238
* update serialization https://github.com/alloy-rs/alloy/blob/54eb6743c2a281cdfc7b989168d83a492aba69da/crates/provider/src/provider/eth_call/params.rs#L64-L69:
If block overrides are configured then we must also serialize all values before: block+stateoverrides

### Additional context

_No response_

Ayushdubey86:
resolves #2490 @mattsse @yash-atreya !!

### DIFF
--- a/crates/provider/src/provider/eth_call/params.rs
+++ b/crates/provider/src/provider/eth_call/params.rs
@@ -86,7 +86,9 @@ impl<N: Network> serde::Serialize for EthCallParams<N> {
 
             seq.serialize_element(self.overrides().unwrap_or(&StateOverride::default()))?;
 
-            seq.serialize_element(self.block_overrides().as_ref().unwrap())?;
+            if let Some(block_overrides) = self.block_overrides() {
+                seq.serialize_element(block_overrides)?;
+            }
         } else if let Some(block) = self.block() {
             // Only block is present
             seq.serialize_element(&block)?;
--- a/crates/provider/src/provider/eth_call/params.rs
+++ b/crates/provider/src/provider/eth_call/params.rs
@@ -84,14 +84,11 @@ impl<N: Network> serde::Serialize for EthCallParams<N> {
         if self.block_overrides().is_some() || self.overrides().is_some() {
             seq.serialize_element(&self.block().unwrap_or_default())?;
 
-            if let Some(state_overrides) = self.overrides() {
-                seq.serialize_element(state_overrides)?;
-            }
+            seq.serialize_element(self.overrides().unwrap_or(&StateOverride::default()))?;
 
-            if let Some(block_overrides) = self.block_overrides() {
-                seq.serialize_element(block_overrides)?;
-            }
+            seq.serialize_element(self.block_overrides().as_ref().unwrap())?;
         } else if let Some(block) = self.block() {
+            // Only block is present
             seq.serialize_element(&block)?;
         }
 
--- a/crates/provider/src/provider/eth_call/params.rs
+++ b/crates/provider/src/provider/eth_call/params.rs
@@ -81,16 +81,14 @@ impl<N: Network> serde::Serialize for EthCallParams<N> {
         let mut seq = serializer.serialize_seq(Some(len))?;
         seq.serialize_element(&self.data())?;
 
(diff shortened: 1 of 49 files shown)


crates/provider/src/provider/eth_call/params.rs
```diff
@@ -52,18 +55,40 @@ where
...
 
 impl<N: Network> serde::Serialize for EthCallParams<N> {
     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let len = if self.overrides().is_some() { 3 } else { 2 };
+        let len = match (self.block_overrides(), self.overrides()) {
+            (Some(_), Some(_)) => 4,
+            (Some(_), None) => 3,
+            (None, Some(_)) => 3,
+            (None, None) => 2,
+        };
```
mattsse, line 78: you can simplify this
we keep the previous one, but if blockoverrides are set then this is always  `len = 4`

crates/provider/src/provider/eth_call/params.rs
```diff
@@ -52,18 +55,40 @@ where
...
         seq.serialize_element(&self.data())?;
 
-        if let Some(overrides) = self.overrides() {
+        if let Some(block_overrides) = self.block_overrides() {
+            seq.serialize_element(block_overrides)?;
             seq.serialize_element(&self.block().unwrap_or_default())?;
-            seq.serialize_element(overrides)?;
+            if let Some(state_overrides) = self.overrides() {
+                seq.serialize_element(state_overrides)?;
+            }
```
mattsse, line 88: these must be in the correct order

block, stateoverrides, blockoverrides

crates/provider/src/provider/eth_call/params.rs
```diff
@@ -52,18 +55,40 @@ where
...
 
         let mut seq = serializer.serialize_seq(Some(len))?;
         seq.serialize_element(&self.data())?;
 
-        if let Some(overrides) = self.overrides() {
+        if let Some(block_overrides) = self.block_overrides() {
+            seq.serialize_element(block_overrides)?;
             seq.serialize_element(&self.block().unwrap_or_default())?;
-            seq.serialize_element(overrides)?;
+            if let Some(state_overrides) = self.overrides() {
```
mattsse, line 86: here we need to unwrap_or_default because blockoverrides must go last

mattsse (changes requested):

crates/provider/src/provider/eth_call/params.rs
Ayushdubey86: op catch

Ayushdubey86:
@mattsse pushed all ! 💯

crates/provider/src/provider/eth_call/params.rs
```diff
@@ -52,19 +55,42 @@ where
...
+        } else if self.overrides().is_some() {
+            3
+        } else {
+            2
+        };
 
         let mut seq = serializer.serialize_seq(Some(len))?;
         seq.serialize_element(&self.data())?;
 
-        if let Some(overrides) = self.overrides() {
```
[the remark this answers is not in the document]
Ayushdubey86: pushed

mattsse (approved):

PR #2496, opened 2025-05-23
chore: handle tron '0x' stateRoot as zero

### LINKED ISSUE #2494: [Feature] Support handling "0x" values for stateRoot in AnyHeader
### Component

rpc

ref: https://github.com/tronprotocol/java-tron/issues/6126

ideally 

https://github.com/alloy-rs/alloy/blob/main/crates/consensus-any/src/block/header.rs#L10-L10

can handle `stateRoot: "0x"` values even if they are invlaid and should just default to 0

```sh
curl --location 'https://api.trongrid.io/jsonrpc' --header 'Content-Type:application/json' --data '{          ? master
    "jsonrpc": "2.0",
    "method": "eth_getBlockByNumber",
    "params": ["latest", true],
    "id": 1
}' | jq .result
```

## TODO
* add custom deserialize fn for: https://github.com/alloy-rs/alloy/blob/main/crates/consensus-any/src/block/header.rs#L19-L20 (e.g. String and check if just 0x, maybe even just using https://github.com/alloy-rs/alloy/blob/main/crates/serde/src/lib.rs#L16-L16 works here) + test case from sample tron response
* get sample response and make sure this can be deserialized into: https://github.com/alloy-rs/alloy/blob/main/crates/network/src/any/mod.rs#L88-L88

### Additional context

_No response_

Ayushdubey86:
resolves #2494

### DIFF
--- a/crates/consensus-any/src/block/header.rs
+++ b/crates/consensus-any/src/block/header.rs
@@ -1,7 +1,8 @@
+#[cfg(feature = "serde")]
+use alloc::string::String;
 use alloy_consensus::{BlockHeader, Header};
 #[cfg(feature = "serde")]
 use alloy_primitives::hex;
-
 use alloy_primitives::{Address, BlockNumber, Bloom, Bytes, B256, B64, U256};
 #[cfg(feature = "serde")]
 use serde::{de::Error, Deserialize, Deserializer};
--- a/crates/consensus-any/src/block/header.rs
+++ b/crates/consensus-any/src/block/header.rs
@@ -368,6 +368,7 @@ fn lenient_state_root<'de, D>(deserializer: D) -> Result<B256, D::Error>
 where
     D: serde::de::Deserializer<'de>,
 {
+    use alloc::string::String;
     use core::str::FromStr;
     use serde::de::Error;
 
--- a/crates/consensus-any/src/block/header.rs
+++ b/crates/consensus-any/src/block/header.rs
@@ -1,7 +1,6 @@
 use alloy_consensus::{BlockHeader, Header};
-use alloy_primitives::{Address, BlockNumber, Bloom, Bytes, B256, B64, U256};
-use alloy_serde::deserialize_state_root;
-
+use alloy_primitives::{hex, Address, BlockNumber, Bloom, Bytes, B256, B64, U256};
+use serde::{de::Error, Deserialize, Deserializer};
 /// Block header representation with certain fields made optional to account for possible
 /// differences in network implementations.
 #[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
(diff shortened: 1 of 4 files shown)


crates/serde/src/lib.rs
```diff
@@ -48,3 +49,17 @@ where
 {
     s.serialize_str(&format!("{x:x}"))
 }
+
+/// Custom deserializer for `state_root` that treats `"0x"` or empty as `B256::ZERO`.
+pub fn deserialize_state_root<'de, D>(deserializer: D) -> Result<B256, D::Error>
```
mattsse, line 54: we can move this to the header file and keep as private fn, 
let's also reference the issue in the comment

crates/consensus-any/src/block/header.rs
mattsse, line 1: here we need a test case using the header part of the tron response, see issue.

and ideally a test for `AnyRpcBlock` as well

mattsse (changes requested):
this

crates/consensus-any/src/block/header.rs
Ayushdubey86: yeah , sort of in wip

Ayushdubey86:
@mattsse can you check this out

crates/consensus-any/src/lib.rs
```diff
@@ -12,3 +12,5 @@ pub use block::AnyHeader;
 
 mod receipt;
 pub use receipt::AnyReceiptEnvelope;
+
+extern crate alloc;
```
mattsse, line 16: must go top of file

crates/consensus-any/src/block/header.rs
```diff
@@ -97,6 +103,31 @@ pub struct AnyHeader {
...
+
+    let hex = s.strip_prefix("0x").unwrap_or(s);
+    let bytes = hex::decode(hex).map_err(D::Error::custom)?;
+
+    if bytes.len() != 32 {
+        return Err(D::Error::custom("stateRoot must be exactly 32 bytes"));
+    }
+
+    let mut buf = [0u8; 32];
+    buf.copy_from_slice(&bytes);
```
mattsse, line 127: isn't this just the b256 fromstr impl

```
 B256::from_str(s).map_err(D::Error::custom)

File: crates/consensus-any/src/block/header.rs [hunk 4]
```diff
@@ -1,6 +1,11 @@
+#[cfg(feature = "serde")]
+use alloc::string::String;
 use alloy_consensus::{BlockHeader, Header};
+#[cfg(feature = "serde")]
+use alloy_primitives::hex;
 use alloy_primitives::{Address, BlockNumber, Bloom, Bytes, B256, B64, U256};
-
+#[cfg(feature = "serde")]
+use serde::{de::Error, Deserialize, Deserializer};
```
mattsse, line 8: please no feature gated imports

use full paths where needed instead

mattsse (changes requested):

mattsse (approved):

PR #2498, opened 2025-05-23
feat: add Extended type with alloy trait impls

gejeduck:
## Motivation

ref https://github.com/paradigmxyz/reth/issues/16411
Move `Extended` type from Reth to Alloy 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -62,6 +62,8 @@ pub use signed::Signed;
 pub mod crypto;
 pub mod error;
 
+pub mod extended;
+
 /// Bincode-compatible serde implementations for consensus types.
 ///
 /// `bincode` crate doesn't work well with optionally serializable serde fields, but some of the
--- a/crates/consensus/src/lib.rs
+++ b/crates/consensus/src/lib.rs
@@ -63,6 +63,7 @@ pub mod crypto;
 pub mod error;
 
 pub mod extended;
+pub use extended::Extended;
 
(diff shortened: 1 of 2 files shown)


mattsse (approved):
this lgtm!

possibly that we're still missing a few things, but this is def useful to have here,
because here we also have a similar use case:

https://github.com/foundry-rs/foundry/blob/69db2b0343b9d57deb7f953e45567162035b2674/crates/anvil/core/src/eth/transaction/mod.rs#L587-L602

crates/consensus/src/lib.rs
```diff
@@ -62,6 +62,8 @@ pub use signed::Signed;
 pub mod crypto;
 pub mod error;
 
+pub mod extended;
```
RomanHodulak, line 65: How about a re-export? `pub use extended::Extended`

Using a module named the same as the struct makes for a stuttering-like syntax. I think other transaction wrappers are also being re-exported?

gejeduck: yes right, done

crates/consensus/src/lib.rs
```diff
@@ -63,6 +63,7 @@ pub mod crypto;
 pub mod error;
 
 pub mod extended;
```
RomanHodulak, line 65: This module should probably be private to disambiguate. It's contents are publicly re-exported now.

PR #2499, opened 2025-05-23
Adding support for signing 7702 authorizations

Maroutis:
## Motivation

EIP-7702 introduces a new delegation flow that lets an EOA authorize a separate delegate to act on its behalf.  
To support this, the Ledger signer needs to be able to encode the TLV-style payload and send it to the device for signing. 

## Solution

- Add a new function `sign_auth` which does the following :
  - Internally builds the TLV payload  
  - Then streams it over the APDU `SIGN_EIP7702_AUTHORIZATION` instruction.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer-ledger/Cargo.toml
+++ b/crates/signer-ledger/Cargo.toml
@@ -26,7 +26,6 @@ workspace = true
 alloy-consensus = { workspace = true, features = ["std"] }
 alloy-primitives.workspace = true
 alloy-signer.workspace = true
-rlp = "0.5"
 
 async-trait.workspace = true
 coins-ledger = { version = "0.12", default-features = false }
@@ -40,6 +39,9 @@ alloy-sol-types = { workspace = true, optional = true }
 alloy-dyn-abi = { workspace = true, optional = true }
 alloy-network.workspace = true
 
+# eip7702
+alloy-eips = { workspace = true, optional = true }
+
 [dev-dependencies]
 alloy-consensus = { workspace = true, features = ["std"] }
 alloy-rlp.workspace = true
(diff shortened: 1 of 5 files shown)


crates/signer-ledger/src/signer.rs
```diff
@@ -278,20 +281,50 @@ impl LedgerSigner {
...
     }
 
+    // / Sign “auth data” per EIP-7702:
+    // / msg = keccak256(0x05 ‖ rlp([chain_id, address, nonce]))
+    #[cfg(feature = "eip7702")]
+    pub async fn sign_auth(
+    &self,
+    chain_id: ChainId,
+    delegate: Address,
+    nonce: u64,
```
mattsse, line 291: this is basically:

https://github.com/alloy-rs/eips/blob/ad74348a0c941401a8a42a7662f920274cbd54ab/crates/eip7702/src/auth_list.rs#L48-L56

which I think we can integrate here?

mattsse (changes requested):
thanks, left some suggestions

not familiar with the ~~trezor api~~ so would need someone to take a look here

~~maybe trezor enjoyer @joshieDo ~~

EDIT: this is ledger, should be able to check this

ref https://github.com/LedgerHQ/app-ethereum/issues/715

crates/signer-ledger/Cargo.toml
```diff
@@ -26,6 +26,7 @@ workspace = true
 alloy-consensus = { workspace = true, features = ["std"] }
 alloy-primitives.workspace = true
 alloy-signer.workspace = true
+rlp = "0.5"
```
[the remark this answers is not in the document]
Maroutis: Thanks @mattsse, I've implemented the changes following your inputs.

crates/signer-ledger/src/signer.rs
```diff
@@ -18,6 +18,8 @@ use alloy_sol_types::{Eip712Domain, SolStruct};
 
 #[cfg(feature = "eip7702")]
 use crate::utils::make_eip7702_tlv;
+#[cfg(feature = "eip7702")]
+use alloy_eips::eip7702::Authorization;
```
mattsse, line 22: I'd also like to full paths in this file instead of feature gating imports

PR #2501, opened 2025-05-25
doc: unhide `SendableTx`

0xferrous:
## Motivation

I understand the doc was made `hidden` because as the doc comments suggest this should be considered an internal type. But this internal type is exposed in the `TxFiller` trait and `FillProvider` type and I think since it is already public type because being returned in public API the docs should still be accessible. The warning in the doc comment should be enough for the readers. 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
reasonable

mattsse:
@0xferrous I think we can easily add a way to do this, mind opening a new issue for this?

PR #2512, opened 2025-05-26
fix(provider): CacheLayer - Add block_id to RequestType::params_hash()

alfiehub:
## Motivation

CacheLayer returns the cached value for the first block for subsequent calls against other blocks because `block_id` isn't part of the cache key.

## Solution

Add `block_id` to `RequestType::params_hash()`, which is used as the cache key. `RequestType::with_block_id` is already called where relevant.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse:
> CacheLayer returns the cached value for the first block for subsequent calls against other blocks 

can you elaborate? unclear to me what this means

mattsse:
I see, 

it's this: https://github.com/alloy-rs/alloy/blob/7069d3cc2193749b1577d0ec3d4871175f1ae1fa/crates/provider/src/layers/cache.rs#L124-L124

however we need to solve this a bit differently I think, because we can't cache any Tags ad must always treat those as missing

mattsse (changes requested):
ah, this should fix it, because we already return early here:

https://github.com/alloy-rs/alloy/blob/7069d3cc2193749b1577d0ec3d4871175f1ae1fa/crates/provider/src/layers/cache.rs#L120-L122

can we add another simple test case for this like

https://github.com/alloy-rs/alloy/blob/7069d3cc2193749b1577d0ec3d4871175f1ae1fa/crates/provider/src/layers/cache.rs#L590-L595

but for getBlockByNumber + latest and use anvil_mine in between?

alfiehub:
Added a new test case @mattsse. It increments the Counter and checks the value using `get_storage_at`. I've also updated `RequestType::params_hash` so that it doesnt rely on the `$req.has_block_tag()` early return. Just let me know if you want me to change it back.

mattsse (approved):
lgtm!

PR #2519, opened 2025-05-27
feat: make build_{eip} functions public

### LINKED ISSUE #2516: [Feature] make build_{eip} functions public
### Component

rpc

These are quite useful if you want to build a specific function

we can make all of these pub:
 
https://github.com/alloy-rs/alloy/blob/8466012fdf9543ac57c689d8b326db7afb146d65/crates/rpc-types-eth/src/transaction/request.rs#L797-L802

and change the error from `&str` to `ValueError<Self>`, 

https://github.com/alloy-rs/alloy/blob/8466012fdf9543ac57c689d8b326db7afb146d65/crates/rpc-types-eth/src/transaction/request.rs#L502-L502

this could be a bit disgusting, so we could maybe solve this with a macro or do something like:

https://github.com/alloy-rs/alloy/blob/8466012fdf9543ac57c689d8b326db7afb146d65/crates/rpc-types-eth/src/transaction/request.rs#L792-L796

but ideally retain the actual error, because more useful

### Additional context

_No response_

stevencartavia:
## Motivation
closes #2516

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/request.rs
+++ b/crates/rpc-types-eth/src/transaction/request.rs
@@ -1,17 +1,20 @@
 //! Alloy basic Transaction Request type.
 
 use crate::{transaction::AccessList, BlobTransactionSidecar, Transaction, TransactionTrait};
+use alloc::{
+    string::{String, ToString},
+    vec,
+    vec::Vec,
+};
 use alloy_consensus::{
-    SignableTransaction, TxEip1559, TxEip2930, TxEip4844, TxEip4844Variant, TxEip4844WithSidecar,
-    TxEip7702, TxEnvelope, TxLegacy, TxType, Typed2718, TypedTransaction,
+    error::ValueError, transaction::Recovered, SignableTransaction, TxEip1559, TxEip2930,
+    TxEip4844, TxEip4844Variant, TxEip4844WithSidecar, TxEip7702, TxEnvelope, TxLegacy, TxType,
+    Typed2718, TypedTransaction,
 };
 use alloy_eips::eip7702::SignedAuthorization;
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -135,6 +135,12 @@ pub struct TransactionRequest {
     pub authorization_list: Option<Vec<SignedAuthorization>>,
 }
 
+macro_rules! tx_err {
+    ($self:expr, $msg:expr) => {
+        ValueError::new($self.clone(), $msg)
+    };
+}
```
mattsse, line 142: hmm, this is a bit problematic because this now always does a clone when used in `ok_or(`

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -449,33 +476,44 @@ impl TransactionRequest {
...
             to: to_address,
             value: self.value.unwrap_or_default(),
-            access_list: self.access_list.unwrap_or_default(),
-            blob_versioned_hashes: self
-                .blob_versioned_hashes
-                .ok_or("Missing 'blob_versioned_hashes' field for Eip4844 transaction.")?,
-            max_fee_per_blob_gas: self
-                .max_fee_per_blob_gas
-                .ok_or("Missing 'max_fee_per_blob_gas' field for Eip4844 transaction.")?,
+            access_list: self.access_list.clone().unwrap_or_default(),
```
mattsse, line 507: do we need the clone here?

mattsse (changes requested):
last nits

PR #2523, opened 2025-05-27
feat(`network`): use `FullSigner` in `EthereumWallet` to sign data

yash-atreya:
## Motivation

Currently, the `EthereumWallet` encapsulates a `dyn TxSigner` which only gives access to transaction signing capabilities and NOT message / hash signing e.g signing eip712 typed data.

## Solution

- `FullSigner` gives access to both tx sign and message signing.
- Wrap `FullSigner` in dedicated type `ArcFullSigner` to disambiguate function calls on the signer that are equivalent but from different traits - e.g the `.address(..)` exists on both `TxSigner` and `Signer` requiring fully qualified syntax at callsite by the user, using `ArcFullSigner` type removes the need for that. 
- Use the `ArcFullSigner` in `EthereumWallet`
- Add `sign_hash_with` method

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/network/src/lib.rs
```diff
@@ -6,6 +6,9 @@
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
+#[cfg(feature = "eip712")]
+use alloy_dyn_abi as _;
```
mattsse: why is this required?

crates/network/src/ethereum/wallet.rs
```diff
@@ -207,3 +262,58 @@ impl<W: NetworkWallet<N>, N: Network> IntoWallet<N> for W {
...
+
+    /// Get the underlying [`FullSigner`] as a reference.
+    pub fn signer(&self) -> &dyn FullSigner<Signature> {
+        self.0.as_ref()
+    }
+
+    /// Get the underlying [`FullSigner`] as a mutable reference.
+    pub fn signer_mut(&mut self) -> &mut dyn FullSigner<Signature> {
+        Arc::get_mut(&mut self.0).expect("ArcFullSigner should be unique")
+    }
```
mattsse: we can't guarantee this

mattsse (changes requested):
I think this makes sense,

some nits, can we also send a patch pr to foundry to make sure this doesn't break anything major?

crates/network/src/ethereum/wallet.rs
yash-atreya: [a0e11ae](https://github.com/alloy-rs/alloy/pull/2523/commits/a0e11ae2b2174fd386ae09bf37fea0595036dbe9)

crates/network/src/lib.rs
yash-atreya: [fdf5b65](https://github.com/alloy-rs/alloy/pull/2523/commits/fdf5b650c0057dddaee23a8ddf8abecb7cda407a)

PR #2530, opened 2025-05-28
chore: added TryParseTransportErrorResult

Rimeeeeee:
added  `TryParseTransportErrorResult` which is the result of trying to parse a transport error into a specific interface


crates/transport/src/error.rs
```diff
@@ -96,6 +98,35 @@ impl TransportErrorKind {
...
+    Decoded(I),
+    /// The error was not decoded but the revert data was extracted.
+    UnknownSelector(Bytes),
+    /// The error was not decoded and the revert data was not extracted.
+    Original(TransportError),
+}
+
+impl<I: SolInterface> TryParseTransportErrorResult<I> {
+    /// Attempts to parse a transport error into a specific interface.
+    pub fn try_parse_transport_error(error: TransportError) -> Self {
```
mattsse, line 114: can we move this directly into 

```rust
impl TransportError {}
```

so that this fn can take self

mattsse (changes requested):
cool smol nit

crates/transport/src/error.rs
```diff
@@ -184,6 +186,63 @@ impl RpcErrorExt for RpcError<TransportErrorKind> {
...
+    UnknownSelector(Bytes),
+    /// The error was not decoded and the revert data was not extracted.
+    Original(RpcError<TransportErrorKind, Box<RawValue>>),
+}
+
+/// Extension trait for TransportError parsing capabilities
+pub trait TransportErrorExt {
+    /// Attempts to parse a transport error into a specific interface.
+    fn try_parse_transport_error<I: SolInterface>(self) -> TryParseTransportErrorResult<I>;
+}
```
mattsse: since this is more of a contract feature, I think we should move this to the alloy-contract crate instead.

to 
https://github.com/alloy-rs/alloy/blob/ca4e24f0bbfb9e416c4aac095e6ef0ac06914ab9/crates/contract/src/lib.rs#L15-L15

mattsse (changes requested):
last orga nit

crates/transport/src/error.rs
Rimeeeeee: makes sense

PR #2535, opened 2025-05-30
feat(json-rpc): add request extensions

merklefruit:
## Motivation

Diving into alloy json-rpc I found that the Tower traits are all implemented for the type [`RequestPacket`](https://github.com/alloy-rs/alloy/blob/main/crates/json-rpc/src/packet.rs#L13).

One big difference that I can see with the regular `http::Request` is the lack of [http extensions](https://docs.rs/http/latest/http/request/struct.Request.html#method.extensions) which can be nice to write custom layers that use request-specific metadata.

One example use case would be a custom redundancy layer (e.g. as an expansion to [`FallbackService`](https://github.com/alloy-rs/alloy/blob/main/crates/transport/src/layers/fallback.rs#L31)) that allows per-request redundancy criteria, for instance to require consensus between N transports, or to use the "highest value" returned by N transports.

## Solution

I've added an optional `extensions` field to the `RequestMeta` inside each request packet. 
It should be enough for any Tower layer to query this to define its extra behavior.

### DIFF
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -1,5 +1,6 @@
 use crate::{common::Id, RpcBorrow, RpcSend};
 use alloy_primitives::{keccak256, B256};
+use http::Extensions;
 use serde::{
     de::{DeserializeOwned, MapAccess},
     ser::SerializeMap,
@@ -9,7 +10,7 @@ use serde_json::value::RawValue;
 use std::{borrow::Cow, marker::PhantomData, mem::MaybeUninit};
 
 /// `RequestMeta` contains the [`Id`] and method name of a request.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct RequestMeta {
     /// The method name.
     pub method: Cow<'static, str>,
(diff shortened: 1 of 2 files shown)


crates/json-rpc/src/request.rs
```diff
@@ -9,20 +10,22 @@ use serde_json::value::RawValue;
...
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug)]
 pub struct RequestMeta {
     /// The method name.
     pub method: Cow<'static, str>,
     /// The request ID.
     pub id: Id,
     /// Whether the request is a subscription, other than `eth_subscribe`.
     is_subscription: bool,
+    /// Optional HTTP extensions for the request.
```
mattsse, line 21: I think the docs here are a bit misleading, while this is a type from the http crate, this doesnt have anything to do with http really, can we add a note to this here and the setter+getter that this can be used to store extradata.

crates/json-rpc/src/request.rs
```diff
@@ -41,8 +44,28 @@ impl RequestMeta {
     pub fn set_subscription_status(&mut self, sub: bool) {
         self.is_subscription = sub;
     }
+
+    /// Returns a reference to the request extensions.
+    pub const fn extensions(&self) -> &Extensions {
+        &self.extensions
+    }
```
mattsse, line 51: I'm fine with enshrining this, but then we should also add a extensions_mut fn

crates/json-rpc/src/request.rs
```diff
@@ -41,8 +44,28 @@ impl RequestMeta {
...
+
+    /// Add an extension to the request.
+    pub fn add_extension<T: Clone + Send + Sync + 'static>(&mut self, ext: T) {
+        self.extensions.insert(ext);
+    }
```
mattsse, line 56: if we have extensions_mut then we can remove this imo

mattsse (changes requested):
reasonable imo, some nits

PR #2536, opened 2025-05-30
improve mock transport error messages

0xgleb:
## Motivation

When writing tests for complex logic with multiple JSON RPC calls it's hard to understand a response to what request is being expected from the mock provider.

## Solution

Include request metadata in the error message displayed when the asserter queue is empty.

## PR Checklist

- [X] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
this is better, ty

PR #2538, opened 2025-06-01
chore: relax receipt fn bounds

mattsse:
these can be relaxed so that these can be called with different inner receipt types


crates/rpc-types-eth/src/transaction/receipt.rs
```diff
@@ -204,6 +183,17 @@ impl<T> TransactionReceipt<T> {
     pub fn into_inner(self) -> T {
         self.inner
     }
+
+    /// Calculates the address that will be created by the transaction, if any.
+    ///
+    /// Returns `None` if the transaction is not a contract creation (the `to` field is set), or if
+    /// the `from` field is not set.
```
DaniPopes: from field is always set

DaniPopes (approved):

PR #2540, opened 2025-06-02
fix: make pollers and `Heartbeat` more reliable

klkvr:
# Motivation

Changes pollers and heartbeat to be more tolerant to errors returned from client. Right now node going down for several seconds would cause pollers or heartbeat tasks to get terminated after several retries which might be critical and thus forces any user reliying on those to implement handling of such cases.

With this PR we're now enver ending the task and keep querying the node indefinitely instead

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


yash-atreya (approved):

DaniPopes:
should be fine because both will still shut down when client drops

PR #2543, opened 2025-06-04
feat: EIP-7918

rkrasiuk:
## Description

Implementation of [EIP-7918: Blob base fee bounded by execution cost](https://eips.ethereum.org/EIPS/eip-7918).

### DIFF
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -5,6 +5,12 @@ use crate::{
     eip7594, eip7691,
 };
 
+/// BLOB_BASE_COST represents the minimum execution gas required to include a blob in a block,
+/// as defined by [EIP-7918 (Decoupling Blob Gas from Execution Gas)](https://eips.ethereum.org/EIPS/eip-7918).
+/// This ensures that even though blob gas and execution gas are decoupled, there is still a base
+/// cost in execution gas to include blobs.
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip7840.rs
```diff
@@ -87,9 +105,23 @@ impl BlobParams {
...
+        let target_blob_gas = self.target_blob_gas_per_block();
+        if block_gas_excess_and_used_sum < target_blob_gas {
+            return 0;
+        }
+
+        if BLOB_BASE_COST as u128 * base_fee_per_gas as u128
+            > DATA_GAS_PER_BLOB as u128 * self.calc_blob_fee(excess_blob_gas)
+        {
+            block_gas_excess_and_used_sum
+                + (blob_gas_used * (self.max_blob_count - self.target_blob_count)
```
rakita: Should `max_blob_count` and `target_blob_count` be from the parent block or the current block (In this function current block or next block)? We are currently using parent block max/target count and I can't deduce what is specified in EIP

rkrasiuk: we are currently using next block max/target, but good question

rakita: My mistake, as you said is next blocks max/target.  It feels more logical it is like that, and probably intended in EIP.

mattsse (commented):
superseded by #2821

PR #2545, opened 2025-06-04
Fix misleading doc comment

agryaznov:
EIP1559: `max_fee_per_gas` is not `base_fee_per_gas` (proofs: [1](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md?plain=1#L212), [2](https://github.com/alloy-rs/alloy/blob/84bbfd79222361a07a8febc102e2844bb326e155/crates/provider/src/utils.rs#L119))


mattsse (approved):
fair enough

PR #2554, opened 2025-06-07
docs: fix incorrect type flag doc for EIP-4844 + minor grammar in CI config

MamunC0der:
crates/consensus/src/receipt/envelope.rs
 type flag 2 → type flag 3
Why: EIP-4844 uses type ID 3. Docs were wrong, now fixed.

.config/zepter.yaml
expose → exposes

A it outside → A is outside
Why: Grammar fixes for clarity in CI config comments.

### DIFF
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -39,7 +39,7 @@ pub enum ReceiptEnvelope<T = Log> {
     /// [EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
     #[cfg_attr(feature = "serde", serde(rename = "0x2", alias = "0x02"))]
     Eip1559(ReceiptWithBloom<Receipt<T>>),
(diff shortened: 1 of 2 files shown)


crates/consensus/src/receipt/envelope.rs
```diff
@@ -39,7 +39,7 @@ pub enum ReceiptEnvelope<T = Log> {
     /// [EIP-1559]: https://eips.ethereum.org/EIPS/eip-1559
     #[cfg_attr(feature = "serde", serde(rename = "0x2", alias = "0x02"))]
     Eip1559(ReceiptWithBloom<Receipt<T>>),
-    /// Receipt envelope with type flag 2, containing a [EIP-4844] receipt.
+    /// Receipt envelope with type flag 3, containing a [EIP-4844] receipt.
```
mattsse, line 42: this is an actual doc fix so I let this one through

PR #2556, opened 2025-06-07
feat(provider): Add eth_sendBundle support to provider

cakevm:
This adds a new method + builder to the provider:
- `send_bundle`: send a MEV bundle

This feature also introduces custom http headers for a http request. For that the [extension feature](https://docs.rs/http/latest/http/struct.Extensions.html) is used and a custom type `HttpHeaderExtension` is introduces. A transport can choose to support it e.g. `transport-http`.

For flexibility `FLASHBOTS_SIGNATURE_HEADER` and `sign_flashbots_payload` are exported, if someone prefers to build the request manually.

Usage:
```rust
// without signing
provider.send_bundle(bundle).await;

// with X-Flashbots-Signature
let signer = PrivateKeySigner::random();
provider.send_bundle(bundle).with_auth(signer).await;
```

Find here a full example how it works without this feature: https://github.com/cakevm/alloy-flashbots-client/blob/main/src/main.rs


gakonst:
Should this be in a separate repo and under an extension trait? alloy-rs/mev?

cakevm:
I think this question goes in the direction of the maintainers, but I would suggest keeping it here in `alloy`. Implementing endpoints in the provider for `rpc-types-mev` is, so far, not significantly different from the other provider implementations (aside from the Flashbots authentication). Since it’s feature-gated, keeping everything in one place could help to keep the maintenance overhead reasonable. This could change if more edge cases arise or the implementation becomes too large.

crates/json-rpc/src/packet.rs
```diff
@@ -125,6 +126,27 @@ impl RequestPacket {
...
         self.requests().iter().map(|req| req.method())
     }
+
+    /// Returns a [`HeaderMap`] from the request if `HttpHeaderExtension` is present;
+    /// otherwise, returns an empty map. Only supported for single requests.
+    pub fn headers(&self) -> HeaderMap {
+        if let Some(single_req) = self.as_single() {
+            if let Some(http_header_extension) =
+                single_req.meta().extensions().get::<HttpHeaderExtension>()
+            {
```
mattsse: I think this makes sense as a workaround because this is a common use case,

can we add helper fn for retrieving the `HttpHeaderExtension`

maybe even as mut so that the user doesnt have to deal with the extension type directly

crates/rpc-client/src/call.rs
```diff
@@ -276,6 +276,19 @@ where
             _pd: PhantomData,
         }
     }
+
+    /// Maps the metadata of the request using the provided function.
+    pub fn map_meta(self, f: impl FnOnce(RequestMeta) -> RequestMeta) -> Self {
```
mattsse, line 281: this is public but can panic, this feels wrong, why do we need this?

cakevm: This is a valid point, better be safe then sorry. I was initially inspired by the `map_params` fn above and just copied it. I guess it was intended as `unreachable!`, but I introduces now an error type here. I think it is reasonable.

This here should be refactored as well, but I like to keep the PR small. But I am ok to pick this up and make the use of `RequestAlreadySentError` here as well:
```rust
pub fn map_params<NewParams: RpcSend>(
    self,
    map: impl Fn(Params) -> NewParams,
) -> RpcCall<NewParams, Resp, Output, Map> {
    let CallState::Prepared { request, connection } = self.state else {
        panic!("Cannot get request after request has been sent");
    };
...
```

crates/json-rpc/src/packet.rs
cakevm: Yes this would be a great addition in the future

crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -0,0 +1,145 @@
...
+
+/// Uses the provided signer to generate a signature for Flashbots authentication.
+/// Returns the value for the `X-Flashbots-Signature` header.
+///
+/// See [here](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication) for more information.
+pub async fn sign_flashbots_payload<S: Signer + Send + Sync>(
+    body: String,
+    signer: &S,
+) -> Result<String, alloy_signer::Error> {
+    let message_hash = keccak256(body.as_bytes()).to_string();
```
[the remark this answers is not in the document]
cakevm: This is not correct. It would fail with "error in signature check...". Guess in has worked with ethers that way

cakevm:
@mattsse PR is ready to be reviewed again. I plan to add other methods now as well and challenge the data model a bit to make it more robust/compatible with different builder implementations. But would do this in separate PRs.

crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -0,0 +1,145 @@
...
+                    .map_err(TransportErrorKind::custom)?;
+                let signature = sign_flashbots_payload(body, &signer)
+                    .await
+                    .map_err(TransportErrorKind::custom)?;
+
+                // Add the Flashbots signature to the request headers
+                let headers: HttpHeaderExtension = HttpHeaderExtension::from_iter([(
+                    FLASHBOTS_SIGNATURE_HEADER.to_string(),
+                    signature,
+                )]);
```
mattsse: what if we introduce a simple wrapper type for this?

and also add a function for this directly on meta?

then we can fully control how additional headers are inserted as extensions and the user doesn't have to deal with the correct types

mattsse (changes requested):

crates/provider/src/ext/mev/with_auth.rs
cakevm: Could you layout an example? I am not sure how such a wrapper type should look like, thought `HashMap<String, String>` would be already something like that.

mattsse: > HashMap<String, String> would be already something like that.

any hashmap<string,string> would then mean headermap, which could be misused.

instead of an alias we can add a new type

cakevm: You would suggest something like this?
```rust
pub struct HttpHeaderExtension {
    headers: HashMap<String, String>,
}
impl HttpHeaderExtension {
  pub fn insert<K, V>(mut self, key: K, value: V) -> Self
      where K: Into<String>, V: Into<String>, {
          self.headers.insert(key.into(), value.into());
          self
  }
}
```

We could maybe also export [`HeaderMap`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html) and related types from `json-rpc` that already has the `http` dep. Feels simpler than building an own type for that. How do you feel about that? I did not do that in first place because I was to afraid to add `http` dep to the `provider` crate, and did not thought about exporting.

cakevm: I did the refactoring to [HeaderMap](https://docs.rs/http/latest/http/header/struct.HeaderMap.html). I am unsure about the re-export, or just add `http` to the provider, since it has this dep anyway through `alloy-json-rpc`. Feels more "standard" to just use the `HeaderMap` type. How do you like it?

crates/json-rpc/src/request.rs
```diff
@@ -9,6 +9,9 @@ use serde::{
 use serde_json::value::RawValue;
 use std::{borrow::Cow, marker::PhantomData, mem::MaybeUninit};
 
+/// A map of HTTP headers that extends the request's headers when supported by the transport.
+pub type HttpHeaderExtension = HeaderMap;
```
mattsse: introducing an alias for this is a bit confusing, now we use two different type identifiers in packet.rs for the same headermap type

crates/json-rpc/src/packet.rs
```diff
@@ -125,6 +126,23 @@ impl RequestPacket {
...
+    /// Returns a [`HeaderMap`] from the request if `HttpHeaderExtension` is present;
+    /// otherwise, returns an empty map. Only supported for single requests.
+    pub fn headers(&self) -> HeaderMap {
+        // If this is a batch request, we cannot return headers.
+        let Some(single_req) = self.as_single() else {
+            return HeaderMap::new();
+        };
+        // If the request has an `HttpHeaderExtension`, return its headers.
+        if let Some(http_header_extension) =
+            single_req.meta().extensions().get::<HttpHeaderExtension>()
```
mattsse: this is just HeaderMap

crates/json-rpc/src/request.rs
cakevm: You are absolute right. Let's use the `HeaderMap` everywhere. I think it feels more "standard" now.

crates/json-rpc/src/packet.rs
cakevm: see above

crates/rpc-client/src/call.rs
```diff
@@ -276,6 +290,19 @@ where
             _pd: PhantomData,
         }
     }
+
+    /// Maps the metadata of the request using the provided function.
+    pub fn map_meta(self, f: impl FnOnce(RequestMeta) -> RequestMeta) -> Result<Self, CallError> {
```
cakevm: @mattsse Could you give this a second look? Is this still the way to go here?

The behavior for `map_params` is currently to panic, because the `provider` guarantees the order of the calls. I originally implemented the same for `map_meta`, but after the review, I switched to handling errors instead.

As mentioned before, I’d prefer not to add too much in this PR, but should I create a follow-up PR to adapt `map_params`?

mattsse: these functions are a bit cursed...

let's mirror how we handle this for map_params, this shouldn't panic if used when constructing the request itself, so this would be okay

crates/rpc-client/src/call.rs
```diff
@@ -13,8 +13,22 @@ use std::{
...
+pub enum CallError {
+    /// Error returned when attempting to modify a request that has already been sent.
+    /// This typically occurs after the RPC call has progressed beyond its initial prepared state.
+    #[error("Cannot modify request after it has been sent")]
+    RequestAlreadySent,
+
+    /// Error returned when expected request is missing in prepared state.
+    #[error("No request found in prepared state")]
+    NoRequestInPrepared,
+}
```
mattsse: let's remove this,

sorry didn't look at the other map fns first

mattsse (changes requested):

crates/rpc-client/src/call.rs
cakevm: Sure, reverted it. Indeed it looks a bit odd in the first place. But less code now :)

crates/transport-http/src/hyper_transport.rs
```diff
@@ -92,16 +92,24 @@ where
...
+        let mut builder = hyper::Request::builder()
+            .method(hyper::Method::POST)
+            .uri(self.url.as_str())
+            .header(header::CONTENT_TYPE, header::HeaderValue::from_static("application/json"));
+
+        // Add any additional headers from the request packet.
+        for (key, value) in req.headers() {
+            if let Some(key) = key {
+                builder = builder.header(key, value);
+            }
```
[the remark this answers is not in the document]
cakevm: Yep, that looks better

mteam88:
note: using the `mev-api` feature requires using alloy-provider instead of the alloy crate directly like this
`alloy-provider = { version = "1.0.16", features = ["mev-api"] }`

because the feature isn't on the alloy crate. For other features there is things like `provider-anvil-api` that are exposed apparently.

PR #2573, opened 2025-06-10
fix: correctly decode jwt keys

klkvr:
## Motivation

This currently breaks fusaka-devnet-1

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense

PR #2577, opened 2025-06-12
fix: ERC20 endpoints return type

pistomat:
## Motivation

In Foundry the ERC20 endpoints return `()`.

## Solution

Make the ERC20 endpoints in alloy also return `()`, otherwise you get an error `Error("invalid type: null, expected 32 bytes, represented as a hex string of length 64, an array of u8, or raw bytes", line: 1, column: 4)`

Added tests to make sure it passes.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/anvil.rs
+++ b/crates/provider/src/ext/anvil.rs
@@ -383,10 +383,12 @@ mod tests {
     };
     use alloy_eips::BlockNumberOrTag;
     use alloy_network::TransactionBuilder;
-    use alloy_primitives::B256;
+    use alloy_primitives::{address, B256};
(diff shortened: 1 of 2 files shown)


crates/provider/src/ext/anvil.rs
```diff
@@ -1055,4 +1057,57 @@ mod tests {
         assert_eq!(res.from, alice);
         assert_eq!(res.to, Some(bob));
     }
+
+    #[tokio::test]
+    async fn test_anvil_deal_erc20() {
```
mattsse, line 1062: these are very useful, ty for that

PR #2578, opened 2025-06-13
feat(rpc-types-eth): add helper methods to AccountInfo

mattsse:
## Summary
- Add `is_empty_code_hash()` method to check if code hash is empty
- Add `code_hash()` method to compute the account's code hash
- Add `is_empty()` method to check if account is empty (zero balance, nonce, and empty code)

### DIFF
--- a/crates/rpc-types-eth/src/account.rs
+++ b/crates/rpc-types-eth/src/account.rs
@@ -24,7 +24,7 @@ impl AccountInfo {
     /// Returns true if the code hash is the Keccak256 hash of the empty string `""`.
     #[inline]
     pub fn is_empty_code_hash(&self) -> bool {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/account.rs
```diff
@@ -20,6 +20,30 @@ pub struct AccountInfo {
     pub code: Bytes,
 }
 
+impl AccountInfo {
+    /// Returns true if the code hash is the Keccak256 hash of the empty string `""`.
+    #[inline]
+    pub fn is_empty_code_hash(&self) -> bool {
+        self.code_hash() == KECCAK256_EMPTY
```
DaniPopes: self.code.is_empty

PR #2579, opened 2025-06-14
Update install_test_binaries.sh

Daulox92:
Comment Update

Old: # Installs geth from https://geth.ethereum.org/downloads

New: # Install geth from https://geth.ethereum.org/downloads

Why: To match the imperative style used elsewhere (e.g., # Install reth). Improves consistency and aligns with standard script documentation tone.

Curl Command Fix

Old: curl -so $NAME.zip

New: curl -sLo "$NAME.zip"

Why:

Added -L to follow redirects (which GitHub links require).

Quoted variable to handle filenames with spaces.

This improves reliability and correctness of the download.

### DIFF
--- a/.github/scripts/install_test_binaries.sh
+++ b/.github/scripts/install_test_binaries.sh
@@ -24,7 +24,7 @@ main() {
     wait
 }
 
-# Installs geth from https://geth.ethereum.org/downloads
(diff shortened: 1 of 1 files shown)


mattsse (commented):
dont think we need this

PR #2583, opened 2025-06-15
chore(rpc-types-mev): improve bundle API flexibility

cakevm:
Changes for `EthSendBundle`, `EthBundleHash` and removal of `SendBundleResponse`:
- `bundle_hash` in `EthBundleHash` is optional. Not all builders return such a bundle hash
- Remove duplicate struct `SendBundleResponse` in favor of `EthBundleHash`
- New `extra_fields` as known from `ChainConfig` allow the user to add builder specific fields 

Again a breaking change, but I have the feeling that it is worth it, for a better usability.

Update: Initially I replaced `Vec` with `HashSet` for fields like `reverting_tx_hashes`, but this would potentially lead to a non-deterministic bundle hash calculation. I reverted this change.

### DIFF
--- a/crates/rpc-types-mev/src/eth_calls.rs
+++ b/crates/rpc-types-mev/src/eth_calls.rs
@@ -1,7 +1,7 @@
 use crate::{u256_numeric_string, Privacy, Validity};
 
 use alloy_eips::{eip2718::Encodable2718, BlockNumberOrTag};
-use alloy_primitives::{keccak256, map::HashSet, Address, Bytes, Keccak256, TxHash, B256, U256};
+use alloy_primitives::{keccak256, Address, Bytes, Keccak256, TxHash, B256, U256};
 use alloy_rpc_types_eth::TransactionIndex;
 use alloy_serde::OtherFields;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -312,15 +316,7 @@ impl EthSendBundle {
...
-    pub bundle_hash: B256,
-}
-
-/// Response from the matchmaker after sending a bundle.
-#[derive(Deserialize, Debug, Serialize, Clone, PartialEq, Eq)]
-#[serde(rename_all = "camelCase")]
-pub struct SendBundleResponse {
-    /// Hash of the bundle bodies.
-    pub bundle_hash: B256,
+    pub bundle_hash: Option<B256>,
```
mattsse, line 323: > EthBundleHash is optional. Not all builders return such a bundle hash

great stuff....

crates/rpc-types-mev/src/eth_calls.rs
```diff
@@ -312,15 +316,7 @@ impl EthSendBundle {
 #[serde(rename_all = "camelCase")]
 pub struct EthBundleHash {
     /// Hash of the bundle bodies.
-    pub bundle_hash: B256,
-}
-
-/// Response from the matchmaker after sending a bundle.
-#[derive(Deserialize, Debug, Serialize, Clone, PartialEq, Eq)]
-#[serde(rename_all = "camelCase")]
-pub struct SendBundleResponse {
```
mattsse, line 321: can we keep this, but mark deprecated, because merging this would cause issues in reth because this is currently used in a trait:

https://github.com/paradigmxyz/reth/blob/fb477d8c28b41df73da6c449a0bbc3d072aca73f/crates/rpc/rpc-api/src/mev.rs#L30-L30

cakevm: Makes sense

mattsse (approved):

cakevm:
I’m not sure, but Clippy used to complain whenever I added #[deprecated] to the struct. I believe the type mapping will achieve the same goal :)

mattsse (approved):

PR #2584, opened 2025-06-15
feat: add CCIP (EIP-3668) support

mattsse:
## Summary

This PR adds support for CCIP (Cross Chain Interoperability Protocol) as defined in [EIP-3668](https://eips.ethereum.org/EIPS/eip-3668). CCIP allows smart contracts to transparently fetch external data during eth_call execution through standardized offchain lookups.

## Changes

- Add  wrapper that extends  with automatic OffchainLookup error handling
- Implement gateway resolution following the EIP-3668 specification
- Add  trait with a default  implementation
- Introduce dedicated  type for future extensibility
- Gate entire CCIP module behind the  feature flag

## Usage

```rust
// Enable CCIP resolution for a call
let result = provider
    .call(tx)
    .ccip(provider.clone())
    .await?;
```

## Implementation Details

- Supports recursive CCIP lookups (max depth: 4 as per spec)
- Handles URL template substitution for  and 
- Validates gateway responses and timeout handling
- Extensible through custom  implementations

## TODO

- [ ] Implement proper callback result decoding (currently returns error)
- [ ] Add integration tests with a mock CCIP gateway
- [ ] Consider adding helper methods to  for timeout/retry configuration


DaniPopes:
wrong issue linked

crates/provider/src/provider/ccip.rs
```diff
@@ -0,0 +1,603 @@
...
+        let call_data = call_data.clone();
+
+        Box::pin(async move {
+            // Parse URL and substitute parameters
+            let url = url
+                .replace("{sender}", &format!("{:?}", sender))
+                .replace("{data}", &format!("0x{}", hex::encode(&call_data)));
+
+            // Determine if GET or POST based on URL length
+            let response = if url.len() > 2048 {
```
gskril: Idk Rust well so could be misinterpreting, but determining GET vs POST should be based on the existence of `{data}` in the URL instead of length. According to [the spec](https://eips.ethereum.org/EIPS/eip-3668):

> If the URL template does not contain the `{data}` substitution parameter, the client MUST send a POST request after replacing the substitution parameters as described above. The POST request MUST be sent with a Content-Type of `application/json`, and a payload matching the following schema:

mattsse: yeah this should be removed

PR #2589, opened 2025-06-17
feat(provider): shorthand method to add multicall with allowFailure

beeb:
## Motivation

At the moment, adding a call to a `MulticallBuilder` while specifying that the call is allowed to revert when used in `aggregate3` is quite cumbersome:

```rust
use alloy::providers::{CallItem, MulticallItem as _};
// ...
let call = mycontract.someFunction(arg);
let mut multicall = provider.multicall().dynamic::<someFunctionCall>();
// add multiple calls
multicall = multicall.add_call_dynamic(CallItem::new(call.target(), call.input()).allow_failure(true));
// ...
let res = multicall.aggregate3().await?;
```

Notably, the `MulticallItem` trait must be in scope to use `.target()` and `.input()` on the call, and the dynamic multicall builder must be typed manually with the call's type.

## Solution

Add two methods: `add3` and `add_dynamic3` which take in a `impl MulticallItem` but have a second parameter to set whether the call can revert.

```rust
let mut multicall = provider.multicall().dynamic();
// add multiple calls
multicall = multicall.add_dynamic3(mycontract.someFunction(arg), true);
// ...
let res = muticall.aggregate3().await?;
```

This allows type inference to work its magic and doesn't require to construct a `CallItem` manually. 

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/multicall/mod.rs
+++ b/crates/provider/src/provider/multicall/mod.rs
@@ -202,6 +202,23 @@ where
         self
     }
 
+    /// Add a dynamic call to the builder while specifying whether it is allowed to fail
+    ///
+    /// This is useful in combination with [`aggregate3`][MulticallBuilder::aggregate3].
+    pub fn add_dynamic3(
+        mut self,
+        item: impl MulticallItem<Decoder = D>,
+        allow_failure: bool,
+    ) -> Self {
+        let target = item.target();
+        let input = item.input();
+
+        let call = CallItem::<D>::new(target, input).allow_failure(allow_failure);
+
+        self.calls.push(call.to_call3_value());
+        self
+    }
+
     /// Add a dynamic [`CallItem`] to the builder
     pub fn add_call_dynamic(mut self, call: CallItem<D>) -> Self {
         self.calls.push(call.to_call3_value());
@@ -286,6 +303,28 @@ where
         let input = item.input();
 
         let call = CallItem::<Item::Decoder>::new(target, input);
+
+        self.add_call(call)
+    }
+
+    /// Appends a [`SolCall`] to the stack while specifying whether it is allowed to fail
+    ///
+    /// This is useful in combination with [`aggregate3`][MulticallBuilder::aggregate3].
+    #[expect(clippy::should_implement_trait)]
+    pub fn add3<Item: MulticallItem>(
+        self,
+        item: Item,
+        allow_failure: bool,
+    ) -> MulticallBuilder<T::Pushed, P, N>
+    where
+        Item::Decoder: 'static,
+        T: TuplePush<Item::Decoder>,
+        <T as TuplePush<Item::Decoder>>::Pushed: CallTuple,
+    {
+        let target = item.target();
+        let input = item.input();
+
+        let call = CallItem::<Item::Decoder>::new(target, input).allow_failure(allow_failure);
 
         self.add_call(call)
     }
--- a/crates/contract/src/multicall.rs
+++ b/crates/contract/src/multicall.rs
@@ -175,6 +175,21 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn add3() {
+        let provider = ProviderBuilder::new()
+            .connect_anvil_with_wallet_and_config(|a| a.fork(FORK_URL))
+            .unwrap();
+
+        let dummy_addr = deploy_dummy(provider.clone()).await;
+
+        // allow failure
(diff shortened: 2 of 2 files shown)


crates/provider/src/provider/multicall/mod.rs
```diff
@@ -202,6 +202,23 @@ where
         self
     }
 
+    /// Add a dynamic call to the builder while specifying whether it is allowed to fail
+    ///
+    /// This is useful in combination with [`aggregate3`][MulticallBuilder::aggregate3].
+    pub fn add_dynamic3(
```
mattsse: this is the same as 

  pub fn add_dynamic(mut self, item: impl MulticallItem<Decoder = D>) -> Self {

but with the additional allow_failure?

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -202,6 +202,23 @@ where
...
+        mut self,
+        item: impl MulticallItem<Decoder = D>,
+        allow_failure: bool,
+    ) -> Self {
+        let target = item.target();
+        let input = item.input();
+
+        let call = CallItem::<D>::new(target, input).allow_failure(allow_failure);
```
mattsse: these all operate on MulticallItem

should we make this an additional trait function

then we maybe don't this function if we can do

`item.into_call()` and `item.into_call_with(allow_failure)` or smth?

crates/provider/src/provider/multicall/mod.rs
beeb: Yes!

crates/provider/src/provider/multicall/mod.rs
beeb: You mean to replace the inner implementation here? Would that replace the new functions I proposed here or be an additional change?

beeb: I guess you mean `.add_call_dynamic(mycontract.someFunction(arg).into_call_with(true))` which would replace `add_dynamic3`

beeb: @mattsse what do you think if I implement the following:

```rust
pub trait MulticallItem {
    /// Decoder for the return data of the call.
    type Decoder: SolCall;

    /// The target address of the call.
    fn target(&self) -> Address;
    /// ABI-encoded input data for the call.
    fn input(&self) -> Bytes;

    /// Converts `self` to a [`CallItem`] while specifying whether it can fail.
    fn into_call(self, allow_failure: bool) -> CallItem<Self::Decoder>
    where
        Self: Sized,
    {
        CallItem::<Self::Decoder>::from(self).allow_failure(allow_failure)
    }
}
```

And

```rust
impl<D, T> From<T> for CallItem<D>
where
    D: SolCall,
    T: MulticallItem,
{
    /// Converts a [`MulticallItem`] into a [`CallItem`]
    ///
    /// By default, it doesn't allow for failure when used in [`aggregate3`][crate::MulticallBuilder::aggregate3].
    /// Call [`allow_failure`][CallItem::allow_failure] on the result to specify the failure behavior.
    fn from(value: T) -> Self {
        CallItem::<D>::new(value.target(), value.input())
    }
}
```

mattsse: yeah, but we can also keep these ones here, doesn't hurt really

beeb: Should I push the above conversions in the PR too?

EDIT: i did

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -282,14 +291,26 @@ where
...
-        let call = CallItem::<Item::Decoder>::new(target, input);
-
+        let call: CallItem<Item::Decoder> = item.into();
         self.add_call(call)
     }
 
+    /// Appends a [`SolCall`] to the stack while specifying whether it is allowed to fail
+    ///
+    /// This is useful in combination with [`aggregate3`][MulticallBuilder::aggregate3].
+    pub fn add3<Item: MulticallItem>(
```
mattsse: I find this naming a bit confusing, because this is the same add just with and additional bool,

these functions now also seem no longer necessary now that we have into_call:

self.add_call(item.into_call(allow_failure))

beeb: Yes I agree, probably fine to remove them now.

beeb: Fixed

PR #2591, opened 2025-06-17
feat(rpc): Convert into RPC transaction for generic inner `Transaction`

RomanHodulak:
## Motivation

The RPC transaction can be created from a generic `alloy_consensus::Transaction`. However, this does not cover the case when the RPC inner transaction needs to be converted, for example `EthereumTxEnvelope<Tx4844>` into `EthereumTxEnvelope<Tx4844Variant>`.

## Solution

Add another generic transaction as the input for the conversion, with a `From` bound and a `convert` call.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/mod.rs
+++ b/crates/rpc-types-eth/src/transaction/mod.rs
@@ -151,7 +151,11 @@ where
     }
 
     /// Converts a consensus `tx` with an additional context `tx_info` into an RPC [`Transaction`].
-    pub fn from_transaction(tx: Recovered<T>, tx_info: TransactionInfo) -> Self {
+    pub fn from_transaction<TxIn>(tx: Recovered<TxIn>, tx_info: TransactionInfo) -> Self
+    where
+        TxIn: TransactionTrait,
+        T: From<TxIn>,
+    {
         let TransactionInfo {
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -151,7 +151,11 @@ where
     }
 
     /// Converts a consensus `tx` with an additional context `tx_info` into an RPC [`Transaction`].
-    pub fn from_transaction(tx: Recovered<T>, tx_info: TransactionInfo) -> Self {
+    pub fn from_transaction<TxIn>(tx: Recovered<TxIn>, tx_info: TransactionInfo) -> Self
```
mattsse, line 154: does this require a type hint for `Transactio::from_transaction`?

if not then we can do this without causing problems, but we should check.

RomanHodulak: Yes, that is true unfortunately. The type must be known at some point and AFAIK cannot be set that `T = TxIn` by default.

So I guess that would cause problems.

RomanHodulak:
Closing, see the discussion.

PR #2594, opened 2025-06-18
feat: more serde compat for `TransactionEnvelope` macro

klkvr:
## Motivation

ref https://github.com/alloy-rs/op-alloy/pull/551#issuecomment-2980123042

## Solution

Adds support for forwarding `serde` attributes defined on variants to the serde implementation (tagged/untagged enums). Adds serde type aliases which use uppercased type bytes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
lgtm!

PR #2603, opened 2025-06-21
feat: added convinient fn decode_2718_exact

### LINKED ISSUE #2602: [Feature] Add decode_2718_exact convenience fn
### Component

consensus, eips, genesis

we often need this pattern:

```
            let mut buf = tx.as_ref();
            let tx = T::decode_2718(&mut buf)?;
            if !buf.is_empty() {
                return Err(alloy_rlp::Error::UnexpectedLength);
            }
```

because `decode_2718` does not care about trailing bytes by design.

so we can introduce a convenience fn

https://github.com/alloy-rs/alloy/blob/d49bf5e93a3b224e5887f8598ec4e85bb8994d6a/crates/eips/src/eip2718.rs#L111-L118

that only accepts &[u8] and not mut because we don't care about advancing the buffer because it is expected to return an error if not empty.

## TODO
* add this helper with default impl

### Additional context

_No response_

Soubhik-10:
closes #2602

### DIFF
--- a/crates/eips/src/eip2718.rs
+++ b/crates/eips/src/eip2718.rs
@@ -124,9 +124,23 @@ pub trait Decodable2718: Sized {
             .unwrap_or_else(|| Self::fallback_decode(buf))
     }
 
-    /// Decode a transaction according to [EIP-2718], ensuring no trailing bytes.This is intended
-    /// to decode a single transaction from the entire buffer without leaving any trailing
-    /// bytes.This is needed because decode_2718 does not care about trailing bytes by design.
+    /// Decode a transaction according to [EIP-2718], ensuring no trailing bytes.
+    ///
+    /// This method decodes a single transaction from the entire buffer and ensures that the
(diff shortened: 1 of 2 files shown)


crates/rpc-types-engine/src/payload.rs
mattsse, line 1: can we keep this unchanged for now,
I have a pr wip that would result in some conflicts here

crates/eips/src/eip2718.rs
```diff
@@ -124,6 +124,16 @@ pub trait Decodable2718: Sized {
             .unwrap_or_else(|| Self::fallback_decode(buf))
     }
 
+    /// Decode a transaction according to [EIP-2718], ensuring no trailing bytes.
```
mattsse, line 127: let's a bit more context here, that this is intended to decode a single transaction from the entire buffer without leaving any trailing bytes

mattsse (changes requested):
nice, smol nit

PR #2607, opened 2025-06-22
feat: add log filtering methods to Filter

mattsse:
ports 

https://github.com/paradigmxyz/reth/blob/7e9f1416047f1dc803a366661a561c1e1b9abd2d/crates/rpc/rpc-eth-types/src/logs_utils.rs#L16-L56

as a new util fn on Filter


yash-atreya (approved):
lgtm!

klkvr (approved):

PR #2612, opened 2025-06-23
feat: added FindStorageSlot 

### LINKED ISSUE #2605: [Feature] Add Future type for Finding erc20 storage slot
### Component

contract

we have this helper fn in anvil:

https://github.com/foundry-rs/foundry/blob/edb9cc089065f1638cd9cd6e5511123c02b61a44/crates/anvil/src/eth/api.rs#L1892-L1897

we can convert this into a standalone future type (`struct StorageSlotFinder`) here in alloy, in a mod next to 

https://github.com/alloy-rs/alloy/blob/83e216ce94cdf1d59f630685ef471a377bc8c0e8/crates/contract/src/eth_call.rs#L28-L28

this future needs the same args as the anvil function but also needs provider.

then this function can have another constructor for the balanceOf use case 

### Additional context

_No response_

Soubhik-10:
potentially closes #2605

### DIFF
--- a/crates/contract/src/error.rs
+++ b/crates/contract/src/error.rs
@@ -36,6 +36,9 @@ pub enum Error {
     /// An error occurred while waiting for a pending transaction.
     #[error(transparent)]
     PendingTransactionError(#[from] PendingTransactionError),
+    /// An error occurred while searching for storage slots of an ERC20 token.
+    #[error("Failed to find storage slot for {0} ERC20 token")]
+    StorageSlotNotFound(String),
 }
 
 impl From<alloy_sol_types::Error> for Error {
--- a/crates/contract/src/storage_slot.rs
+++ b/crates/contract/src/storage_slot.rs
@@ -1,0 +1,135 @@
+use alloy_network::TransactionBuilder;
+use alloy_primitives::{Address, Bytes, B256, U256};
+use alloy_provider::Provider;
+use alloy_rpc_types_eth::{
+    state::{AccountOverride, EvmOverrides, StateOverridesBuilder},
+    TransactionRequest,
+};
+use alloy_sol_types::{sol, SolValue};
+
+/// A future type for finding erc20 storage slot.
+#[derive(Debug, Clone)]
+pub struct StorageSlotFinder<P> {
+    provider: P,
+    token_address: Address,
+    calldata: Bytes,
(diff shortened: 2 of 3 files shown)


crates/contract/src/error.rs
```diff
@@ -36,6 +36,9 @@ pub enum Error {
     /// An error occurred while waiting for a pending transaction.
     #[error(transparent)]
     PendingTransactionError(#[from] PendingTransactionError),
+    /// An error occurred while searching for storage slots of an ERC20 token.
+    #[error("Failed to find storage slot for {0} ERC20 token")]
+    StorageSlotNotFound(String),
```
mattsse: I think we should roll an additional error type for this use case

crates/contract/src/storage_slot.rs
```diff
@@ -0,0 +1,134 @@
...
+use alloy_rpc_types_eth::{
+    state::{AccountOverride, EvmOverrides, StateOverridesBuilder},
+    TransactionRequest,
+};
+use alloy_sol_types::{sol, SolValue};
+
+/// A future type for finding erc20 storage slot.
+#[derive(Debug, Clone)]
+pub struct StorageSlotFinder<P> {
+    provider: P,
```
mattsse: we also need to make this generic over N: Network, so that we can use `N::TransactionRequest::default`, only then we can use this in combination wiht the AnyNetwork

mattsse (changes requested):
great start, some nits,

crates/contract/src/storage_slot.rs
```diff
@@ -0,0 +1,165 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum StorageSlotFinderError {
+    /// An error occurred while searching for storage slots of an ERC20 token.
+    #[error("Failed to find storage slot for {0} ERC20 token")]
+    StorageSlotNotFound(String),
+}
+
+impl<P, N> StorageSlotFinder<P, N>
+where
+    P: Provider,
```
mattsse: I think we need Provider<N> here

mattsse (approved):
@Soubhik-10 this previously failed due to a tiny but fatal encoding error

```
let calldata = IERC20::balanceOfCall { target: user }.target.abi_encode().into();
```

notice the .target.

PR #2615, opened 2025-06-23
feat: add connect_reqwest to ProviderBuilder, http_with_client to Cli…

QEDK:
…entBuilder, new_http_with_client to RpcClient

## Motivation
The default `reqwest::Client` implementation is not suitable for usage in high-performance applications due to the lack of connection pooling, low timeout and keepalive durations.

## Solution
Allows any user to pass a custom `reqwest::Client` at the time of `Provider` construction.

## PR Checklist

- [ ] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/builder.rs
+++ b/crates/provider/src/builder.rs
@@ -502,6 +502,18 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
         self.connect_client(client)
     }
 
+    /// Build this provider with a pre-built Reqwest client.
+    #[cfg(any(test, feature = "reqwest"))]
+    pub fn connect_reqwest(self, client: reqwest::Client, url: reqwest::Url) -> F::Provider
+    where
+        L: ProviderLayer<crate::RootProvider<N>, N>,
+        F: TxFiller<N> + ProviderLayer<L::Provider, N>,
+        N: Network,
+    {
+        let client = ClientBuilder::default().http_with_client(client, url);
+        self.connect_client(client)
+    }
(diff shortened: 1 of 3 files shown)


mattsse (changes requested):
I think we're a missing a function that takes a closure with the reqwest::ClientBuilder as well?

then you can do `with_reqwest(|builder| {
  ....
})`

QEDK:
> I think we're a missing a function that takes a closure with the reqwest::ClientBuilder as well?
> 
> then you can do `with_reqwest(|builder| { .... })`

Implemented, please check and lmk!

crates/provider/src/builder.rs
```diff
@@ -502,6 +502,31 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
...
+        N: Network,
+        C: Into<reqwest::Client>,
+    {
+        let client = ClientBuilder::default().http_with_client(client.into(), url);
+        self.connect_client(client)
+    }
+
+    /// Build this provider with a provided Reqwest client builder.
+    #[cfg(any(test, feature = "reqwest"))]
+    pub fn with_reqwest<B>(self, builder: B, url: reqwest::Url) -> F::Provider
```
mattsse: we should swap the order here so that the closure is the last arg

QEDK: Done, resolving.

yash-atreya (approved):
lgtm!

PR #2619, opened 2025-06-25
chore: add serde for `BlobScheduleBlobParams`

kevaundray:
## Motivation

The main motivation is to make `reth::ChainSpec` derive serde, so we can serialize it in the zkVM context. The ChainSpec however includes a `BlobScheduleBlobParams` field

## Solution

Derive serde for `BlobScheduleBlobParams`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7892.rs
+++ b/crates/eips/src/eip7892.rs
@@ -17,6 +17,7 @@ pub enum BlobScheduleEntry {
 
 /// Blob parameters configuration for a chain, including scheduled updates.
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


mattsse:
@kevaundray this seems reasonable

was closing this an accident?

kevaundray:
> @kevaundray this seems reasonable
> 
> was closing this an accident?

Ah it was on purpose, I checked the ChainSpec and there are other fields like `dyn HardFork` which I think might require more thinking on making them serializable, so I closed this until I had a proper plan

mattsse:
I'm also fine with restructing the chainspec traits or impl serde manually

PR #2623, opened 2025-06-26
feat(rpc): Add generic `TxReq` to `Bundle`

RomanHodulak:
## Motivation

The `TransactionRequest` is currently hard-coded to a concrete type. This means that the `Bundle` and all components that depend on it will not work for custom transaction requests.

## Solution

Add generic `TxReq` with `TransactionRequest` as a default.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
I think this also doesn't hurt although the bundle endpoint is more or less just an L1 thing

but no downside in doing this @klkvr ?

PR #2624, opened 2025-06-26
feat(network): Add method to remove nonce from transaction using `TransactionBuilder`

RomanHodulak:
## Motivation

To remove reliance on concrete type `TransactionRequest` and use only the abstraction `TransactionBuilder` in its place, currently there is no way to remove the nonce.

This feature is required for RPC methods like `eth_call`. Lack of nonce makes the EVM set it.

## Solution

Add methods to `TransactionBuilder` that take the nonce out of the transaction.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network/src/any/builder.rs
+++ b/crates/network/src/any/builder.rs
@@ -24,6 +24,10 @@ impl TransactionBuilder<AnyNetwork> for WithOtherFields<TransactionRequest> {
         self.deref_mut().set_nonce(nonce)
     }
 
+    fn take_nonce(&mut self) -> Option<u64> {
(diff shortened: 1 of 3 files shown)


RomanHodulak:
Yeah I'd say let's do op-alloy right away.

popzxc:
> this is technically breaking 

Just got hit by that (looks like [not only I](https://github.com/scroll-tech/reth/pull/268)) 😅 
Given that `alloy` is now past 1.0, it would be really nice to rely on semver compatibility.

PR #2626, opened 2025-06-27
feat: add recover_signer_unchecked_with_buf to `SignerRecoverable` trait

lean-apple:
## Motivation

Closes https://github.com/paradigmxyz/reth/issues/17068.

## Solution

- Add `recover_signer_unchecked_with_buf` to `SignerRecoverable` 
- Add `recover_signer_unchecked_with_buf` impls for `Signed<T>` and `EthereumTxEnvelope<Eip4844>`

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -5,7 +5,7 @@ use crate::{
 use alloy_eips::{
     eip2718::Eip2718Result, eip2930::AccessList, eip7702::SignedAuthorization, Typed2718,
 };
-use alloy_primitives::{Bytes, Sealed, Signature, TxKind, B256, U256};
+use alloy_primitives::{keccak256, Bytes, Sealed, Signature, TxKind, B256, U256};
 use alloy_rlp::BufMut;
 use core::hash::{Hash, Hasher};
 #[cfg(not(feature = "std"))]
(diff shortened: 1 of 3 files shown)


crates/consensus/src/signed.rs
```diff
@@ -465,6 +465,16 @@ where
         let signature_hash = self.signature_hash();
         crate::crypto::secp256k1::recover_signer_unchecked(self.signature(), signature_hash)
     }
+
+    fn recover_signer_unchecked_with_buf(
+        &self,
+        buf: &mut Vec<u8>,
+    ) -> Result<alloy_primitives::Address, crate::crypto::RecoveryError> {
+        buf.clear();
```
mattsse, line 473: this makes sense

mattsse (approved):
lgtm

@RomanHodulak for vis, this is also something we need to add to op-alloy

crates/consensus/src/transaction/recovered.rs
```diff
@@ -248,6 +248,14 @@ pub trait SignerRecoverable {
...
     fn recover_signer_unchecked(&self) -> Result<Address, RecoveryError>;
 
+    /// Same as [`SignerRecoverable::recover_signer_unchecked`] but receives a buffer to operate on
+    /// for encoding. This is useful during batch recovery of historical transactions to avoid
+    /// allocating a new buffer for each transaction.
+    fn recover_signer_unchecked_with_buf(
+        &self,
+        buf: &mut alloc::vec::Vec<u8>,
+    ) -> Result<Address, RecoveryError>;
+
```
klkvr: should we have a default implementation here simply calling `self.recover_signer()`?

mattsse: yeah good call

PR #2634, opened 2025-06-29
feat: Add dynamic crypto backend for ecrecover

kevaundray:
## Motivation

One motivation for this is to allow downstream users to plug in their own implementation of certain functions. The downstream user we have in mind are zkVM guest program implementers where they could add precompiles using this strategy.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -443,6 +443,9 @@ mod tests {
 
         /// Test that when crypto-backend feature is enabled but no provider is installed,
         /// it falls back to the default implementation
+        /// 
+        /// Note: this test is unreliable because if the CryptoProvider is set, it will use the MockProvider
+        /// which doesn't have ideal behaviour
         #[test]
         #[cfg(any(feature = "secp256k1", feature = "k256"))]
         fn test_fallback_to_default_when_no_provider() {
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -6,6 +6,9 @@ use alloy_primitives::U256;
 #[cfg(any(feature = "secp256k1", feature = "k256"))]
 use alloy_primitives::Signature;
 
+#[cfg(feature = "crypto-backend")]
+pub use backend::{install_default_provider, CryptoProvider, ProviderAlreadySetError};
+
 /// Opaque error type for sender recovery.
 #[derive(Debug, Default, thiserror::Error)]
 #[error("Failed to recover the signer")]
@@ -400,10 +403,7 @@ mod tests {
 
     #[cfg(feature = "crypto-backend")]
     mod backend_tests {
-        use crate::crypto::{
-            backend::{try_get_provider, CryptoProvider},
-            RecoveryError,
-        };
+        use crate::crypto::{backend::CryptoProvider, RecoveryError};
         use alloc::sync::Arc;
         use alloy_primitives::{Address, Signature, B256};
 
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -98,8 +98,7 @@ pub mod backend {
         }
     }
 
-    #[cfg(feature = "std")]
-    impl std::error::Error for CryptoProviderAlreadySetError {}
+    impl core::error::Error for CryptoProviderAlreadySetError {}
 
     /// Install the default crypto provider.
     ///
@@ -111,14 +110,14 @@ pub mod backend {
     ) -> Result<(), CryptoProviderAlreadySetError> {
         #[cfg(feature = "std")]
         {
-            DEFAULT_PROVIDER.set(provider.clone()).map_err(|_| {
+            DEFAULT_PROVIDER.set(provider).map_err(|provider| {
                 // Return the provider we tried to install in the error
                 CryptoProviderAlreadySetError { provider }
             })
         }
         #[cfg(not(feature = "std"))]
         {
-            DEFAULT_PROVIDER.set(Box::new(provider.clone())).map_err(|_| {
+            DEFAULT_PROVIDER.set(Box::new(provider)).map_err(|provider| {
                 // Return the provider we tried to install in the error
                 CryptoProviderAlreadySetError { provider }
             })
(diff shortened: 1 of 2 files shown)


crates/consensus/src/crypto.rs
```diff
@@ -286,4 +388,92 @@ mod tests {
...
+                assert!(Arc::ptr_eq(&err.provider, &provider2));
+            }
+        }
+
+
+        /// Test that when crypto-backend feature is enabled but no provider is installed,
+        /// it falls back to the default implementation
+        /// 
+        /// Note: this test is unreliable because if the CryptoProvider is set, it will use the MockProvider
+        /// which doesn't have ideal behaviour
```
kevaundray: Noting this note. We should probably remove this since we cannot guarantee that another test will not set the CryptoProvider

kevaundray: removed in d106417b1e441c97faf1ff675c6ad258387c645a

crates/consensus/src/crypto.rs
```diff
@@ -73,6 +175,13 @@ pub mod secp256k1 {
         sig[32..64].copy_from_slice(&signature.s().to_be_bytes::<32>());
         sig[64] = signature.v() as u8;
 
+        // Try dynamic backend first when crypto-backend feature is enabled
+        #[cfg(feature = "crypto-backend")]
+        if let Some(provider) = super::backend::try_get_provider() {
+            return provider.recover_signer_unchecked(&sig, &hash.0);
+        }
```
kevaundray, line 230: If crypto-baackend feature-flag is set, then this will call whatever backend is set here.

To install your own backend, you can call `install_default_provider` in your own codebase. `install_default_provider` can only be called once, calling it more than once will return an error

crates/consensus/src/crypto.rs
```diff
@@ -44,6 +47,105 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
...
+    }
+
+    impl core::fmt::Display for ProviderAlreadySetError {
+        fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            write!(f, "crypto provider already installed")
+        }
+    }
+
+    #[cfg(feature = "std")]
+    impl std::error::Error for ProviderAlreadySetError {}
```
mattsse: this can be regular core::error then we dont need the feature gate

crates/consensus/src/crypto.rs
```diff
@@ -44,6 +47,105 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
...
+    ///
+    /// This sets the global default provider used by the high-level crypto functions.
+    /// Returns an error containing the provider that was attempted to be installed if one is
+    /// already set.
+    pub fn install_default_provider(
+        provider: Arc<dyn CryptoProvider>,
+    ) -> Result<(), ProviderAlreadySetError> {
+        #[cfg(feature = "std")]
+        {
+            DEFAULT_PROVIDER.set(provider.clone()).map_err(|_| {
```
mattsse: I think this clone is a bit redundant here because the arg is already Arc

or we just accept CryptoProvider and Arc it when we install it

crates/consensus/src/crypto.rs
mattsse, line 230: this seems okay

crates/consensus/src/crypto.rs
```diff
@@ -44,6 +47,105 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
...
+    /// Global default crypto provider.
+    #[cfg(feature = "std")]
+    static DEFAULT_PROVIDER: OnceLock<Arc<dyn CryptoProvider>> = OnceLock::new();
+
+    #[cfg(not(feature = "std"))]
+    static DEFAULT_PROVIDER: OnceBox<Arc<dyn CryptoProvider>> = OnceBox::new();
+
+    /// Error returned when attempting to install a provider when one is already installed.
+    /// Contains the provider that was attempted to be installed.
+    pub struct ProviderAlreadySetError {
```
mattsse: I'd like to rename this to CryptoProviderAlreadySetError

because provider is already quite overloaded within alloy (rpc provider)

mattsse (changes requested):
only have nits, otherwise this lgtm

but we can add some additional docs for the backend type

crates/consensus/src/crypto.rs
kevaundray: Ah will change!

crates/consensus/src/crypto.rs
kevaundray: A different strategy would be to make it such that the crypto-backend was exactly like the other two backends. It would need to have a `recover_signer_unchecked` function instead of a method.

The downside of this approach is that the fallback would need to be defined in the crypto-backend module or we could just return an error, if no backend has been installed

crates/consensus/src/crypto.rs
kevaundray: The [set](https://doc.rust-lang.org/std/sync/struct.OnceLock.html#method.set) method returns `T` so I use that for the map_err closure instead

crates/consensus/src/crypto.rs
```diff
@@ -44,6 +47,104 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
...
+            DEFAULT_PROVIDER.set(provider).map_err(|provider| {
+                // Return the provider we tried to install in the error
+                CryptoProviderAlreadySetError { provider }
+            })
+        }
+        #[cfg(not(feature = "std"))]
+        {
+            DEFAULT_PROVIDER.set(Box::new(provider)).map_err(|provider| {
+                // Return the provider we tried to install in the error
+                CryptoProviderAlreadySetError { provider: *provider }
```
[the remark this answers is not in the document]
kevaundray: This doesn't look correct, when we dereference the Box with *provider, it should move the value out of the Box, automatically deallocating the Box itself.

This is not true if the type is copyable (its not in this case) but if it was, then when the closure is dropped, the Box should also be dropped

mattsse (approved):
this lgtm

mind taking a look at this as well @klkvr @DaniPopes

crates/consensus/src/crypto.rs
mattsse: > if no backend has been installed

this I'd like to avoid because this will be very unpleasant for users that just want to do some simple rpc stuff

klkvr (approved):

PR #2635, opened 2025-06-29
make tx build fns public

zemse:
## Motivation

Currently when I have `TransactionRequest` and I want to sign it using a LocalSigner:

```rs
 let mut tx = tx_req
    .build_typed_tx()
    .unwrap() // TypedTransaction
    .eip1559()
    .unwrap() // &Tx1559
    .clone();

let signature = wallet.sign_transaction_sync(&mut tx)?;
``` 

## Solution

Could be simplified as

```rs
let mut tx = tx_req.build_1559(); // Tx1559
let signature = wallet.sign_transaction_sync(&mut tx)?;
```

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/request.rs
+++ b/crates/rpc-types-eth/src/transaction/request.rs
@@ -380,7 +380,7 @@ impl TransactionRequest {
     ///
     /// Returns an error if required fields are missing.
     /// Use `complete_legacy` to check if the request can be built.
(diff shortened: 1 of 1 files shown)


onbjerg (approved):
makes sense

mattsse:
hmm, preferred would be #2519

because returning a vec as error is a bit horrible and changing that later would be breaking.

PR #2636, opened 2025-06-29
feat: add dedicated error for SECP256K1N_HALF error

### LINKED ISSUE #2633: [Feature] Add dedicated error for SECP256K1N_HALF error
### Component

signers

https://github.com/alloy-rs/alloy/blob/3f38234b9954dfd2eeb61a336c142d0a8940712c/crates/consensus/src/crypto.rs#L86-L89

currently returns an empty error,

we should introduce a dedicated error for this and use for recoveryerror:

https://github.com/alloy-rs/alloy/blob/3f38234b9954dfd2eeb61a336c142d0a8940712c/crates/consensus/src/crypto.rs#L27-L27

## TODO
* introduce error type that can convey what the error is and use that in this case

### Additional context

_No response_

stevencartavia:
## Motivation
closes #2633

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -6,6 +6,11 @@ use alloy_primitives::U256;
 #[cfg(any(feature = "secp256k1", feature = "k256"))]
 use alloy_primitives::Signature;
 
+/// Error for signature S.
+#[derive(Debug, thiserror::Error)]
+#[error("signature S value is greater than `secp256k1n / 2`")]
+pub struct InvalidSignatureS;
+
 /// Opaque error type for sender recovery.
 #[derive(Debug, Default, thiserror::Error)]
 #[error("Failed to recover the signer")]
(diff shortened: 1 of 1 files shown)


crates/consensus/src/crypto.rs
```diff
@@ -85,7 +90,7 @@ pub mod secp256k1 {
     /// If the S value is too large, then this will return a `RecoveryError`
     pub fn recover_signer(signature: &Signature, hash: B256) -> Result<Address, RecoveryError> {
         if signature.s() > SECP256K1N_HALF {
-            return Err(RecoveryError::new());
+            return Err(RecoveryError::from_source(super::InvalidSignatureS));
```
onbjerg: theres a `use super::*` so we don't need to qualify it here

mattsse (approved):

PR #2638, opened 2025-06-30
feat: add BuilderBlockValidationV5 for relay for Fusaka

### LINKED ISSUE #2637: [Feature] Add BuilderBlockValidationV5 types for relay block validation for Fusaka.
### Component

rpc

Fusaka introduces `BlobsBundleV2` containing cell proofs. We need to create a new `BuilderBlockValidationV5` API to support `BlobsBundleV2` . This type will be used in adding a new `validate_builder_submission` API in reth.

### Additional context

_No response_

varun-doshi:
## Motivation

Fixes #2637 

## Solution

Add type `SignedBidSubmissionV5` for Fulu

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -951,12 +951,14 @@ impl<'de> serde::Deserialize<'de> for BlobsBundleV2 {
         }
         let raw = BlobsBundleRaw::deserialize(deserializer)?;
 
-        if raw.proofs.len() == raw.commitments.len() + CELLS_PER_EXT_BLOB {
+        if raw.proofs.len() == raw.blobs.len() * CELLS_PER_EXT_BLOB
+            && raw.commitments.len() == raw.blobs.len()
+        {
             Ok(Self { commitments: raw.commitments, proofs: raw.proofs, blobs: raw.blobs })
         } else {
             Err(serde::de::Error::invalid_length(
                 raw.proofs.len(),
-                &format!("{}", raw.commitments.len() + CELLS_PER_EXT_BLOB).as_str(),
+                &format!("{}", raw.commitments.len() * CELLS_PER_EXT_BLOB).as_str(),
             ))
         }
     }
@@ -1885,9 +1887,9 @@ mod tests {
         let commitments = vec![Bytes48::default()];
 
         let blobs_bundle_v2 = BlobsBundleV2 {
-            proofs: vec![Bytes48::default(); commitments.len() + CELLS_PER_EXT_BLOB],
+            proofs: vec![Bytes48::default(); commitments.len() * CELLS_PER_EXT_BLOB],
             commitments,
-            blobs: vec![],
+            blobs: vec![Blob::default()],
         };
 
         let serialized = serde_json::to_string(&blobs_bundle_v2).unwrap();


crates/rpc-types-beacon/src/relay.rs
```diff
@@ -181,6 +203,8 @@ pub struct SignedBidSubmissionV4 {
 #[cfg_attr(feature = "ssz", derive(ssz_derive::Decode, ssz_derive::Encode))]
 #[cfg_attr(feature = "ssz", ssz(enum_behaviour = "transparent"))]
 pub enum SubmitBlockRequest {
+    /// Fulu [`SignedBidSubmissionV5`].
+    Fulu(SignedBidSubmissionV5),
```
mattsse, line 207: this will be a bit problematic because v4 and v5 are now have the same format.

so I'm not sure how to handle this yet.

perhaps we exclude this from this enum for now, because deserializing electra v4 would endup as fulu variant here for now.

and technically we don't strictly need this type @bharath-123 because in both ssz and json it has the same representation (blobsbundlev1 and v2 are the same there)

rkrasiuk: needs a custom deser by looking at `proofs` vs `commitments` lengths. bundle v2 should have `CELLS_PER_EXT_BLOB` (128) more proof entries than commitments whereas in blobs bundle v1 `proofs` and `commitments` length must be the same

mattsse: this means we should add a custom deserde to BlobsBundleV2?

rkrasiuk: imo we need to add the length checks on both v1 and v2 deser, but that will not automatically solve our issues here, so we'd need to add a custom deser for this type too

rkrasiuk: imo we need to add the length checks on both v1 and v2 deser, but that will not automatically solve our issues here, so we'd need to add a custom deser for this type too

bharath-123: ya but I believe as @rkrasiuk pointed `BlobsBundleV2` will have more proofs and I am not sure if there will be any future feature additions to it. Definitely sounds right to implement deserde for this too.

crates/rpc-types-engine/src/payload.rs
```diff
@@ -837,6 +838,30 @@ pub struct BlobsBundleV1 {
...
+    {
+        #[derive(serde::Deserialize)]
+        struct BlobsBundleRaw {
+            commitments: Vec<alloy_consensus::Bytes48>,
+            proofs: Vec<alloy_consensus::Bytes48>,
+            blobs: Vec<alloy_consensus::Blob>,
+        }
+        let raw = BlobsBundleRaw::deserialize(deserializer)?;
+
+        if raw.proofs.len() == raw.commitments.len() {
```
mattsse: we also need one more check for proofs.len == blobs.len

crates/rpc-types-engine/src/payload.rs
```diff
@@ -912,6 +937,30 @@ pub struct BlobsBundleV2 {
...
+    {
+        #[derive(serde::Deserialize)]
+        struct BlobsBundleRaw {
+            commitments: Vec<alloy_consensus::Bytes48>,
+            proofs: Vec<alloy_consensus::Bytes48>,
+            blobs: Vec<alloy_consensus::Blob>,
+        }
+        let raw = BlobsBundleRaw::deserialize(deserializer)?;
+
+        if raw.proofs.len() == raw.commitments.len() + CELLS_PER_EXT_BLOB {
```
mattsse: are we missing another length check here @rkrasiuk ?

varun-doshi: Will wait for confirmation

mattsse (approved):
lgtm, pending @rkrasiuk

crates/rpc-types-engine/src/payload.rs
```diff
@@ -912,6 +937,31 @@ pub struct BlobsBundleV2 {
...
```
mattsse: ah this isn't quite accurate, we need to check that `proofs.len() = blobs.len() * CELLS_PER_EXT_BLOB`

and also ensure that the commitments.len == blobs.len()

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1787,6 +1837,80 @@ mod tests {
...
+    }
+
+    #[test]
+    #[cfg(feature = "serde")]
+    #[cfg(not(debug_assertions))]
+    fn serde_blobsbundlev2_not_empty_pass() {
+        let commitments = vec![Bytes48::default()];
+
+        let blobs_bundle_v2 = BlobsBundleV2 {
+            proofs: vec![Bytes48::default(); commitments.len() + CELLS_PER_EXT_BLOB],
```
bharath-123:
```suggestion
            proofs: vec![Bytes48::default(); commitments.len() * CELLS_PER_EXT_BLOB],
```

crates/rpc-types-engine/src/payload.rs
varun-doshi: Ahh yes good catch

rkrasiuk (approved):

mattsse (approved):

PR #2642, opened 2025-07-01
feat: add provider-mev-api top level feature re-export

mteam88:
## Motivation
Using the mev-api feature today requires using alloy-provider instead of the alloy crate directly like this
alloy-provider = { version = "1.0.16", features = ["mev-api"] }

because the feature isn't re-exported into the alloy crate. For other features there is things like provider-anvil-api that are exposed apparently.

## Solution
Add the reexport so `provider-mev-api` feature can be used instead

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -189,6 +189,11 @@ provider-anvil-node = [
     "alloy-provider?/anvil-node",
     "node-bindings",
 ]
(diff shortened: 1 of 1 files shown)


mteam88:
https://github.com/alloy-rs/alloy/pull/2556#issuecomment-3025229916

mattsse (approved):

PR #2645, opened 2025-07-02
feat: added helper AnvilApi future type for oneshot impersonations

### LINKED ISSUE #2641: [Feature] Add helper AnvilApi future type for oneshot impersonations
### Component

rpc

anvil API has send_impersonate_transaction which is a helper that bypasses any tx signing and sends the tx as is.

ideally we have a way to also impersonate and undo imperosnate at the same time.

for this we can create a new future type that takes the tx request, if the tx request has a from field,
we impersonate first, then send the request, then undo impersonate

## TODO
* add this helper function that does the enable/disable and is ideally generic over a future that is executed in between

### Additional context

e.g.

```
/// Executes a transaction request as an impersonated account
pub async fn send_impersonated_tx<P: Provider + AnvilApi<Ethereum>>(
    provider: &P,
    tx_request: TransactionRequest,
    fund_amount: Option<U256>,
) -> Result<()> {
    // Extract from address
    let from = tx_request
        .from
        .ok_or_else(|| eyre::eyre!("Transaction request must have 'from' field set"))?;


    let impersonate_future = provider.anvil_impersonate_account(from);

    try_join!(fund_future, impersonate_future)?;

    // By using anvil_send_impersonated_transaction, we can use WalletProviders
    let tx_hash = provider.anvil_send_impersonated_transaction(tx_request).await?;

    // Get receipt and stop impersonation
    let receipt_future = provider.get_transaction_receipt(tx_hash);
    let stop_impersonate_future = provider.anvil_stop_impersonating_account(from);

    try_join!(receipt_future, stop_impersonate_future)?;

    Ok(())
}
```

maybe @Soubhik-10 ?

Soubhik-10:
closes #2641


crates/contract/src/impersonate.rs
```diff
@@ -0,0 +1,146 @@
+use alloy_network::Network;
+use alloy_primitives::U256;
+use alloy_provider::{ext::AnvilApi, Provider};
+use alloy_rpc_types_eth::TransactionRequest;
+use alloy_transport::TransportError;
+use futures::try_join;
```
mattsse: hmm, looks like we can actually move this to where AnvilApi is defined, this doesnt necessarily need to added in the contract crate?

crates/contract/src/impersonate.rs
```diff
@@ -0,0 +1,146 @@
...
+        } else {
+            impersonate_future.await?;
+        }
+
+        // Send transaction
+        let tx_hash = provider.anvil_send_impersonated_transaction(tx_request.clone()).await?;
+
+        // Wait for receipt and stop impersonation
+        let receipt_future = provider.get_transaction_receipt(tx_hash);
+        let stop_impersonate_future = provider.anvil_stop_impersonating_account(from);
```
mattsse: let's make this optional by adding a bool as field

crates/contract/src/impersonate.rs
```diff
@@ -0,0 +1,146 @@
...
+///
+/// let call = ImpersonatedCall::new(provider, tx, Some(U256::from(1e18 as u64)));
+/// let receipt = call.send_impersonated_tx().await?;
+/// println!("Impersonated tx succeeded in block: {:?}", receipt.block_number);
+/// # Ok(())
+/// # }
+/// ```
+
+#[derive(Debug, Clone)]
+pub struct ImpersonatedCall<P, N>
```
mattsse: actually, maybe we dont need a dedicated call for this and instead can add this as a utility trait function to AnvilAp, that takes the request and a ```struct ImpersonateConfig { fund, stop_impersonate}```

mattsse (changes requested):
cool, functionally this looks great already

crates/provider/src/ext/anvil.rs
```diff
@@ -398,6 +409,46 @@ where
...
+            self.anvil_stop_impersonating_account(from).await?;
+        }
+
+        Ok(pending)
+    }
+}
+
+/// Configuration for impersonated transactions, including optional funding and whether to stop
+/// impersonation.
+#[derive(Debug, Clone)]
```
mattsse, line 445: let's impl default default for this, with None and true, and add a few nice builder style functions to configure these, then this can be used like

`ImpersonateConfig::default().keep_impersonate().fund(U256::MAX)`

crates/provider/src/ext/anvil.rs
```diff
@@ -192,6 +202,7 @@ impl<N, P> AnvilApi<N> for P
 where
     N: Network,
     P: Provider<N>,
+    N: Network<TransactionRequest = TransactionRequest>,
```
mattsse: do we need this?

I think we can use:

https://github.com/alloy-rs/alloy/blob/c808022d571aaffb97a3429a0fdba6fc06e8f7aa/crates/network/src/transaction/builder.rs#L126-L127

mattsse (approved):
very nice

PR #2674, opened 2025-07-08
refactor(rpc): Add handwritten bounds on generic `TxReq` using `serde` attributes

RomanHodulak:
## Motivation

`Deserialize` derive macro adds `Default` bound on `TxReq` when `serde(default)` attributes are used.

## Solution

Add handwritten bounds on generic `TxReq` using `serde` attributes.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/simulate.rs
+++ b/crates/rpc-types-eth/src/simulate.rs
@@ -12,7 +12,16 @@ pub const MAX_SIMULATE_BLOCKS: u64 = 256;
 /// executed.
 #[derive(Clone, Debug)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+#[cfg_attr(
+    feature = "serde",
+    serde(
+        rename_all = "camelCase",
+        bound(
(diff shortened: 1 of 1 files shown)


mattsse (commented):
should we just implement default for these types?

RomanHodulak:
That is what I thought and done in https://github.com/alloy-rs/alloy/pull/2662

But upon inspecting the `serde::Deserialize` derive macro and code it generates I found out that it still adds a bound on `TxReq` by itself if you use these `serde(default)` attributes.

https://github.com/serde-rs/serde/blob/babafa54d283fb087fa94f50a2cf82fa9e582a7c/serde_derive/src/de.rs#L217

crates/rpc-types-eth/src/simulate.rs
```diff
@@ -15,13 +15,19 @@ pub const MAX_SIMULATE_BLOCKS: u64 = 256;
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
```
mattsse: I think this is equivalent to `serde(bound = "")` on struct level

klkvr (approved):

PR #2676, opened 2025-07-08
Add method to fetch ENS Reverse registrar addr

abhijeetbhagat:
## Motivation

There should be a way to get the ENS Reverse Registrar address using ENS Registry.

## Solution

Solution is to add a method on the ProviderEnsExt trait in addition to the `get_resolver()` method.

## PR Checklist

- [X] Added Tests
- [X] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/ens/src/lib.rs
+++ b/crates/ens/src/lib.rs
@@ -94,7 +94,7 @@ mod contract {
             function resolver(bytes32 node) view returns (address);
 
             /// returns the owner of this node
-            function owner(bytes32 node) external returns (address);
+            function owner(bytes32 node) view returns (address);
         }
 
         /// ENS Resolver interface.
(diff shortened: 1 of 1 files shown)


crates/ens/src/lib.rs
```diff
@@ -143,6 +156,10 @@ mod provider {
             error_name: &str,
         ) -> Result<EnsResolverInstance<&P, N>, EnsError>;
 
+        /// Returns the reverse registrar for the specified node. The `&str` is only used for error
+        /// messages.
```
mattsse: this doesnt accept a &str, so this suggestions makes sense

mattsse (changes requested):
reasonable,

smol nit

mattsse (approved):

PR #2679, opened 2025-07-09
feat(rpc-types-engine): custom ssz decode to distinguish Fulu and Electra payload

cakevm:
For SSZ when using `SubmitBlockRequest::from_ssz_bytes(...)`, the enum default to `Fulu` without the custom deserialization for an `Electra` payload. Same as for the serde it compares the amount of proofs/commits to determine the version.

https://github.com/alloy-rs/alloy/blob/fe4bfe2be3d78345d32c8bbd542123e2c099a146/crates/rpc-types-beacon/src/relay.rs#L205-L209

Test with `#[cfg(not(debug_assertions))]` are not executed in the CI, locally they are fine. Same concept as the serde tests. Run:
`cargo test -p alloy-rpc-types-engine --features ssz ssz_blobsbundlev2 --release`

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -1010,15 +1010,17 @@ impl ssz::Decode for BlobsBundleV2 {
         let proofs_offset = u32::from_le_bytes([bytes[4], bytes[5], bytes[6], bytes[7]]) as usize;
         let blobs_offset = u32::from_le_bytes([bytes[8], bytes[9], bytes[10], bytes[11]]) as usize;
 
-        // Ensures no overlapping fields and no out-of-bounds access. Return error if any of these
-        // invalid conditions are true:
+        // Validates offsets to ensure no overlapping fields and no out-of-bounds access.
+        // Note: We use < instead of <= to allow empty fields where consecutive offsets are equal.
+        // For example, an empty bundle will have all offsets = 12 (end of header).
+        // Return error if any of these invalid conditions are true:
         // 1. commitments_offset < 12: would overlap with header
-        // 2. proofs_offset <= commitments_offset: proofs would overlap or come before commitments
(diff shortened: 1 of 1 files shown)


crates/rpc-types-engine/src/payload.rs
```diff
@@ -992,6 +992,68 @@ impl<'de> serde::Deserialize<'de> for BlobsBundleV2 {
...
+impl ssz::Decode for BlobsBundleV2 {
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, ssz::DecodeError> {
+        // Need at least 12 bytes for the 3 offset values
+        if bytes.len() < 12 {
+            return Err(ssz::DecodeError::InvalidByteLength { len: bytes.len(), expected: 12 });
+        }
```
mattsse: I think instead of doing this, we can do the same thing we do for the deserialize impl and use a private helper type that we then check after deserde

this is more expensive when used in untagged, but tbh I dont really want to maintain these ssz checks here and this should no longer be that problematic once fulu activates

mattsse (changes requested):
@rkrasiuk for vis

crates/rpc-types-engine/src/payload.rs
cakevm: You are right, I overshot here. It is now much more readable and maintainable.

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1992,6 +2027,73 @@ mod tests {
         assert!(deserialized.is_err());
     }
 
+    #[test]
+    #[cfg(feature = "ssz")]
+    #[cfg(not(debug_assertions))]
```
mattsse, line 2032: I really want to move to heap alloc blobs -.-

mattsse (approved):
lgtm

PR #2685, opened 2025-07-10
chore: refine optional rpc-types features

koushiro:
## Motivation

Re-export some types from `alloy-rpc-types-any` crate in `alloy` crate level

## Solution

- add optional `rpc-types-any` feature in `alloy` crate level
- update `rpc-types-*` features to only depend on `rpc-types` and `alloy-rpc-types?/*` features

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -165,12 +165,18 @@ provider-anvil-api = [
 provider-debug-api = [
     "providers",
     "alloy-provider?/debug-api",
+    "rpc-types-debug",
     "rpc-types-trace",
 ]
 provider-engine-api = [
     "providers",
     "alloy-provider?/engine-api",
     "rpc-types-engine",
+]
+provider-mev-api = [
+    "providers",
+    "alloy-provider?/mev-api",
+    "rpc-types-mev",
 ]
 provider-net-api = ["providers", "alloy-provider?/net-api"]
 provider-trace-api = [
@@ -189,11 +195,6 @@ provider-anvil-node = [
     "alloy-provider?/anvil-node",
     "node-bindings",
 ]
-provider-mev-api = [
-    "providers",
-    "alloy-provider?/mev-api",
-    "rpc-types-mev",
-]
 
 # pubsub
 pubsub = [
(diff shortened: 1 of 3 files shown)


crates/alloy/Cargo.toml
```diff
@@ -210,36 +211,17 @@ rpc-client = ["rpc", "transports", "transport-http", "dep:alloy-rpc-client"]
...
-rpc-types-anvil = [
-    "rpc-types",
-    "alloy-rpc-types?/anvil",
-    "alloy-provider?/anvil-api",
-]
+rpc-types-admin = ["rpc-types", "alloy-rpc-types?/admin"]
+rpc-types-anvil = ["rpc-types", "alloy-rpc-types?/anvil"]
+rpc-types-any = ["rpc-types", "alloy-rpc-types?/any"]
 rpc-types-beacon = ["rpc-types", "alloy-rpc-types?/beacon"]
-rpc-types-debug = ["rpc-types", "alloy-provider?/debug-api"]
```
mattsse, line 224: why are the provider features removed here?

I believe this would be breaking, can we limit this pr to just the any feature?

koushiro: Why does the `rpc-types-*` features depend on the `alloy-provider?/*-api`. It's so weird.
Intuitively, the APIs should depend on the types, not the types on the APIs.

mattsse: they don't depend on it, but they enable the api if the provider is also enabled, what's the issue with that?

koushiro: Why do you need `rpc-types-*` features to enable the api? If you want to enable the api, you should enable `provider-*-api` features instead of `rpc-types-*` features, it should only be related to types.

mattsse: it would be breaking now,

and these only enable these extra ext trait support if alloy-provider is already included, I dont think this is an issue

koushiro: I reverted some changes, but it still doesn't make sense to me, 
because currently `rpc-types-mev` feature doesn't enable the ext trait of provider like other `rpc-types-*` features did

mattsse (approved):

mattsse:
unable to make edits here so I re-submitted this with fixed features

Issue #2667, opened 2025-07-07
[Feature] provider extension for tenderly RPC methods


siosw:
### Component

provider, pubsub

### Describe the feature you would like

tenderly RPCs expose a number of custom methods, it would be nice to have a provider extension to support them.

from what i can tell this would require
- adding a `TenderlyApi` trait to the provider crate
- adding a `rpc-types-tenderly` crate containing the tenderly rpc types

### Additional context

i have most of this implemented already and would love to upstream if you are interested

mattsse:
@siosw happy to accept this, yes!

mattsse:
I think we can close this one now

PR #2699, opened 2025-07-12
feat: tenderly provider ext

siosw:
## Motivation
add tenderly provider extension #2667 

## Solution
for now this only adds the support for the simulation but wanted to get this up for discussion

I'm a little unsure about the `TenderlyLogInput` deserde, is there a smarter way of implementing this that I'm missing?

I set the provider tests to be ignored since they require access to a tenderly endpoint

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/provider/src/ext/tenderly.rs
```diff
@@ -0,0 +1,203 @@
...
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
+pub trait TenderlyApi<N: Network>: Send + Sync {
+    /// Simulates a transaction as it would execute on the given block, allowing overrides of state
+    /// variables and balances of all accounts
+    async fn tenderly_simulate_transaction(
+        &self,
+        tx: N::TransactionRequest,
+        block: BlockNumberOrTag,
+        state_overrides: Option<StateOverride>,
+        block_overrides: Option<BlockOverrides>,
```
mattsse, line 36: to make this a bit nice I suggest we do something like

https://github.com/alloy-rs/alloy/blob/d9ec782cc1e3a7850f44f72884aea8f6b9b11ca8/crates/provider/src/provider/eth_call/mod.rs#L147-L158

here, I think we could even reuse this type here?

crates/provider/src/ext/tenderly.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    /// Simulates a transaction as it would execute on the given block, allowing overrides of state
+    /// variables and balances of all accounts
+    async fn tenderly_simulate_bundle(
+        &self,
+        txs: &[N::TransactionRequest],
+        block: BlockNumberOrTag,
+        state_overrides: Option<StateOverride>,
+        block_overrides: Option<BlockOverrides>,
+    ) -> TransportResult<Vec<TenderlySimulationResult>>;
```
mattsse, line 47: for this we could mimic what we have for ethcallmany:

https://github.com/alloy-rs/alloy/blob/d9ec782cc1e3a7850f44f72884aea8f6b9b11ca8/crates/provider/src/provider/eth_call/call_many.rs#L16-L26

mattsse (changes requested):
supportive,

suggesting to introduce future types for the simulate calls, like we use for ethcall and ethcallmany, this gives a nicer api with only (txrequest) as argument and then setter functions for the call future

siosw:
> supportive,
> 
> suggesting to introduce future types for the simulate calls, like we use for ethcall and ethcallmany, this gives a nicer api with only (txrequest) as argument and then setter functions for the call future

thanks for taking a look. agree it would be nicer to have the same api as eth call & reuse the type. 
i stumbled onto some unexpected behaviour of `EthCall`, basically 
```
let call: EthCall<Ethereum, String> = EthCall::new(provider.weak_client(), "tenderly_simulateTransaction", tx.clone()).block(block.into());
let res = call.await;
```
ends up calling `eth_estimateGas` 
i think this happens whenever the method string is not equal to `eth_call`
is there something I'm missing here?

siosw:
ah ok i guess it's this: https://github.com/alloy-rs/alloy/blob/1f157f03af5d9472044d6972ad84056f89db7d4c/crates/provider/src/provider/eth_call/mod.rs#L101

do you think we should make ethcall support more rpc calls? or create a similar type specifically for the tenderly apit?
in any case the current ethcall impl seems confusing to me because it makes it seem like it acceps arbitrary rpc calls.

siosw:
@mattsse i got it working with `EthCall` & `EthCallMany` but it requires some changes to the types.
there are two issues with this approach that i've been running into:
- tenderly requires an explicit block number for simulation which doesn't match the `EthCall` behavior
- when submitting more than one bundle to `EthCallMany` all bundles other than the first one are silently dropped

maybe this means we will have to implement tenderly specific future types for this? wdyt?

crates/provider/src/layers/batch.rs
mattsse, line 1: we cant do these changes, because this would be breaking

crates/rpc-types-tenderly/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+/// Tenderly RPC simulation result.
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TenderlySimulationResult {
+    /// The final status of the transaction, typically indicating success or failure.
+    pub status: bool,
+    /// The amount of gas used by the transaction.
+    pub gas_used: U64,
+    /// The total amount of gas used when this transaction was executed in the block.
+    pub cumulative_gas_used: U64,
```
mattsse: these should use primitives u64 and use alloy_serde::quantitiy

crates/rpc-types-tenderly/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// EIP-1559 transaction
+    #[serde(rename = "0x2")]
+    Eip1559,
+    /// EIP-4844 transaction
+    #[serde(rename = "0x3")]
+    Eip4844,
+    /// EIP-7702 transaction
+    #[serde(rename = "0x4")]
+    Eip7702,
+}
```
mattsse: this is just TxTypes

crates/rpc-types-tenderly/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+        #[serde(rename_all = "camelCase")]
+        struct RawLogInput {
+            value: serde_json::Value,
+            r#type: String,
+            name: String,
+            indexed: bool,
+        }
+
+        let raw = RawLogInput::deserialize(deserializer)?;
+        let ty: DynSolType = raw.r#type.parse().unwrap();
```
mattsse: this we cant do in deserde

crates/rpc-types-tenderly/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// Call type.
+    pub r#type: TenderlyCallType,
+    /// Origin address of the call.
+    pub from: Address,
+    /// Target address of the call.
+    pub to: Address,
+    /// Gas used by the call.
+    pub gas: U64,
+    /// Gas used by the call.
+    pub gas_used: U64,
```
mattsse: same

crates/rpc-types-tenderly/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    pub raw: Log,
+}
+
+/// Log inputs decoded by the tenderly node.
+#[derive(Clone, Debug)]
+pub struct TenderlyLogInput {
+    /// Value of the input.
+    pub value: DynSolValue,
+    /// Type of the input.
+    pub r#type: DynSolType,
```
mattsse: I'd rather have this as symmetrical json values and then we have helpers for extracting the sol value and not do this during serde

mattsse (changes requested):
I see,

okay after thinking about this, let's just go with all the params as args, this is easier and we dont need to change any existing code for this.

siosw:
@mattsse anything i can do to get this over the hump?

mattsse:
nope lgtm, ty! merging for next release

PR #2706, opened 2025-07-14
chore(eips): remove max blobs per tx from `BlobParams` ser/deser

rkrasiuk:
## Description

`maxBlobsPerTx` field was removed from `BlobParams`.  For new Fusaka limit we will set it in the same way we currently handle `blob_base_cost` (see https://github.com/alloy-rs/alloy/pull/2543).

Ref https://github.com/ethereum/EIPs/pull/9981


mattsse (commented):
superseded by #2820

PR #2707, opened 2025-07-14
feat: Allow sharing Ledger transport in LedgerSigner

### LINKED ISSUE #2509: [Feature] Allow sharing Ledger transport in `LedgerSigner`
### Component

signers

The core problem is not being able to have multiple `LedgerSigner`s pointing to different derivation path accounts since the first instantiation will create and hold the transport preventing a new `LedgerSigner` from being instantiated.

There are few ways we can do fix this I think:
1. `new_with_transport` that takes `Arc<Mutex<Ledger>>`
2. instantiate the transport only during the specific operations, which will make it drop at the end of the call and allow multiple signers
3. make a ledger specific `NetworkWallet`(which as I understand is the way to define a wallet with multiple accounts) that can allow multiple derivation paths

I think #1 is probably the most flexible?

Example code that fails:
```rust
use alloy::signers::ledger::{HDPath, LedgerSigner};

#[tokio::main]
    tracing_subscriber::fmt().init();

    let ledger = LedgerSigner::new(HDPath::LedgerLive(0), None).await?;
    let address = ledger.get_address().await?;
    println!("address: {address:?}");

    let ledger2 = LedgerSigner::new(HDPath::LedgerLive(1), None).await?;
    let address2 = ledger2.get_address().await?;
    println!("address2: {address2:?}");

    Ok(())
}
```

### Additional context

_No response_

varun-doshi:
## Motivation

Fixes #2509 

## Solution

Implement `new_with_transport`.
Changes LedgerSigner to have a Arc in transport

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer-ledger/src/signer.rs
+++ b/crates/signer-ledger/src/signer.rs
@@ -173,10 +173,10 @@ impl LedgerSigner {
         chain_id: Option<ChainId>,
         transport: Arc<Mutex<Ledger>>,
     ) -> Result<Self, LedgerError> {
-        let mut transport_guard = transport.lock().await;
-        let address =
-            Self::get_address_with_path_transport(&mut *transport_guard, &derivation).await?;
-        drop(transport_guard);
+        let address = {
+            let transport_guard = transport.lock().await;
+            Self::get_address_with_path_transport(&transport_guard, &derivation).await?
+        };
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this does make sense

PR #2709, opened 2025-07-14
fix: Remove unnecessary clone() calls in Genesis::extend_accounts tests

reallesee:
Removed unnecessary clone() calls when passing arguments to extend_accounts() method, as it takes ownership via IntoIterator trait, avoiding unnecessary memory allocations.


mattsse:
this doesnt actually compile so we need this clone

PR #2712, opened 2025-07-14
chore: add helper to collect rpc logs

mattsse:
smol helper for collecting all rpc logs in a block:

https://github.com/paradigmxyz/reth/blob/b9c63f6a10bfbf89b7b3585acf34750d3dbf7493/crates/rpc/rpc-eth-types/src/receipt.rs#L51-L65

cc @lean-apple


lean-apple:
Thanks, always more cleaner than the macro in `reth`

Issue #2704, opened 2025-07-14
[Bug] Multicall decoding for empty response


Corn3333:
### Component

provider, pubsub

### What version of Alloy are you on?

1.0.20

### Operating System

Windows

### Describe the bug

I use this code to make a call via Multicall to get pool reserves:

```rust
pub async fn multicall_reserves<T: Provider>(
    provider: &T,
    data: &Vec<Address>
) -> MulticallResult<()> {    

    let call_items = data.iter()
        .map(|address| {
            let input = getReservesCall {}.abi_encode().into();
            CallItem::new(*address, input).allow_failure(true)
        })
        .collect::<Vec<CallItem<getReservesCall>>>();
    
    let results = multicall(&provider, call_items).await?;

    for result in results {
        println!("{:?}", result);
    }


    Ok(())
}

async fn multicall<T, C, R>(provider: &T, call_items: Vec<CallItem<C>>) -> MulticallResult<Vec<Result<R, Failure>>>
where
    T: Provider,
    C: SolCall<Return = R> + 'static,
{
    let multicall = provider
        .multicall()
        .dynamic::<C>()
        .extend_calls(call_items);
    
    multicall.aggregate3().await
}
``` 

If the address is a valid Uniswap V2 pool, then everything is fine. But if I make a request to other addresses, I get an error:
```rust
Err(DecodeError(Overrun))
``` 

For the study I chose the following pools:
```rust
    let pools: Vec<Address> = vec![
        address!("0x26021f1b02f23ae36540c20c1615a6665e092cbc"), // UniV2
        address!("0xb0a62eac9222b18f545c86978aaf09c0b75ce764"), // UniV2
        address!("0xc4704f13d5e08b27b039d53873e813dd2fad99d9"), // UniV2

        address!("0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"), // WETH
        address!("0x71C7789175fF712975f86Eb82A3Bf65b6E4F159c"), // WETH Deployer(EOA)
        address!("0x7E75b6876a9feE811EC67B55385BA5A1491D11f3"), // UniV3
    ];
``` 

Then I wrote my own Multicall implementation:
```rust

pub async fn test_multicall<T: Provider>(
    provider: &T,
    data: &Vec<Address>
) -> AnyResult<()> {
    use alloy_provider::bindings::IMulticall3::{aggregate3Call, Call3};

    let mut calls = Vec::new();

    for address in data {
        calls.push(Call3 {
[...]

PR #2724, opened 2025-07-17
fix: return abi decoding errors as multicall failures

Corn3333:
## Motivation
Fixes https://github.com/alloy-rs/alloy/issues/2704

## Solution
```rust
If res.success == true
``` 
check: 
```rust
if res.returnData is empty.
``` 

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/multicall/inner_types.rs
+++ b/crates/provider/src/provider/multicall/inner_types.rs
@@ -162,7 +162,7 @@ impl<D: SolCall> CallTuple for Dynamic<D> {
             if res.success {
                 ret.push(
                     D::abi_decode_returns(&res.returnData)
-                        .or_else(|_| Err(Failure { idx, return_data: res.returnData.clone() })),
+                        .map_err(|_| Failure { idx, return_data: res.returnData.clone() }),
                 )
             } else {
                 ret.push(Err(Failure { idx, return_data: res.returnData.clone() }));
--- a/crates/provider/src/provider/multicall/inner_types.rs
+++ b/crates/provider/src/provider/multicall/inner_types.rs
@@ -159,14 +159,8 @@ impl<D: SolCall> CallTuple for Dynamic<D> {
     ) -> Result<Self::Returns> {
         let mut ret = vec![];
         for (idx, res) in results.iter().enumerate() {
-            if res.success {
-                if res.returnData.is_empty() {
-                    ret.push(Err(Failure { idx, return_data: res.returnData.clone() }));
-                } else {
-                    ret.push(Ok(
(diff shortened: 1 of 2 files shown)


mattsse (commented):
wdyt @DaniPopes @yash-atreya 

I think we can always assume that empty response is an error?

Corn3333:
> wdyt @DaniPopes @yash-atreya
> 
> I think we can always assume that empty response is an error?

We can try to decode, if there is an error => then wrap it in Failure.
But this causes optimization to suffer

crates/provider/src/provider/multicall/inner_types.rs
```diff
@@ -160,9 +160,13 @@ impl<D: SolCall> CallTuple for Dynamic<D> {
         let mut ret = vec![];
         for (idx, res) in results.iter().enumerate() {
             if res.success {
```
DaniPopes, line 162: this should just be `if res.success && let Ok() = D::...` instead of checking for empty return data

Corn3333: When implementing your option, I get an error:
```rust
error[E0658]: `let` expressions in this position are unstable
   --> crates\provider\src\provider\multicall\inner_types.rs:162:31
    |
162 |             if res.success && let Ok(decode) = D::abi_decode_returns(&res.returnData) {
    |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information
``` 

But I found a beautiful solution

crates/provider/src/provider/multicall/tuple.rs
```diff
@@ -145,7 +145,8 @@ macro_rules! impl_tuple {
 
                 Ok(($(
                     match &results[$idx].success {
-                        true => Ok($ty::abi_decode_returns(&results[$idx].returnData).map_err(MulticallError::DecodeError)?),
+                        true => $ty::abi_decode_returns(&results[$idx].returnData)
+                            .or(Err(Failure { idx: $idx, return_data: results[$idx].returnData.clone() })),
```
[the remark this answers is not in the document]
Corn3333:
```rust
.or() 
```
returns the Result type. It makes no sense to additionally wrap it in Ok()

mattsse (approved):
lgtm, pending @DaniPopes

crates/provider/src/provider/multicall/inner_types.rs
DaniPopes: ah this repo hasn't moved to edition 2024 yet, this is fine

PR #2736, opened 2025-07-23
Multicall send support

### LINKED ISSUE #2733: [Feature] Support for sending transactions via Multicall (not just eth_call)
### Component

provider, pubsub

I’ve been experimenting with the Multicall3 helper in Alloy and noticed that it currently only supports performing readonly calls (i.e. using eth_call). In my use case, I’d like to batch multiple state‐changing operations into a single transaction so that I can submit them atomically.

For example, I’d love to do something like:
``` rust
let multicall = provider.multiclal()
    .add_call(contract_a.method1(...))
    .add_call(contract_b.method2(...));
// … then send as a single on‐chain transaction
let pending_tx = multicall.send().await?;
```

Right now, however, Multicall3 seems hard‑wired to invoke eth_call, and there’s no obvious way to switch it over to a sendRawTransaction.

### Additional context

_No response_

nitininhouse:
## Summary

Adds `send_*` methods to `MulticallBuilder` to support sending multicall operations as transactions instead of just read-only calls.

## Changes

- Add `send_aggregate()`
- Add `send_try_aggregate()` 
- Add `send_aggregate3()`
- Add `send_aggregate3_value()`
- Add `send_block_and_aggregate()`
- Add `send_try_block_and_aggregate()`

## Usage
Users can now send multicall operations as transactions:

```rust
let multicall = provider.multicall()
    .add_call(contract_a.method1(...))
    .add_call(contract_b.method2(...));

// Send as a single on-chain transaction
let pending_tx = multicall.send_aggregate().await?;
```
Closes #2733

### DIFF
--- a/crates/provider/src/provider/multicall/mod.rs
+++ b/crates/provider/src/provider/multicall/mod.rs
@@ -624,6 +624,7 @@ where
         let tryBlockAndAggregateReturn { blockNumber, blockHash, returnData } = output;
         Ok((blockNumber.to::<u64>(), blockHash, T::decode_return_results(&returnData)?))
     }
+
     /// Sends the `tryBlockAndAggregate` function as a transaction  
     pub async fn send_try_block_and_aggregate(
         &self,
@@ -632,6 +633,7 @@ where
         let call = self.to_try_block_and_aggregate_call(require_success);
         self.build_and_send(call, None).await
     }
+
     /// Helper for building the transaction request for the given call type input.
     fn build_request<M: SolCall>(
         &self,


crates/provider/src/provider/multicall/mod.rs
```diff
@@ -486,12 +498,22 @@ where
...
 
     /// Encodes the calls for the `aggregate3` function and returns the populated transaction
     /// request.
     pub fn to_aggregate3_request(&self) -> N::TransactionRequest {
         self.build_request(self.to_aggregate3_call(), None)
     }
+    /// Sends the `aggregate3Value` function as a transaction
+    pub async fn send_aggregate3_value(&self) -> Result<PendingTransactionBuilder<N>> {
+    let total_value = self.calls.iter().map(|c| c.value).fold(U256::ZERO, |acc, x| acc + x);
+    let call = aggregate3ValueCall { calls: self.calls.to_vec() };
```
mattsse: could we also add a fn for to_aggregate3ValueCall

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -554,6 +576,16 @@ where
         let result = T::decode_return_results(&returnData)?;
         Ok((blockNumber.to::<u64>(), blockHash, T::try_into_success(result)?))
     }
+    /// Sends the `blockAndAggregate` function as a transaction
+    pub async fn send_block_and_aggregate(&self) -> Result<PendingTransactionBuilder<N>> {
+    let calls = self
+        .calls
+        .iter()
+        .map(|c| Call { target: c.target, callData: c.callData.clone() })
+        .collect::<Vec<_>>();
```
mattsse: same here

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -571,7 +603,19 @@ where
...
+pub async fn send_try_block_and_aggregate(
+    &self,
+    require_success: bool,
+) -> Result<PendingTransactionBuilder<N>> {
+    let calls = self
+        .calls
+        .iter()
+        .map(|c| Call { target: c.target, callData: c.callData.clone() })
+        .collect::<Vec<_>>();
+    let call = tryBlockAndAggregateCall { requireSuccess: require_success, calls: calls.to_vec() };
```
mattsse: same

crates/provider/src/provider/multicall/mod.rs
nitininhouse: done

crates/provider/src/provider/multicall/mod.rs
nitininhouse: done

crates/provider/src/provider/multicall/mod.rs
nitininhouse: done

nitininhouse:
Thank you, @mattsse! Looking forward to seeing it merged.

PR #2740, opened 2025-07-25
feat(transport/layers): redundancy layer

thedevbirb:
## Motivation

While building a reliable service such as a sequencer like [taiko-mk1](https://github.com/chainbound/taiko-mk1), we discovered the necessity to run with multiple, redundant L1/L2 EL clients, so that the service can still function in case some of the RPCs go down or lag behind.

## Solution

The PR introduces a new `Tower` layer/service called `RedundancyLayer`/`RedundancyService` that allows to create an Alloy `RpcClient` using multiple underlying RPC clients. Since multiple responses can be returned, the concept of a `RedundancyStrategy` is also introduced: depending on the type of request, a user might be interested in having the first successful answer (e.g. sending a transaction), or the "best" answer returned (e.g. the highest block available).

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation

### DIFF
--- a/crates/transport/src/layers/mod.rs
+++ b/crates/transport/src/layers/mod.rs
@@ -13,3 +13,6 @@ pub use retry::{RateLimitRetryPolicy, RetryBackoffLayer, RetryBackoffService, Re
 /// FallbackLayer
 mod fallback;
 pub use fallback::{FallbackLayer, FallbackService};
+
+pub mod redundancy;
+pub use redundancy::{RedundancyLayer, RedundancyService};
--- a/crates/transport/src/layers/redundancy.rs
+++ b/crates/transport/src/layers/redundancy.rs
@@ -1,0 +1,441 @@
+//! Redundancy layer for blockchain client failover and response optimization.
+//!
+//! This module provides a transparent redundancy system that enables clients to:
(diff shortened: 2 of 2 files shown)


thedevbirb:
> Could you perhaps give a brief overview how this is different from the `FallbackService`

Good question. For full transparency, I acknowledge that right now the `FallbackService` is a better implementation of the `RedundancyService` when only used with `FirstRpcSuccess` strategy.

However, the end-goal would be achieving a more extensible version of such service. Right now it features:
- An additional strategy, `HighestValue`, helpful for any response type which can features a partial ordering.
- The ability to customise strategies _per request_, allowing more flexibility in its usage.

We've thought of additional strategies, like:
- `Consensus` -- you require `m` out of `n` equal responses, otherwise you fail.
- `Merge` -- you merge all responses into one. An example of this being merging response from an endpoint which returns you mempool transactions.

Do you think these additional functionalities would be of interest to you?

Overall, I understand that right now having two similar solutions doesn't make too much sense. I've converted it the PR to draft and plan to experiment with absorbing the `FallbackProvider` into this service while retaining the scoring logic which is quite useful.

PR #2741, opened 2025-07-26
feat: added type for debug_storageRangeAt

Rimeeeeee:
ref [#17615](https://github.com/paradigmxyz/reth/pull/17615)

### DIFF
--- a/crates/rpc-types-debug/src/debug.rs
+++ b/crates/rpc-types-debug/src/debug.rs
@@ -2,6 +2,7 @@
 
 use alloc::{collections::btree_map::BTreeMap, vec::Vec};
 use alloy_primitives::{Bytes, StorageKey, B256};
+use derive_more::{AsMut, AsRef, Deref, DerefMut};
 use serde::{Deserialize, Serialize};
 
 /// Represents the result of a storage slot query.
@@ -15,7 +16,9 @@ pub struct StorageResult {
 }
 
 /// Wrapper type for a map of storage slots.
-#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(
+    Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize, AsRef, Deref, AsMut, DerefMut,
+)]
 #[serde(rename_all = "camelCase")]
 pub struct StorageMap(pub BTreeMap<B256, StorageResult>);
 
--- a/crates/rpc-types-debug/src/debug.rs
+++ b/crates/rpc-types-debug/src/debug.rs
@@ -1,8 +1,31 @@
 //! Types for the `debug` API.
 
-use alloc::vec::Vec;
-use alloy_primitives::Bytes;
+use alloc::{collections::btree_map::BTreeMap, vec::Vec};
+use alloy_primitives::{Bytes, StorageKey, StorageValue, B256};
 use serde::{Deserialize, Serialize};
+
+/// Represents the result of a storage slot query.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageResult {
+    /// The key of the storage slot
+    pub key: B256,
+    /// The storage key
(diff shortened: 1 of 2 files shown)


crates/rpc-types-debug/src/debug.rs
```diff
@@ -1,9 +1,32 @@
...
 
-use alloc::vec::Vec;
-use alloy_primitives::Bytes;
+use alloc::{collections::btree_map::BTreeMap, vec::Vec};
+use alloy_primitives::{Bytes, StorageKey, StorageValue, B256};
 use serde::{Deserialize, Serialize};
 
+/// Represents the result of a storage slot query.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageResult {
```
mattsse, line 11: I believe these should be camelCase?

crates/rpc-types-debug/src/debug.rs
```diff
@@ -1,9 +1,32 @@
...
+    /// The value stored at the slot
+    pub value: StorageValue,
+}
+
+/// Represents a map of storage slots.
+pub type StorageMap = BTreeMap<B256, StorageResult>;
+
+/// Represents the result of a storage range query.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageRangeResult {
```
mattsse: same here

crates/rpc-types-debug/src/debug.rs
```diff
@@ -1,9 +1,32 @@
...
+    /// The key of the storage slot
+    pub key: B256,
+    /// The storage key
+    pub storage_key: StorageKey,
+    /// The value stored at the slot
+    pub value: StorageValue,
+}
+
+/// Represents a map of storage slots.
+pub type StorageMap = BTreeMap<B256, StorageResult>;
```
mattsse: let's introduce a dedicated wrapper type for this, this way we can later add more functionality

crates/rpc-types-debug/src/debug.rs
```diff
@@ -1,9 +1,32 @@
...
 
+/// Represents the result of a storage slot query.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct StorageResult {
+    /// The key of the storage slot
+    pub key: B256,
+    /// The storage key
+    pub storage_key: StorageKey,
+    /// The value stored at the slot
+    pub value: StorageValue,
```
mattsse: I think this isn't quite equivalent to the reference

https://github.com/ethereum/go-ethereum/blob/16117eb7cddc4584865af106d2332aa89f387d3d/eth/api_debug.go#L199-L203

crates/rpc-types-debug/src/debug.rs
```diff
@@ -1,25 +1,27 @@
...
-    pub value: StorageValue,
+    pub value: B256,
 }
 
-/// Represents a map of storage slots.
-pub type StorageMap = BTreeMap<B256, StorageResult>;
+/// Wrapper type for a map of storage slots.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StorageMap(pub BTreeMap<B256, StorageResult>);
```
mattsse: we can also derive asref asmut and refer for this

like we do here

https://github.com/alloy-rs/alloy/blob/ff758f86622c343457c4fe856a88033819f11d6e/crates/consensus/src/transaction/recovered.rs#L11-L11
if this isnt a dep, we can add

https://github.com/alloy-rs/alloy/blob/ff758f86622c343457c4fe856a88033819f11d6e/crates/consensus/src/transaction/recovered.rs#L8-L8

mattsse (changes requested):
last pedantic suggestions,

we also want from+into

mattsse (approved):

PR #2751, opened 2025-07-31
feat: add TxHashRef trait and implementations

mattsse:
## Summary

Introduces a new `TxHashRef` trait that provides a standard interface for accessing transaction hashes without requiring ownership. The trait assumes the implementing type already owns or has computed the transaction hash.

## Changes

- Added `TxHashRef` trait in `transaction/mod.rs` with a single method `tx_hash(&self) -> &TxHash`
- Implemented `TxHashRef` for:
  - `Recovered<T>` where `T: TxHashRef` - delegates to inner type
  - `WithEncoded<T>` where `T: TxHashRef` - accesses value field
  - `Signed<T>` where `T: RlpEcdsaEncodableTx` - uses existing hash() method
  - `EthereumTxEnvelope<Eip4844>` where `Eip4844: RlpEcdsaEncodableTx` - uses existing tx_hash() method

This provides a consistent way to access transaction hashes across different transaction wrapper types.


crates/consensus/src/transaction/mod.rs
```diff
@@ -557,6 +557,30 @@ where
...
+///
+/// This trait is implemented by types that contain or can provide a reference to a
+/// transaction hash ([`TxHash`]). It provides a standard interface for accessing
+/// transaction hashes without requiring ownership.
+pub trait TxHashRef {
+    /// Returns a reference to the transaction hash.
+    ///
+    /// This assumes the implementing type already owns or has computed the transaction hash.
+    fn tx_hash(&self) -> &TxHash;
+}
```
klkvr, line 572: should this have `auto_impl` for references?

PR #2755, opened 2025-08-04
fix: poller breaks if server drops the filter

prestwich:
Adds a new check in the filter poller to see if the server has dropped 

## Motivation

FilterPoller will error indefinitely if the server reboots or drops the filter for any reason

## Solution

- Add a new finished state
- move to finished when the server's error response method includes `filter not found`
    - I have checked that the string `filter not found` is used by reth, geth, and nethermind
- drive-by: implement fused stream

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
makes sense, ty!

PR #2756, opened 2025-08-05
feat: implement `Drop` for `MnemonicBuilder` make sure passphrase and password will be cleaned up

chiro-hiro:
## Motivation

I want to ensure that MnemonicBuilder clears the passphrase and password as soon as the closure block ends.

## Solution

Borrow `&mut String` and zeroize each byte with `.fill(0)`

Implement `Drop` for `Entropy` https://github.com/summa-tx/coins/pull/142

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer-local/Cargo.toml
+++ b/crates/signer-local/Cargo.toml
@@ -32,7 +32,10 @@ k256.workspace = true
 rand.workspace = true
 thiserror.workspace = true
 async-trait.workspace = true
-zeroize = { workspace = true, optional = true, features = ["zeroize_derive"] }
+zeroize = { workspace = true, optional = true, features = [
+    "zeroize_derive",
+    "alloc",
+] }
 
 # keystore
 eth-keystore = { version = "0.5.0", default-features = false, optional = true }


chiro-hiro:
Should we use [zeroize](https://crates.io/crates/zeroize)?

mattsse (dismissed):
I see, this makes sense.

mattsse:
@chiro-hiro I think we can/should use zeroize here, it is probably already in the dep tree, so let's use that

chiro-hiro:
Hi @mattsse, I have some updates.

1. Added `zeroize@1.8.1` at workspace level for future use.
2. Added `zeroize` as an optional dependency for `signer-local`, enabled via the `zeroize` feature.
3. Automatically enable the `zeroize` feature when the `mnemonic` feature is used.

For now, I'm only apply `zeroize` for `MnemonicBuilder`. The wider adoption of `zeroize` may need to be a separated PR.

mattsse (approved):
lgtm!

PR #2757, opened 2025-08-05
fix: with_avg_unit_cost

jorgemmsilva:
The current function takes ownership of `self` and never returns, making it unusable.

This fix simply returns `Self` so the caller can continue to use it


mattsse (approved):
ah yes

ty

PR #2759, opened 2025-08-05
fix(docs): correct typos in EIP and ENS references

radik878:
Fixed typo in crates/eips/README.md:

- Replaced Proprosal with correct Proposal

Fixed ENS domain typo in crates/ens/src/lib.rs:

- Replaced "deployd.eth" with correct "deployed.eth"

Ensures proper naming and accurate documentation/tests.

### DIFF
--- a/crates/eips/README.md
+++ b/crates/eips/README.md
@@ -1,6 +1,6 @@
 # alloy-eips
 
-Ethereum Improvement Proprosal (EIP) implementations.
+Ethereum Improvement Proposal (EIP) implementations.
 
 Contains constants, helpers, and basic data structures for consensus EIPs.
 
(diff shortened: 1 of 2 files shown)


crates/eips/README.md
```diff
@@ -1,6 +1,6 @@
 # alloy-eips
 
-Ethereum Improvement Proprosal (EIP) implementations.
+Ethereum Improvement Proposal (EIP) implementations.
```
mattsse, line 3: this one seems fine

crates/ens/src/lib.rs
```diff
@@ -321,7 +321,7 @@ mod tests {
         let provider = ProviderBuilder::new()
             .connect_http("https://reth-ethereum.ithaca.xyz/rpc".parse().unwrap());
 
-        let name = "deployd.eth";
+        let name = "deployed.eth";
```
mattsse: this we should keep because this actually checks the address

mattsse (changes requested):

crates/ens/src/lib.rs
radik878: > this we should keep because this actually checks the address

Reverted this change, boss

PR #2763, opened 2025-08-06
chore: serialize return data with prefix

### LINKED ISSUE #2761: [Bug] Return default Geth trace `return_value` *with* `0x` prefix
### Component

rpc

_No response_

### Operating System

None

### Describe the bug

Since https://github.com/ethereum/go-ethereum/pull/31216 ; the returned value of the default Geth trace now includes the `0x` prefix.

It still is excluded here: https://github.com/alloy-rs/alloy/blob/ce5f562e9617dc04a28e58f30b0b442c7b705c98/crates/rpc-types-trace/src/geth/mod.rs#L62-L63

This might be a breaking change, but given the link to geth's repo above the struct definition, I think we should align.

mattsse:
closes #2761

### DIFF
--- a/crates/rpc-types-trace/src/geth/mod.rs
+++ b/crates/rpc-types-trace/src/geth/mod.rs
@@ -59,7 +59,6 @@ pub struct DefaultFrame {
     /// How much gas was used.
     pub gas: u64,
     /// Output of the transaction
-    #[serde(serialize_with = "alloy_serde::serialize_hex_string_no_prefix")]
     pub return_value: Bytes,
     /// Recorded traces of the transaction
     pub struct_logs: Vec<StructLog>,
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/geth/mod.rs
```diff
@@ -749,6 +748,19 @@ mod tests {
     use super::*;
     use similar_asserts::assert_eq;
 
+    #[test]
+    fn test_return_data_prefix() {
+        let raw = r#"{
+  "failed": false,
+  "gas": 0,
+  "returnValue": "",
```
ngotchac, line 756: Just noting, I think `geth` returns `0x` for empty `return_value`

PR #2766, opened 2025-08-06
feat: add headerinfo helper type

mattsse:
This is the bare minimum we can extract without additional context (e.g. specid, blobcontext)

but is almost identical to the evm `BlockEnv`,

with this we can make it easier to fill the BlockEnv that we need for the evm,

with the specid as additional context this can be easily converted into the full blockenv then

this is an additional header trait fn but is default impl'ed, so this doesnt break.

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -5,7 +5,10 @@ use alloc::{
     string::{String, ToString},
     vec::Vec,
 };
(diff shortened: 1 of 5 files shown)


crates/rpc-types-engine/src/payload.rs
```diff
@@ -1385,6 +1385,49 @@ impl ExecutionPayload {
         &mut self.as_v1_mut().transactions
     }
 
+    /// Extracts essential information into one container type.
+    pub fn header_info(&self) -> HeaderInfo {
+        self.as_v1();
```
klkvr:
```suggestion
```

klkvr (approved):

Issue #2776, opened 2025-08-12
[Feature] Move CallKind enum from inspectors to alloy


mattsse:
### Component

rpc

### Describe the feature you would like

we have: 

https://github.com/paradigmxyz/revm-inspectors/blob/557abd1a72fcef1ee610f79eec3c4371c20fcdc6/src/tracing/types.rs#L459-L479

ideally we move this to the trace rpc crate because this is useful

## TODO
* move enum and conversions to geth/call

nadtech-hub:
May I take on it?

PR #2779, opened 2025-08-12
feat(rpc-types-trace): move `CallKind` from `revm-inspector`

moricho:
## Motivation
part of: https://github.com/alloy-rs/alloy/issues/2776

## Solution
This PR imports `CallKind` implementation and some conversions from `revm-inspector`, and use it for the `typ` field in `CallFrame` instead of `String`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -42,7 +42,7 @@ pub struct CallFrame {
     pub value: Option<U256>,
     /// The type of the call.
     #[serde(rename = "type")]
-    pub typ: String,
+    pub typ: CallKind,
```
mattsse: I'd like to keep this as string for now, because this would be a breaking change

we can make this change for the next breaking release

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -61,21 +61,21 @@ impl CallFrame {
...
     /// Returns true if this is a static call
     pub fn is_static_call(&self) -> bool {
-        self.typ == "STATICCALL"
+        self.typ == CallKind::StaticCall
     }
 
     /// Returns true if this is a auth call
     pub fn is_auth_call(&self) -> bool {
-        self.typ == "AUTHCALL"
+        self.typ == CallKind::AuthCall
```
mattsse, line 79: we can keep this checks if we add another PartialEq<String|&str> impl for CallKind

crates/rpc-types-trace/src/geth/call.rs
moricho: Got it, I'll keep this as String and change it to CallKind later
https://github.com/alloy-rs/alloy/pull/2779/commits/62342fc4981f7967ae302b56fd155148c6b1c4ad

mattsse (approved):

Issue #2778, opened 2025-08-12
[Feature] add blob fee helpers to payload type


mattsse:
### Component

rpc

### Describe the feature you would like

ref https://github.com/paradigmxyz/reth/blob/main/crates/ethereum/evm/src/lib.rs#L316-L320

same as the header impls but for the payload type https://github.com/alloy-rs/alloy/blob/c197dca34ea1c2d866336cc4cb2c4807f20b62de/crates/rpc-types-engine/src/payload.rs#L1169-L1169

https://github.com/alloy-rs/alloy/blob/c197dca34ea1c2d866336cc4cb2c4807f20b62de/crates/consensus/src/block/header.rs#L643-L673

lean-apple:
Hi, may I or do you want to go for it @iTranscend ?

mattsse:
assigned, should be quick and separate from #17824 because we still need a release for integrating this in reth

PR #2782, opened 2025-08-13
feat: complete execution payload getter methods

lean-apple:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/2778.

## Solution

Copy same functions than on the header https://github.com/alloy-rs/alloy/blob/c197dca34ea1c2d866336cc4cb2c4807f20b62de/crates/consensus/src/block/header.rs#L643-L681

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -660,7 +660,7 @@ pub trait BlockHeader {
     ///
     /// Returns `None` if the `blob_params` are `None`.
     fn maybe_next_block_excess_blob_gas(&self, blob_params: Option<BlobParams>) -> Option<u64> {
-        self.next_block_excess_blob_gas(blob_params?)
+        blob_params.and_then(|params| self.next_block_excess_blob_gas(params))
     }
 
     /// Returns the blob fee for the next block according to the EIP-4844 spec.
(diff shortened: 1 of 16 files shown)


crates/rpc-types-beacon/src/payload.rs
```diff
@@ -112,8 +112,8 @@ struct BeaconPayloadAttributes {
 
 /// Optimism Payload Attributes
 #[serde_as]
-#[derive(Serialize, Deserialize)]
-struct BeaconOptimismPayloadAttributes {
+#[derive(Debug, Serialize, Deserialize)]
+pub struct BeaconOptimismPayloadAttributes {
```
lean-apple: @mattsse, @emhane , this object was spotted by clippy as never used, should we keep it ? if not, can remove in another pr.

emhane: idk, @mattsse is this type wip? seems like this type should be moved to op-alloy at least

crates/consensus/src/block/header.rs
```diff
@@ -660,7 +660,7 @@ pub trait BlockHeader {
     ///
     /// Returns `None` if the `blob_params` are `None`.
     fn maybe_next_block_excess_blob_gas(&self, blob_params: Option<BlobParams>) -> Option<u64> {
-        self.next_block_excess_blob_gas(blob_params?)
+        blob_params.and_then(|params| self.next_block_excess_blob_gas(params))
```
mattsse: can we leave this unchanged?

no need to change this

crates/rpc-types-beacon/src/payload.rs
mattsse: unrelated to this pr, please leave unchanged

likely that we can remove this

mattsse (changes requested):
cool,

only cosmetic nits

PR #2790, opened 2025-08-14
feat: add debug state dump struct

Pana:
## Motivation
To dev the debug_dumpBlock and debug_accountRange RPC method, it is required first add the rpc type here

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-debug/src/state_dump.rs
+++ b/crates/rpc-types-debug/src/state_dump.rs
@@ -46,6 +46,8 @@ pub struct StateDump {
     pub accounts: BTreeMap<Address, AccountState>,
     /// Next can be set to represent that this dump is only partial, and Next
     /// is where an iterator should be positioned in order to continue the dump.
+    ///
+    /// Note: this uses base64 because this is based on the default go-ethereum behaviour that by default serializes in base64 <https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/core/state/dump.go#L66-L68>
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde_as(as = "Option<Base64>")]
     pub next: Option<Bytes>,


crates/rpc-types-debug/src/state_dump.rs
```diff
@@ -0,0 +1,99 @@
...
+pub struct StateDump {
+    /// The root hash of the state trie
+    pub root: B256,
+    /// A map of accounts, indexed by address
+    pub accounts: BTreeMap<Address, AccountState>,
+    /// Next can be set to represent that this dump is only partial, and Next
+    /// is where an iterator should be positioned in order to continue the dump.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<Base64>")]
+    pub next: Option<Bytes>,
```
mattsse: can you elaborate on why this is base64?

mattsse (commented):
nice

I only have one question regarding the base64 serde, why is this not just hex?

crates/rpc-types-debug/src/state_dump.rs
Pana: In geth's implementation this filed is serialized as base64, I suspect that this RPC method may not have been well designed and its implementation is somewhat ad hoc. However, to maintain compatibility, we also chose to use base64 encoding here.

mattsse: can you reference the geth code for this, can't find base64 here

https://github.com/ethereum/go-ethereum/blob/85077be58edea572f29c3b1a6a055077f1a56a8b/core/state/dump.go#L66-L68

Pana: geth use []byte for the next field, in Go, []byte default will use base64 encoding for json. hexutil.Bytes will encode as hex

mattsse: great stuff :D

mattsse (approved):

PR #2793, opened 2025-08-14
[feat] Introduced helper iterator for CallFrame

### LINKED ISSUE #2775: [Feature] Add helper iterator for Callframe
### Component

rpc

Callframe is a nested/recursive structure

https://github.com/alloy-rs/alloy/blob/6520d9efd2cf07e94fc64300287c5ab41329e5e3/crates/rpc-types-trace/src/geth/call.rs#L11-L36

it would be nice to have an iterator type that iterates over all callframes but allows skipping child calls

## TODO
* add helper iterator that keeps track of the parent <> child relation ship and can skip child calls
* this can also be solved with some stack based approach

### Additional context

_No response_

developeruche:
## Motivation

Fixes #2775 

## Solution
Introduced a New Iterator type; 
`CallFrameIter`, also enabling a utility method for skipping children.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/builder.rs
+++ b/crates/provider/src/builder.rs
@@ -309,6 +309,14 @@ impl<L, F, N> ProviderBuilder<L, F, N> {
         self.layer(CallBatchLayer::new())
     }
 
+    /// Aggregate multiple `eth_call` requests with block number queries done by calling Arbsym
+    /// precompile.
+    ///
+    /// See [`CallBatchLayer`] for more information.
+    pub fn with_arbitrum_call_batching(self) -> ProviderBuilder<Stack<CallBatchLayer, L>, F, N> {
+        self.layer(CallBatchLayer::new().arbitrum_compat())
+    }
+
     // --- Build to Provider ---
 
     /// Finish the layer stack by providing a root [`Provider`], outputting


crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -173,6 +173,56 @@ impl FlatCallConfig {
...
+
+impl<'a> Iterator for CallFrameIter<'a> {
+    type Item = (&'a CallFrame, Option<&'a CallFrame>);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let (frame, parent) = self.stack.pop()?;
+
+        // Determine if we should skip children for this frame
+        let should_skip =
+            self.skip_depth.map(|depth| depth == self.stack.len() + 1).unwrap_or(false);
```
mattsse: this feels a bit weird, because this doesn't work for the intended use case which is

let next = iter.next();
if next.<> iter.skip_childred

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -173,6 +173,56 @@ impl FlatCallConfig {
...
+    /// This applies only to the *next* `next` call,
+    /// and will prevent traversal into its children.
+    pub fn skip_children(&mut self) {
+        // Mark the current depth so we can avoid pushing its children
+        self.skip_depth = Some(self.stack.len());
+    }
+}
+
+impl<'a> Iterator for CallFrameIter<'a> {
+    type Item = (&'a CallFrame, Option<&'a CallFrame>);
```
mattsse: I'd like to introduce a helper type for this,

that then also borrows `CallFrameIter` itself, this way we can directly control how the iter behaves

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -173,6 +173,74 @@ impl FlatCallConfig {
...
+    /// Creates a new iterator starting from `root`.
+    pub fn new(root: &'a CallFrame) -> Self {
+        Self { stack: vec![(root, None)], last_frame: None }
+    }
+
+    /// Skips children for the most recently yielded item.
+    /// Note: this would panic if there are no parent owning the children to
+    /// skip. `next` must be called before `skip_children`.
+    pub fn skip_children(&mut self) {
+        let parent = self.last_frame.unwrap();
```
developeruche: @mattsse would it be cool to prop an error here rather than unwrapping

mattsse: I guess we can just use an let some here?

mattsse (commented):
great

one more nit re unwrap

developeruche:
> great
> 
> one more nit re unwrap

@mattsse, sorry for the delay. For some reason, your review didn't drop a notification.

grandizzy:
@developeruche would you still be on to drive this PR to the finish line? thank you!

developeruche:
> @developeruche would you still be on to drive this PR to the finish line? thank you!

Yeah, I am! Still waiting on a review

mattsse:
Closing in favor of #3768 so this feature lands through a single PR path. Thank you for the work here.

PR #2794, opened 2025-08-14
[Feature] Implement support different signatures in envelope macro

### LINKED ISSUE #2718: [Feature] Support different signatures in envelope macro
### Component

consensus, eips, genesis

currently we enforce the signature type on serialize:

https://github.com/alloy-rs/alloy/blob/576a3ab3677642f6e5056e9127613bc4632b37d8/crates/consensus/src/signed.rs#L535-L535

because we need the hash

https://github.com/alloy-rs/alloy/blob/576a3ab3677642f6e5056e9127613bc4632b37d8/crates/consensus/src/signed.rs#L546-L546

because the signature type is enforced by `RlpEcdsaDecodableTx`

https://github.com/alloy-rs/alloy/blob/576a3ab3677642f6e5056e9127613bc4632b37d8/crates/consensus/src/signed.rs#L118-L123

https://github.com/alloy-rs/alloy/blob/576a3ab3677642f6e5056e9127613bc4632b37d8/crates/consensus/src/transaction/rlp.rs#L110-L113

Ideally Signed can obtain a hash from the tx with any signature,
this could require another helper trait that we could auto impl for all RlpEcdsaDecodableTx

something similar to https://github.com/alloy-rs/alloy/blob/576a3ab3677642f6e5056e9127613bc4632b37d8/crates/network/src/transaction/signer.rs#L82-L82

cc @allnil 

### Additional context

_No response_

developeruche:
## Motivation
Closes #2718

## Solution
- Introduced `TxHashable` trait
- decoupled `tx_hash_with_type` from `RlpEcdsaEncodableTx`
- Implemented the `TxHashable` where needed.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/hashable.rs
+++ b/crates/consensus/src/transaction/hashable.rs
@@ -1,0 +1,7 @@
+use alloy_primitives::TxHash;
+
+/// Generic trait to get a transaction hash from any signature type
+pub trait TxHashable<S> {
+    /// Calculate the transaction hash for the given signature and type.
+    fn tx_hash_with_type(&self, signature: &S, ty: u8) -> TxHash;
+}
--- a/crates/consensus/src/transaction/rlp.rs
+++ b/crates/consensus/src/transaction/rlp.rs
@@ -1,5 +1,4 @@
-use crate::Signed;
-use alloc::vec::Vec;
+use crate::{transaction::hashable::TxHashable, Signed};
 use alloy_eips::{
     eip2718::{Eip2718Error, Eip2718Result},
     Typed2718,
@@ -11,7 +10,7 @@ use alloy_rlp::{Buf, BufMut, Decodable, Encodable, Header};
 #[doc(hidden)]
 #[doc(alias = "RlpEncodableTx", alias = "RlpTxEncoding")]
 #[auto_impl::auto_impl(&, Arc)]
-pub trait RlpEcdsaEncodableTx: Sized + Typed2718 {
+pub trait RlpEcdsaEncodableTx: Sized + Typed2718 + TxHashable<Signature> {
     /// Calculate the encoded length of the transaction's fields, without a RLP
     /// header.
     fn rlp_encoded_fields_length(&self) -> usize;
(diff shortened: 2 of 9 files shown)


mattsse (commented):
thanks, this is def something we want to support,

need a few more eyes on this, will get back to this next week

but all of this checks out

crates/consensus/src/transaction/rlp.rs
```diff
@@ -11,7 +10,7 @@ use alloy_rlp::{Buf, BufMut, Decodable, Encodable, Header};
 #[doc(hidden)]
 #[doc(alias = "RlpEncodableTx", alias = "RlpTxEncoding")]
 #[auto_impl::auto_impl(&, Arc)]
-pub trait RlpEcdsaEncodableTx: Sized + Typed2718 {
+pub trait RlpEcdsaEncodableTx: Sized + Typed2718 + TxHashable<Signature> {
```
klkvr: this would be a breaking change as is. I'm wondering if we could avoid any changes to this trait and instead have a blanket impl for `TxHashable<alloy_primitives::Signature>` for any type implementing `RlpEcdsaEncodableTx`?

developeruche: We can't use a blanket `impl` like this:

```rust
impl<T> TxHashable<Signature> for T
where
    T: RlpEcdsaEncodableTx,
{
    /// Calculate the transaction hash for the given signature and type.
    fn tx_hash_with_type(&self, signature: &Signature, _ty: u8) -> alloy_primitives::TxHash {
        let mut buf = Vec::with_capacity(self.rlp_encoded_length_with_signature(signature));
        self.rlp_encode_signed(signature, &mut buf);
        keccak256(&buf)
    }
}
```

because the various transaction types require different implementations of the `tx_hash_with_type` function.

Overriding this in the transaction type module is also not an option, since the Rust compiler responds with an error like:

```
conflicting implementations of trait TxHashable<alloy_primitives::Signature> for type TxEip1559
conflicting implementation for TxEip1559
```

Another alternative would be to move the `tx_hash_with_type` function into the `TxHashable` trait and provide a default implementation there. This technically works, but it’s somewhat messy because it introduces additional trait coupling (`+ TxHashable<Signature>`) everywhere `RlpEcdsaEncodableTx` is used as a generic type and when the `EIP4844` generic type is created.

The updated code is in the next commit. This approach avoids breaking changes, but personally, I still prefer the other version. 🙂

@klkvr

klkvr: what I meant is that we could keep the tx_hash methods in `RlpEcdsaEncodableTx` and simply delegate to them in the blanket impl, I've re-added the impl in a10f60b7f

crates/consensus/src/transaction/hashable.rs
```diff
@@ -0,0 +1,25 @@
+use alloy_eips::Typed2718;
+use alloy_primitives::{Signature, TxHash};
+
+use crate::transaction::RlpEcdsaEncodableTx;
+
+/// Generic trait to get a transaction hash from any signature type
+pub trait TxHashable<S>: Typed2718 {
```
mattsse, line 7: ah that makes sense

PR #2800, opened 2025-08-19
feat(providers): pause heartbeat when no transactions are pending

DaniPopes:
Two things to consider:
- We want to stop polling the underlying block number stream when paused
- We want to avoid fetching all the blocks that were included while we're paused

Semantically this should be equivalent to making the block stream `None`, and then re-initializing it to `Some` once we have new transactions.
We don't want to do this since it's more complicated and for e.g. websockets we would need to handshake the subscription again, while I think we want to keep it alive while we're paused.

I don't know how sound this implementation is, so I would like thorough reviews. I've tested it manually, it looks like it works fine on a toy example where I sleep after sending a transaction, then send another one.

Fixes https://github.com/alloy-rs/alloy/issues/1318.

### DIFF
--- a/crates/provider/src/blocks.rs
+++ b/crates/provider/src/blocks.rs
@@ -45,6 +45,9 @@ impl Paused {
     }
 
     /// Waits until the paused state is changed to `false`.
+    ///
+    /// Returns `true` if the method actually waited for the paused state to become unpaused,
+    /// or `false` if it was already unpaused when called.
     async fn wait(&self) -> bool {
         if !self.is_paused() {
             return false;


crates/provider/src/blocks.rs
```diff
@@ -123,13 +133,17 @@ impl<N: Network> NewBlocks<N> {
                 yield known_block;
             }
 
+            // If we're paused, wait until we're unpaused.
+            // Once unpaused, reset `self.next_yield` to ignore the blocks that were included while we were paused.
+            let unpaused = self.check_paused().await;
```
mattsse: if this is paused once, how will this be awoken?

mattsse: this is a bit problematic because we have different stream types

if we stop polling the subscription, we will just accumulate a bunch of messages is buffers

mattsse (changes requested):
this isn't the right fix for this because this doesnt have a wake mechanism and we also need to consider that we need to "pause" the underlying stream, the websocket connection, where we need to still drain incoming messages

I tried this before here by refactoring the stream type

https://github.com/alloy-rs/alloy/pull/2719

so we should move the pause logic inside the impl stream

crates/provider/src/blocks.rs
DaniPopes: right, I would've guessed they would just get discarded after filling the buffer?

DaniPopes: Don't know how it works, i can add a tokio Notify?

mattsse: yeah that should work, or a tokio::watch channel

rn, the poll_fn does not register the task's waker which means that tokio won't poll this task again automatically

crates/provider/src/blocks.rs
```diff
@@ -19,6 +26,35 @@ const MAX_RETRIES: usize = 3;
...
+        }
+    }
+
+    /// Waits until the paused state is changed to `false`.
+    async fn wait(&self) -> bool {
+        if !self.is_paused() {
+            return false;
+        }
+        self.notify.notified().await;
+        debug_assert!(!self.is_paused());
```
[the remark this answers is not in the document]
DaniPopes: the `notified().await` only ever resumes once it's unpaused, so this comment is wrong

crates/provider/src/blocks.rs
prestwich: the oldest ones get discarded and the receivers will get a [lagging error](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html#lagging)

crates/provider/src/blocks.rs
```diff
@@ -123,13 +164,17 @@ impl<N: Network> NewBlocks<N> {
...
+            let unpaused = self.paused.wait().await;
+
             // Get the tip.
             let Some(block_number) = numbers_stream.next().await else {
                 debug!("polling stream ended");
                 break 'task;
             };
             trace!(%block_number, "got block number");
-            if self.next_yield == NO_BLOCK_NUMBER {
+            if self.next_yield == NO_BLOCK_NUMBER || unpaused {
```
mattsse, line 177: right, if we unpause we need to reset the yield tracker

mattsse (approved):
lgtm,

this should finally do it

PR #2804, opened 2025-08-21
fix: alloy-signer Error::message and replace invalid Error::other string usages

Bashmunta:
Reason: String does not implement std::error::Error, so Error::message constructing Self::Other(err.to_string().into()) could not compile. The same issue affected call sites passing &str/String to Error::other, which expects Into<Box<dyn Error + Send + Sync + 'static>>.

Goal: Make message-based error construction compile-safe and consistent across the codebase, while preserving existing API ergonomics.

Changes:
- Added private MessageError in crates/signer/src/error.rs and updated Error::message to wrap messages into Box<dyn Error>.
- Replaced string/format! arguments to Error::other with Error::message:


DaniPopes:
It clearly does compile, so I don't know what you're on about mate

PR #2808, opened 2025-08-21
feat(rpc): add cost() default function to ReceiptResponse trait

### LINKED ISSUE #2807: [Feature] Add fn fn cost to receiptresponse
### Component

rpc

receiptresponse has 

https://github.com/alloy-rs/alloy/blob/ba6777d7c1b42114d0a686c95aeff0d2ac8edf4c/crates/network-primitives/src/traits.rs#L33-L37

if you multiply both, you get the tx cost, so we can add this as default trait fn

# TODO
* add fn cost -> u128

u128 should be sufficient here

### Additional context

_No response_

dharmvr1:
## Motivation
`ReceiptResponse` currently exposes `gas_used()` and `effective_gas_price()`, but computing the total transaction cost requires manual multiplication each time.  
Since this is a common operation, adding a default method improves ergonomics and readability.

## Solution
- Added a default `cost()` method to `ReceiptResponse` that returns:
  ```rust
  self.gas_used() as u128 * self.effective_gas_price()

closes #2807

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -36,6 +36,11 @@ pub trait ReceiptResponse {
     /// Effective gas price.
     fn effective_gas_price(&self) -> u128;
 
+    /// Total cost of this transaction = gas_used * effective_gas_price.
+    fn cost(&self) -> u128 {
+        self.gas_used() as u128 * self.effective_gas_price()
+    }
+
     /// Blob gas used by the eip-4844 transaction.
     fn blob_gas_used(&self) -> Option<u64>;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm

dharmvr1:
Hi @mattsse, I wanted to confirm — should we add a unit test for this kind of helper function? Since it’s just a simple multiplication, it might already be indirectly covered by higher-level tests, but I can add a minimal test if you think it’s useful.

mattsse:
this is fine and also not a breaking change, ty!

PR #2819, opened 2025-08-25
refactor: avoid breaking `next_block_excess_blob_gas` API

klkvr:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -105,7 +105,19 @@ impl BlobParams {
     /// Calculates the `excess_blob_gas` value for the next block based on the current block
     /// `excess_blob_gas` and `blob_gas_used`.
     #[inline]
+    #[deprecated(note = "Use `next_block_excess_blob_gas_osaka` instead")]
     pub const fn next_block_excess_blob_gas(
+        &self,
+        excess_blob_gas: u64,
+        blob_gas_used: u64,
+    ) -> u64 {
+        self.next_block_excess_blob_gas_osaka(excess_blob_gas, blob_gas_used, 0)
+    }
+
+    /// Calculates the `excess_blob_gas` value for the next block based on the current block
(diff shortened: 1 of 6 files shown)


crates/eips/src/eip7840.rs
```diff
@@ -105,10 +105,22 @@ impl BlobParams {
     /// Calculates the `excess_blob_gas` value for the next block based on the current block
     /// `excess_blob_gas` and `blob_gas_used`.
     #[inline]
+    #[deprecated(note = "Use `next_block_excess_blob_gas_osaka` instead")]
```
mattsse: I'd like to mark this deprecated just before the next reth release, otherwise this will introduce a bunch of warnings in existing repos

crates/eips/src/eip7840.rs
```diff
@@ -105,10 +105,22 @@ impl BlobParams {
...
         excess_blob_gas: u64,
         blob_gas_used: u64,
+    ) -> u64 {
+        self.next_block_excess_blob_gas_osaka(excess_blob_gas, blob_gas_used, 0)
+    }
+
+    /// Calculates the `excess_blob_gas` value for the next block based on the current block
+    /// `excess_blob_gas`, `blob_gas_used` and `base_fee_per_gas`.
+    #[inline]
+    pub const fn next_block_excess_blob_gas_osaka(
```
mattsse, line 121: should we name this `next_block_excess_blob_gas_eip7840` then this isnt necessarily restricted to the osaka hardfork?

however osaka is perhaps clearer for the avg user, but the avg doesnt really deal with that anyway

mattsse (changes requested):
smol request and 1 suggestion,

we can change all reth usage without needing the deprecate helper, I'd enable this just before the next reth release

PR #2848, opened 2025-09-04
chore: remove outdated TODO

neowangreal:
The TODO comment on line 257 of `crates/provider/src/layers/batch.rs` states that block ID is not implemented, but the code immediately below (line 258) actually implements the block ID checking correctly:

```rust
if params.block().is_some_and(|block| block != BlockId::latest()) {
    return false;
}
```

This implementation correctly prevents batching when the block ID is not


DaniPopes:
It's not implemented as in it could handle it better

PR #2851, opened 2025-09-07
fix: skip receipt fetch if more confirmations are requested

### LINKED ISSUE #2816: [Bug] get receipt might ignore confirmations
### Component

rpc

_No response_

### Operating System

None

### Describe the bug

we have this fix to prevent race conditions with tx monitoring on short blocktimes:

https://github.com/alloy-rs/alloy/blob/5a215f58052e942160bcc23b4071dc985ddd63f7/crates/provider/src/heart.rs#L243-L247

however this can sidestep the `config.confirmation` setting

we should not do this if confirmation is > 1, then this fix isnt necessary anyway

mattsse:
closes #2816

we dont need this workaround if more than 1 confirmations are required and cant even use this here

### DIFF
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -9,7 +9,7 @@ use alloy_primitives::{
     TxHash, B256,
 };
 use alloy_transport::{utils::Spawnable, TransportError};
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):
lgtm

PR #2854, opened 2025-09-08
feat(rpc-client): add connect convenience method

mattsse:
## Summary

Adds a new convenience method `connect` to `BuiltInConnectionString` that combines parsing a connection string and connecting to it in a single async call.

## Changes

- Added `connect` async method that:
  - Takes a connection string as `&str`
  - Parses it using `from_str`
  - Immediately connects using `connect_boxed`
  - Returns `Result<BoxTransport, TransportError>`

This simplifies the common pattern:
```rust
// Before
BuiltInConnectionString::from_str(s)?.connect_boxed().await?

// After
BuiltInConnectionString::connect(s).await?
```

### DIFF
--- a/crates/rpc-client/src/builtin.rs
+++ b/crates/rpc-client/src/builtin.rs
@@ -45,6 +45,25 @@ impl TransportConnect for BuiltInConnectionString {
 }
 
 impl BuiltInConnectionString {
+    /// Parse a connection string and connect to it in one go.
+    ///
+    /// This is a convenience method that combines `from_str` and `connect_boxed`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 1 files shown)


crates/rpc-client/src/builtin.rs
```diff
@@ -45,6 +45,25 @@ impl TransportConnect for BuiltInConnectionString {
...
+    ///
+    /// ```
+    /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// use alloy_rpc_client::BuiltInConnectionString;
+    ///
+    /// let transport = BuiltInConnectionString::parse_and_connect("http://localhost:8545").await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn parse_and_connect(s: &str) -> Result<BoxTransport, TransportError> {
```
DaniPopes: how about just fn connect

DaniPopes (approved):

PR #2856, opened 2025-09-09
Avoid panic in Drop for GethInstance; use best-effort shutdown

sashaodessa:
### Description
- Replace expect-based kill in Drop with a best-effort termination flow.
- Check process state via try_wait(); kill and wait only if still running.
- Ignore errors in Drop to prevent double-panic and aborts.

#### Why
- Panicking in Drop is unsafe, especially during unwinding, and can abort the process.
- Improves robustness of test and library usage.

#### Changes
- Update `crates/node-bindings/src/nodes/geth.rs`:
  - Use `try_wait()` → conditional `kill()` and `wait()`.
  - Avoid panics in the destructor.

### DIFF
--- a/crates/node-bindings/src/nodes/geth.rs
+++ b/crates/node-bindings/src/nodes/geth.rs
@@ -169,7 +169,21 @@ impl GethInstance {
 
 impl Drop for GethInstance {
     fn drop(&mut self) {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
I think we can keep using regular kill here

PR #2858, opened 2025-09-09
fix(serde): handle overflow in quantity deserialization for Arbitrum gas prices

### LINKED ISSUE #2842: [Bug] Arbitrum Gas Price Deserialize Faliure
### Component

serde

_No response_

### Operating System

None

### Describe the bug

# Bug

When Arbitrum One was running under the Arbitrum Classic stack (before Nitro), the gas price was an unbounded bid, so when requesting those transactions via RPC, you may obtain a very high amount in the `gasPrice` property.

More info: https://docs.arbitrum.io/for-devs/troubleshooting-building#why-do-some-old-transactions-have-extremely-high-gas-prices-when-querying-them

The issue is this means that `block` requests for many earlier arbitrum blocks literally will not deserialize, making indexing arbitrum blocks gracefully with alloy-rs `AnyRpcBlock` or `AnyRpcTransaction` impossible without additional custom logic.

Ideally we could fix this in the `AnyRpcTransaction` deserialization level such that any number like this that wildly overflows is trimmed back down to either a flat `0`, or we convert the pure `u128`  into an `Option<u128>` to represent the failure state.

_I have attempted to solve this over a few hours but didn't manage to get anywhere beyond adding tests to recreate the bug and I believe narrow it down to the `crates/serde/src/quantity.rs` file

# Example

```
curl -X POST 'https://rpc.poolz.finance/arbitrum' \
  -d '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_getBlockByNumber",
    "params": ["0xfb3988", true]
  }'
```

## Response

You will notice in the response we get a value like this, which appears to be across all major RPC node providers and so not particularly avoidable.

```json
"gasPrice":"0x30783134626639633464372e3333333333333333333333333333333333333333",
```

avorylli:
Fixes deserialization failures when processing Arbitrum Classic blocks with extremely high gas prices that overflow u128.

**Problem:**
- Arbitrum Classic used unbounded gas price bids
- Values like `0x30783134626639633464372e3333333333333333333333333333333333333333` cause u128 overflow
- Deserialization panics instead of gracefully handling overflow

**Solution:**
- Modified `ConvertRuint::from_ruint()` to return 0 on overflow instead of panicking
- Added `overflow_default()` method to trait for configurable overflow handling
- Implemented for all primitive types (u8, u16, u32, u64, u128, bool)

Closes #2842

### DIFF
--- a/crates/consensus/src/transaction/eip2930.rs
+++ b/crates/consensus/src/transaction/eip2930.rs
@@ -28,7 +28,7 @@ pub struct TxEip2930 {
     /// As ethereum circulation is around 120mil eth as of 2022 that is around
     /// 120000000000000000000000000 wei we are safe to use u128 as its max number is:
     /// 340282366920938463463374607431768211455
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
+    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity::gas_price"))]
     pub gas_price: u128,
     /// A scalar value equal to the maximum
     /// amount of gas that should be used in executing


mattsse (commented):
we cant really approach this like this, because we'd need to specialize the actual serde impl manually and cant change the regular quantity fn

jaggad:
> we cant really approach this like this, because we'd need to specialize the actual serde impl manually and cant change the regular quantity fn

if we're aiming for "I think we could make this u128::Max and maybe store it in an otherfield as u256 if it's for AnyBlock", how would you go about approaching it?

mattsse:
the only way to do this really is manual serde impl on the block type

avorylli:
@mattsse 
So I  created a gas_price module in quantity.rs that handles overflow specifically for gas prices

And updated the transaction types to use it and reverted the ConvertRuint changes.

I hope this should fix the Arbitrum issue without breaking anything else.

jaggad:
I think this looks pretty good.

@avorylli @mattsse What do we think is needed to get this over the line, it's a pretty high value addition for our use case right now. If we think it can merge soon I'll wait, otherwise I might try and build from this fork?

jaggad:
@mattsse Is there anything holding up this PR from being merged from your perspective?

desfero:
@mattsse we would also greatly benefit if this PR get's merged and new released pushed 🙏

DaniPopes:
Closed in favor of #3010. This requires too many changes to fix the AI slop and has unrelated changes.

PR #2860, opened 2025-09-10
Avoid panic on TempDir cleanup in node-bindings utils

viktorking7:
- Replace `temp_dir.close().unwrap()` with best-effort `let _ = temp_dir.close();` on Unix.
- Prevents unexpected panics during temp directory cleanup in library code.
- Preserves existing function signatures and all call sites.


mattsse:
test failures unrelated

PR #2864, opened 2025-09-11
feat(signer-local): add mnemonic builder helpers and iterator

mattsse:
Adds convenience methods and iterator support to `MnemonicBuilder` for easier signer generation from mnemonic phrases.

## Changes

- Added `from_phrase_nth` and `try_from_phrase_nth` for creating signers at specific derivation indices
- Added `from_phrase_first` and `try_from_phrase_first` convenience methods for the first signer (index 0)
- Implemented `IntoIterator` for `MnemonicBuilder` to generate multiple signers with incrementing indices
- Added `MnemonicSignerIter` iterator type that yields signers for derivation paths `m/44'/60'/0'/0/{index}`

## Usage

```rust
// Create signer at specific index
let signer = MnemonicBuilder::from_phrase_nth(phrase, 5);

// Create first signer (index 0)
let signer = MnemonicBuilder::from_phrase_first(phrase);

// Iterate over multiple signers
let builder = MnemonicBuilder::default().phrase(phrase);
for signer in builder.into_iter().take(10) {
    let signer = signer?;
    // Use signer...
}
```

The iterator requires `W: Wordlist + Clone` and generates signers deterministically from the same mnemonic phrase.

### DIFF
--- a/crates/signer-local/src/mnemonic.rs
+++ b/crates/signer-local/src/mnemonic.rs
@@ -66,6 +66,43 @@ impl MnemonicBuilder<English> {
     pub fn from_phrase<P: Into<String>>(phrase: P) -> Self {
         Self::english().phrase(phrase)
     }
+
+    /// Creates a new [`MnemonicBuilder`] with the [`English`] wordlist and the given phrase and
+    /// returns the signer with the derivation path ` "m/44'/60'/0'/0/{index}"`
+    pub fn try_from_phrase_nth<P: Into<String>>(
+        phrase: P,
+        index: u32,
+    ) -> Result<PrivateKeySigner, LocalSignerError> {
+        Self::from_phrase(phrase).index(index)?.build()
+    }
+
+    /// Creates a new [`MnemonicBuilder`] with the [`English`] wordlist and the given phrase and
+    /// returns the signer with the derivation path ` "m/44'/60'/0'/0/{index}"`
+    ///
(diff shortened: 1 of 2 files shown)


crates/signer-local/src/mnemonic.rs
```diff
@@ -182,6 +219,36 @@ impl<W: Wordlist> MnemonicBuilder<W> {
...
+    builder: MnemonicBuilder<W>,
+    current_index: u32,
+}
+
+impl<W: Wordlist + Clone> Iterator for MnemonicSignerIter<W> {
+    type Item = Result<PrivateKeySigner, LocalSignerError>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // Use the builder to generate signer at current index
+        let result = self.builder.clone().index(self.current_index).and_then(|b| b.build());
```
DaniPopes, line 234: can avoid cloning and `current_index` field if you add a &mut self method like `increment_index`

```rust
let result = self.builder.build()?;
self.builder.increment()?;
Some(result)
```

mattsse: hmm this is a bit tricky because derivation path can be empty.

this more behaves like a using the default path with new index, this seems easier than handling derivation path as well
for most usecases (testing) this should be sufficient

DaniPopes (approved):

PR #2865, opened 2025-09-11
chore: add reqwest-default-tls feature

mattsse:
the reqwest feature always pulls in openssl which is bad because there's no other way to just enable basic reqwest without default tls


Pana:
There is no doc about this change, eg the doc at this link https://alloy.rs/rpc-providers/http-provider/ is broken now

Issue #2873, opened 2025-09-14
[Bug] Cannot be used on platforms lacking atomic, such as RISC-V


zhang-wenchao:
### Component

consensus, eips, genesis

### What version of Alloy are you on?

alloy-consensus v1.0.30

### Operating System

Other (please specify)

### Describe the bug

```
note: found an item that was configured out
--> /home/life/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/lib.rs:229:9
|
229 | pub mod sync;
| ^^^^
note: the item is gated here
--> /home/life/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/lib.rs:228:1
|
228 | #[cfg(all(not(no_rc), not(no_sync), target_has_atomic = "ptr"))]
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
= note: this error originates in the attribute macro auto_impl::auto_impl (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0433]: failed to resolve: could not find sync in alloc
--> /home/life/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-consensus-1.0.30/src/transaction/rlp.rs:13:1
|
13 | #[auto_impl::auto_impl(&, Arc)]
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ could not find sync in alloc
```

Adding Arc here is unnecessary. If necessary, adding portable-atomic can be controlled through the feature. This makes many embedded platforms and wasm unusable

zhang-wenchao:
#1657 #1395

grandizzy:
@zhang-wenchao thanks, are you still willing to drive the PR to the finish line?

DaniPopes:
https://github.com/alloy-rs/alloy/pull/2874#issuecomment-3456557764

PR #2874, opened 2025-09-14
refactor: remove arc, not available on embedded systems or wasm.

zhang-wenchao:
#2873

## Motivation
Arc should not be added to the underlying library, which makes the underlying library less universal.

## Solution
Can be deleted.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -564,7 +564,6 @@ impl<'a> arbitrary::Arbitrary<'a> for Header {
 }
 
 /// Trait for extracting specific Ethereum block data from a header
-#[auto_impl::auto_impl(&, Arc)]
 pub trait BlockHeader {
     /// Extracts essential information into one container type.
     fn header_info(&self) -> HeaderInfo {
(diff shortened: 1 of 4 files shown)


zhang-wenchao:
> this is alloc, and should be wasm compat,
> 
> where are you running into this
> 
> we cant easily remove this because this is a breaking change

`riscv32i-unknown-none-elf` target and other no atomic platform.

Perhaps I can submit a version that avoids including Arc when the platform doesn't support atomic.

This makes it easy to check `#[cfg(target_has_atomic = "ptr")]` .

What do you think of this modification?

DaniPopes:
You can `cfg_attr(target_has_atomic = "ptr", ...)`

zhang-wenchao:
I've run into a problem.

`#[auto_impl(keep_default_for(&, Arc))]`

This can't be written as 

`#[cfg_attr(target_has_atomic = "ptr",auto_impl(keep_default_for(&, Arc)))]`

DaniPopes:
You can split the macro into 2 calls, one for & and cfg attr for Arc

DaniPopes:
We can't actually do this because its a breaking change and we can't conditionally disable arc

PR #2878, opened 2025-09-15
Avoid panic in AnvilInstance::drop; fail soft on kill error

vtjl10:
### Summary
Make teardown more robust by preventing a panic in `Drop` when killing the Anvil process fails.

### Changes
- Replace `self.child.kill().expect("could not kill anvil")` with a fallible `kill()` and soft error logging via `eprintln!`.

### Rationale
- Panics in `Drop` can abort the process during unwinding and complicate shutdown. Soft-failing improves reliability and debuggability.

### DIFF
--- a/crates/node-bindings/src/nodes/anvil.rs
+++ b/crates/node-bindings/src/nodes/anvil.rs
@@ -133,7 +133,9 @@ impl Drop for AnvilInstance {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


mattsse (approved):
makes sense

vtjl10:
> makes sense

@mattsse all checks have passed

PR #2879, opened 2025-09-15
Replace redundant ok().unwrap() with explicit invariant in ConvertRuint

maradini77:
Clarify intent and remove redundancy in `quantity.rs` conversions.

- Replace `try_into().ok().unwrap()` with `unwrap_or_else(|_| unreachable!(...))` in:
  - `ConvertRuint::into_ruint`
  - `ConvertRuint::from_ruint`

- Eliminates unnecessary `Option` hop and improves diagnostics.
- Communicates the invariant without requiring a `Debug` bound on the error (which `expect` would need).

### DIFF
--- a/crates/serde/src/quantity.rs
+++ b/crates/serde/src/quantity.rs
@@ -314,12 +314,12 @@ mod private {
 
         #[inline]
         fn into_ruint(self) -> Self::Ruint {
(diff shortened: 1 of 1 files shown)


mattsse (commented):
the unwrap is fine here

PR #2880, opened 2025-09-15
fix(signer-aws): return structured error instead of panic on parity recovery failure

Snezhkko:
- Replace panic! in parity recovery with SignatureRecoveryFailed error in GCP and AWS signers.
- Update helper to return Result and propagate through sign_digest_inner.
- This change aligns with the project’s error-handling conventions (cryptographic ops return Result), prevents unexpected process crashes on operational issues (mismatched key, invalid digest, malformed signature), and keeps APIs consistent across signers.


mattsse (approved):
this is better than panicking

PR #2881, opened 2025-09-15
feat(rpc-trace): add tx_index for traceCall

jsvisa:
## Motivation

Add `txIndex` for the `debug_traceCall` api, ref https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#tracecallconfig and https://github.com/ethereum/go-ethereum/pull/28460

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/geth/mod.rs
+++ b/crates/rpc-types-trace/src/geth/mod.rs
@@ -740,8 +740,8 @@ pub struct GethDebugTracingCallOptions {
     pub block_overrides: Option<BlockOverrides>,
     /// The transaction index to trace within the block.
     /// If set, the state at the given index will be used to trace the call.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub tx_index: Option<U256>,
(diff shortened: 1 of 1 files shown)


crates/rpc-types-trace/src/geth/mod.rs
```diff
@@ -737,12 +738,16 @@ pub struct GethDebugTracingCallOptions {
     /// The block overrides to apply
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub block_overrides: Option<BlockOverrides>,
+    /// The transaction index to trace within the block.
+    /// If set, the state at the given index will be used to trace the call.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tx_index: Option<U256>,
```
mattsse: can we make this an u64 with quantity::opt?

jsvisa: good point

mattsse (approved):
lgtm

I'm blocking this a bit because would cause issues with current reth versions

PR #2884, opened 2025-09-16
feat(provider): eth_getLogs with batch and count options

wetkeyboard:
Ref https://github.com/foundry-rs/foundry/pull/11638#pullrequestreview-3224941067


crates/provider/src/provider/trait.rs
```diff
@@ -67,6 +67,40 @@ pub type FilterPollerBuilder<R> = PollerBuilder<(U256,), Vec<R>>;
...
 /// [`TransactionBuilder`]: alloy_network::TransactionBuilder
 /// [`DebugApi`]: crate::ext::DebugApi
+
+#[derive(Clone, Debug, Default)]
+pub struct LogOptions {
+    /// If provided, fetches logs in batches of this size when querying a block range.
+    pub batch_size: Option<u64>,
+    /// If provided, stops fetching when this many logs have been collected.
+    pub max_count: Option<usize>,
+}
```
mattsse: this doesn't really belong here

can we move all of this to logs.rs

crates/provider/src/provider/trait.rs
```diff
@@ -664,6 +698,84 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
     }
 
+    /// Enhanced log retrieval with batch and count options.
+    async fn get_logs_with_options(
+        &self,
+        filter: &Filter,
+        options: &LogOptions,
+    ) -> TransportResult<Vec<Log>> {
+        // Validate options
+        if options.batch_size.is_some_and(|x| x == 0) {
```
mattsse: all of this is way to verbose

what we should try instead is have a dedicated call type, similar to what we do for call

https://github.com/alloy-rs/alloy/blob/20de20c78285fd0eb832161caa264031f45e9811/crates/provider/src/provider/trait.rs#L191-L191

and then have the logic for streaming, batching in there

mattsse (changes requested):
not opposed to this, but let's try this as a dedicated future type instead

wetkeyboard:
Updated, please take another look @mattsse , now the usage like below:

```
  // Simple log retrieval (unchanged)
  let logs = provider.get_logs(&filter).await?;

  // New enhanced API with batch support
  // Basic usage (equivalent to get_logs)
  let logs = provider.logs(filter).await?;

  // With batching (fetch logs in chunks of 1000 blocks)
  let logs = provider.logs(filter).with_batch_size(1000).await?;

  // With count limit (stop after 500 logs)
  let logs = provider.logs(filter).with_max_count(500).await?;

  // Combined options
  let logs = provider.logs(filter)
      .with_batch_size(1000)
      .with_max_count(500)
      .await?;
```

0xferrous:
can we get this going again?

Issue #2885, opened 2025-09-16
[Bug] Build fails when with serde version


nakul1010:
### Component

consensus, eips, genesis

### What version of Alloy are you on?

1.0.31

### Operating System

None

### Describe the bug

My `Cargo.toml`

```toml
alloy = { version = "1.0.31", features = [
    "full",
    "node-bindings",
    "rpc-types-debug",
    "rpc-types-trace",
    "json-rpc",
    "rpc-client",
    "signer-aws",
    "signer-keystore",
    "signer-ledger",
    "signer-mnemonic",
    "signer-trezor",
    "signer-yubihsm",
    "eips",
    "sol-types",
    "contract",
    "serde",
] }

serde = "1.0.223"
serde_json = "1.0"
```

Error when running `cargo build`

```bash
error[E0433]: failed to resolve: could not find `__private` in `serde`
   --> /Users/nakul/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-consensus-1.0.31/src/transaction/envelope.rs:162:24
    |
162 | #[derive(Clone, Debug, TransactionEnvelope)]
    |                        ^^^^^^^^^^^^^^^^^^^ could not find `__private` in `serde`
    |
    = note: this error originates in the derive macro `TransactionEnvelope` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Already tried `cargo clean &&  rm Cargo.lock  && cargo build`

Ev3rlasting:
I am having the same issue. 
```
    Checking alloy-consensus v1.0.31
error[E0433]: failed to resolve: could not find `__private` in `serde`
   --> /Users/ray/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/alloy-consensus-1.0.31/src/transaction/envelope.rs:162:24
```

nakul1010:
To compile the code, I needed to pin specific versions: 
```toml
alloy = { version = "=1.0.31", features = [
    "full",
    "node-bindings",
    "rpc-types-debug",
    "rpc-types-trace",
    "json-rpc",
    "rpc-client",
    "signer-aws",
    "signer-keystore",
    "signer-ledger",
    "signer-mnemonic",
    "signer-trezor",
    "signer-yubihsm",
    "eips",
    "sol-types",
    "contract",
    "serde",
] }
serde = "=1.0.223"
serde_json = "=1.0.145"
```

kallydev:
[...]

PR #2886, opened 2025-09-16
fix: use serde private mod patch notation

grandizzy:
## Motivation
- ref https://github.com/serde-rs/serde/pull/2980
- closes https://github.com/alloy-rs/alloy/issues/2885

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -137,7 +137,7 @@ tracing-subscriber = "0.3"
 tracing-futures = "0.2"
 
 # serde
-serde = { version = "1.0.223", default-features = false, features = [
+serde = { version = "1.0.225", default-features = false, features = [
     "derive",
     "alloc",
 ] }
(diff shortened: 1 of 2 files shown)


yash-atreya (approved):
lgtm!

context:

> I think it is low cost and decent benefit. If there is any logic in there that someone wants access to, they need to find someone to maintain it officially so that their users do not experience repeated breakage from serde updates.

from https://github.com/serde-rs/serde/pull/2980#issuecomment-3293788179

mattsse (approved):
let's get this out to get unblocked

@klkvr 

I think we could also just remove the manual deserialize impl entirely?

https://github.com/alloy-rs/alloy/blob/main/crates/tx-macros/src/serde.rs#L228-L263

because I believe deriving the deserialize here would generate the same code but with a worse error message, although this would be preferred over having to maintain this

PR #2893, opened 2025-09-17
feat: add pagination support to watch_logs()

### LINKED ISSUE #2889: [Feature] Support pagination in `DynProvider::watch_logs()`
### Component

provider, pubsub

`DynProvider::watch_logs()`([code](https://github.com/alloy-rs/alloy/blob/20de20c78285fd0eb832161caa264031f45e9811/crates/provider/src/provider/erased.rs#L197)) doesn't support pagination at the moment, which might lead to memory allocation problems when fetching events for very huge block ranges. That would be great to introduce the possibility of specifying the page size so the client side shouldn't care about manually implementing it each time.

### Additional context

_No response_

neowangreal:
Fixes #2889

Add `watch_logs_paginated()` method to prevent memory issues when fetching logs from large block ranges.

### Changes
- Add `PaginatedLogsPoller` that fetches logs in configurable page sizes
- Add `Provider::watch_logs_paginated(filter, page_size)` method
- Implement pagination logic that divides block ranges into smaller chunks

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -483,6 +483,14 @@ where
         self.inner.watch_logs(filter).await
     }
 
+    fn watch_logs_paginated(
+        &self,
+        filter: &Filter,
+        page_size: u64,
+    ) -> crate::provider::PaginatedLogsPoller {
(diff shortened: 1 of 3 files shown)


DaniPopes:
Thanks but closing as issue is rejected

PR #2896, opened 2025-09-17
json-rpc: propagate MapAccess::next_key errors in notification deserializer

VolodymyrBg:
Replace while let Ok(Some(key)) = map.next_key() with while let Some(key) = map.next_key()? in crates/json-rpc/src/notification.rs to avoid swallowing key deserialization errors. This improves error diagnostics on malformed input. The rest of the repository already uses the recommended pattern with immediate error propagation, so this change aligns the implementation with the established style.


mattsse (commented):
we want to error here

PR #2901, opened 2025-09-18
chore(`ci`): enable CodeQL

zerosnacks:
## Motivation

This PR introduces CodeQL code scanning initially just focused on Github actions as it is fast to run.

https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning-with-codeql

Results are reported privately in the `security` tab.

## Solution

This workflow was derived from the default workflow example Github provides enhanced with concurrency cancel in progress, updated cron to run daily and allow workflow dispatch. Trigger on cron, pull requests and pushes to master.

This can later be expanded to cover Rust (currently in preview mode & will require custom build / caches to run efficiently).

PR #2903, opened 2025-09-18
chore(`ci`): enable CodeQL as part of `ci.yml`

zerosnacks:
## Motivation

This PR introduces CodeQL code scanning as part of `ci.yml` initially just focused on Github actions as it is fast to run.

https://docs.github.com/en/code-security/code-scanning/introduction-to-code-scanning/about-code-scanning-with-codeql

Results are reported privately in the `security` tab.

## Solution

This workflow was derived from the default workflow example Github provides enhanced with concurrency cancel in progress, updated cron to run daily and allow workflow dispatch. Trigger on cron, pull requests and pushes to master.

This can later be expanded to cover Rust (currently in preview mode & will require custom build / caches to run efficiently).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,6 +216,35 @@ jobs:
   deny:
     uses: ithacaxyz/ci/.github/workflows/deny.yml@main
 
+  codeql:
+    name: Analyze (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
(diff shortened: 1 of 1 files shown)


zerosnacks:
I think this should work and I think it is preferable to integrate it as part of the regular CI flow; at least for libraries. 

Initial setup was done in https://github.com/alloy-rs/alloy/pull/2901#issuecomment-3306420170 hence no new comment here from `github-advanced-security`.

PR #2910, opened 2025-09-21
fix(provider): require context in eth_callMany params

cyborg42:
## Problem

State overrides are ignored when the `context` parameter is missing from `eth_callMany` requests.

## Motivation

When the `context` parameter is missing from `eth_callMany`, the `overrides` parameter gets parsed as the `context` parameter due to positional parameter parsing. This causes the state overrides to be completely ignored, leading to unexpected behavior.

## Solution

- Change `context` parameter type from `Option<StateContext>` to `StateContext` (required)
- This ensures proper parameter positioning and prevents misinterpretation of the overrides

## Backward Compatibility

To maintain API compatibility, the existing `pub const fn context(&self) -> Option<&StateContext>` method is preserved. This can be updated to return `&StateContext` in a future breaking change release.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


cyborg42:
@mattsse fixed in Serialize

crates/provider/src/provider/eth_call/params.rs
```diff
@@ -182,16 +182,21 @@ impl<'req> EthCallManyParams<'req> {
 
 impl serde::Serialize for EthCallManyParams<'_> {
     fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-        let len = if self.overrides().is_some() { 3 } else { 2 };
+        let (len, context, overrides) = match (self.context(), self.overrides()) {
+            (Some(context), Some(overrides)) => (3, Some(context), Some(overrides)),
+            (Some(context), None) => (2, Some(context), None),
+            (None, Some(overrides)) => (3, Some(&StateContext::default()), Some(overrides)),
+            (None, None) => (1, None, None),
+        };
```
mattsse: checks out

mattsse:
this does not compile

cant apply changes manually because opened from main,

recommend to not use the match statement here and instead use is_some to adjust the lend and also for the serialization of the context

crates/provider/src/provider/eth_call/params.rs
cyborg42: > checks out

Oops.. this can compile under rust 1.90, I've already fixed it under rust 1.86 now

cyborg42:
> this does not compile
> 
> cant apply changes manually because opened from main,
> 
> recommend to not use the match statement here and instead use is_some to adjust the lend and also for the serialization of the context

Fixed under rust 1.86 now

PR #2921, opened 2025-09-24
fix(jwt): report parent directory path in try_create_random

sashass1315:
The CreateDir error in JwtSecret::try_create_random used the target file path instead of the parent directory, producing misleading diagnostics (“failed to create dir <file-path>”). Since the error variant and message explicitly refer to a directory, the path should be the actual directory that failed to be created. This change improves error clarity without affecting behavior or APIs.


mattsse (approved):
seems okay

PR #2924, opened 2025-09-24
docs: Document why sequential block fetching is preferred over join_all

neowangreal:
While reviewing the code, I noticed a TODO suggesting` join_all` for concurrent block fetching and thought it could be a good performance optimization. After investigating, I found that while we do use `try_join_all `successfully elsewhere in the codebase (like `hashes_to_blocks` in utils.rs), this streaming context actually requires sequential fetching for proper per-block retry logic, cache size management, and RPC server backpressure control. I've replaced the ambiguous TODO with clear documentation explaining why the current approach is optimal, preventing future developers from going down the same investigation path.

### DIFF
--- a/crates/provider/src/blocks.rs
+++ b/crates/provider/src/blocks.rs
@@ -192,7 +192,11 @@ impl<N: Network> NewBlocks<N> {
             };
 
             // Then try to fill as many blocks as possible.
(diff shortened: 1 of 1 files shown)


DaniPopes:
This is not entirely right, we could have a smarter batching fetcher

PR #2925, opened 2025-09-25
feat: generate TypedTransaction enum in envelope macro

### LINKED ISSUE #2923: [Feature] Generate TypedTransaction enum in envelope macro
### Component

consensus, eips, genesis

it would be nice if we could generate

https://github.com/alloy-rs/alloy/blob/e69183f07ab4a051fb35b0fc3e1d7f54ffd9faa0/crates/consensus/src/transaction/typed.rs#L45-L61

via the envelope macro as well (opt-in)

which is similar to the regular envelope, but likely requires another attribute to support providing the path to the raw tx type `envelop(typed=TxEip772)` for example.

this also needs to support a different serde scheme, also new envelope attribute

https://github.com/alloy-rs/alloy/blob/e69183f07ab4a051fb35b0fc3e1d7f54ffd9faa0/crates/consensus/src/transaction/typed.rs#L35-L42

### Additional context

_No response_

stevencartavia:
## Motivation
closes #2923

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/tx-macros/src/expand.rs
```diff
@@ -681,4 +685,214 @@ impl Expander {
...
+                        if let Some(syn::GenericArgument::Type(inner_ty)) = args.args.first() {
+                            return quote! { #inner_ty };
+                        }
+                    }
+                }
+            }
+        }
+        // Fallback to the original type if we can't extract from Signed<T>
+        quote! { #ty }
+    }
```
klkvr: we should add a way to define a custom inner type here

e.g op deposit tx here https://github.com/alloy-rs/op-alloy/blob/e0db0e171846f72b0c77f24528d72feab594d193/crates/consensus/src/transaction/envelope.rs#L40

something like
```rust
#[envelope(ty = 126, typed = TxDeposit)]
Deposit(Sealed<TxDeposit>),
```

also maybe we could figure out a way to handle flattened envelopes here nicely via a new trait associating an envelope with a typed transaction

crates/tx-macros/src/expand.rs
```diff
@@ -681,4 +685,214 @@ impl Expander {
...
+        let arbitrary_cfg = &self.arbitrary_cfg;
+
+        // Generate variants for typed transaction - extract inner types from Signed wrappers
+        let variants = self.variants.all.iter().map(|v| {
+            let name = &v.name;
+            let serde_attrs = if let Some(attrs) = &v.serde_attrs {
+                quote! { #attrs }
+            } else {
+                quote! {}
+            };
```
klkvr: those serde attrs will likely not work for inner type so we can't use them like this

crates/tx-macros/src/expand.rs
```diff
@@ -681,4 +685,214 @@ impl Expander {
...
+        );
+
+        // Generate serde derives only if serde is enabled
+        let serde_derives = if self.serde_enabled {
+            quote! {
+                #[cfg_attr(#serde_cfg, derive(#alloy_consensus::private::serde::Serialize, #alloy_consensus::private::serde::Deserialize))]
+            }
+        } else {
+            quote! {}
+        };
```
klkvr: instead of simply deriving we want to take a similar approach as here https://github.com/alloy-rs/alloy/blob/9590610d613825a67df35a0ecef6824175b9fb4a/crates/tx-macros/src/serde.rs#L36-L38

and correctly handle untagged legacy tx. resulting impl should be similar to https://github.com/alloy-rs/alloy/blob/9590610d613825a67df35a0ecef6824175b9fb4a/crates/consensus/src/transaction/typed.rs#L605

crates/tx-macros/src/expand.rs
```diff
@@ -681,4 +685,214 @@ impl Expander {
...
+
+        quote! {
+            #[doc = #doc_comment]
+            #[derive(Clone, Debug, PartialEq, Eq, Hash)]
+            #serde_derives
+            #arbitrary_derives
+            pub enum #typed_name<Eip4844> {
```
klkvr: this will always generate a type with EIp4844 which wrong. we should use `self.generics` here

crates/tx-macros/src/expand.rs
stevencartavia: > fully remove EthereumTypedTransaction definition in favor of macro usage
> maybe we could figure out a way to handle flattened envelopes here nicely via a new trait associating an envelope with a typed transaction

not done yet, some progress

klkvr (approved):

PR #2928, opened 2025-09-25
feat: Add `UnsupportedTransactionType` error

RomanHodulak:
## Motivation

Custom nodes can deal with conversions or processing of transactions of a type that they can't deal with.

An error like this is usually defined locally in those projects.

## Solution

Add `UnsupportedTransactionType` error that can convert into `TransactionBuilderError`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
cool

PR #2933, opened 2025-09-26
refactor(beacon-events): use BeaconBlockHeader in light client finality

GarmashAlex:
- Replaced local Beacon and Beacon2 structs in LightClientFinalityData with crate::header::BeaconBlockHeader for both attested and finalized headers.
- Removed the now-unused Beacon and Beacon2 definitions and cleaned an unused B256 import.
- Ensures symmetry with LightClientOptimisticData, which already uses BeaconBlockHeader.
- Aligns with Beacon API schemas where both headers share the same structure.
- Reduces duplication, avoids confusing naming (Beacon2), and improves maintainability and type consistency.

### DIFF
--- a/crates/rpc-types-beacon/src/events/light_client_finality.rs
+++ b/crates/rpc-types-beacon/src/events/light_client_finality.rs
@@ -28,12 +28,18 @@ pub struct AttestedHeader {
     pub beacon: BeaconBlockHeader,
 }
 
+/// Backwards-compatible alias for the previously local `Beacon` header type.
+pub type Beacon = BeaconBlockHeader;
+
 /// Contains the `Beacon2` header that was finalized.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct FinalizedHeader {
     /// The `Beacon2` object representing the block header.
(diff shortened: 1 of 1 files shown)


crates/rpc-types-beacon/src/events/light_client_finality.rs
```diff
@@ -24,50 +25,14 @@ pub struct LightClientFinalityData {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct AttestedHeader {
     /// The `Beacon` object representing the block header.
-    pub beacon: Beacon,
-}
-
-/// Represents the header of a beacon block.
-#[serde_as]
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct Beacon {
```
mattsse, line 33: could we keep aliases for this, so that this change doesn't break?

GarmashAlex: > could we keep aliases for this, so that this change doesn't break?

Yes Boss!

mattsse (approved):

PR #2935, opened 2025-09-26
fix: JWT Token Validation Logic in AuthService

maradini77:
## Bug Description

Critical logical error in JWT token validation that could lead to authentication vulnerabilities.

### 1. **Inverted Validation Logic**
- **Before**: Old tokens were considered valid, new tokens were rejected
- **After**: Tokens are properly validated based on time difference within latency buffer

### 2. **Incorrect Token Creation**
- **Before**: Tokens created with future timestamp (`iat = now + 60s`)
- **After**: Tokens created with current timestamp (`iat = now`)

### 3. **Improved Documentation**
- Added clear English comments explaining validation logic
- Enhanced method documentation for better maintainability

### DIFF
--- a/crates/transport-http/src/layers/auth.rs
+++ b/crates/transport-http/src/layers/auth.rs
@@ -61,11 +61,15 @@ impl<S> AuthService<S> {
 
     /// Validate the token in the request headers.
     ///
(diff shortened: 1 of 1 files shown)


mattsse (approved):
ah yeah this is right

PR #2940, opened 2025-09-27
Add Default and Per-Request HTTP Timeouts to Reqwest Transport

viktorking7:
## Description
- **What**: Introduces 30s timeouts for HTTP requests in the reqwest transport.
- **Why**: Prevents indefinitely hanging requests under network/server issues; improves resilience and resource predictability.
- **Changes**:
  - Default client timeout via `ClientBuilder` in `ReqwestConnect::get_transport`.
  - Per-request timeout safeguard in `Http<Client>::do_reqwest`.


mattsse (commented):
this can be configured already and we shouldnt change default behaviour

PR #2943, opened 2025-09-28
chore(beacon): remove redundant DisplayFromStr on header.block_number

radik878:
- Remove #[serde_as(as = "DisplayFromStr")] from ExecutionPayloadHeader.block_number (type String).
- For String this adapter is a no-op: serde already serializes/deserializes JSON strings natively. Keeping it adds noise and suggests a conversion that does not occur.
- All other DisplayFromStr uses remain on numeric/bignum types to enforce beacon “quoted decimal” semantics.


mattsse (changes requested):
can we change the type to u64 instead?

radik878:
> can we change the type to u64 instead?

changed, thanks for review mattsse

mattsse (approved):

PR #2955, opened 2025-09-29
feat: add size guard for hyper HTTP responses (reject oversized bodies)

maradini77:
### PR Description
- Introduces a 10 MiB upper bound for Hyper-based HTTP responses using `Content-Length`.
- Rationale: Prevents excessive memory usage and potential DoS via extremely large JSON-RPC responses.
- Behavior:
  - If `Content-Length` is present and exceeds the limit, return an error early.
  - If `Content-Length` is absent, behavior is unchanged.
- Scope: Minimal and self-contained in `crates/transport-http/src/hyper_transport.rs`. No API changes.

### DIFF
--- a/crates/transport-http/src/hyper_transport.rs
+++ b/crates/transport-http/src/hyper_transport.rs
@@ -132,7 +132,7 @@ where
                 )));
             }
         }
-        
+
         // Unpack data from the response body. We do this regardless of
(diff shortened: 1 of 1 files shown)


onbjerg (changes requested):
this would already break some existing jsonrpc apis such as traces, as they can get very large

PR #2957, opened 2025-09-30
fix(tx-macros): correct hex alias comment and replace unwrap with safe pattern

rnkrtt:
Fixed misleading comment about hex alias generation direction and replaced unsafe unwrap() with safe if-let pattern in serde.rs. The comment incorrectly stated `0x0 for 0x00` when the code actually generates `0x01 for 0x1` (adding leading zero). Also improved code safety by handling potential None case from chars().last().


mattsse (commented):
this unwrap is fine because we already checked length

PR #2960, opened 2025-09-30
Normalize active transport count in FallbackService to avoid zero-active configuration

viktorking7:
### Description
- **Summary**: Clamp `active_transport_count` to [1, transports.len()] when transports are present; 0 if empty.
- **Why**: Prevents a silent misconfiguration where no requests are dispatched if `active_transport_count` is 0.
- **How**: Normalize in `FallbackService::new` and add a brief comment.


onbjerg (changes requested):
this is incorrect, setting to 0 is perfectly valid.

https://github.com/viktorking7/alloy/blob/dc4d622969249ff81a4eac9a47f6c7f6c06e109c/crates/transport/src/layers/fallback.rs#L84-L94

PR #2961, opened 2025-09-30
fix(beacon): use u64 for gas equality errors

VolodymyrBg:
- Change IncorrectGasLimit.actual and IncorrectGasUsed.actual from B256 to u64, and update variant docs to reference gas limit/used.
- Gas fields are u64 across execution payloads and beacon serde (e.g., rpc-types-engine ExecutionPayloadV1/V2/V3), so using B256 was inconsistent and misleading.
- This prevents downstream type mismatches and makes error reporting accurate.


DaniPopes:
is this used anywhere? its not actually an rpc type @mattsse

PR #2974, opened 2025-10-03
feat: add `transactionReceipts` into SubscriptionKind

10gic:
Add transactionReceipts for receipts notification. For more information, see: https://github.com/ethereum/go-ethereum/pull/32697

### DIFF
--- a/crates/rpc-types-eth/src/pubsub.rs
+++ b/crates/rpc-types-eth/src/pubsub.rs
@@ -1,7 +1,7 @@
 //! Ethereum types for pub-sub
 
-use crate::{Filter, Header, Log, Transaction};
-use alloc::{boxed::Box, format};
+use crate::{Filter, Header, Log, Transaction, TransactionReceipt};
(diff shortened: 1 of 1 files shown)


10gic:
Hi @DaniPopes, @mattsse, @klkvr, hope you're doing well. When you have a chance, could you take a look at this PR? Any feedback would be great. Thanks!

crates/rpc-types-eth/src/pubsub.rs
```diff
@@ -141,6 +191,72 @@ impl From<bool> for Params {
...
+                                        hash_vec.len(),
+                                        MAX_TX_HASHES
+                                    )));
+                                }
+                            }
+
+                            transaction_hashes = Some(hashes);
+                        }
+                        _ => {
+                            map.next_value::<serde_json::Value>()?;
```
mattsse: why is this using serde value here?

10gic: Now, I just return an error if there are unknown fields.

PR #2976, opened 2025-10-03
provider/heart: prevent underflow in confirmation height calculation

sashaodessa:
This change replaces the `u64` computation `block_height + confirmations - 1` with `block_height.saturating_add(confirmations.saturating_sub(1))` in `crates/provider/src/heart.rs` to prevent underflow when `confirmations == 0`. The fix preserves existing behavior for `confirmations >= 1`, eliminates potential panics in debug builds and silent wraparound in release, and improves robustness of confirmation scheduling.


onbjerg (changes requested):
In practice this can never underflow because you cannot send a transaction *at* the genesis block, so the block height will always be at least 1. In the case that you send a tx in the first block, if `required_confmrations` is 0, `confirmed_at` would be `1 - 0 - 1 = 0` which is fine.

PR #2996, opened 2025-10-07
fix: Use injected #alloy_rlp path in generated typed_decode error mapping

GarmashAlex:
Replaced bare reference to alloy_rlp::Error::Custom with the injected #alloy_rlp::Error::Custom in the TransactionEnvelope derive expansion. This aligns with how other paths are injected and prevents potential resolution issues for downstream crates that don’t directly depend on alloy-rlp under the alloy_rlp name. Ensures consistency and robustness of the generated code across varying dependency configurations.


klkvr (approved):
good catch, ty

PR #2998, opened 2025-10-07
perf(serde): avoid allocation in B256 hex serialization

prestoalvarez:
Description
- What: Replace serialize_str(&format!("{x:x}")) with collect_str(&format_args!("{x:x}")) in crates/serde/src/lib.rs.
- Why: Eliminates an intermediate String allocation on a hot serialization path, improving performance and reducing allocator pressure.
- How: Use Serializer::collect_str with format_args! to write directly to the serializer without allocating.

### DIFF
--- a/crates/serde/src/lib.rs
+++ b/crates/serde/src/lib.rs
@@ -8,6 +8,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
+
 use alloy_primitives::{hex, B256};
 use serde::Serializer;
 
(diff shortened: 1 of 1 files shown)


crates/serde/src/lib.rs
```diff
@@ -8,8 +8,6 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
```
DaniPopes, line 10:
```suggestion
extern crate alloc;

```

PR #2999, opened 2025-10-07
refactor(network/any): remove redundant From<&AnyTxType> for u8 impl

prestoalvarez:
Description
- What: Drop the duplicate conversion impl `From<&AnyTxType> for u8` from crates/network/src/any/unknowns.rs.
- Why: `AnyTxType` is Copy; having both by-value and by-ref conversions is redundant and bloats the API surface without added semantics.


klkvr:
this conversion is fine and that'd be a breaking change

PR #3003, opened 2025-10-07
refactor(node-bindings): remove redundant bounds checks in extract_value

kilavvy:
## Changes
- Removed unnecessary `start <= line.len() && end <= line.len()` bounds checks in `extract_value` utility function
- Simplified control flow by directly returning the trimmed slice

## Motivation
The bounds checks are redundant because:
- `start` is always valid (derived from `find()` position + key length)
- `end` is always valid (via `find().map().unwrap_or(line.len())`)


DaniPopes:
Technically you're right but this doesn't matter at all

PR #3004, opened 2025-10-07
refactor(transport): avoid variable shadowing in RetryBackoffService::call

kilavvy:
## Changes
- Renamed local `inner` clone to `new_inner` in `RetryBackoffService::call` to avoid shadowing the replaced `inner`.

## Motivation
- The previous naming reused `inner` for two different values (clone vs replaced), which hindered readability and reasoning about the code.


onbjerg (changes requested):
this doesnt matter

PR #3006, opened 2025-10-08
chore(serde): use explicit alloc::format! and remove unused import

reject-i:
Remove unused format import and replace format! with alloc::format! for consistency with storage.rs and better no_std explicitness


klkvr:
it's fine as is

PR #3009, opened 2025-10-08
fix(network): remove Default from EthereumWallet; construct valid state in new()

Galoretka:
Problem: EthereumWallet implemented Default, producing an empty wallet with a zero default address. Calling default_signer() on such a value panicked, violating documented expectations that a wallet always has at least one signer and a valid default.

Changes:
- Dropped Default derive from EthereumWallet to prevent invalid construction.
- Refactored EthereumWallet::new to initialize signers and default directly, ensuring a valid state from the start.
- Documented the invariant (wallet always has ≥1 signer, default must be present).
- Added a debug_assert in default_signer() to enforce the invariant in debug builds.

Rationale: Aligns implementation with docs and prevents panics arising from invalid default construction while preserving public API behavior for valid wallets.

### DIFF
--- a/crates/network/src/ethereum/wallet.rs
+++ b/crates/network/src/ethereum/wallet.rs
@@ -6,7 +6,10 @@ use std::{fmt::Debug, sync::Arc};
 use super::Ethereum;
 
 /// A wallet capable of signing any transaction for the Ethereum network.
-#[derive(Clone, Default)]
+///
+/// Invariant: this wallet always contains at least one signing credential and the
+/// `default` address MUST be present in `signers`.
+#[derive(Clone)]
 pub struct EthereumWallet {
     default: Address,
     signers: AddressHashMap<Arc<dyn TxSigner<Signature> + Send + Sync>>,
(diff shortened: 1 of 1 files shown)


crates/network/src/ethereum/wallet.rs
```diff
@@ -36,9 +39,12 @@ impl EthereumWallet {
...
     {
-        let mut this = Self::default();
-        this.register_default_signer(signer);
-        this
+        let mut signers: AddressHashMap<Arc<dyn TxSigner<Signature> + Send + Sync>> =
+            AddressHashMap::default();
+        let address = signer.address();
+        signers.insert(address, Arc::new(signer));
+
+        Self { default: address, signers }
```
DaniPopes, line 47: this change achieves nothing

crates/network/src/ethereum/wallet.rs
```diff
@@ -6,7 +6,10 @@ use std::{fmt::Debug, sync::Arc};
 use super::Ethereum;
 
 /// A wallet capable of signing any transaction for the Ethereum network.
-#[derive(Clone, Default)]
+///
+/// Invariant: this wallet always contains at least one signing credential and the
+/// `default` address MUST be present in `signers`.
+#[derive(Clone)]
```
DaniPopes, line 12: this is not a public facing invariant that matters. it's also technically valid for one to call default and register_default_signer manually.

crates/network/src/ethereum/wallet.rs
```diff
@@ -91,6 +97,7 @@ impl EthereumWallet {
 
     /// Get the default signer.
     pub fn default_signer(&self) -> Arc<dyn TxSigner<Signature> + Send + Sync + 'static> {
+        debug_assert!(self.signers.contains_key(&self.default), "default signer must exist");
```
DaniPopes, line 100: this doesn't do anything more than the panic when the `get` returns `None`

PR #3014, opened 2025-10-09
chore(consensus): remove dummy once_cell import from lib.rs

viktorking7:
### Motivation
Remove a dummy import used solely to silence unused dependency warnings. This improves build hygiene and makes dependency usage explicit.


klkvr:
it's fine

PR #3021, opened 2025-10-12
perf(rpc-types-engine): avoid allocation in serde error formatting

anim001k:
## Summary

Optimize error formatting in `BlobsBundleV1` deserialization to avoid unnecessary string allocation.

## Changes

- Replace `format!("{}", raw.commitments.len()).as_str()` with `format_args!("{}", raw.commitments.len())` in `serde::de::Error::invalid_length`
- Eliminates temporary `String` allocation in deserialization path


DaniPopes:
this is not necessary

PR #3028, opened 2025-10-14
feat(provider): add subscribe_noparams helper and use in admin

### LINKED ISSUE #3027: [Feature] Add a helper subscribe function to trait
### Component

provider, pubsub

creating subscriptions is currently a big weird ref https://github.com/alloy-rs/alloy/issues/2983

https://github.com/alloy-rs/alloy/blob/4c3db1395784ba178077733a76db52841870a9f4/crates/provider/src/ext/admin.rs#L75-L82

we should add a way to do this in one step by introducing a helper trait fn that does this

## TODO
* add helper function to Provider trait for this
* add an example to alloy-examples

### Additional context

_No response_

rnkrtt:
Add Provider::subscribe_noparams for non-standard, no-params subscriptions and refactor AdminApi::subscribe_peer_events to use it. 

Fixes #3027

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -1160,7 +1160,7 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
     /// # async fn example(provider: impl alloy_provider::Provider) -> Result<(), Box<dyn std::error::Error>> {
     /// use futures::StreamExt;
     ///
-    /// let sub = provider.subscribe_noparams::<alloy_rpc_types_admin::PeerEvent>("admin_peerEvents").await?;
+    /// let sub = provider.subscribe_to::<alloy_rpc_types_admin::PeerEvent>("admin_peerEvents").await?;
     /// let mut stream = sub.into_stream().take(5);
     /// while let Some(event) = stream.next().await {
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -1148,6 +1148,38 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// let mut stream = sub.into_stream().take(5);
+    /// while let Some(event) = stream.next().await {
+    ///    println!("peer event: {event:#?}");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "pubsub")]
+    #[auto_impl(keep_default_for(&, &mut, Rc, Arc, Box))]
+    fn subscribe_noparams<R>(&self, method: &'static str) -> GetSubscription<NoParams, R>
```
mattsse: I wonder if we can come up with a better name for this because I find `subscribe_noparams` a bit confusing

subscribe is already taken


maybe just `subscribe_to`?

rnkrtt:
> pedantic naming nit

Renamed to `subscribe_to` - much better name, thanks!

PR #3030, opened 2025-10-15
fix(provider): correct EthCallParams serialize length when no options set

sashass1315:
Previously reported seq length was 2 while emitting only 1 element when no block/overrides/block_overrides were set.
Now computes length as:
- 4 when block_overrides present
- 3 when overrides present
- 2 when only block present
- 1 when none present

Aligns with existing tests and avoids potential serde sequence length mismatches with non-JSON serializers.


mattsse (approved):
checks out

PR #3036, opened 2025-10-16
feat: add methods to build EIP-7594 sidecars with default and custom settings

frankudoags:
## Motivation

sidecar builder was lacking a method to build for 7594 with default kzg settings

## Solution

- new fn to build with default kzg(mainnet) moved old functionality to a helper fn that accepts settings
- mark old fn as deprecated

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/eips/src/eip4844/builder.rs
mattsse: let's self self.build_7594_with_settings here so that we dont have duplicated code

mattsse (changes requested):
smol nit

mattsse (approved):
sweet

PR #3039, opened 2025-10-17
fix(signer): propagate semver parse errors instead of unwrap

GarmashAlex:
Replace unwrap() with error propagation when parsing semver VersionReq in both the Ledger and Trezor signers so malformed version requirement strings no longer cause panics and instead flow through the existing error variants (LedgerError::SemVerError and TrezorError::Semver); this removes dead error variants, improves robustness, and maintains backward-compatible API behavior while aligning error handling across both signers.


mattsse (approved):
sure why not

PR #3045, opened 2025-10-18
Remove redundant clone in fallback transport

MamunC0der:
Simplified FallbackService::make_request by dropping the extra clone() on ScoredTransport, avoiding needless copies with no behavioral change. Verified via cargo check --package alloy-transport.


mattsse (approved):
I see, we already cloned this one earlier

PR #3049, opened 2025-10-19
fix: add subscribe id into subscribe_block

fenghaojiang:
## Motivation

The original result from subscribe block is missing subscription id so we cannot unsubscribe it. 

## Solution

Adding the subscription id


mattsse (commented):
this is breaking unfortunately, the better approach here would be to specialize the stream type so that this can be accesses

PR #3050, opened 2025-10-19
node-bindings(geth): remove duplicate --miner.etherbase arg in Clique mode

kilavvy:
### Description
- Summary: Remove a duplicated CLI argument passed to geth when Clique is enabled.
- Motivation: The same `--miner.etherbase` flag was added twice, which is redundant and can confuse readers and tooling.
- Change: Keep a single `--miner.etherbase` assignment.

### DIFF
--- a/crates/node-bindings/src/nodes/geth.rs
+++ b/crates/node-bindings/src/nodes/geth.rs
@@ -487,7 +487,6 @@ impl Geth {
                 let extra_data_bytes =
                     [&[0u8; 32][..], clique_addr.as_ref(), &[0u8; 65][..]].concat();
                 genesis.extra_data = extra_data_bytes.into();
-                // extra: etherbase is set once below after clique address resolution
             }
 
             let clique_addr = self.clique_address().ok_or_else(|| {


onbjerg (changes requested):
this just changes a comment, what is the actual change here?

zerosnacks:
Closing, PR has no relevant changes

Feel free to leave a comment if this is inaccurate

kilavvy:
> Closing, PR has no relevant changes
> 
> Feel free to leave a comment if this is inaccurate

@zerosnacks Not just a comment change. In Clique mode we were appending `--miner.etherbase` twice: once inside the genesis initialization branch and again a few lines below after resolving the Clique address. This PR removes the first occurrence and keeps the latter (single source of truth), adding a clarifying comment. There’s no behavioral change when both values are the same, but it removes a redundant duplicate flag and avoids future divergence or ambiguity.

zerosnacks:
@kilavvy can you point me to the exact lines? I don't see it being appended twice.

kilavvy:
> @kilavvy can you point me to the exact lines? I don't see it being appended twice.

@zerosnacks It is appended twice in Clique mode. First inside at lines 491–495, and again at lines 508–511 after resolving the Clique address. This PR removes the first occurrence and keeps the latter.

onbjerg:
The PR only has 3 removed lines and they are all comments, please recheck if you pushed the change you intended

kilavvy:
> The PR only has 3 removed lines and they are all comments, please recheck if you pushed the change you intended

@onbjerg Could you please reopen the PR so I can push the intended fix? I realized the change to remove the duplicate `--miner.etherbase` append didn’t make it into the previous diff. I’ve updated the code locally and will push as soon as the PR is reopened.

kilavvy:
@onbjerg updated

crates/node-bindings/src/nodes/geth.rs
```diff
@@ -487,11 +487,7 @@ impl Geth {
                 let extra_data_bytes =
                     [&[0u8; 32][..], clique_addr.as_ref(), &[0u8; 65][..]].concat();
                 genesis.extra_data = extra_data_bytes.into();
-
-                // we must set the etherbase if using clique
-                // need to use format! / Debug here because the Address Display impl doesn't show
-                // the entire address
-                cmd.arg("--miner.etherbase").arg(format!("{clique_addr:?}"));
+                // extra: etherbase is set once below after clique address resolution
```
onbjerg:
```suggestion
```

DaniPopes:
Thanks but clique has been deprecated since Geth v1.14 (Apr 24, 2024). How did you find this / what relevance does this have today?

zerosnacks:
> First inside at lines 491–495, and again at lines 508–511 after resolving the Clique address.

I see, for context: https://github.com/alloy-rs/alloy/blob/4817fc3d3f6dba9c70164a385af87b63ced8621d/crates/node-bindings/src/nodes/geth.rs#L491-L511

mattsse (approved):
I see, since was indeed supplied twice

PR #3052, opened 2025-10-19
chore: enable debug_code_by_hash test

neowangreal:
reth v1.8.2 supports this now, removing the TODO


neowangreal:
Original test expected mainnet contract data on empty dev node. Fixed by testing with empty code hash and non-existent hashes to verify API availability without needing contract deployment.

PR #3063, opened 2025-10-23
Feat/ens functions

### LINKED ISSUE #3076: [Feature] Ens Avatar Resolution
### Component

other

ENS has a provided a standard in its avatar resolution [ENSIP-12](https://docs.ens.domains/ensip/12/). In short sometimes avatar can be ERC721 / ERC1155, or an image URL. Clients like viem, already implement this. Hence it would be good if alloy also covers this.

### Additional context

End result i imagine is a function in the ens crates called ```lookup_avatar(name: &str,gateways: Option<HashMap<String,String>) -> Result<String,EnsError>``` , it returns a image url to the avatar, and maps it to the custom gateway url if passed.

dipanshuhappy:
## Motivation

This pull request resolves #3076. In short, avatar images needs to be resolved in a certain way.

## Solution

Essentially the algorithm is:
1. Get avatar value from txt record fetch
2. If avatar is given in context of nft uri then 
    a. Resolve NFT URI 
    b. Resolve NFT Image with custom gateways url passed for ipfs or arweave
    c. Return final image url
3. Else image url is given then resolve with custom gateways passed for ipfs or arweave.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       # Only run tests on latest stable and above
       - name: Install cargo-nextest
         if: ${{ matrix.rust != '1.86' }} # MSRV
-        uses: taiki-e/install-action@e7ef886cf8f69c25ecef6bbc2858a42e273496ec # v2
+        uses: taiki-e/install-action@e43a5023a747770bfcb71ae048541a681714b951 # v2
         with:
           tool: nextest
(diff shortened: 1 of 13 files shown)


grandizzy:
thank you, per the comment in ticket, we won't be supporting this feature in alloy

PR #3068, opened 2025-10-23
fix: bpo config parsing

theochap:
## Description

In `alloy-genesis`, the bpo hardforks are matched by incorrect literal strings which broke the computation of the `blob_schedule_blob_params`. This PR fixes that by ensuring we're matching against the names provided by the `alloy-hardforks` crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,7 @@ alloy-eip2930 = { version = "0.2.0", default-features = false }
 alloy-eip7702 = { version = "0.6.1", default-features = false }
 
 # hardforks
-alloy-hardforks = "0.2.0"
+alloy-hardforks = "0.3.5"
 
 # ethereum
 ethereum_ssz_derive = "0.9"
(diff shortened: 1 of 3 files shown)


theochap:
Doing `alloy_hardforks::EthereumHardfork::Bpo1.name().to_string()` yields `Bpo1` and not `bpo1` in `alloy-hardforks v0.3.5`

klkvr:
yep but expected value in the config is `bpo1`, right? based on https://eips.ethereum.org/EIPS/eip-7892

so comparing it with `Bpo1` seems wrong?

theochap:
> yep but expected value in the config is `bpo1`, right? based on https://eips.ethereum.org/EIPS/eip-7892
> 
> so comparing it with `Bpo1` seems wrong?

Agreed, I'll fix that in the config then. Thanks for pointing it out. Maybe it'd be worth unifying the naming conventions because I'd expect `alloy_hardforks::EthereumHardfork::Bpo1.name()` to yield the correct name for the blob schedule config.

theochap:
Closing this one given the discussion above

PR #3069, opened 2025-10-23
fix(transport): decrement request counter on retry limit exceeded

maradini77:
The `RetryBackoffService` has a counter leak when retry attempts are exhausted. When `rate_limit_retry_number` exceeds `max_rate_limit_retries`, the service returns an error without decrementing the `requests_enqueued` counter. This causes the queue size metric to become inaccurate, leading to incorrect backoff calculations in subsequent requests.

Added `fetch_sub(1, Ordering::SeqCst)` before the early return when retry limit is exceeded. This ensures the counter is properly decremented in all error paths, maintaining accurate queue metrics and preventing excessive backoff delays.


mattsse (approved):
ah cool, very nice find

PR #3070, opened 2025-10-23
refactor(json-rpc): improve `RequestPacket::is_empty` implementation

anim001k:
## Description

The current `RequestPacket::is_empty` implementation uses `self.len() == 0` which is less idiomatic than using pattern matching and `Vec::is_empty()` directly.

## Solution

- Replace `self.len() == 0` with pattern matching on enum variants
- Use `batch.is_empty()` for `Batch` variant instead of computing length
- Maintain identical behavior while improving code clarity and following Rust idioms


onbjerg (changes requested):
this is basically the same thing

PR #3073, opened 2025-10-26
docs(local-signer): add `with_chain_id` to match crate docs example

gap-editor:
### description

The crate-level docs demonstrate using `with_chain_id(Some(...))` on `LocalSigner`, but the method didn’t exist on the type. This PR adds a minimal builder-style helper:

```rust
pub fn with_chain_id(mut self, chain_id: Option<ChainId>) -> Self
```

It aligns the API with the docs, improves ergonomics, and doesn’t alter existing behavior (the mutable `set_chain_id` remains available).


mattsse (commented):
we don't need this because this is already a trait fn

PR #3077, opened 2025-10-27
refactor(rpc-types-eth): remove duplicate block range filtering logic

prestoalvarez:
## Description

Remove code duplication in `FilteredParams::filter_block_range` by delegating to the existing `Filter::matches_block_range` method.

## Changes

- Replace manual block range validation logic in `FilteredParams::filter_block_range` with delegation to `Filter::matches_block_range`
- Maintain identical behavior while reducing code duplication


mattsse (approved):
this is indeed duplicated

PR #3085, opened 2025-10-27
feat: implement CallBuilder::legacy() to prefer legacy by setting gas_price

VolodymyrBg:
Implement CallBuilder::legacy() so that it prefers building legacy transactions without breaking abstractions. The method sets gas_price when it is not present using max_fee_per_gas as a reasonable fallback and returns self, letting the underlying transaction request handle type selection and conflict trimming. This keeps behavior consistent with preferred_type and provider gas filling logic and respects EIP-2930, 4844, and 7702 semantics.

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -424,16 +424,6 @@ impl<P: Provider<N>, D: CallDecoder, N: Network> CallBuilder<P, D, N> {
         self
     }
 
-    /// Uses a Legacy transaction instead of an EIP-1559 one to execute the call
-    pub fn legacy(mut self) -> Self {
-        if self.request.gas_price().is_none() {
-            if let Some(max_fee) = self.request.max_fee_per_gas() {
-                self.request.set_gas_price(max_fee);
-            }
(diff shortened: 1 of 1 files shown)


crates/contract/src/call.rs
```diff
@@ -425,8 +425,13 @@ impl<P: Provider<N>, D: CallDecoder, N: Network> CallBuilder<P, D, N> {
...
-    pub fn legacy(self) -> Self {
-        todo!()
+    pub fn legacy(mut self) -> Self {
+        if self.request.gas_price().is_none() {
+            if let Some(max_fee) = self.request.max_fee_per_gas() {
+                self.request.set_gas_price(max_fee);
+            }
+        }
+        self
     }
```
mattsse: I think this fn is a potential footgun and we should actually remove this because this might not end up using legacy all the time

VolodymyrBg: > I think this fn is a potential footgun and we should actually remove this because this might not end up using legacy all the time

Done, boss

PR #3088, opened 2025-10-28
chore: expose WebSocketConfig in RPC client and provider for non-WASM builds

cakevm:
Exposing the `WebSocketConfig` from the alloy-provider to not require any extra dependency to use it.

Re-export it to make it available from: `alloy_provider` -> `alloy_rpc_client` -> `alloy_transport_ws`

Example usage: `use alloy_provider::{WsConnect, WebSocketConfig};`


mattsse (approved):
makes sense

PR #3094, opened 2025-10-28
transport: move alloy_primitives to dev-deps

wedjob0X:
remove the dummy use alloy_primitives as _; from alloy/crates/transport/src/lib.rs
move alloy-primitives from [dependencies] to [dev-dependencies] since only tests need it
ran cargo check and cargo test -p alloy-transport locally


wedjob0X:
> please run cargo +nightly fmt
> 
> and avoid opening prs from main

https://github.com/alloy-rs/alloy/pull/3106

PR #3095, opened 2025-10-28
fix: saturate gas price in deser for unknown networks

sistemd:
## Motivation

I noticed that certain transactions on Arbitrum One fail to deserialize. Searching through existing issues I came across [this](https://github.com/alloy-rs/alloy/issues/2842), describing the exact error I was facing. Seeing that a pull request fixing this issue has already been merged, I updated `alloy` to the most recent version at the time (v1.0.41) hoping it would fix the deserialization problem — but it didn't. Since I wasn't sure whether the issue was now on my side, I decided to find the root cause and attempt to fix it.

What I found is that the pull request that closed the issue linked above made changes to a code path that was not being executed when working with Arbitrum One, with the main junction being [this](https://github.com/alloy-rs/alloy/blob/f487c2d6ecc416f909c6c4b48873037c968c5ba1/crates/network/src/any/either.rs#L210-L215) enumeration. The saturation logic only applies to the `AnyTxEnvelope::Ethereum` variant, which does not apply to Arbitrum One.

Take this command output for example:

```sh
❯ curl -s $ARBONE_RPC_PROVIDER_URL \
-d '{"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":["0xfb3988", true],"id":1}' | jq -r '.result.transactions[].type'

0x78
```
0x78 is a transaction type that does not correspond to an Ethereum transaction. Because of the type, the transaction wouldn't match any of [these](https://github.com/alloy-rs/alloy/blob/f487c2d6ecc416f909c6c4b48873037c968c5ba1/crates/consensus/src/transaction/envelope.rs#L169-L191) variants (which is where type definitions lead us from `AnyTxEnvelope::Ethereum`) and the newly added saturation logic would not kick in. This leaves us with the `AnyTxEnvelope::Unknown` variant and its deserialization logic, which still does not perform any saturation on gas prices over `u128::MAX`. 

## Solution

I took the same patch from [this](https://github.com/alloy-rs/alloy/pull/3010) pull request and applied it to [this](https://github.com/alloy-rs/alloy/blob/f487c2d6ecc416f909c6c4b48873037c968c5ba1/crates/rpc-types-eth/src/transaction/mod.rs#L456-L464) helper struct. As the documentation of its parent module says, this is where the `gasPrice` field will end up if it doesn't get consumed by the inner transaction (which won't be the case here).

After making these changes, I tested our indexer again to make sure that the problem was resolved and it was.

Please let me know if I've missed anything or if further changes are required.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/rpc-types-eth/src/transaction/mod.rs
+++ b/crates/rpc-types-eth/src/transaction/mod.rs
@@ -26,7 +26,7 @@ mod receipt;
 pub use receipt::TransactionReceipt;
 
 pub mod request;
-pub use request::{TransactionInput, TransactionInputKind, TransactionRequest};
+pub use request::{FillTransaction, TransactionInput, TransactionInputKind, TransactionRequest};
 
 /// Serde-bincode-compat
 #[cfg(all(feature = "serde", feature = "serde-bincode-compat"))]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
         rust:
           - "stable"
           - "nightly"
-          - "1.86" # MSRV
+          - "1.88" # MSRV
         flags:
           # No features
           - "--no-default-features -F reqwest-rustls-tls"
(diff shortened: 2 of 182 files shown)


crates/rpc-types-eth/src/transaction/mod.rs
```diff
@@ -458,7 +476,7 @@ mod tx_serde {
             default,
             rename = "gasPrice",
             skip_serializing_if = "Option::is_none",
-            with = "alloy_serde::quantity::opt"
+            with = "self"
```
DaniPopes: you can avoid all of this by using U256 and saturating at conversion points since this is a private struct

sistemd: https://github.com/alloy-rs/alloy/pull/3095/commits/8ce8a3d52a5b50e9f0cf16fac91888445be6009b

jaggad:
@sistemd I filed the issue for https://github.com/alloy-rs/alloy/issues/2842 and have also upgraded the system that expected to be fixed by this last PR being merged.

However the issue remains on my end as well. I just came digging to see if I could fine what the root cause was as well since the latest change was no good and saw your PR here.

crates/rpc-types-eth/src/transaction/mod.rs
jaggad: @DaniPopes just wanting to follow up here on status of this and if you're happy with the latest changes for this PR to be merged

jaggad:
Would it be possible to get this merged in the near future? Would like to depend on the upstream alloy-rs repo if possible

PR #3108, opened 2025-10-29
Remove redundant block hash iteration in provider heartbeat

wedjob0X:
reuse the collected block transaction hashes to build the B256HashSet and query unconfirmed, avoiding a second iteration and allocation
import the Provider trait from crate::provider so the trait methods remain in scope explicitly


DaniPopes:
this is worse

PR #3111, opened 2025-10-30
fix: remove redundant `Option` from `as_receipt` and related methods in `ReceiptEnvelope`

maradini77:
---

This PR simplifies the API of `ReceiptEnvelope` by changing the return type of the method `as_receipt` (and similar methods) to directly return a reference, instead of always returning `Some` inside an `Option`.

- All enum variants contain a `ReceiptWithBloom<Receipt<T>>`, so the methods are infallible.
- Removes unnecessary `.unwrap()` usages, improves readability and usage ergonomics.

---


mattsse:
Closing - not accepting low-value changes.

PR #3114, opened 2025-10-30
refactor(provider): simplify `HyperClient` init with `layer` method

### LINKED ISSUE #2988: [Feature] Make AuthLayer easier to use
### Component

provider, pubsub, transports

Now #1314 added `AuthLayer` based on hyper transport client to refresh the bearer tokens.
But we have to construct a `HyperClient` from the most inner `TokioExecutor::new()`, then `hyper_util::client::legacy::Client` then `tower::ServiceBuilder` after that we could call `HyperClient::from_service`.

Is it possible to simplify this procedure, we don't like to talk to tower in order to use `AuthLayer`, maybe we could do things like this
```rust
let secret = JwtSecret::random();
HyperClient::new().layer(AuthLayer::new(secret))
```

### Additional context

_No response_

mablr:
## Motivation

Close #2988

Make `HyperClient` composable.

## Solution

- Introduced a new `layer` method in `HyperClient` to apply middleware layers, enhancing composability.
- Updated documentation for the new method with an example usage.
- Removed unnecessary service construction in provider tests which basically demonstrates the benefit of this new `layer` method.

## PR Checklist

- [x] Added Tests (Updated existing one)
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/src/hyper_transport.rs
+++ b/crates/transport-http/src/hyper_transport.rs
@@ -11,7 +11,7 @@ use hyper::{
 };
 use hyper_util::client::legacy::Error;
 use std::{future::Future, marker::PhantomData, pin::Pin, task};
-use tower::Service;
+use tower::{Layer, Service};
 use tracing::{debug, debug_span, trace, Instrument};
 
 #[cfg(feature = "hyper-tls")]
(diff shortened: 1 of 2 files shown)


crates/transport-http/src/hyper_transport.rs
```diff
@@ -78,6 +78,28 @@ impl<B, S> HyperClient<B, S> {
...
+    /// # Example
+    ///
+    /// ```ignore
+    /// #use alloy_transport_http::HyperClient;
+    /// #use alloy_transport_http::AuthLayer;
+    /// #use alloy_rpc_types_engine::JwtSecret;
+    ///
+    /// let secret = JwtSecret::random();
+    /// let client = HyperClient::new()
+    ///     .layer(AuthLayer::new(secret));
```
mattsse, line 95: sweet

mattsse (approved):
great, very nice ux improvement

PR #3119, opened 2025-10-31
fix: Preserve inner retry service state across requests

viktorking7:
Removed the redundant `to_vec()` clones inside `crates/provider/src/provider/multicall/mod.rs`, so the already collected call vectors are reused without extra allocations or copies. Also fixed a more serious issue in `crates/transport/src/layers/retry.rs`: the retry layer used to swap the inner service with a fresh clone on every request, discarding any accumulated state. The inner service now lives inside an `Arc<Mutex<_>>`, letting every retry reuse the same instance and preserving connection pools, caches, and other stateful data.


mattsse (commented):
we dont need to change this because this is cloned per request

PR #3121, opened 2025-10-31
docs(network): add usage examples for try_into_either and try_map_unknown

avorylli:
Add examples to try_into_either and try_map_unknown methods in AnyTxEnvelope.

These methods lacked documentation examples, making it difficult for users to understand when and how to use them. 
The examples demonstrate converting unknown transactions to custom types and extracting specific fields.


mattsse (approved):
cool, this is better

PR #3122, opened 2025-11-01
feat: add map sidecar fns

mattsse:
more map fns


grandizzy:
omg, aws update

mattsse:
yeah I think we can bump msrv now

PR #3126, opened 2025-11-02
refactor(transport-ws): replace tracing macro_use with explicit imports

Doryu001:
remove the legacy #[macro_use] extern crate tracing; from transport-ws
import only the macros that are actually used (trace! and error!) in each module


mattsse (commented):
not necessary

PR #3127, opened 2025-11-02
Remove unused Debug import in transport error module

wedjob0X:
The crates/transport/src/error.rs module derives Debug on its types, so the explicit std::fmt::Debug import was unused and flagged during review.
Change: Dropped the redundant fmt::Debug import; the derive-generated impls remain untouched.


mattsse (commented):
not necessary

PR #3128, opened 2025-11-03
chore(ens): bumped ens crate with appropriate features

dipanshuhappy:
## Motivation

Without enables these features we can't use it in the ```alloy::Provider```

## Solution

Just enabled in the ```Cargo.toml```
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/alloy/Cargo.toml
+++ b/crates/alloy/Cargo.toml
@@ -30,7 +30,7 @@ alloy-core.workspace = true
 alloy-consensus = { workspace = true, optional = true }
 alloy-contract = { workspace = true, optional = true }
 alloy-eips = { workspace = true, optional = true }
-alloy-ens = { workspace = true, optional = true, features = ["contract","provider"] }
+alloy-ens = { workspace = true, optional = true, features = ["contract", "provider"] }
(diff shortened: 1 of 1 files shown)


dipanshuhappy:
> these should not be enabled on workspace level and instead forward features in the alloy-meta crate instead

Got it

dipanshuhappy:
> these should not be enabled on workspace level and instead forward features in the alloy-meta crate instead

I have shifted it to the meta crates

crates/alloy/Cargo.toml
```diff
@@ -30,7 +30,7 @@ alloy-core.workspace = true
 alloy-consensus = { workspace = true, optional = true }
 alloy-contract = { workspace = true, optional = true }
 alloy-eips = { workspace = true, optional = true }
-alloy-ens = { workspace = true, optional = true }
+alloy-ens = { workspace = true, optional = true, features = ["contract","provider"] }
```
mattsse:
```suggestion
alloy-ens = { workspace = true, optional = true, features = ["contract", "provider"] }
```

PR #3132, opened 2025-11-03
fix: BlobParams bincode deserialization

leruaa:
## Motivation

Trying to deserialize a `ChainConfig` with bincode fail where blob schedule is specified (see the updated unit test).

The error is: 

```
UnexpectedVariant { type_name: "Option<T>", allowed: Range { min: 0, max: 1 }, found: 6 }
```

## Solution

Don't skip serializing (I'm not sure if it's the right fix: skipping the field entirely also works).

Edit: looks like tests are more happy when skipping the field  entirely

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -181,6 +181,7 @@ mod serde_impl {
         max_blob_count: u64,
         #[serde(rename = "target")]
         target_blob_count: u64,
+        #[serde(skip)]
         min_blob_fee: Option<u128>,
     }
 


crates/eips/src/eip7840.rs
```diff
@@ -181,7 +181,7 @@ mod serde_impl {
         max_blob_count: u64,
         #[serde(rename = "target")]
         target_blob_count: u64,
-        #[serde(skip_serializing)]
+        #[serde(skip)]
         min_blob_fee: Option<u128>,
```
mattsse, line 185: ah yeah we can keep this symmetrical

mattsse (approved):
all my homies hate bincode

PR #3138, opened 2025-11-04
feat(genesis): add parent hash field in Genesis

xzav3r:
## Motivation

Some blockchain networks use custom genesis block heights where the genesis block may have a non-zero parent hash (e.g., when starting from a specific block height in an existing chain). The current `Genesis` struct does not support specifying the parent hash, which limits compatibility with such networks.

This change adds support for optional `parent_hash` field in the genesis configuration, allowing users to specify a parent hash when it's needed for custom genesis scenarios.

## Solution

- Added `parent_hash: Option<B256>` field to the `Genesis` struct with appropriate serde attributes:
  - `default`: Allows deserialization without the field (defaults to `None`)
  - `skip_serializing_if = "Option::is_none"`: Omits the field when serializing if it's `None`

- Added `with_parent_hash()` builder method for convenient construction, consistent with other optional fields

- Added comprehensive test coverage (`test_parent_hash_serialization`) to verify:
  - Serialization/deserialization of `parent_hash` when set
  - Field is omitted when `None` during serialization
  - Deserialization works correctly when the field is missing (defaults to `None`)
  - Deserialization works correctly when the field is present

- Updated existing test expectations to match the new field structure

## PR Checklist

- [x] Added Tests

- [ ] Added Documentation

- [ ] Breaking changes

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -69,6 +69,9 @@ pub struct Genesis {
     /// The genesis block number
     #[serde(default, skip_serializing_if = "Option::is_none", with = "alloy_serde::quantity::opt")]
     pub number: Option<u64>,
+    /// The parent hash
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub parent_hash: Option<B256>,
 }
 
 impl Genesis {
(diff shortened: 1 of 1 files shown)


xzav3r:
Hi maintainers, I allow you to edit this PR if needed. Please feel free to make any necessary changes or improvements. 
I'm looking forward to having this merged when convenient. Thanks for your review!

xzav3r:
> not opposed to this, because doesn't hurt
> 
> but perhaps a better solution would be to add otherfields here, so that more things can be captured

https://github.com/ethereum/go-ethereum/blob/15ff378a8927eed211589bcf375aa5c528209b71/core/genesis.go#L72

Because I noticed that the `Genesis` in geth also has this field, it is actually a standard field. So aligning with geth might be a better solution and do not have to add any other fileds. What do you think?

PR #3141, opened 2025-11-04
fix(pubsub): avoid BiBTreeMap remove/insert in SubscriptionManager::notify

Fibonacci747:
Replace remove_by_left/insert with get_by_left in SubscriptionManager::notify().

ActiveSubscription::notify(&self, Box<RawValue>) does not require mutable or owned access to the subscription; borrowing via get_by_left is sufficient. EthNotification carries an owned Box<RawValue>, so passing it to notify remains correct.

This change eliminates unnecessary map removals/insertions (and potential allocations), reduces contention and churn on BiBTreeMap, and preserves behavior.


mattsse (approved):
i see, makes sense

PR #3145, opened 2025-11-05
Chore(rpc-types-anvil): derive default for MineOptions

Doryu001:
Derive Default for MineOptions and tag the Options variant with #[default] drop the redundant manual Default implementation while preserving existing behaviour


mattsse (commented):
not something we need to change

PR #3147, opened 2025-11-05
feat: add Transaction bound to Network::TxEnvelope

### LINKED ISSUE #3133: [Feature] add `Transaction` bound to `Network::TxEnvelope`
### Component

network, json-rpc

Can the `Network::TxEnvelope` associated type have a `Transation` bound?

I have something which implements Network::TxEnvelpe and I want to get a `to` and `value` from it.

### Additional context

_No response_

0xChaddB:
Add Transaction bound to Network::TxEnvelope so callers can access transaction methods
Fixes #3133


mattsse (approved):
we can safely do this because we expect this is to be the case

PR #3151, opened 2025-11-06
feat: added amsterdam to blob params

Soubhik-10:
cc @mattsse @Rimeeeeee

### DIFF
--- a/crates/eips/src/eip7840.rs
+++ b/crates/eips/src/eip7840.rs
@@ -78,6 +78,18 @@ impl BlobParams {
         }
     }
 
+    /// Returns [`BlobParams`] configuration activated with Amsterdam hardfork.
+    pub const fn amsterdam() -> Self {
+        Self {
+            target_blob_count: eip7691::TARGET_BLOBS_PER_BLOCK_ELECTRA,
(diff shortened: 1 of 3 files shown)


crates/eips/src/eip7840.rs
```diff
@@ -78,6 +78,18 @@ impl BlobParams {
...
+    /// Returns [`BlobParams`] configuration activated with Amsterdam hardfork.
+    pub const fn amsterdam() -> Self {
+        Self {
+            target_blob_count: eip7691::TARGET_BLOBS_PER_BLOCK_ELECTRA,
+            max_blob_count: eip7691::MAX_BLOBS_PER_BLOCK_ELECTRA,
+            update_fraction: eip7691::BLOB_GASPRICE_UPDATE_FRACTION_PECTRA,
+            min_blob_fee: eip4844::BLOB_TX_MIN_BLOB_GASPRICE,
+            max_blobs_per_tx: eip7594::MAX_BLOBS_PER_TX_FUSAKA,
+            blob_base_cost: BLOB_BASE_COST,
+        }
```
mattsse, line 90: I think these are likely incorrect because the asmterdam fields are still tbd

onbjerg:
What is the blocker here?

mattsse:
this is breaking

PR #3152, opened 2025-11-06
chore: add amsterdam support

mattsse:
ref https://github.com/ethpandaops/bal-devnets/blob/master/network-configs/devnet-1/metadata/genesis.json


crates/genesis/src/lib.rs
onbjerg, line 316: let's mark this non exhaustive so future additions are non breaking

PR #3153, opened 2025-11-06
feat: Make AnyTransactionReceipt a dedicated type 

### LINKED ISSUE #3149: [Feature] Make AnyTransactionReceipt a dedicated type
### Component

consensus, eips, genesis

we have this alias 

https://github.com/alloy-rs/alloy/blob/808b88df4d627e98b3a3ea55ab54d96d264a23bb/crates/rpc-types-any/src/transaction/receipt.rs#L7-L7

whichs makes it hard to add convenience conversion functions

we want this to be like

https://github.com/alloy-rs/alloy/blob/808b88df4d627e98b3a3ea55ab54d96d264a23bb/crates/network/src/any/mod.rs#L216-L216

instead

## TODO
* introduce dedicated struct in network/any/transaction
* keep the existing type alias for now, because replacing this would be a breaking change

### Additional context

@0xChaddB

0xChaddB:
Transforms `AnyTransactionReceipt` from a type alias to a dedicated struct following the same pattern as `AnyRpcTransaction`.

Fixes #3149


crates/network/src/any/mod.rs
```diff
@@ -529,6 +531,265 @@ impl Typed2718 for AnyRpcTransaction {
     }
 }
 
+/// A wrapper for [`AnyTransactionReceipt`] that allows for handling unknown receipt types.
+#[derive(Clone, Debug, From, PartialEq, Eq, Deserialize, Serialize)]
+pub struct AnyTransactionReceipt(pub WithOtherFields<TransactionReceipt<AnyReceiptEnvelope<Log>>>);
```
mattsse: this file is already quite large, can we move this a dedicated file instead? like transaction.rs

mattsse (changes requested):
very nice

mattsse (approved):

PR #3157, opened 2025-11-07
fix: use IgnoredAny for unknown fields in PubSubItem deserializer

phrwlk:
Replace serde_json::Value with serde::de::IgnoredAny when discarding unknown fields in PubSubItem’s custom Deserialize (crates/json-rpc/src/notification.rs).
This avoids unnecessary allocations and JSON tree construction for ignored fields.
Matches existing approach in response module (response/error.rs, response/mod.rs).

### DIFF
--- a/crates/json-rpc/src/response/mod.rs
+++ b/crates/json-rpc/src/response/mod.rs
@@ -284,8 +284,7 @@ where
                             id = Some(map.next_value()?);
                         }
                         Field::Unknown => {
-                            let _: serde::de::IgnoredAny = map.next_value()?; // ignore unknown
-                                                                              // fields
+                            let _: serde::de::IgnoredAny = map.next_value()?; // ignore
                         }
                     }
                 }


crates/json-rpc/src/response/mod.rs
```diff
@@ -284,7 +284,8 @@ where
                             id = Some(map.next_value()?);
                         }
                         Field::Unknown => {
-                            let _: serde::de::IgnoredAny = map.next_value()?; // ignore
+                            let _: serde::de::IgnoredAny = map.next_value()?; // ignore unknown
+                                                                              // fields
```
DaniPopes: please revert

DaniPopes (changes requested):

crates/json-rpc/src/response/mod.rs
phrwlk: > please revert

sorry, fixed

mattsse (approved):

crates/json-rpc/src/response/mod.rs
mattsse: was my bad edit, sorry @phrwlk

phrwlk: > was my bad edit, sorry @phrwlk

@mattsse it's ok, you're doing a huge amount of work

PR #3168, opened 2025-11-08
remove unused itertools import from wallet provider tests

0xLogicalx:
drop the dangling `itertools::Itertools` import from the wallet provider test module keep the test suite free from `unused_imports` warnings enforced by the workspace lints


mattsse (commented):
not worth doing

PR #3169, opened 2025-11-09
perf(network-primitives): use iter::empty() instead of array literal in txns()

DeVikingMark:
Replace `[].iter()` with `core::iter::empty()` in `BlockTransactions::txns()`  to avoid unnecessary stack allocations. 

Added `EitherIter` helper type to unify different iterator types returned by the method.


mattsse (commented):
this isnt needed

PR #3172, opened 2025-11-10
fix: encode TraceFilter after/count as QUANTITY

VolodymyrBg:
This change updates the TraceFilter fields after and count to use Ethereum JSON-RPC QUANTITY (hex) encoding via alloy_serde::quantity::opt. OpenEthereum/Parity trace_filter expects these pagination fields as hex quantities; previously they were serialized as decimal JSON numbers and would fail to parse hex input, leading to incompatibilities with nodes. Aligning these fields with the existing fromBlock/toBlock handling ensures correct wire format for both serialization and deserialization and brings consistency across the RPC types.

### DIFF
--- a/crates/rpc-types-trace/src/filter.rs
+++ b/crates/rpc-types-trace/src/filter.rs
@@ -28,8 +28,10 @@ pub struct TraceFilter {
     #[serde(default)]
     pub mode: TraceFilterMode,
     /// Output offset
+    #[serde(with = "alloy_serde::quantity::opt")]
     pub after: Option<u64>,
     /// Output amount
+    #[serde(with = "alloy_serde::quantity::opt")]
     pub count: Option<u64>,
 }
 


mattsse (commented):
I believe those should actually be numbers


> trace_filter expects these pagination fields as hex quantities

can you back this up?

VolodymyrBg:
> I believe those should actually be numbers
> 
> > trace_filter expects these pagination fields as hex quantities
> 
> can you back this up?

In Ethereum JSON-RPC, numbers must be QUANTITY (hex). The trace_filter parameters in Parity/OpenEthereum follow the same rules, so after/count is also QUANTITY.
References:
EIP-1474 (QUANTITY rules): https://eips.ethereum.org/EIPS/eip-1474#quantity
OpenEthereum trace_filter: https://openethereum.github.io/JSONRPC-trace-module#trace_filter

mattsse:
this examples shows

>   "after": 1000,
  "count": 100

VolodymyrBg:
> this examples shows
> 
> > "after": 1000,
> > "count": 100

fair

PR #3174, opened 2025-11-10
feat: Improve PrivacyHint deserialization error to include offending value

Bashmunta:
Replace generic serde error with Error::unknown_variant and add a static list of allowed hints. This improves diagnostics and aligns with existing repository style that includes precise context in error messages, without changing functional behavior beyond error text.


mattsse (approved):
fine

PR #3177, opened 2025-11-11
feat: support send_transaction_sync

### LINKED ISSUE #3175: [Feature] support send_transaction_sync
### Component

rpc

similar to

https://github.com/alloy-rs/alloy/blob/35bc323cc23911e52cb1ec8f6a58ebb5d5869b27/crates/provider/src/provider/trait.rs#L932-L937

but returns the txreceipt immediately

this new trait fn should return a helper future type that awaits the response, but also holds the raw tx and the tx hash so that it can be recovered

see also:

https://github.com/alloy-rs/alloy/blob/35bc323cc23911e52cb1ec8f6a58ebb5d5869b27/crates/provider/src/provider/trait.rs#L958-L977

## TODO

* add helper Future type (e.g. `SendTransactionSync<N> {raw, txhash, fut}`) where fut is basically

https://github.com/alloy-rs/alloy/blob/35bc323cc23911e52cb1ec8f6a58ebb5d5869b27/crates/provider/src/provider/trait.rs#L881-L881

we also need a new error type that then returns the raw tx and the hash

* we also need to specialize this for:
https://github.com/alloy-rs/alloy/blob/35bc323cc23911e52cb1ec8f6a58ebb5d5869b27/crates/provider/src/fillers/mod.rs#L635-L652 

### Additional context

ref https://github.com/foundry-rs/foundry/issues/12518

@stevencartavia

stevencartavia:
## Motivation
closes #3175

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


crates/provider/src/heart.rs
```diff
@@ -435,6 +435,135 @@ impl Future for PendingTransaction {
...
+/// A synchronous transaction sender that returns transaction receipt immediately.
+///
+/// This future combines transaction submission and receipt fetching into a single operation,
+/// reducing the number of RPC calls and providing immediate access to transaction metadata.
+#[must_use = "this type does nothing unless you await it"]
+pub struct SendTransactionSync<N: Network> {
+    /// The raw transaction bytes that were sent.
+    pub raw: alloy_primitives::Bytes,
+    /// The transaction hash.
+    pub tx_hash: TxHash,
```
mattsse: we can make these private

crates/provider/src/heart.rs
```diff
@@ -435,6 +435,135 @@ impl Future for PendingTransaction {
...
 }
 
+#[cfg(not(target_family = "wasm"))]
+/// Future type for SendTransactionSync on non-wasm targets.
+type SendTransactionSyncFuture<N> = std::pin::Pin<
+    Box<
+        dyn Future<Output = Result<<N as Network>::ReceiptResponse, SendTransactionSyncError>>
+            + Send,
+    >,
+>;
```
mattsse: this file is already quite large, let's move these types to a new file instead

crates/provider/src/provider/trait.rs
```diff
@@ -976,6 +976,91 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// ```no_run
+    /// # async fn example<N: alloy_network::Network>(provider: impl alloy_provider::Provider<N>, tx: N::TransactionRequest) -> Result<(), Box<dyn std::error::Error>> {
+    /// let sync_tx = provider.send_transaction_sync(tx).await?;
+    ///
+    /// // Access transaction metadata immediately
+    /// println!("Transaction hash: {}", sync_tx.tx_hash());
+    /// println!("Raw transaction: {}", sync_tx.raw());
+    ///
+    /// // Wait for the receipt
+    /// let receipt = sync_tx.await?;
```
mattsse: this should be possible with one await I think,

so we can change the trait fn to just fn send_transaction_sync(

crates/provider/src/provider/trait.rs
```diff
@@ -976,6 +977,92 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn send_transaction_sync(
+        &self,
+        tx: N::TransactionRequest,
+    ) -> Result<N::ReceiptResponse, SendTransactionSyncError> {
+        let sync_tx = self.send_transaction_sync_internal(SendableTx::Builder(tx)).await.map_err(
+            |transport_error| SendTransactionSyncError::SubmissionFailed {
+                raw: alloy_primitives::Bytes::default(), // Empty for unsigned transactions
```
mattsse: this we can't do, so we need an error that can represent this case

crates/provider/src/provider/trait.rs
```diff
@@ -976,6 +977,92 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+                // For unsigned transactions, raw bytes are not meaningful since the node
+                // signs and potentially modifies the transaction. We use empty bytes.
+                let raw = Bytes::default();
+
+                // Create future for receipt fetching
+                let provider = self.root().clone();
+                let raw_clone = raw.clone();
+                let fut = Box::pin(async move {
+                    let pending_builder = PendingTransactionBuilder::new(provider, tx_hash);
+                    pending_builder.get_receipt().await.map_err(|error| {
```
mattsse: this is basically just send_transaction_internal + await the receipt, right?

mattsse (approved):

crates/provider/src/provider/trait.rs
```diff
@@ -976,6 +976,62 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+        tx: SendableTx<N>,
+    ) -> TransportResult<N::ReceiptResponse> {
+        // Make sure to initialize heartbeat before we submit transaction, so that
+        // we don't miss it if user will subscriber to it immediately after sending.
+        let _handle = self.root().get_heart();
+
+        match tx {
+            SendableTx::Builder(mut tx) => {
+                alloy_network::TransactionBuilder::prep_for_submission(&mut tx);
+                let receipt = self.client().request("eth_sendTransactionSync", (tx,)).await?;
```
hazelnutcloud, line 1025: `eth_sendTransactionSync` does not exist and is not part of EIP-7966 btw

hai-rise: 😭

kien-rise: the correct name is `eth_sendRawTransactionSync` 🫠

crates/provider/src/provider/trait.rs
```diff
@@ -976,6 +976,62 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    /// [`ProviderLayer`]: crate::ProviderLayer
+    /// [`TxFiller`]: crate::fillers::TxFiller
+    #[doc(hidden)]
+    async fn send_transaction_sync_internal(
+        &self,
+        tx: SendableTx<N>,
+    ) -> TransportResult<N::ReceiptResponse> {
+        // Make sure to initialize heartbeat before we submit transaction, so that
+        // we don't miss it if user will subscriber to it immediately after sending.
+        let _handle = self.root().get_heart();
```
hazelnutcloud, line 1020: This is also not needed as no subscription will be made

PR #3179, opened 2025-11-11
chore: Avoid cloning EIP-4844 sidecar during request build

Doryu001:
Issue: build_4844_with_sidecar always cloned the BlobTransactionSidecar, triggering redundant allocation and violating redundant_clone. Fix: Move the sidecar out of self.sidecar with take(), preserving behavior while removing the unnecessary clone. Verification: cargo check (pending – cargo unavailable on this machine). Recommend running cargo check and cargo test -p alloy-rpc-types-eth once available.


mattsse (approved):
this sia ppropriate

Issue #2826, opened 2025-08-27
[Feature] configure error code for dropped filters in `PollerStream`


siosw:
### Component

rpc

### Describe the feature you would like

when using `PollerStream` with rpc providers, filters often get dropped at some point. this should cause the poller to be stopped but [the current logic](https://github.com/alloy-rs/alloy/blob/33f17b8dd7b820bdd5cf271b067ce2f2be79c8f1/crates/rpc-client/src/poller.rs#L373) for this is not very robust. 
tenderly for example just returns an error code and no message, this makes it hard to implement logic for resetting the filter.

it would be nice to allow consumers of `PollerStream` to optionally provide the response codes that should cause the poller to stop, if received.

nadtech-hub:
✋

grandizzy:
@nadtech-hub are you still willing to take this one? thanks!

nadtech-hub:
@grandizzy Yes, I am

dipanshuhappy:
> [@grandizzy](https://github.com/grandizzy) Yes, I am

Hey @nadtech-hub, so i was working on #3150, curious if you are also adding a function parameter to the builder struct that runs if it causes error.

nadtech-hub:
Hi @dipanshuhappy my original intention was to add map error function defined by user to PollerStream in order to let user interpret rpc an error.

nadtech-hub:
@dipanshuhappy you can raise PR before me cuz I think this feature is insufficient

dipanshuhappy:
> [@dipanshuhappy](https://github.com/dipanshuhappy) you can raise PR before me cuz I think this feature is insufficient

Hey @nadtech-hub, I didn't realise how tricky this would be 😄 . I am currently implemented a ```with_reconnect``` on the PollerBuilder, and without adding new generics to PollerStream I had been reaching a dead end. For now I went ahead with using a type ```serde::RawValue``` but still on it. Ideally tomorrow i should be able to wrap it up and make a PR

nadtech-hub:
json value is a good option, I was trying crate either

dipanshuhappy:
> json value is a good option, I was trying crate either

Been stuck with implementing the function callback. Seems like ```weak_client()``` is not working in the function closure i created. Still debugging that.

dipanshuhappy:
> json value is a good option, I was trying crate either

Hey, just made a PR on it. #3178

nadtech-hub:
> > json value is a good option, I was trying crate either
[...]

PR #3182, opened 2025-11-12
feat: error code for dropped filters

nadtech-hub:
This should resolve [#2826](https://github.com/alloy-rs/alloy/issues/2826)

### DIFF
--- a/crates/rpc-client/src/poller.rs
+++ b/crates/rpc-client/src/poller.rs
@@ -1,5 +1,6 @@
 use crate::WeakClient;
 use alloy_json_rpc::{RpcRecv, RpcSend};
+use alloy_primitives::map::HashSet;
 use alloy_transport::utils::Spawnable;
 use futures::{ready, stream::FusedStream, Future, FutureExt, Stream, StreamExt};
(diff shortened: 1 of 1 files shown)


siosw:
hey @nadtech-hub thanks for the PR, this isn't quite what I was going for when writing the issue,
i was imagining something like

```rust
if resp.message.contains("filter not found") || self.terminal_error_codes.contains(resp.code) {
    // stop the poller
}
```

where the list of error codes is something the user of the poller can set

nadtech-hub:
> hey @nadtech-hub thanks for the PR, this isn't quite what I was going for when writing the issue, i was imagining something like
> 
> ```rust
> if resp.message.contains("filter not found") || self.terminal_error_codes.contains(resp.code) {
>     // stop the poller
> }
> ```
> 
> where the list of error codes is something the user of the poller can set

do you need particular ones to be printed out?

nadtech-hub:
> ```rust
> terminal_error_codes
> ```

so you need certain RPC error codes. @siosw see if changes are sufficient now

PR #3184, opened 2025-11-12
refactor: remove unnecessary Encode bound in parse_request_payload

sashass1315:
Tightened the generic bound in crates/rpc-types-beacon/src/requests.rs for parse_request_payload from Vec<T>: Decode + Encode to T: Decode because the function only calls Vec::::from_ssz_bytes and never encodes, making Encode redundant; this reduces unnecessary constraints and improves clarity without changing behavior, since Vec<T> implements Decode when T: Decode.


onbjerg (approved):
lgtm, would like @mattsse to take a pass too though because this is likely used in reth

PR #3187, opened 2025-11-13
docs: add documentation for build_unsigned and build methods

forkfury:
Adds documentation for build_unsigned and build methods in AnyNetwork TransactionBuilder implementation. Documents limitations regarding Ethereum-only transaction types and error conditions to improve API clarity.


mattsse (approved):
fines, this doesnt hurt and is actually useful ctx here

PR #3192, opened 2025-11-13
fix: normalize recovery bytes in flashbots signature

### LINKED ISSUE #3191: [Bug] Inconsistent signature handling for sign_flashbots_payload
### Component

consensus, eips, genesis

1.1.1

### Operating System

None

### Describe the bug

`alloy_primitives::Signature` normalizes the recovery ID as soon as it’s created (`Signature::from_raw_array` calls `normalize_v`). Internally it keeps `v` as a boolean `y-parity`, and every recovery path (`recover_signer`) rebuilds the [r || s || v] array by copying r, s, and writing signature.v() as u8. The canonical way here is always to feed a `v ∈ {0,1}` into libsecp256k1 (also matching go-ethereum’s `SigToPub`).

In `crates/provider/src/ext/mev/with_auth.rs`, `sign_flashbots_payload` takes the signer output and immediately calls `signature.as_bytes()`. That accessor is for RLP/EIP-155 encoding, so it writes `v = 27 + parity` (see `Signature::as_bytes`). 

## Impact
The flashbots signature does not follow the canonical encoding. The server may use `recover_signer` (in rust) or geth's `crypto.SigToPub` to recover the pubkey. However both handlers do not accept the signature generated by `sign_flashbots_payload`.

## Fix
replace `message_hash.as_bytes()` in `sign_flashbots_payload` with correct `y-parity`.
```rust
let signature = signer.sign_message(...).await?;
let mut sig_bytes = [0u8; 65];
sig_bytes[..32].copy_from_slice(&signature.r().to_be_bytes::<32>());
sig_bytes[32..64].copy_from_slice(&signature.s().to_be_bytes::<32>());
sig_bytes[64] = signature.v() as u8;
hex::encode_prefixed(sig_bytes);
```

0xJchen:
This PR changes the signature encoding in `sign_flashbots_payload` to meet the canonical way of formatting sigs. It serializes flashbots signatures with a normalized recovery byte (`v = 0/1`), instead of using `Signature::as_bytes()`, which produced legacy 27/28 values

Fixes #3191

### DIFF
--- a/crates/provider/src/ext/mev/with_auth.rs
+++ b/crates/provider/src/ext/mev/with_auth.rs
@@ -145,6 +145,6 @@ mod tests {
         .unwrap();
         let body = "sign this message".to_string();
         let signature = sign_flashbots_payload(body.clone(), &signer).await.unwrap();
-        assert_eq!(signature, "0xd5F5175D014F28c85F7D67A111C2c9335D7CD771:0x983dc7c520db0d287faff3cd0aef81d5a7f4ffd3473440d3f705da16299724271f660b6fe367f455b205bc014eff3e20defd011f92000f94d39365ca0bc786721b");
+        assert_eq!(signature, "0xd5F5175D014F28c85F7D67A111C2c9335D7CD771:0x983dc7c520db0d287faff3cd0aef81d5a7f4ffd3473440d3f705da16299724271f660b6fe367f455b205bc014eff3e20defd011f92000f94d39365ca0bc7867200");
     }
 }


crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -122,7 +122,13 @@ pub async fn sign_flashbots_payload<S: Signer + Send + Sync>(
 ) -> Result<String, alloy_signer::Error> {
     let message_hash = keccak256(body.as_bytes()).to_string();
     let signature = signer.sign_message(message_hash.as_bytes()).await?;
-    Ok(format!("{}:{}", signer.address(), hex::encode_prefixed(signature.as_bytes())))
+
+    // Normalized recovery byte (0/1) following the canonical signature encoding
+    let mut sig_bytes = [0u8; 65];
+    sig_bytes[..32].copy_from_slice(&signature.r().to_be_bytes::<32>());
+    sig_bytes[32..64].copy_from_slice(&signature.s().to_be_bytes::<32>());
+    sig_bytes[64] = signature.v() as u8;
```
mattsse, line 130: I see, I feel like want this as a function natively on Signature as well

cc @cakevm

cakevm:
Makes sense to me as well. I learned that `recover_signer` just works with both 27/28 and 0/1. In the Flashbots codebase I found they normalize it e.g. [here](https://github.com/flashbots/rpc-endpoint/blob/a7e8aa9beb22eb89bde8d60b87d3adbf6739b572/adapters/flashbots/signature.go#L38) for golang. So I am here in favour to follow the standard and include this PR. 
I also tested it with the current Flashbots/Buildernet RPC and it is accepted as valid.

PS: @0xJchen you may like to check the test `test_sign_flashbots_payload`, it should now end with 00 not 1b=27.

crates/provider/src/ext/mev/with_auth.rs
cakevm: Yep, it would be great if there is a fn in `Signature` for this

PR #3193, opened 2025-11-13
chore(node-bindings/anvil): unify startup timeout with shared NODE_STARTUP_TIMEOUT

phrwlk:
Unifies Anvil’s startup timeout default with the crate-wide NODE_STARTUP_TIMEOUT used by geth and reth, removing the local duplicate default to prevent drift. Keeps the existing Anvil builder override as milliseconds for backward compatibility and explicitly documents the units to eliminate ambiguity.

### DIFF
--- a/crates/node-bindings/src/nodes/anvil.rs
+++ b/crates/node-bindings/src/nodes/anvil.rs
@@ -1,6 +1,6 @@
 //! Utilities for launching an Anvil instance.
 
-use crate::NodeError;
+use crate::{NodeError, NODE_STARTUP_TIMEOUT};
 use alloy_hardforks::EthereumHardfork;
 use alloy_network::EthereumWallet;
 use alloy_primitives::{hex, Address, ChainId};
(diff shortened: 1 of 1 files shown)


crates/node-bindings/src/nodes/anvil.rs
```diff
@@ -449,16 +447,15 @@ impl Anvil {
...
         let mut reader = BufReader::new(stdout);
+        let timeout = self.timeout.map(Duration::from_millis).unwrap_or(NODE_STARTUP_TIMEOUT);
 
         let mut private_keys = Vec::new();
         let mut addresses = Vec::new();
         let mut is_private_key = false;
         let mut chain_id = None;
         let mut wallet = None;
         loop {
-            if start + Duration::from_millis(self.timeout.unwrap_or(ANVIL_STARTUP_TIMEOUT_MILLIS))
```
mattsse, line 459: okay, this is actually a fix, so I allow it

mattsse (approved):

PR #3194, opened 2025-11-14
fix(provider): use fillable parameter in ChainIdFiller::fill

avorylli:
ChainIdFiller::fill now uses the fillable parameter directly instead of calling fill_sync again. 

This matches the pattern used by other fillers and ensures the prepare() result is properly utilized.


mattsse:
Closing - trivial refactor, not accepting low-value changes.

PR #3195, opened 2025-11-14
fix(rpc-types-engine): remove const from CancunPayloadFields::new

oooLowNeoNooo:
Remove const modifier from CancunPayloadFields::new since Vec parameters cannot be used in const fn contexts. Add clippy allow attribute to suppress the false positive warning about making the function const.


mattsse (commented):
not needed

PR #3202, opened 2025-11-14
feat: obtain the transaction hash if eth_sendrawSync

### LINKED ISSUE #3201: [Feature] Make it easy to obtain the transaction hash if eth_sendrawSync failed
### Component

provider, pubsub

ref https://eips.ethereum.org/EIPS/eip-7966

is supposed to return an error with the tx hash:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": 5,
    "message": "The transaction was added to the mempool but wasn't processed due to a missing nonce or insufficient funds. Please double-check your account and try again.",
    "data": "0x1234abcd..."
  }
}
```

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/provider/src/provider/trait.rs#L881-L884

if this returns an error:

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/transport/src/error.rs#L11-L11

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/json-rpc/src/error.rs#L5-L9

then we want an additional helper to parse the txhash from the data

https://github.com/alloy-rs/alloy/blob/2f84d1c3923c8c685e82fa14cf4d513acb59bfb0/crates/json-rpc/src/error.rs#L135-L141

## TODO

* add helper fn to `RpcError`
* add example how to recover the txhash to fn docs

### Additional context

_No response_

stevencartavia:
## Motivation
closes #3201

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/json-rpc/src/error.rs
+++ b/crates/json-rpc/src/error.rs
@@ -154,17 +154,11 @@ impl<E> RpcError<E, Box<RawValue>> {
     /// Parses the error data field as a hex string of specified length.
     ///
     /// Returns `Some(T)` if the data contains a valid hex string of the expected length.
-    fn parse_data<T: std::str::FromStr>(&self, expected_hex_len: usize) -> Option<T> {
+    fn parse_data<T: std::str::FromStr>(&self) -> Option<T> {
         let error_payload = self.as_error_resp()?;
         let data = error_payload.data.as_ref()?;
         let data_str = data.get().trim_matches('"').trim();
-        let hex_str = data_str.strip_prefix("0x").unwrap_or(data_str);
-
-        if hex_str.len() == expected_hex_len {
-            hex_str.parse().ok()
-        } else {
-            None
-        }
+        data_str.parse().ok()
     }
 
     /// Extracts a transaction hash from the error data field.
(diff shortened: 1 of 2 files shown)


crates/json-rpc/src/error.rs
```diff
@@ -148,3 +149,55 @@ impl<E, ErrResp> RpcError<E, ErrResp> {
...
+    /// // Simulate an EIP-7966 error response
+    /// let json = r#"{"code":5,"message":"insufficient funds","data":"0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"}"#;
+    /// let error_payload: ErrorPayload = serde_json::from_str(json).unwrap();
+    /// let rpc_error: RpcError<(), _> = RpcError::ErrorResp(error_payload);
+    ///
+    /// if let Some(tx_hash) = rpc_error.extract_transaction_hash() {
+    ///     println!("Transaction hash: {}", tx_hash);
+    /// }
+    /// ```
+    pub fn extract_transaction_hash(&self) -> Option<B256> {
```
mattsse: what we can do here is somthing like

```
fn pase_data<T:FromStr>
```

and then rename this to just 

```
fn tx_hash_data
```

that calls this

crates/json-rpc/src/error.rs
```diff
@@ -148,3 +149,55 @@ impl<E, ErrResp> RpcError<E, ErrResp> {
...
+    /// use alloy_json_rpc::{RpcError, ErrorPayload};
+    /// use alloy_primitives::B256;
+    ///
+    /// // Simulate an EIP-7966 error response
+    /// let json = r#"{"code":5,"message":"insufficient funds","data":"0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"}"#;
+    /// let error_payload: ErrorPayload = serde_json::from_str(json).unwrap();
+    /// let rpc_error: RpcError<(), _> = RpcError::ErrorResp(error_payload);
+    ///
+    /// if let Some(tx_hash) = rpc_error.extract_transaction_hash() {
+    ///     println!("Transaction hash: {}", tx_hash);
```
mattsse, line 183: these are great, but we should also add this to the _sync fns because we wont to highlight this there

mattsse (changes requested):
cool, some suggestions

mattsse (approved):
lgtm

I removed the string error variant, that we dont need here

PR #3203, opened 2025-11-15
docs(rpc-types-engine): correct ExecutionPayloadV3 spec URL and Execu…

jenniferzelo:
Corrects `ExecutionPayloadV3` doc to reference the Cancun V3 spec instead of Shanghai V2 spec, as V3 was introduced in the Cancun hardfork 
Fixes `ExecutionPayload::try_into_block` doc that incorrectly referenced `ExecutionPayloadV1` when the method operates on the `ExecutionPayload` enum.


mattsse (approved):
fine

PR #3204, opened 2025-11-15
fix(provider): prevent u64 overflow when required_confirmations is 0

maradini77:
Fixes a potential infinite hang when `required_confirmations` is set to `0` in `PendingTransactionConfig`. The calculation `block_height + confirmations - 1` would overflow to `u64::MAX`, causing the watcher to never trigger.


DaniPopes:
block height is always non zero

PR #3206, opened 2025-11-16
fix: eliminate ambiguity regarding missing subscriptions

Forostovec:
This change corrects the error semantics for PubSub subscription retrieval by modifying GetSub to carry an Option over its oneshot channel, allowing the service to send Some when a subscription exists and None when it does not. The frontend now maps None to a custom “subscription not found” TransportError and preserves BackendGone strictly for oneshot cancellations that reflect a true backend or service failure. This aligns behavior with the service’s documented intent, enables consumers to properly distinguish between missing subscriptions and infrastructure failures, and maintains compatibility with existing provider code paths without expanding the transport error enum.

### DIFF
--- a/crates/pubsub/src/frontend.rs
+++ b/crates/pubsub/src/frontend.rs
@@ -42,10 +42,9 @@ impl PubSubFrontend {
             backend_tx
                 .send(PubSubInstruction::GetSub(id, tx))
                 .map_err(|_| TransportErrorKind::backend_gone())?;
-            rx.await.map_err(|_| TransportErrorKind::backend_gone())?.map_or_else(
-                || Err(TransportErrorKind::custom_str("subscription not found")),
-                |sub| Ok(sub),
-            )
+            rx.await
+                .map_err(|_| TransportErrorKind::backend_gone())?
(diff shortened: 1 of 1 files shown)


crates/pubsub/src/service.rs
```diff
@@ -123,14 +123,10 @@ impl<T: PubSubConnect> PubSubService<T> {
...
-    /// `tx` is dropped. This notifies the waiter that the subscription does
-    /// not exist.
-    fn service_get_sub(&self, local_id: B256, tx: oneshot::Sender<RawSubscription>) {
-        if let Some(rx) = self.subs.get_subscription(local_id) {
-            let _ = tx.send(rx);
-        }
+    /// If the subscription exists, the waiter is sent `Some` broadcast receiver. If
+    /// the subscription does not exist, the waiter is sent `None`.
+    fn service_get_sub(&self, local_id: B256, tx: oneshot::Sender<Option<RawSubscription>>) {
+        let _ = tx.send(self.subs.get_subscription(local_id));
```
mattsse, line 129: this is better

PR #3211, opened 2025-11-18
Fix/fallback sequential for sync methods

### LINKED ISSUE #3186: [Bug] FallbackLayer breaks stateful RPC methods such as eth_sendRawTransactionSync due to parallel fan-out
### Component

transports

1.1.0

### Operating System

macOS (Apple Silicon)

### Describe the bug

FallbackLayer performs parallel fan-out to all transports for every request.
This behavior is correct for idempotent/stateless RPC methods, but breaks stateful methods such as `eth_sendRawTransactionSync` (EIP-7966), or any other RPC with side effects or synchronous waiting semantics.

The result is incorrect responses, especially when handling "already known" errors, and violates the semantics of these RPCs.

## Problem

With multiple transports:
1. Transport A receives the tx, processes it, and returns a success receipt
2. Transport B receives the tx, but because the hash already exists, returns "already known"
3. "already known" returns faster and is not a transport error, so FallbackLayer picks it
4. The actual correct receipt from transport A is discarded
5. Client receives "already known" instead of the real receipt

This makes eth_sendRawTransactionSync unusable with FallbackLayer.

## Reproduction Steps
1. Configure FallbackLayer with two HTTP transports, connected to RPC endpoints implementing `eth_sendRawTransactionSync`
2. Submit a valid signed tx
3. Observe that Alloy returns "already known"

CC @Syzygy106 as you have contributed eth_sendRawTransactionSync recently.

Syzygy106:
## Motivation

Fixes #3186

The `FallbackLayer` implementation was using parallel fan-out for all RPC methods, which caused issues with methods that have non-deterministic results when executed concurrently across multiple nodes.

Specifically, `eth_sendRawTransactionSync` (EIP-7966) waits for a transaction receipt. When executed in parallel, for example:
- **Node A (slow)**: Processes the transaction and returns the receipt after 100ms
- **Node B (fast)**: Recognizes the transaction as "already known" and returns an error after 10ms

The `FallbackLayer` would incorrectly return the fast "already known" error instead of waiting for the actual receipt from the slower node.

## Solution

Introduced selective execution strategies based on result determinism:

1. **Added `requires_sequential_execution()` function** - Identifies RPC methods that return non-deterministic results when executed in parallel:
   - `eth_sendRawTransactionSync` (EIP-7966)
   - `eth_sendTransactionSync`

2. **Added `make_request_sequential()` method** - Implements sequential fallback that tries transports one-by-one until success, respecting the ranking system.

3. **Modified `make_request()`** - Routes non-deterministic methods to sequential execution while keeping parallel execution for all other methods (including `eth_sendRawTransaction`, which returns a deterministic transaction hash).

**Key Design Decision**: Not all stateful methods require sequential execution. Methods like `eth_sendRawTransaction` are stateful but return deterministic results (transaction hashes), making parallel execution safe and beneficial for performance.

### Representation - 
```text
/// # Before Fix (Parallel Execution):
/// Time       Transport A               Transport B
/// ─────────────────────────────────────────────────────────
/// 0ms        🚀 START                  🚀 START (parallel!)
///            │                         │
/// 10ms       │ Processing...           ✓ "already_known"
///            │                         └──> Returns ❌ WRONG!
/// 50ms       ✓ Receipt {status: 0x1}
///            └──> Discarded (too late)
/// 
///
/// # After Fix (Sequential Execution - Success Case):
/// 
/// Time       Transport A               Transport B
/// ─────────────────────────────────────────────────────────
/// 0ms        🚀 START
///            │
/// 10ms       │ Processing...           (not called yet)
///            │
/// 50ms       ✓ Receipt {status: 0x1}  (not called - A succeeded!)
///            └──> Returns ✅ CORRECT!
/// 
///
/// # After Fix (Sequential Execution - Fallback Case):
/// 
/// Time       Transport A               Transport B
/// ─────────────────────────────────────────────────────────
/// 0ms        🚀 START
///            │
/// 10ms       │ Processing...           (waiting...)
///            │
/// 50ms       ❌ Connection timeout
///            │
/// 51ms       (A failed, trying B...)  🚀 START
///                                      │
/// 61ms                                 │ Processing...
///                                      │
/// 101ms                                ✓ Receipt {status: 0x1}
///                                      └──> Returns ✅ CORRECT!
```

## PR Checklist

- [x] Added Tests
  - `test_non_deterministic_method_uses_sequential_fallback` - Verifies `eth_sendRawTransactionSync` uses sequential execution and returns correct receipt
  - `test_deterministic_method_uses_parallel_execution` - Verifies `eth_sendRawTransaction` continues using parallel execution for performance
- [x] Added Documentation
  - Comprehensive inline comments explaining the logic
  - Function documentation for new methods
- [ ] Breaking changes

### DIFF
--- a/crates/transport/src/layers/fallback.rs
+++ b/crates/transport/src/layers/fallback.rs
@@ -46,18 +46,21 @@ impl<S: Clone> FallbackService<S> {
     /// Uses the default set of sequential methods (eth_sendRawTransactionSync,
     /// eth_sendTransactionSync).
     pub fn new(transports: Vec<S>, active_transport_count: usize) -> Self {
-        Self::with_sequential_methods(
+        Self::new_with_sequential_methods(
             transports,
             active_transport_count,
             default_sequential_methods(),
         )
     }
 
-    /// Create a new fallback service with custom sequential methods.
+    /// Create a new fallback service from a list of transports.
     ///
-    /// The `sequential_methods` parameter specifies which RPC methods require sequential execution
-    /// due to non-deterministic results in parallel execution.
-    pub fn with_sequential_methods(
+    /// The `active_transport_count` parameter controls how many transports are used for requests
+    /// at any one time.
+    ///
+    /// Uses the given set of sequential methods (eth_sendRawTransactionSync,
+    /// eth_sendTransactionSync).
+    pub fn new_with_sequential_methods(
         transports: Vec<S>,
         active_transport_count: usize,
         sequential_methods: HashSet<String>,
@@ -73,6 +76,21 @@ impl<S: Clone> FallbackService<S> {
             active_transport_count,
             sequential_methods: Arc::new(sequential_methods),
         }
+    }
+
+    /// Inserts the sequential method into the set.
+    pub fn append_sequential_method(mut self, sequential_method: impl Into<String>) -> Self {
+        let mut methods = Arc::unwrap_or_clone(self.sequential_methods);
+        methods.insert(sequential_method.into());
+        self.sequential_methods = Arc::new(methods);
+        self
+    }
+
+    /// Configures the `sequential_methods` parameter specifies which RPC methods require sequential
+    /// execution due to non-deterministic results in parallel execution.
+    pub fn with_sequential_methods(mut self, sequential_methods: HashSet<String>) -> Self {
+        self.sequential_methods = Arc::new(sequential_methods);
+        self
     }
 
     /// Log the current ranking of transports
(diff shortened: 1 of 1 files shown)


crates/transport/src/layers/fallback.rs
```diff
@@ -79,14 +146,31 @@ where
...
-    /// while keeping them available.
+    /// while ensuring correctness for methods that return different results in parallel.
     async fn make_request(&self, req: RequestPacket) -> Result<ResponsePacket, TransportError> {
+        // Check if this method returns non-deterministic results and needs sequential execution
+        let method = match &req {
+            RequestPacket::Single(r) => r.method(),
+            RequestPacket::Batch(reqs) => reqs.first().map_or("", |first| first.method()),
+        };
+
+        if requires_sequential_execution(method) {
```
mattsse: I think we should change this to `.any()` because even if there's a single sequential fn in a batch we should perform this sequentially

crates/transport/src/layers/fallback.rs
```diff
@@ -68,6 +68,73 @@ impl<S: Clone> FallbackService<S> {
...
+    matches!(
+        method,
+        // EIP-7966: eth_sendRawTransactionSync - waits for receipt
+        // Parallel issue: returns receipt from first node, "already known" from others
+        "eth_sendRawTransactionSync" |
+        // eth_sendTransactionSync - same as above but for unsigned transactions
+        // Parallel issue: returns receipt from first node, "already known" from others
+        "eth_sendTransactionSync"
+    )
+}
```
mattsse: can we make this a hashset in the fallbacklayer itself, with these values as the default, then a user can install addiitonal ones on demand

crates/transport/src/layers/fallback.rs
```diff
@@ -144,6 +228,58 @@ where
...
     }
+
+    /// Make a sequential request for methods with non-deterministic results.
+    ///
+    /// This method tries each transport one at a time, in order of their score.
+    /// It returns the first successful response, or an error if all transports fail.
+    ///
+    /// This approach ensures methods like `eth_sendRawTransactionSync` return the correct
+    /// receipt instead of "already known" errors from parallel execution.
+    async fn make_request_sequential(
```
mattsse, line 213: this makes sense

crates/transport/src/layers/fallback.rs
Syzygy106: > I think we should change this to `.any()` because even if there's a single sequential fn in a batch we should perform this sequentially

Yep, that's make sense

crates/transport/src/layers/fallback.rs
Syzygy106: > can we make this a hashset in the fallbacklayer itself, with these values as the default, then a user can install addiitonal ones on demand

ok, let me think about it

PR #3215, opened 2025-11-18
refactor: remove redundant trait bounds in CloneTransport impl

conomist:
What Changed
crates/transport/src/boxed.rs
Lines: 79-82
Removed + Send + Sync from the where clause of impl<T> CloneTransport for T
The Transport trait already includes Send + Sync + 'static as supertraits (defined in trait.rs:36-44), making these bounds redundant when specified again in the implementation block.


mattsse (commented):
these are required

PR #3218, opened 2025-11-19
add conditional compilation for Windows-specific test imports

0xLogicalx:
Fix clippy unused import warnings by adding conditional compilation attributesto platform-specific test imports.

### DIFF
--- a/crates/provider/src/provider/web3_signer.rs
+++ b/crates/provider/src/provider/web3_signer.rs
@@ -69,21 +69,15 @@ impl<P: Provider<N> + Clone, N: Network> Web3Signer<P, N> {
     }
 }
 
-#[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/web3_signer.rs
```diff
@@ -71,10 +71,15 @@ impl<P: Provider<N> + Clone, N: Network> Web3Signer<P, N> {
 
 #[cfg(test)]
 mod tests {
+    #[cfg(not(windows))]
```
onbjerg: just hoist this to the module level since there are no windows tests in here

0xLogicalx: Hoisted the #[cfg(not(windows))] to the module level so the whole tests module is skipped on Windows.

PR #3221, opened 2025-11-19
feat(provider): add `fill_transaction` method to Provider trait

### LINKED ISSUE #3220: [Feature] Add eth_fillTransaction to provider trait
### Component

provider, pubsub

now that we have the rpc type we can add it to the trait fn as well, T should be N::Transaction I think

cc @mablr 

### Additional context

_No response_

mablr:
## Motivation

Close #3220 

## Solution

add `fill_transaction` method to Provider trait

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,38 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.2](https://github.com/alloy-rs/alloy/releases/tag/v1.1.2) - 2025-11-20
+
+### Bug Fixes
+
+- Use BlockHash for ForkedNetwork.fork_block_hash ([#3224](https://github.com/alloy-rs/alloy/issues/3224))
+- [rpc-types] Correct BeaconExecutionPayloadV3 doc ([#3216](https://github.com/alloy-rs/alloy/issues/3216))
+- Eliminate ambiguity regarding missing subscriptions ([#3206](https://github.com/alloy-rs/alloy/issues/3206))
+- Normalize recovery bytes in flashbots signature ([#3192](https://github.com/alloy-rs/alloy/issues/3192))
+- Increase average CU cost from `17` to `20` per updated Alchemy docs  ([#3208](https://github.com/alloy-rs/alloy/issues/3208))
+- Trezor derivation path ([#3148](https://github.com/alloy-rs/alloy/issues/3148))
+
+### Dependencies
+
+- [deps] Bump taiki-e/install-action from 2.62.45 to 2.62.49 ([#3173](https://github.com/alloy-rs/alloy/issues/3173))
+
+### Documentation
+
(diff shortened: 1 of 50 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -1084,6 +1084,18 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
 
+    /// Fills a transaction with missing fields using default values.
+    ///
+    /// This method prepares a transaction by populating missing fields such as gas limit,
+    /// gas price, or nonce with appropriate default values. The response includes both the
+    /// RLP-encoded signed transaction and the filled transaction.
+    async fn fill_transaction(
+        &self,
+        tx: N::TransactionRequest,
+    ) -> TransportResult<FillTransaction> {
```
mablr: Unsure about bounds here :/

mattsse: ah this should be

https://github.com/alloy-rs/alloy/blob/c3b38a874057a6a87e3a4389178b4ac0643e169a/crates/network/src/ethereum/mod.rs#L19-L19

N::Unsigned I believe

right @klkvr ?

mablr: hum that's unfortunate because setting `TransportResult<FillTransaction<N::UnsignedTx>>` fails because it doesn't implement `Debug`.

But there is a lint that enforces `Debug` impl for `FillTransaction` :/

mablr: Btw `N::Unsigned` is incompatible with `RpcSend`.

`FillTransaction<N::TransactionResponse>` seems to work, is this suitable ?

klkvr: I think this actually needs to be `N::TxEnvelope`

for missing bounds we can probably enforce them directly on `Provider::fill_transaction`

maybe even directly on `Network` but that would likely be breaking (even though envelope should always be `RpcSend`)

mattsse: @mablr let's try Envelope and add an anvil test, this now supports it so we can see if it works
for missing debug, if required we can def add this to the trait bound on Network

mablr: ok just seen your comment, i'll do that 👍 

(ignore the last commit for the moment)

mablr: done 👍 

Added `RcpObject` trait bound on Network, it doesn't seem to break anything.
Added a simple test as well.

crates/network/src/lib.rs
```diff
@@ -62,7 +62,7 @@ pub trait Network: Debug + Clone + Copy + Sized + Send + Sync + 'static {
 
     /// The network transaction envelope type.
     #[doc(alias = "TransactionEnvelope")]
-    type TxEnvelope: Eip2718Envelope + Transaction + Debug;
+    type TxEnvelope: Eip2718Envelope + Transaction + Debug + RpcObject;
```
klkvr: this is technically breaking and I think we can just enforce this bound here directly:
https://github.com/alloy-rs/alloy/blob/12cf8658f79fab04b86557036fcc8f81d372676b/crates/provider/src/provider/erased.rs#L383-L386

like `fn fill_transaction(...) where N::TxEnvelope: RpcSend`

mablr: Ok i’ll revert the last change

PR #3222, opened 2025-11-20
Remove heartbeat check and return Err on unsigned tx

hazelnutcloud:
## Motivation

https://github.com/alloy-rs/alloy/pull/3177#discussion_r2544577731
https://github.com/alloy-rs/alloy/pull/3177#discussion_r2544666131

## Solution

- Removed heartbeat check as no subscription is made with `sendRawTransactionSync`. Receipt is immediately returned from RPC.
- Removed call to `sendTransactionSync`. EIP-7966 does not support unsigned transactions. Return LocalUsage error instead.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -1059,15 +1059,9 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
         &self,
         tx: SendableTx<N>,
     ) -> TransportResult<N::ReceiptResponse> {
-        // Make sure to initialize heartbeat before we submit transaction, so that
-        // we don't miss it if user will subscriber to it immediately after sending.
-        let _handle = self.root().get_heart();
-
         match tx {
-            SendableTx::Builder(mut tx) => {
-                alloy_network::TransactionBuilder::prep_for_submission(&mut tx);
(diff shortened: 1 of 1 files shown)


crates/provider/src/provider/trait.rs
```diff
@@ -1059,15 +1059,9 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
         tx: SendableTx<N>,
     ) -> TransportResult<N::ReceiptResponse> {
-        // Make sure to initialize heartbeat before we submit transaction, so that
-        // we don't miss it if user will subscriber to it immediately after sending.
-        let _handle = self.root().get_heart();
-
         match tx {
-            SendableTx::Builder(mut tx) => {
-                alloy_network::TransactionBuilder::prep_for_submission(&mut tx);
-                let receipt = self.client().request("eth_sendTransactionSync", (tx,)).await?;
```
mattsse, line 1069: oh I missed that this was also removed,
this we need to add back

mattsse (changes requested):
also needs fmt

mattsse:
supersed by https://github.com/alloy-rs/alloy/pull/3297

PR #3227, opened 2025-11-21
Fix block stream retry logic for sequential fetching

sashaodessa:
retry each block number in `NewBlocks` before advancing to the next; break out of the fill loop only after exhausting retries for that block

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,7 @@
 [profile.default]
 retries = { backoff = "exponential", count = 2, delay = "2s", jitter = true }
 slow-timeout = { period = "30s", terminate-after = 4 }
+
+[[profile.default.overrides]]
+filter = "test(can_launch_reth_custom_ports)"
+retries = { backoff = "exponential", count = 5, delay = "3s", jitter = true }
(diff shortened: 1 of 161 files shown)


crates/provider/src/blocks.rs
```diff
@@ -193,30 +193,35 @@ impl<N: Network> NewBlocks<N> {
...
+                            break None;
+                        }
+                        Ok(None) => {
+                            error!(number, "failed to fetch block (doesn't exist)");
+                            break None;
+                        }
                     }
                 };
+                let Some(block) = block else {
+                    break 'fill;
```
mattsse, line 223: I think this should just continue here?

sashaodessa: > I think this should just continue here?

@mattsse We need to break out of the loop once the per-number retries are exhausted. Continuing here would skip over the missing block, leaving a hole at self.next_yield; newer blocks would pile up in the cache but couldn’t be yielded until that hole is filled, so we’d waste RPC calls and still block the stream. The previous implementation also break’d on an unrecoverable failure, so this keeps the same behaviour while making the retry budget actually per block.

mattsse:
Closing as superseded by #3698, which merged on February 17, 2026 and already fixes this retry logic on main.

PR #3232, opened 2025-11-22
feat: include DecodeError in SszDecodeError display

Bashmunta:
SszDecodeError previously formatted only the request_type and omitted the underlying ssz::DecodeError from the display output and error chaining, which made diagnosing SSZ decoding issues difficult. This change updates the error message to include the inner cause and annotates the second field with #[source] so that Error::source() returns the underlying DecodeError. The wording is also adjusted from “requests_type” to “request_type” for consistency with the surrounding enum. This aligns with existing practices in the codebase where inner causes are surfaced (e.g., Multicall DecodeError) and improves observability without altering the variant shape or call sites.

### DIFF
--- a/crates/rpc-types-beacon/src/requests.rs
+++ b/crates/rpc-types-beacon/src/requests.rs
@@ -169,8 +169,8 @@ mod ssz_requests_conversions {
         #[error("unknown request_type prefix: {0}")]
         UnknownRequestType(u8),
         /// Remaining bytes could not be decoded as SSZ requests_data.
-        #[error("ssz error decoding request_type {0}: {1}")]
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this makes sense

mattsse:
@Bashmunta this doesn't compile just yet

Bashmunta:
> @Bashmunta this doesn't compile just yet

gonna work on it

Bashmunta:
> @Bashmunta this doesn't compile just yet

A resolution like this will be alright? 
`#[error("ssz decode error for request_type {0}: {1:?}")]`
Sorry for long answer

mattsse (approved):

PR #3236, opened 2025-11-23
feat: switch parsing to split_once/strip_suffix to avoid allocations

VolodymyrBg:
Replaced multiple split(...).collect::<Vec<_>>() usages in TxpoolInspectSummaryVisitor::visit_str() with split_once and strip_suffix, eliminating unnecessary intermediate Vec allocations and making the parsing intent explicit. The refactor preserves existing error messages and behavior while tightening validation by requiring the final segment to end exactly with " wei", which matches the serializer’s format and improves robustness without altering expected successful inputs or tests.


DaniPopes:
doesnt matter

PR #3239, opened 2025-11-24
ci: bump actions/checkout to v6

rejected-l:
Bumps to v6 for Node.js 24 support. Requires runner v2.329.0+.

https://github.com/actions/checkout/releases/tag/v6.0.0


zerosnacks:
Thanks but we have Dependabot configured for this

PR #3243, opened 2025-11-24
fix(throttle): preserve Tower backpressure by swapping inner with mem::replace

Bashmunta:
Use std::mem::replace instead of a clone in ThrottleService::call() so that the request executes exactly the same instance of inner on which poll_ready() was called. This preserves Tower backpressure and does not lose permits (unlike calling on a clone).


mattsse (commented):
yeah this doesnt matter

PR #3248, opened 2025-11-24
fix: guard test-only imports to eliminate Windows clippy warnings

justinbay:
gate TxKind under the features that actually use it in crates/consensus/src/transaction/envelope.rs, scope provider test imports so Windows builds skip the reth-dependent code in crates/provider/src/provider/trait.rs, disable the web3_signer test module on Windows to avoid unused imports when reth isn’t built


mattsse (commented):
we can keep as is

PR #3256, opened 2025-11-26
Scope provider test-only imports and trim redundant block cloning

0xLogicalx:
gate provider test utilities on non-Windows builds to avoid unused-import warnings, pull in alloy_consensus::Transaction only, where needed so trait methods resolve cleanly, reduce clone overhead in block RPC tests by cloning just the header payload, guard web3-signer tests with cfg(all(test, not(windows))) so they only build where dependencies exist


DaniPopes:
please stop doing this, its not helping anyone

PR #3267, opened 2025-11-28
chore: remove unnecessary Unpin bound on decoder in EthCall futures

Snezhkko:
The Unpin bound on the generic decoder type D in IntoFuture for EthCall and Future for EthCallFut was unnecessary because D is only ever referenced via &D and is neither pinned nor moved across await boundaries. The inner RPC future is the only pinned entity and is handled via pin! on the local binding. Removing the bound widens generic compatibility for non-Unpin decoders without affecting safety or behavior. Cross-checking provider’s EthCall implementation shows no analogous bound is required, confirming that this constraint was overly strict rather than intentional API design.

### DIFF
--- a/crates/contract/src/eth_call.rs
+++ b/crates/contract/src/eth_call.rs
@@ -121,7 +121,7 @@ where
 
 impl<'coder, D, N> std::future::IntoFuture for EthCall<'coder, D, N>
 where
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this seems okay

PR #3273, opened 2025-11-29
feat(provider): add `verify_flashbots_signature` function

cakevm:
## Motivation

Implement the counterpart for `sign_flashbots_payload` and allow to verify a signature. The fn is exposed for everyone to use.
This allows also to have a roundtrip test and verify both directions. From https://github.com/alloy-rs/alloy/pull/3192 originates to have different variations for the recovery bytes. 

## Solution

I added `sign_flashbots_payload` fn. I changed the signature for `sign_flashbots_payload` to accept `body: impl AsRef<[u8]>` since ownership was not required and this is backward compatible to `String` but gives most flexibility. This helps to have for both fn's a similar signature. For `verify_flashbots_signature` this has the advantage for verify without the need to check for valid UTF-8 (`&str`), and just use `&[u8]`.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/mev/with_auth.rs
+++ b/crates/provider/src/ext/mev/with_auth.rs
@@ -108,7 +108,7 @@ where
                 // Generate the Flashbots signature for the request body
                 let body = serde_json::to_string(&self.inner.request())
                     .map_err(TransportErrorKind::custom)?;
-                let signature = sign_flashbots_payload(body.as_bytes(), &signer)
+                let signature = sign_flashbots_payload(body, &signer)
                     .await
                     .map_err(TransportErrorKind::custom)?;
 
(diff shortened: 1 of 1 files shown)


crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -131,20 +153,112 @@ pub async fn sign_flashbots_payload<S: Signer + Send + Sync>(
...
+/// Verifies a Flashbots signature and returns the recovered signer address.
+///
+/// The signature format is `{address}:{signature_hex}` as produced by
+/// [`sign_flashbots_payload`]. Both normalized (v=0/1) and EIP-155 (v=27/28)
+/// recovery bytes are supported.
+///
+/// See [Flashbots docs](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication) for more information.
+pub fn verify_flashbots_signature(
+    signature_header: &str,
+    body: impl AsRef<[u8]>,
```
onbjerg: imo let's do `&[u8]` instead, we can always make it more generic later. the current impl monomorphizes pretty badly

crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -131,20 +153,112 @@ pub async fn sign_flashbots_payload<S: Signer + Send + Sync>(
     Ok(format!("{}:{}", signer.address(), hex::encode_prefixed(sig_bytes)))
 }
 
+/// Verifies a Flashbots signature and returns the recovered signer address.
+///
+/// The signature format is `{address}:{signature_hex}` as produced by
+/// [`sign_flashbots_payload`]. Both normalized (v=0/1) and EIP-155 (v=27/28)
+/// recovery bytes are supported.
+///
+/// See [Flashbots docs](https://docs.flashbots.net/flashbots-auction/advanced/rpc-endpoint#authentication) for more information.
```
onbjerg, line 178: can you add an example as well?

crates/provider/src/ext/mev/with_auth.rs
cakevm: Many thanks for the hint. Did not thought about that. Makes sense. Corrected it for both fn to accept &[u8]. So both have a similar signature.

crates/provider/src/ext/mev/with_auth.rs
cakevm: I added for both examples. Both have asserts that they could fail in doc tests if not working. Second the first one is async and need a async fn around it to work.

cakevm:
I gave it a second thought and reverted the changes to `sign_flashbots_payload`. I did not wanted to blow the scope here too much and leave it as it is. I am only adding now `verify_flashbots_signature` and improving the docs/examples.

PR #3274, opened 2025-11-29
fix(consensus): silence unused generic param in Recovered::try_convert

Galoretka:
Rename the unused generic parameter E to _E in Recovered::try_convert and the deprecated try_convert_transaction. This keeps behavior and the public API intact and aligns with similar try_convert methods elsewhere.


Galoretka:
> yeah we dont actually need this, so we could remove entirely

you're fast really, thanks for review

Galoretka:
faster than light

Issue #3264, opened 2025-11-27
[Feature] Introduce MinedTransactionInfo


mattsse:
### Component

provider, pubsub

### Describe the feature you would like

basically a subset of

https://github.com/alloy-rs/alloy/blob/1cee6d808d9fec09f41ba1dea7d49af8da20ed67/crates/consensus/src/transaction/meta.rs#L5-L9

that includes

block number +  hash + index

## TODO
* add new meta helper struct
* introduce helper trait for `TransactionResponse` to obtain this

this is mostly for convenience to deal with multiple optional fields

dipanshuhappy:
Happy to work on this

onbjerg:
closed in https://github.com/alloy-rs/alloy/pull/3275

PR #3275, opened 2025-11-30
[Feature] Introduce MinedTransactionInfo

developeruche:
## Motivation
Cloeses #3264

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network-primitives/src/traits.rs
+++ b/crates/network-primitives/src/traits.rs
@@ -148,10 +148,10 @@ pub trait TransactionResponse: Transaction {
         }
     }
 
-    /// Returns the [`InclusionInfo`] if the transaction has been mined.
+    /// Returns the [`InclusionInfo`] if the transaction has been included.
     ///
     /// Returns `None` if this transaction is still pending (missing block number, hash, or index).
-    fn mined_info(&self) -> Option<InclusionInfo> {
+    fn inclusion_info(&self) -> Option<InclusionInfo> {
         Some(InclusionInfo {
             block_hash: self.block_hash()?,
             block_number: self.block_number()?,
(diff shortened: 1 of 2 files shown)


crates/network-primitives/src/tx_meta.rs
```diff
@@ -0,0 +1,12 @@
+use alloy_primitives::B256;
+
+/// Additional fields in the context of a block that contains this _mined_ transaction.
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
```
onbjerg:
```suggestion
#[derive(Debug, Clone, Default, Eq, PartialEq)]
```

i dont think this should be copy

crates/network-primitives/src/traits.rs
```diff
@@ -147,6 +147,17 @@ pub trait TransactionResponse: Transaction {
             ty => Some(ty),
         }
     }
+
+    /// Returns the [`InclusionInfo`] if the transaction has been mined.
+    ///
+    /// Returns `None` if this transaction is still pending (missing block number, hash, or index).
+    fn mined_info(&self) -> Option<InclusionInfo> {
```
onbjerg:
```suggestion
    /// Returns the [`InclusionInfo`] if the transaction has been included.
    ///
    /// Returns `None` if this transaction is still pending (missing block number, hash, or index).
    fn inclusion_info(&self) -> Option<InclusionInfo> {
```

crates/network-primitives/src/tx_meta.rs
```diff
@@ -0,0 +1,12 @@
+use alloy_primitives::B256;
+
+/// Additional fields in the context of a block that contains this _mined_ transaction.
```
onbjerg:
```suggestion
/// Additional fields in the context of a block that contains an included transaction.
```

onbjerg (changes requested):

PR #3286, opened 2025-12-02
chore(node-bindings): fix incorrect tool name in Geth documentation

jenniferzelo:
Changed references from `geth-cli` to `geth` in the documentation for `Geth::port()` and `Geth::block_time()` methods.  The tool name is `geth`, not `geth-cli`, as defined in the codebase constant.


mattsse (commented):
we can keep this as is

Issue #3288, opened 2025-12-03
[Bug] wasm32-wasip1 panics when using transport `spawn`


Robert-MacWha:
### Component

transports

### What version of Alloy are you on?

1.1.2

### Operating System

Linux

### Describe the bug

For wasm build targets (and with the `wasm-bindgen` feature flag) alloy uses wasm-bindgen-futures and wasmtimer.  However, for wasm32-wasip1 targets, these two crates are unavailable since they're specifically designed for integration with JS environments.

The correct solution for wasm32-wasip1 would be using tokio, same as in non-wasm environments.  Tokio has [limited wasm support](https://docs.rs/tokio_wasi/latest/tokio/#wasm-support) but enough for alloy's needs.

Robert-MacWha:
I implemented a fix for this here: https://github.com/alloy-rs/alloy/pull/3289

PR #3289, opened 2025-12-03
Wasm32 wasip support

Robert-MacWha:
## Motivation

Fixes https://github.com/alloy-rs/alloy/issues/3288

## Solution

Switching around various target flags so `wasm32-wasip1` uses tokio instead of `wasm-bindgen-futures` or `wasmtimer`.

I'm not sure how best to handle the `wasm-bindgen` feature flag.  We can leave it as only valid for wasm32-unknown-unknown targets (in which case it'll require a modification to `crates/transport/src/utils.rs`) or can remove it entirely.  Depends on the original motivation for introducing the flag I guess.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport/src/utils.rs
+++ b/crates/transport/src/utils.rs
@@ -32,7 +32,8 @@ pub trait Spawnable {
     /// Spawn the future as a task.
     ///
     /// In wasm32-unknown-unknown this will be a `wasm-bindgen-futures::spawn_local` call,
-    /// in wasm32-wasip1 and native it will be a `tokio::spawn` call.
+    /// in wasm32-wasip1 it will be a `tokio::task::spawn_local` call,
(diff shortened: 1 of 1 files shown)


DaniPopes:
makes sense, thanks
nit: can you use the same cfg in both "branches" and just negate it in the "else"? `all(wasm, unknown) {}; not(all(wasm, unknown))`

Robert-MacWha:
Updated.  I also tried running the failing workflows locally and after the commit they all seemed to pass.  If it's still failing I'll take a closer look.  I'm also not using rust nightly, so that might be the difference.

DaniPopes (approved):
thanks!

PR #3291, opened 2025-12-03
docs: add an example in alloy meta crate

onbjerg:
## Motivation

We want to give a small sampler of how to use Alloy, while also being able to link out to more examples.

## Solution

Add a simple example of using a provider, and a simple example of using a provider using a different network type.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,49 @@ For a more fine-grained control over the features you wish to include, you can a
 
 A comprehensive list of available features can be found on [docs.rs](https://docs.rs/crate/alloy/latest/features) or in the [`alloy` crate's `Cargo.toml`](https://github.com/alloy-rs/alloy/blob/main/crates/alloy/Cargo.toml).
 
+## Examples
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,6 +33,49 @@ For a more fine-grained control over the features you wish to include, you can a
...
+
+Here's a simple example of connecting to an Ethereum node and querying the latest block:
+
+```rust
+use alloy::providers::{Provider, ProviderBuilder};
+
+# async fn example() -> Result<(), Box<dyn std::error::Error>> {
+// Create a provider with the HTTP transport using the `reqwest` crate.
+let rpc_url = "https://eth.llamarpc.com".parse()?;
+let provider = ProviderBuilder::new().connect_http(rpc_url);
```
DaniPopes: can use .connect().await to avoid specifying http (detected)

PR #3296, opened 2025-12-03
feat: make BuiltInConnectionString::connect configurable

### LINKED ISSUE #3294: [Feature] make `BuiltInConnectionString::connect` configurable
### Component

transports

https://github.com/alloy-rs/alloy/blob/54db44a9442804ff60cec407a187098a78806db0/crates/rpc-client/src/builtin.rs#L67-L73

ideally we another fn that can provide additional config settings

e.g. 

https://github.com/alloy-rs/alloy/blob/54db44a9442804ff60cec407a187098a78806db0/crates/transport-ws/src/native.rs#L24-L33

or for http headers, etc.

## TODO
* introduce a general purpose settings/config struct with a subset of connection settings
* introduce `BuiltInConnectionString::connect_with` that accepts the settings and forwards the relevant one to the transport

### Additional context

_No response_

stevencartavia:
## Motivation
closes #3294

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-client/src/builtin.rs
+++ b/crates/rpc-client/src/builtin.rs
@@ -20,62 +20,6 @@ pub enum BuiltInConnectionString {
     Ipc(std::path::PathBuf),
 }
 
-/// Configuration for connecting to built-in transports.
-///
-/// Provides a flexible way to configure various aspects of the connection,
-/// including authentication, retry behavior, and transport-specific settings.
-#[derive(Clone, Debug, Default)]
-#[non_exhaustive]
-pub struct ConnectionConfig {
-    /// Authorization header for authenticated connections.
-    pub auth: Option<alloy_transport::Authorization>,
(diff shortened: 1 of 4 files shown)


crates/rpc-client/src/builtin.rs
```diff
@@ -20,6 +20,62 @@ pub enum BuiltInConnectionString {
     Ipc(std::path::PathBuf),
 }
 
+/// Configuration for connecting to built-in transports.
+///
+/// Provides a flexible way to configure various aspects of the connection,
+/// including authentication, retry behavior, and transport-specific settings.
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub struct ConnectionConfig {
```
mattsse: this def is out of order and in between a struct and impl block

let moves this below all things BuiltInConnectionString

PR #3297, opened 2025-12-04
chore: move heartbeat handle to unsigned tx variant

mattsse:
supersedes https://github.com/alloy-rs/alloy/pull/3222


mattsse:
actually we should keep this in case the tx fails

PR #3304, opened 2025-12-07
refactor: remove unnecessary Encodable7594 bound in RlpEcdsaDecodableTx impls

jenniferzelo:
The `RlpEcdsaDecodableTx` implementations for `TxEip4844Variant<T>` and `TxEip4844WithSidecar<T>` only call decode methods (`T::decode_7594`),  never encode methods, making the `Encodable7594` bound redundant.


mattsse (approved):
makes sense

mattsse:
this is actually required

PR #3306, opened 2025-12-07
fix: Mark external mainnet ENS tests as ignored by default

anim001k:
Guard mainnet-dependent ENS tests with `#[ignore]` so they run only intentionally and do not break CI when external RPC is unavailable.


DaniPopes:
this is fine we have a lot of CI that may fail on external rpcs

PR #3314, opened 2025-12-09
fix: Enforce ValueTx for multicall batches with non-zero value

ANtutov:
Previously MulticallError::ValueTx was never emitted and aggregate/aggregate3 silently ignored non-zero value on inner calls. This makes multicall semantics around payable calls ambiguous and leaves ValueTx as effectively dead code. The change enforces the intended invariant by returning ValueTx from the shared build_and_call/build_and_send helpers whenever a batch contains calls with non-zero value but no total value is provided, and updates the error message to point users towards aggregate3_value/send_aggregate3_value for payable batches. New tests cover aggregate and aggregate3 with non-zero value to lock in the behavior.


mattsse (commented):
dont think we need this

PR #3315, opened 2025-12-09
Remove redundant Send/Sync bound from async sign_typed_data

splinter012:
Drop unnecessary Send + Sync bound from Signer::sign_typed_data to match the sync variant and accept non-Send/Sync payloads.


DaniPopes:
not the same, it makes the future non send

PR #3318, opened 2025-12-10
docs(provider): fix debug trace call option type

Forostovec:
Doc comments for the debug_trace_call* and debug_trace_call_many* helpers incorrectly referenced GethDebugTracingOptions while the actual signatures use GethDebugTracingCallOptions. This change updates the documentation to match the real type so that users are not misled when configuring trace call options. No behavior or RPC types are changed, only the comments are corrected.


Forostovec:
just a sec, got a bit confused

Forostovec:
> can you also add backticks? e.g. `` [`GethDebugTracingOptions`] ``

Hi, thank you for review, i've added backticks for GethDebugTracingCallOptions. and for GethDebugTracingOptions to keep consistency?

PR #3320, opened 2025-12-10
Remove no-op convert_to_hashes helper from provider utils

anim001k:
- drop the unused convert_to_hashes helper and its call sites in provider get_block flow
- rely on RPC `full` flag for hash-only responses instead of redundant post-processing
- keep block-hash retrieval behavior unchanged while simplifying utils


mattsse (commented):
we can keep this

PR #3321, opened 2025-12-10
feat: allow fusaka sidecars in the tx request

prestwich:
## Motivation

RPC types were not updated to allow fusaka in the `TxRequest` which means the tx building process can't accept fusaka sidecars.

## Solution

Modify the `TxRequest` and add a `TransactionBuilder7594` trait

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/network-primitives/src/tx_builders.rs
+++ b/crates/network-primitives/src/tx_builders.rs
@@ -55,7 +55,7 @@ pub trait TransactionBuilder7594: Default + Sized + Send + Sync + 'static {
     /// Get the max fee per blob gas for the transaction.
     fn max_fee_per_blob_gas(&self) -> Option<u128>;
 
-    /// Set the max fee per blob gas  for the transaction.
+    /// Set the max fee per blob gas for the transaction.
     fn set_max_fee_per_blob_gas(&mut self, max_fee_per_blob_gas: u128);
 
     /// Gets the EIP-7594 blob sidecar of the transaction.
--- a/crates/rpc-types-eth/src/transaction/request.rs
+++ b/crates/rpc-types-eth/src/transaction/request.rs
@@ -1094,12 +1094,10 @@ impl TransactionBuilder7702 for TransactionRequest {
 }
 
 impl TransactionBuilder7594 for TransactionRequest {
-    /// Get the max fee per blob gas for the transaction.
     fn max_fee_per_blob_gas(&self) -> Option<u128> {
         self.max_fee_per_blob_gas
     }
 
-    /// Set the max fee per blob gas  for the transaction.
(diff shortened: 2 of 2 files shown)


crates/network-primitives/src/tx_builders.rs
```diff
@@ -50,6 +50,27 @@ pub trait TransactionBuilder7702: Default + Sized + Send + Sync + 'static {
     }
 }
 
+/// Transaction builder type supporting EIP-7594 transaction fields.
+pub trait TransactionBuilder7594: Default + Sized + Send + Sync + 'static {
+    /// Get the max fee per blob gas for the transaction.
+    fn max_fee_per_blob_gas(&self) -> Option<u128>;
+
+    /// Set the max fee per blob gas  for the transaction.
```
onbjerg:
```suggestion
    /// Set the max fee per blob gas for the transaction.
```

smol extra space

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -1086,6 +1093,26 @@ impl TransactionBuilder7702 for TransactionRequest {
     }
 }
 
+impl TransactionBuilder7594 for TransactionRequest {
+    /// Get the max fee per blob gas for the transaction.
+    fn max_fee_per_blob_gas(&self) -> Option<u128> {
+        self.max_fee_per_blob_gas
+    }
+
+    /// Set the max fee per blob gas  for the transaction.
```
onbjerg: same
```suggestion
    /// Set the max fee per blob gas for the transaction.
```

prestwich:
i removed the rustdoc from the impl, as unnecessary to reproduce there

prestwich:
CI is flaking here?

crates/provider/src/fillers/gas.rs
```diff
@@ -332,7 +332,7 @@ mod tests {
         let tx = TransactionRequest {
             to: Some(address!("d8dA6BF26964aF9D7eEd9e03E53415D37aA96045").into()),
             max_fee_per_blob_gas: Some(0),
-            sidecar: Some(sidecar),
+            sidecar: Some(sidecar.into()),
```
mattsse, line 335: this is technically breaking but this likely only affects like 5 users that should migrate anyway

mattsse (approved):
looks like the only breaking change is on the pub field level and since this is an option it doesnt break : None inits, and for those users where this breaks they should migrate anyway

crates/provider/src/fillers/gas.rs
prestwich: yeah, the RPC spec had a breaking change, so the RPC implementation needs a breaking change 😮‍💨

PR #3330, opened 2025-12-11
feat: added ExecutionPayloadV4 and engine_newPayloadV5,engine_getPayloadV6

Rimeeeeee:
Added `ExecutionPayloadV4` and `engine_newPayloadV5`,` engine_getPayloadV6`.
 [ref ](https://github.com/paradigmxyz/reth/issues/20262)
cc: @mattsse

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -1370,6 +1370,7 @@ impl ssz::Decode for ExecutionPayloadV4 {
 }
 
 #[cfg(feature = "ssz")]
+#[cfg(feature = "amsterdam")]
 impl ssz::Encode for ExecutionPayloadV4 {
     fn is_ssz_fixed_len() -> bool {
         false


crates/eips/Cargo.toml
```diff
@@ -27,6 +27,7 @@ workspace = true
 alloy-eip2124.workspace = true
 alloy-eip2930.workspace = true
 alloy-eip7702.workspace = true
+alloy-eip7928 = { workspace = true, features = ["rlp", "borsh"] }
```
mattsse: I dont think we need all features here?

crates/rpc-types-engine/src/payload.rs
```diff
@@ -1166,6 +1188,211 @@ impl TryFrom<BlobsBundleV2> for BlobTransactionSidecarEip7594 {
...
+    /// Converts [`alloy_consensus::Block`] to [`ExecutionPayloadV4`] using the given block hash.
+    ///
+    /// See also [`ExecutionPayloadV3::from_block_unchecked`].
+    // pub fn from_block_unchecked<T, H>(block_hash: B256, block: &Block<T, H>) -> Self
+    // where
+    //     T: Encodable2718,
+    //     H: BlockHeader,
+    // {
+    //     Self {
+    //         block_access_list: alloy_rlp::encode(block.body.block_access_list.clone().unwrap())
```
mattsse: let's just duplicate these and foraward to the inner V3, we can take care of the bal header hash separately

Cargo.toml
```diff
@@ -89,9 +89,10 @@ alloy-chains = { version = "0.2", default-features = false }
 alloy-eip2124 = { version = "0.2.0", default-features = false }
 alloy-eip2930 = { version = "0.2.3", default-features = false }
 alloy-eip7702 = { version = "0.6.3", default-features = false }
+alloy-eip7928 = { version = "0.1.0", default-features = false, git = "https://github.com/alloy-rs/eips.git" }
```
mattsse: I just published this

mattsse (changes requested):
nice, few nits

crates/rpc-types-beacon/src/payload.rs
```diff
@@ -435,9 +435,71 @@ pub mod beacon_payload_v3 {
...
+        D: Deserializer<'de>,
+    {
+        BeaconExecutionPayloadV4::deserialize(deserializer).map(Into::into)
+    }
+}
+
 /// Represents all possible payload versions.
 #[derive(Debug, Serialize)]
 #[serde(untagged)]
+#[cfg(not(feature = "amsterdam"))]
```
mattsse, line 502: instead of feature gating the type, can we instead feature gate the variant instead?

crates/consensus-any/src/block/header.rs
```diff
@@ -98,13 +99,376 @@ pub struct AnyHeader {
     pub requests_hash: Option<B256>,
 }
 
+#[cfg(not(feature = "amsterdam"))]
+impl AnyHeader {
```
mattsse: do we need this feature gated impl or can we instead feature gate the the access_list_hash usage instead?

PR #3338, opened 2025-12-12
Remove redundant Clone bound from Network trait

0xlupin:
Drop the redundant Clone supertrait from Network, keeping Copy + Send + Sync + 'static intact. Simplify implementor requirements with no behavioral change


DaniPopes:
Doesn't matter it's a ZST anyway

PR #3351, opened 2025-12-15
fix: header Sealable hash_slow recursion

maximevtush:
Calling `hash_slow` via `Sealable` on `Header` recursed into itself, causing an immediate stack overflow.


DaniPopes:
no

PR #3365, opened 2025-12-17
Clarify debug log on block height regression

vtjl10:
Adjust the debug log in the `NewBlocks` polling loop so that when `block_number < self.next_yield` it logs a “block height regressed” message with both current and expected heights. This makes it clear that the node returned an older block height (potential reorg or desync) instead of merely “not advancing”, improving observability and reducing confusion during incident analysis.


DaniPopes:
not true

PR #3367, opened 2025-12-17
Add a layer to alloy-transport-http that allows propagating trace information

prestwich:
## Motivation

Allow distributed tracing to track alloy requests through compatible RPC endpoints, by allowing users to add support for the `traceparent` header to the alloy http transport stack

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-http/Cargo.toml
+++ b/crates/transport-http/Cargo.toml
@@ -76,6 +76,9 @@ jwt-auth = [
     "dep:jsonwebtoken",
 ]
 traceparent = [
+    "dep:tower",
+    "dep:tracing",
+    "dep:alloy-json-rpc",
     "dep:opentelemetry-http",
     "dep:opentelemetry",
     "dep:tracing-opentelemetry",


crates/transport-http/Cargo.toml
```diff
@@ -43,8 +43,13 @@ hyper-tls = { workspace = true, optional = true }
...
 jsonwebtoken = { workspace = true, optional = true }
 
+# trace propagation
+opentelemetry-http = { version = "0.31.0", optional = true}
+tracing-opentelemetry = { version = "0.32.0", optional = true}
+opentelemetry = { version = "0.31.0", optional = true }
+
 [features]
-default = ["reqwest", "reqwest-default-tls"]
+default = ["reqwest", "reqwest-default-tls", "traceparent"]
```
zerosnacks, line 52: nit: `trace-parent` is less easy to misread as `transparent` which has specific meaning in Rust core & Serde

prestwich: [traceparent](https://www.w3.org/TR/trace-context/#traceparent-header) is the name of the W3C standard

crates/json-rpc/src/packet.rs
```diff
@@ -133,19 +133,15 @@ impl RequestPacket {
         self.requests().iter().map(|req| req.method())
     }
 
-    /// Retrieves the [`HeaderMap`] from the request metadata if available;
-    /// otherwise, returns an empty map. This functionality is only supported for single requests.
+    /// Retrieves the combined headers from all requests in the packet. If
+    /// multiple requests contain the same header, the last one wins.
     pub fn headers(&self) -> HeaderMap {
```
prestwich, line 138: This seems like a weird oversight in the first implementation?

crates/provider/src/ext/mev/with_auth.rs
```diff
@@ -120,8 +120,7 @@ where
                 )]);
 
                 // Patch the existing RPC call with the new headers
-                let rpc_call = self.inner.map_meta(|meta| {
-                    let mut meta = meta;
+                let rpc_call = self.inner.map_meta(|mut meta| {
```
prestwich, line 123: drive-by nit

crates/transport-http/src/layers/mod.rs
```diff
@@ -1,7 +1,13 @@
-//! tower http like layer implementations that work over the http::Request type.
-#![cfg(all(not(target_family = "wasm"), feature = "hyper"))]
+//! Additional layers that may be useful for HTTP transport users. Typically
+//! these inject headers into the HTTP requests.
+#![cfg(not(target_family = "wasm"))]
```
prestwich, line 3: cleaned up this whole spec, as `hyper` is specified as a requirement of `jwt-auth` so we do not need it in any of the `cfg` here

crates/transport-http/src/hyper_transport.rs
```diff
@@ -195,6 +195,6 @@ where
     fn call(&mut self, req: RequestPacket) -> Self::Future {
         let this = self.clone();
         let span = debug_span!("HyperTransport", url = %this.url);
-        Box::pin(this.do_hyper(req).instrument(span))
+        Box::pin(this.do_hyper(req).instrument(span.or_current()))
```
prestwich, line 198: this is a drive-by fix. when instrumenting futures, if the span instrumented is not enabled, then the connection to the currently-active span will be broken. I.e. everything that happens in the future will be disassociated with the trace context that caused it to happen

I suspect this occurs many places in the codebase but haven't checked. grepping for `.instrument(` calls and then checking if the inner span is debug or lower should do it

crates/transport-http/src/reqwest_transport.rs
```diff
@@ -88,6 +88,6 @@ impl Service<RequestPacket> for Http<reqwest::Client> {
     fn call(&mut self, req: RequestPacket) -> Self::Future {
         let this = self.clone();
         let span = debug_span!("ReqwestTransport", url = %this.url);
-        Box::pin(this.do_reqwest(req).instrument(span))
+        Box::pin(this.do_reqwest(req).instrument(span.or_current()))
```
prestwich, line 91: see other comment

onbjerg (approved):
this makes sense to me

PR #3368, opened 2025-12-17
docs: fix finalized checkpoint event topic name

Fibonacci747:
Updated the FinalizedCheckpointEvent doc comment to reference the finalized_checkpoint beacon node eventstream topic. The struct already models the finalized checkpoint event, but the comment incorrectly mentioned the deposit topic.


DaniPopes:
wrong

PR #3369, opened 2025-12-17
Remove redundant Clone bound from Network trait

conomist:
drop the explicit Clone bound from Network since Copy already implies it.
simplify trait constraints without altering behavior; avoid redundant bounds.


DaniPopes:
no

PR #3381, opened 2025-12-18
refactor(rpc-types-mev): simplify conversions and remove redundant closures

soniseth0:
Refactors MEV types to use more idiomatic Rust patterns: replace redundant closure `|item| item.into_owned()` with direct `Cow::into_owned` method reference in bincode compat conversion, replace `as_ref().map(|b| *b)` with idiomatic `copied()` for `Option<u64>` in `Inclusion::max_block_number()`

### DIFF
--- a/crates/rpc-types-mev/src/mev_calls.rs
+++ b/crates/rpc-types-mev/src/mev_calls.rs
@@ -183,7 +183,7 @@ impl Inclusion {
 
     /// Returns the block number of the last block the bundle is valid for.
     #[inline]
-    pub fn max_block_number(&self) -> Option<u64> {
+    pub const fn max_block_number(&self) -> Option<u64> {
         self.max_block
     }
 }


crates/rpc-types-mev/src/mev_calls.rs
```diff
@@ -184,7 +184,7 @@ impl Inclusion {
     /// Returns the block number of the last block the bundle is valid for.
     #[inline]
     pub fn max_block_number(&self) -> Option<u64> {
-        self.max_block.as_ref().map(|b| *b)
+        self.max_block.copied()
```
onbjerg:
```suggestion
        self.max_block
```

crates/rpc-types-mev/src/mev_calls.rs
```diff
@@ -184,7 +184,7 @@ impl Inclusion {
     /// Returns the block number of the last block the bundle is valid for.
     #[inline]
     pub fn max_block_number(&self) -> Option<u64> {
```
onbjerg:
```suggestion
    pub const fn max_block_number(&self) -> Option<u64> {
```

PR #3383, opened 2025-12-19
Fix incorrect debug log timing in RPC call preparation

vtjl10:
The debug log message "sending request" was emitted before the request was actually serialized and sent. This could mislead developers during debugging, as the log would indicate a request was sent even when serialization failed locally, making it harder to distinguish between network issues and serialization errors.


DaniPopes:
doesn't matter

PR #3407, opened 2025-12-22
feat(alloy-eip4361): add alloy-eip4361 crate for EIP-4361 Sign-In with Ethereum

### LINKED ISSUE #3096: [Feature] verifySiweMessage
### Component

provider, pubsub

Implementation of [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361). I ideally would imagine a function imported from utils, or even calling from the provider can also be called. ```provider.verify_siwe_message(message,signature)```

### Additional context

Depending if this would be supported, i would love to work on generateSiweNonce and createSiweMessage.

mfw78:
Adds `alloy-eip4361` and `alloy-eip1271` crates for [EIP-4361](https://eips.ethereum.org/EIPS/eip-4361) Sign-In with Ethereum support.

## Features

- **alloy-eip4361**: Parse, construct, and verify SIWE messages
  - Winnow-based parser following EIP-4361 ABNF grammar
  - Typestate `MessageBuilder` for compile-time validated construction
  - EIP-191 signature verification for EOA wallets
  - Optional EIP-1271 verification for contract wallets (via provider)
  - `no_std` compatible with optional `std`, `rand`, `serde` features

- **alloy-eip1271**: EIP-1271 signature verification utilities
  - `Eip1271` extension trait on `B256` for contract signature verification

Closes #3096


crates/eip1271/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+                Ok(false)
+            }
+            Err(alloy_contract::Error::ZeroData(..)) => {
+                // Empty return data - address is not a contract or doesn't implement EIP-1271
+                Ok(false)
+            }
+            Err(e) => Err(e.into()),
+        }
+    }
+}
```
0xferrous: I think this would serve better as a `ProviderEip1271Ext` like `ProviderEnsExt` instead, so implementing it for generic Provider P instead of B256.

mfw78: Good shout - refactored to `ProviderEip1271Ext` following the `ProviderEnsExt` pattern. Also made the signature param accept `impl AsRef<[u8]>` for ergonomics.

crates/siwe/Cargo.toml
```diff
@@ -0,0 +1,66 @@
+[package]
+name = "alloy-siwe"
```
mattsse: idk if this commonly referred to as siwe, but id prefer if we could name this alloy-eip4361 instead to be more consistent

mattsse (changes requested):
I feel like support for the message format could be a standalone  thing that we just include rpc support for, but I dont think including this really hurts

so I only really have a pedantic naming nit

crates/siwe/Cargo.toml
mfw78: Done – renamed to `alloy-eip4361` for consistency.

PR #3413, opened 2025-12-24
fix(geth): respect custom genesis in clique mode

phrwlk:
## Motivation

In the if is_clique branch:
```rust
if let Some(genesis) = &mut self.genesis {
    let clique_config = CliqueConfig { period: Some(0), epoch: Some(8) };
    genesis.config.clique = Some(clique_config);
    // ...
    genesis.extra_data = extra_data_bytes.into();
}
// ...
self.genesis = Some(Genesis::clique_genesis(
    self.chain_id.ok_or(NodeError::ChainIdNotSet)?,
    clique_addr,
));
```

- First, the existing self.genesis is modified, and then it is definitely overwritten with the new Genesis::clique_genesis(...).
- As a result, any changes in if let Some(genesis) = &mut self.genesis are never used when initializing the node — this is essentially dead code and a potential logical inconsistency with API expectations (if the user wanted to set their own 
genesis).

## Solution

Made Clique initialization reuse an existing user-provided genesis by only injecting the Clique config and extra data in that case, and falls back to constructing a default Genesis::clique_genesis only when no genesis was set on the builder.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/geth.rs
+++ b/crates/node-bindings/src/nodes/geth.rs
@@ -470,35 +470,27 @@ impl Geth {
 
         // use geth init to initialize the datadir if the genesis exists
         if is_clique {
-            let clique_addr = self.clique_address();
-            if let Some(genesis) = &mut self.genesis {
-                // set up a clique config with an instant sealing period and short (8 block) epoch
-                let clique_config = CliqueConfig { period: Some(0), epoch: Some(8) };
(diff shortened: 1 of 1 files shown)


DaniPopes:
clique has been deprecated for years

PR #3421, opened 2025-12-30
feat(rpc-client): allow setting terminal error codes on pollers

siosw:
## Motivation
https://github.com/alloy-rs/alloy/issues/2826

## Solution
continuation of #3182
allows terminal error codes to be set on the `Poller` that will cause it to terminate.
if no codes are set we use the existing logic to use the error message as a fallback

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


siosw:
@mattsse just gently pinging here if this is smth you'd still be looking to get in

mattsse (approved):
doesnt hurt really

mattsse:
sorry this got lost over the new years -.-

PR #3428, opened 2025-12-31
fix: update `SidecarBuilder::build` to allow 7594

prestwich:
Changes behavior of `build()` to require explicit output typing 

This is breaking but _not_ semver breaking, as the interface is compatible and only downstream type annotations need to change

## Motivation

`build` will silently produce bad results post-fusaka. 

## Solution

Forces users to explicitly specify what sort of blob they want
Provide a framework for future blob types as non-breaking changes

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/eips/src/eip4844/builder.rs
+++ b/crates/eips/src/eip4844/builder.rs
@@ -2,14 +2,12 @@ use crate::eip4844::{
     utils::WholeFe, Blob, BYTES_PER_BLOB, FIELD_ELEMENTS_PER_BLOB, FIELD_ELEMENT_BYTES_USIZE,
 };
 use alloc::vec::Vec;
+use core::cmp;
 
 #[cfg(any(feature = "kzg", feature = "arbitrary"))]
 use crate::eip4844::BlobTransactionSidecar;
 #[cfg(feature = "kzg")]
-use crate::eip4844::Bytes48;
-#[cfg(feature = "kzg")]
 use crate::{eip4844::env_settings::EnvKzgSettings, eip7594::BlobTransactionSidecarEip7594};
-use core::cmp;
 
 /// Describes types that can be built from blobs and KZG settings.
 ///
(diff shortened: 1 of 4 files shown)


crates/eips/src/eip4844/builder.rs
```diff
@@ -369,36 +432,46 @@ impl<T: SidecarCoder> SidecarBuilder<T> {
...
+    ) -> Result<BlobTransactionSidecar, c_kzg::Error> {
+        BlobTransactionSidecar::build(self.inner.blobs, settings)
     }
 
     /// Build the sidecar from the data, with default (Ethereum Mainnet)
     /// settings.
     #[cfg(feature = "kzg")]
-    pub fn build(self) -> Result<BlobTransactionSidecar, c_kzg::Error> {
-        self.build_with_settings(EnvKzgSettings::Default.get())
+    pub fn build_4844(self) -> Result<BlobTransactionSidecar, c_kzg::Error> {
```
mattsse, line 453: I think we can get away with this because everyone should migrate to this anyway

crates/eips/src/eip4844/builder.rs
```diff
@@ -11,6 +11,69 @@ use crate::eip4844::Bytes48;
...
 
+/// Describes types that can be built from blobs and KZG settings.
+///
+/// This trait primarily exists to allow inference of the output type of
+/// [`SidecarBuilder::build`], and should generally not be implemented outside
+/// of this crate.
+#[cfg(feature = "kzg")]
+pub trait BuildableSidecar {
+    /// Build the sidecar from the blobs and KZG settings.
+    fn build(blobs: Vec<Blob>, settings: &c_kzg::KzgSettings) -> Result<Self, c_kzg::Error>
```
mattsse: should we name this build_with_settings and add a default `build` here?

prestwich: seems reasonable. I was also thinking about moving this logic onto the type itself, so that you can call `BlobTransactionSidecar::build` without this trait in scope

prestwich: (this trait then delegates to the method on the type)

prestwich: okay sooooo actually this partially exists, with code duplication. and that api takes `Vec<c_kzg::Blob>`

For consistency and correctness, it seems like we should

- change the API on the sidecar types to use `Vec<Blob>` 
- have the construction code live on the sidecar
- dedup the construction code in `builder.rs` by making it call type methods

crates/eips/src/eip4844/builder.rs
prestwich: Cargo team considers generalizing to be semver `minor`

https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible

crates/eips/src/eip4844/builder.rs
prestwich: okay, I made the changes above. This is breaking for anyone using the methods on the sidecar types themselves. Previous api on those methods appears to have been a mistake, as the code is duplicated from elsewhere, and uses only outside types in the public API

mattsse (approved):

PR #3429, opened 2026-01-02
perf(transport): reduce lock contention in fallback metrics

snazzysam933:
moved atomic counters outside the rwlock in transport metrics to avoid holding the write lock when updating simple stats. previously code grabbing the lock for the entire **track_success/track_failure** operation including counter increments and timestamp updates causing unnecessary contention when multiple transports get hit concurrently.


DaniPopes:
doesn't matter

PR #3441, opened 2026-01-05
chore: update trezor dep

lean-apple:
## Motivation

Ideally remove https://github.com/foundry-rs/foundry/blob/a38549ca4a0b1a934763a9c8560bb02b6f3b12bd/deny.toml#L11. if we can update protobuf dep

## Solution

update trezor-client

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
looks alright

https://github.com/trezor/trezor-firmware/commits/main/rust/trezor-client

PR #3447, opened 2026-01-06
feat:  added custom estimator to blobgasfilter

### LINKED ISSUE #3443: [Feature] Add custom estimator to blobgasfilter
### Component

rpc

currently we do:

https://github.com/alloy-rs/alloy/blob/main/crates/provider/src/fillers/gas.rs#L228-L234

in BlobGasfiller

we should do something like:

https://github.com/alloy-rs/alloy/blob/main/crates/provider/src/utils.rs#L29-L32

that is obj safe so we can also do customize this behaviour in the filler

perhaps @Rimeeeeee or @Soubhik-10 ?

### Additional context

_No response_

Rimeeeeee:
fixes #3443

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,7 +152,7 @@ either = { version = "1.15", default-features = false }
 http = "1.1.0"
 itertools = { version = ">=0.13, <=0.14", default-features = false }
 jsonwebtoken = "9.3.0"
-lru = "0.13"
+lru = "0.16"
 once_cell = { version = "1.21", default-features = false }
 parking_lot = "0.12.3"
 pin-project = "1.1"
--- a/crates/consensus/src/transaction/eip4844.rs
+++ b/crates/consensus/src/transaction/eip4844.rs
@@ -29,7 +29,7 @@ use alloy_eips::eip7594::{BlobTransactionSidecarEip7594, BlobTransactionSidecarV
 #[cfg_attr(feature = "serde", serde(untagged))]
 #[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
 #[doc(alias = "Eip4844TransactionVariant")]
-pub enum TxEip4844Variant<T = BlobTransactionSidecar> {
+pub enum TxEip4844Variant<T = BlobTransactionSidecarVariant> {
     /// A standalone transaction with blob hashes and max blob fee.
     TxEip4844(TxEip4844),
     /// A transaction with a sidecar, which contains the blob data, commitments, and proofs.
@@ -106,6 +106,16 @@ impl<T> From<TxEip4844> for TxEip4844Variant<T> {
 impl From<(TxEip4844, BlobTransactionSidecar)> for TxEip4844Variant<BlobTransactionSidecar> {
     fn from((tx, sidecar): (TxEip4844, BlobTransactionSidecar)) -> Self {
         TxEip4844WithSidecar::from_tx_and_sidecar(tx, sidecar).into()
+    }
+}
+
+impl From<TxEip4844WithSidecar<BlobTransactionSidecar>>
+    for TxEip4844Variant<BlobTransactionSidecarVariant>
+{
+    fn from(tx: TxEip4844WithSidecar<BlobTransactionSidecar>) -> Self {
+        let (tx, sidecar) = tx.into_parts();
+        let sidecar_variant = BlobTransactionSidecarVariant::Eip4844(sidecar);
+        TxEip4844WithSidecar::from_tx_and_sidecar(tx, sidecar_variant).into()
     }
 }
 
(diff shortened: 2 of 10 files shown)


crates/provider/src/fillers/gas.rs
```diff
@@ -187,9 +190,74 @@ impl<N: Network> TxFiller<N> for GasFiller {
...
+
+    /// Estimates the base fee per blob gas given the latest max_fee_per_blob_gas.
+    pub async fn estimate<P: Provider<N>, N: Network>(
+        self,
+        provider: &P,
+        tx: &<N as Network>::TransactionRequest,
+    ) -> TransportResult<u128>
+    where
+        N::TransactionRequest: TransactionBuilder4844,
+    {
```
mattsse: instead of replicating this, what we can do is introducing a trait with an estimate fn that takes in the latest base_fee_per_blob_gas and the `blob_gas_used_ratio` as arguments, then this can be even more customized

crates/provider/src/fillers/gas.rs
```diff
@@ -187,9 +190,74 @@ impl<N: Network> TxFiller<N> for GasFiller {
...
+            .finish()
+    }
+}
+
 /// Filler for the `max_fee_per_blob_gas` field in EIP-4844 transactions.
 #[derive(Clone, Copy, Debug, Default)]
-pub struct BlobGasFiller;
+pub struct BlobGasFiller {
+    /// The blob gas estimator to use.
+    pub estimator: BlobGasEstimator,
```
mattsse, line 278: then we can add `Arc<dyn BlobGasEstimator>` here

crates/provider/src/fillers/gas.rs
```diff
@@ -219,19 +287,7 @@ where
...
-            if max_fee_per_blob_gas >= BLOB_TX_MIN_BLOB_GASPRICE {
-                return Ok(max_fee_per_blob_gas);
-            }
-        }
-
-        provider
-            .get_fee_history(2, BlockNumberOrTag::Latest, &[])
-            .await?
-            .base_fee_per_blob_gas
-            .last()
```
mattsse, line 232: and plug it in here

crates/provider/src/fillers/gas.rs
```diff
@@ -187,9 +190,74 @@ impl<N: Network> TxFiller<N> for GasFiller {
...
+        };
+
+        match self {
+            Self::Default => Ok(base_fee),
+            Self::Scaled(scale) => Ok(base_fee.saturating_mul(scale)),
+        }
+    }
+}
+
+impl fmt::Debug for BlobGasEstimator {
```
mattsse, line 260: I think we can derive this

crates/provider/src/fillers/gas.rs
```diff
@@ -187,9 +191,79 @@ impl<N: Network> TxFiller<N> for GasFiller {
...
 }
 
+/// An estimator function for blob gas fees.
+pub type BlobGasEstimatorFunction = fn(u128, f64) -> u128;
+
+/// A trait responsible for estimating blob gas values
+pub trait BlobGasEstimatorFn: Send + Sync + Unpin {
+    /// Estimates the blob gas fee given the base fee per blob gas
+    /// and the blob gas usage ratio.
+    fn estimate(&self, base_fee_per_blob_gas: u128, blob_gas_used_ratio: f64) -> u128;
```
mattsse: for the ratio we can give &[f64] so that estimators have a historical view over the last blocks

crates/provider/src/fillers/gas.rs
```diff
@@ -225,13 +299,18 @@ where
...
+        let base_fee_per_blob_gas = provider
             .get_fee_history(2, BlockNumberOrTag::Latest, &[])
             .await?
             .base_fee_per_blob_gas
             .last()
             .ok_or(RpcError::NullResp)
-            .copied()
+            .copied()?;
+
+        let blob_gas_used_ratio = 1.0;
```
mattsse: this we need to get from the fee_hsitory response

crates/provider/src/fillers/gas.rs
```diff
@@ -187,9 +191,79 @@ impl<N: Network> TxFiller<N> for GasFiller {
...
+}
+
+/// Blob Gas estimator variants
+#[derive(Default, Clone)]
+pub enum BlobGasEstimator {
+    /// Uses the builtin estimator
+    #[default]
+    Default,
+    /// Uses a custom estimator
+    Scaled(Arc<dyn BlobGasEstimatorFn>),
```
mattsse: we can name this custom

but we can also add a helper fn to BlobGasEstimator::scaled that creates a new custom instace that multiplies the gas like `fn scaled(scale) -> Self::custom(|gas, _| gas * scale)`

mattsse (changes requested):

juchiast:
FYI this will break compilation for code that pass the unit struct as a value, such as:

```rust
let filler = JoinFill::new(
    GasFiller,
    JoinFill::new(
        BlobGasFiller,
        JoinFill::new(
            NonceFiller::new(nonce_manager.clone()),
            ChainIdFiller::default(),
        ),
    ),
);
```

PR #3450, opened 2026-01-06
fix: graceful shutdown for reth and geth instances

snazzysam933:
Nodes were being killed immediately on drop, which isn't great - can corrupt databases or leave things in a bad state. Now sending SIGTERM first and giving them 10s to clean up before force killing. Standard stuff but makes tests more reliable.

### DIFF
--- a/crates/node-bindings/src/utils.rs
+++ b/crates/node-bindings/src/utils.rs
@@ -37,7 +37,7 @@ impl GracefulShutdown {
             }
         }
 
-        child.kill().expect(&format!("could not kill {}", process_name));
+        child.kill().unwrap_or_else(|_| panic!("could not kill {}", process_name));
     }
 }
 


crates/node-bindings/src/nodes/reth.rs
```diff
@@ -128,8 +131,31 @@ impl RethInstance {
...
+        let start = Instant::now();
+
+        while start.elapsed() < timeout {
+            match child.try_wait() {
+                Ok(Some(_)) => return,
+                Ok(None) => std::thread::sleep(Duration::from_millis(100)),
+                Err(_) => break,
+            }
+        }
     }
```
mattsse: this is the same fn, right?

can we unify this with a helper type like

```rust
struct Shutdown {
  graceful_timeout, pid
}
```

so that we can reuse this for both reth and geth?

snazzysam933: Make sense!

snazzysam933:
The cargo deny failure. After asking CloudCode it tells` it's detecting that `bincode` dependency is unmaintained (RUSTSEC-2025-0141). This PR only modifies process shutdown logic and doesn't touch dependencies.`

mattsse (approved):

PR #3451, opened 2026-01-07
feat(provider): cache get_balance and get_account methods

TracyMock21:
Adds caching support for `get_balance` and `get_account` methods in `CacheProvider`, following the same pattern as existing cached methods like `get_transaction_count`.


mattsse:
Closing — this doesn't appear to be a well-motivated contribution.

PR #3454, opened 2026-01-07
chore: update alloy-trie

Soubhik-10:


mattsse (approved):

mattsse:
this is equivalent because the txs are already 2718 encoded

PR #3457, opened 2026-01-07
[rpc-client] Prevent zero poll interval from truncation

maximevtush:
Fixes a bug where `set_poll_interval` could store `0ms` when sub-millisecond `Duration` values were provided, since `Duration::as_millis()` truncates to zero.


mattsse (commented):
imo is fine to not check for 0 here

PR #3466, opened 2026-01-07
feat(rpc-types-eth): add Params::from_json_value

mattsse:
Extracts the deserialization logic from the `Deserialize` impl into a reusable `from_json_value` method.


mattsse:
likely useful for https://github.com/paradigmxyz/reth/pull/20777

onbjerg:
```
let p: Params = serde_json::from_value(val)?;
```

should give you the same result and uses the de/ser impl already present, I don't think adding this fn is very useful

mattsse:
gives the same result but has another value deserde roundtrip

PR #3470, opened 2026-01-08
fix: Fix fake_exponential for use with large inputs

piersy:
## Motivation

TL;DR The pre-existing implementation differs from geth and broke Celo's zk proving, by returning invalid values (due to internal overflow) whe large numerators are used.

The Celo blockchain faced an execution mismatch between the go op-stack and the rust succinct-stack that prevented us from being able to correctly prove execution. This was caused by the alloy fake_exponential implementation returning an invalid value when a large numerator was provided

The inputs in question were:
* factor: 1
* numerator: 299453931
* denominator: 5007716

This happened to Celo and not other chains because we have not yet updated to correctly process the Fusaka bpo hardforks which results in higher (and incorrect) blob base fee values being calculated (since the update fraction is smaller). This is fine for us however, because unlike optimism we don't charge additional l2 gas based on the l1 gas fees and so we don't make use of the blob base fee value, however it is still included in the state by the L1 info transaction, and it is calculated using `fake_exponential` and differs from the result given by geth. This PR seeks to make the fake_exponential implementation match that of geth where possible without changing the return type of u128.

## Solution

Now arbitrarily sized values are used for the calculation, however if
the result does not fit in a u128 then the function will panic instead
of returning an invalid value.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [?] Breaking changes


mattsse (commented):
still need a bit until I fully understand this

but wouldn't we get the same if we convert the values to u256 internally and then do a `to::<u128>()` call on the final value?

piersy:
> but wouldn't we get the same if we convert the values to u256 internally and then do a to::<u128>() call on the final value?

I'll check, that does seem to be a neater approach, originally I used the big_int before I realised how slow large inputs would make it, but given that we are restricted to a numerator/denominator ratio of 88 probably u256 would be fine.

piersy:
Hi @mattsse.

You could still face issues even if u256 were used internally in the following case

* factor\*numerator\*denominator is bigger than u256, in some cases the _final result_ would also exceed u128 but in many, as long as the denominator is big enough the end result could fit in u128 even though the internal calculation would have overflowed u256.
 
Those are pretty big numbers so you could go for using U256 internally and panicking if an overflow is encountered, but it wouldn't behave the same as using `num_bigint` internally for all inputs.

piersy:
Hi @mattsse, in case you're interested I made some benchmarks, it seems that for realistic values _(bpo2 update fraction and 193634108 excess blob gas a value taken from eth mainnet)_ that the u256 version is about 6 times faster. But the speedup is relatively small ~6µs per call.

```
fake_exponential/bigint_realistic
                        time:   [6.9590 µs 7.8870 µs 9.1500 µs]
Found 18 outliers among 100 measurements (18.00%)
  2 (2.00%) high mild
  16 (16.00%) high severe

fake_exponential/u256_realistic
                        time:   [1.2633 µs 1.3071 µs 1.3842 µs]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
```

See [here](https://github.com/piersy/alloy/blob/482c33b7ff32f41e0c9c1edbfb4268f17bcd53f2/crates/eips/benches/fake_exponential.rs#L71-L90)

piersy:
I also had a look at changing the return type to U256, and the resulting diff was smaller than I expected, but I guess those changes would likely break a ton of dependent packages.

See the change [here](https://github.com/piersy/alloy/commit/1bea4b5e4802c9dac85b9527bfeba7088e9f84cb)

palango:
@mattsse Could you please have another look?

piersy:
Hi @mattsse, @DaniPopes, @klkvr, @onbjerg, @zero & @grandizzy is there any chance someone could take a look at this? Thanks!

palango:
@mattsse could you please have a look?

mattsse:
sorry this took forever ....

completed with #3961

PR #3481, opened 2026-01-09
chore(eips): remove unused Blob import when kzg feature is disabled

marukai67:
Removes dead code warning by conditionally importing Blob only when the kzg feature is enabled.


onbjerg (changes requested):
this pr just moves an import around, it does not add a feature gate or anything

PR #3502, opened 2026-01-12
perf(transport): avoid cloning full transport vector per request

Bashmunta:
FallbackService was cloning the entire Vec<ScoredTransport<S>> on every request just to sort it and then truncate to the active transport count. This meant we were cloning all transports, even those that were never used for a given call, which is wasteful when the pool is large. The new implementation sorts references to the existing transports and only clones the top N entries that are actually used, preserving behavior while reducing per-request allocations and clone overhead.


DaniPopes:
Doesn't change

PR #3503, opened 2026-01-12
Extract common trim logic for trezor encoding.

M0ndarynka:
Extract duplicated leading zero byte trimming logic into a shared helper function.


mattsse (commented):
not again

PR #3504, opened 2026-01-12
Clean up Trezor signer: drop redundant clone and reuse trim helper

0xcharry:
Remove an unnecessary DerivationType clone in TrezorSigner::new and reuse the stored value. Factor leading-zero byte trimming into a helper and call it from the trezor conversion helpers to avoid duplicated slicing code.


mattsse (commented):
we can keep it as is

PR #3506, opened 2026-01-12
feat: add getPayloadHacked endpoints

Rjected:
## Motivation

This is an experimental set of engine methods that accept just a list of txs to use to build a block

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/CHANGELOG.md
+++ b/crates/provider/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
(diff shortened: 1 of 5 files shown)


crates/rpc-types-engine/src/lib.rs
```diff
@@ -56,6 +56,10 @@ pub const CAPABILITIES: &[&str] = &[
     "engine_getPayloadV2",
     "engine_getPayloadV3",
     "engine_getPayloadV4",
+    "engine_getPayloadV1Hacked",
+    "engine_getPayloadV2Hacked",
+    "engine_getPayloadV3Hacked",
+    "engine_getPayloadV4Hacked",
```
mattsse: I think these should be excluded here because these arent part of the CL capabilities handshake?

Rjected: oops yep

PR #3509, opened 2026-01-12
[transport] Remove redundant Send + Sync bounds from CloneTransport

marukai67:
Removes redundant Send + Sync bounds from the return type of CloneTransport::clone_box().


mattsse (commented):
we can keep these

PR #3511, opened 2026-01-13
feat(engine): add TestingApi with testing_buildBlockV1 support

mattsse:
## Summary

Adds support for the `testing_` RPC namespace from reth, specifically `testing_buildBlockV1`.

## Changes

- Add `TestingBuildBlockRequestV1` type to `alloy-rpc-types-engine` in new `testing.rs` module
- Add `TestingApi` trait to `alloy-provider` with:
  - `build_block<R>` - generic method accepting custom method name and response type
  - `build_block_v1` - convenience method for `testing_buildBlockV1` returning `ExecutionPayloadEnvelopeV5`
- Add `TESTING_BUILD_BLOCK_V1` capability constant

## References

- reth implementation: https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-api/src/testing.rs
- Spec: https://github.com/marcindsobczak/execution-apis/blob/main/src/testing/testing_buildBlockV1.md

### DIFF
--- a/crates/provider/src/ext/testing.rs
+++ b/crates/provider/src/ext/testing.rs
@@ -9,9 +9,6 @@ use alloy_rpc_types_engine::{ExecutionPayloadEnvelopeV5, TestingBuildBlockReques
 use alloy_transport::TransportResult;
 
 /// Extension trait that gives access to Testing API RPC methods.
-///
-/// Note:
-/// > The provider should use a JWT authentication layer.
 #[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
 #[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
 pub trait TestingApi<N>: Send + Sync {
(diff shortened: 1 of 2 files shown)


crates/provider/src/ext/testing.rs
```diff
@@ -0,0 +1,53 @@
...
+use crate::Provider;
+use alloy_json_rpc::RpcRecv;
+use alloy_network::Network;
+use alloy_rpc_types_engine::{ExecutionPayloadEnvelopeV5, TestingBuildBlockRequestV1};
+use alloy_transport::TransportResult;
+
+/// Extension trait that gives access to Testing API RPC methods.
+///
+/// Note:
+/// > The provider should use a JWT authentication layer.
```
Rjected: is this true? thought this is just through the regular rpc api

Rjected (commented):
just one comment otherwise lgtm

crates/provider/src/ext/testing.rs
mattsse: it is not, updated

PR #3513, opened 2026-01-13
refactor(crypto): eliminate signature encoding duplication

sashaodessa:
The signature-to-bytes encoding logic (converting `Signature` to `[u8; 65]` for recovery backends) was duplicated in multiple places within `crypto.rs`. This encoding has a critical detail: the recovery ID must be `0/1` (not `27/28`), and having multiple copies increases the risk of future bugs if one copy is modified incorrectly.

Extracted the encoding logic into a single helper function `signature_to_recoverable_bytes()` and replaced all duplicated code blocks with calls to this function. This establishes a single source of truth for the encoding format and documents the recovery ID requirement in one place.

### DIFF
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -199,6 +199,17 @@ pub mod secp256k1 {
     #[cfg(feature = "secp256k1")]
     use super::impl_secp256k1 as imp;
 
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we can keep this as is

PR #3514, opened 2026-01-13
refactor(rpc-types-eth): remove redundant clones in block tests

0xlupin:
Eliminates unnecessary memory allocations in test code by optimizing clone operations.


mattsse (approved):
sure

PR #3539, opened 2026-01-17
perf(signer-ledger): eliminate allocations in derivation path parsing

0xxFloki:
Optimizes `LedgerSigner::path_to_bytes` to eliminate unnecessary string allocations in the hot path when parsing BIP-32 derivation paths.


mattsse (commented):
we can keep as is

PR #3546, opened 2026-01-19
refactor: deduplicate signature hash calculation in `Signed`

majtte:
Extracted shared `signature_hash` computation from `recover_with_buf` and `recover_unchecked_with_buf` into a single helper to avoid duplication.


mattsse (commented):
no

PR #3554, opened 2026-01-20
Remove unused imports in consensus envelope tests

conomist:
Drop the unused-import allowance in the consensus envelope tests.
Remove unused SignerRecoverable import and gate TxKind behind k256 so the module compiles lint-clean without suppressions.


mattsse (commented):
you see that this doesnt compile right

PR #3561, opened 2026-01-21
fix(docs): exclude signer-trezor from docs.rs builds

### LINKED ISSUE #3560: docs.rs builds fail since v1.4.1
It seems docs.rs builds fail since v1.4.1, so the latest available docs version for alloy metacrate is v1.4.0:
https://docs.rs/crate/alloy/latest/builds/2823765

The individual crates are OK.

gakonst:
## Summary

Fixes #3560 - docs.rs builds fail since v1.4.1

## Problem

The `trezor-client` crate causes a rustc ICE (Internal Compiler Error) on nightly during macro expansion:

```
assertion failed: !deep_decl.is_glob_import()
thread 'rustc' panicked at compiler/rustc_resolve/src/imports.rs:378:13
```

This prevents docs.rs from building documentation for the alloy metacrate since docs.rs uses `--all-features`.

## Solution

Replace `all-features = true` in the docs.rs metadata with an explicit feature list that includes all features **except** `signer-trezor`.

Once the upstream trezor-client issue is fixed or a rustc fix lands, we can revert to `all-features = true`.


gakonst:
Closing - the rustc ICE has been fixed in latest nightly. Will investigate the proper solution.

PR #3568, opened 2026-01-21
chore: "reapply "chore: added slot num in payload attributes

mattsse:
Reverts alloy-rs/alloy#3567

blocked until we have a new header field


mattsse:
Closing as superseded by main. The slot_number field and the missing-slot_number serde coverage are already present on main, so there is no remaining delta to update here.

PR #3570, opened 2026-01-21
feat: parse data field on error for eth_simulateV1 if present in response

antonydenyer:
## Motivation
Many clients return the revert reasons in the data parameter when a simulation reverts, making the data returned available. 

## Solution

Add param to wire up serialisation with appropriate defaults

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


crates/rpc-types-eth/src/simulate.rs
```diff
@@ -190,6 +190,9 @@ pub struct SimulateError {
     pub code: i32,
     /// Message error
     pub message: String,
+    /// Data for the error e.g revert reason
+    #[cfg_attr(feature = "serde", serde(default))]
+    pub data: String,
```
mattsse: can we make this optionally instead

antonydenyer: https://github.com/alloy-rs/alloy/pull/3570/changes/58e115f7a1af71a3ecc860c54846fb7e6fb46541 

Changed type to `Option<Bytes>` to be inline with `eth_call` and revert reason

gakonst:
To make this non-breaking for downstream consumers like reth, could you also add `#[derive(Default)]` to `SimulateError`? That would allow us to use `SimulateError { code, message, ..Default::default() }` which won't break when new fields are added.

crates/rpc-types-eth/src/simulate.rs
```diff
@@ -180,7 +180,7 @@ impl<TxReq> SimulatePayload<TxReq> {
 }
 
 /// The error response returned by the `eth_simulateV1` method.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
onbjerg: please remove the `Default` derive. you can keep `#[serde(default)]` on the `data` field

antonydenyer: https://github.com/alloy-rs/alloy/pull/3570/changes/27e783ef8164d3f8513c9d11c927dfbad32fb0ca

good spot, left over from original implementation

mattsse:
thanks @antonydenyer something went rouge here hehe

antonydenyer:
Any news on this, folks?

PR #3583, opened 2026-01-23
feat: add `gas_spent` field to receipts

jenpaff:
EIP-7778 introduces a new gas_spent field in receipts to track per-transaction gas after refunds (what users actually pay). This separates it from cumulative_gas_used, which now represents gas before refunds for block limit enforcement. This PR adds the gas_spent field to TransactionReceipt and the ReceiptResponse trait, following the same pattern as blob_gas_used with optional serde handling for backward compatibility.


onbjerg (approved):

mattsse (commented):
this has been excluded from the eip so we can close

PR #3588, opened 2026-01-24
perf(signer-ledger): remove redundant allocations and avoid division by zero

0xxFloki:
Removes unnecessary allocations and fixes potential edge case in `alloy-signer-ledger`.


onbjerg:
unnecessary. the pr mentions an edge case but does not go into detail what it is, if you found an edge case, please open up an issue

PR #3594, opened 2026-01-26
DO NOT MERGE Bal devnet2

mattsse:


crates/consensus/src/receipt/receipts.rs
```diff
@@ -151,8 +168,12 @@ impl<T: Decodable> Receipt<T> {
         let cumulative_gas_used = Decodable::decode(buf)?;
         let logs_bloom = Decodable::decode(buf)?;
         let logs = Decodable::decode(buf)?;
+        let gas_spent = if buf.is_empty() { None } else { Some(Decodable::decode(buf)?) };
```
klkvr: this is likely wrong because `buf` might contain the next receipt's body. we need to take a slice of the buf here https://github.com/alloy-rs/alloy/blob/e51986ff9012b348c1fea56ae5c8b7dad57d16fe/crates/consensus/src/receipt/receipts.rs#L166-L172

PR #3595, opened 2026-01-26
feat: Adding an ECDSA verify

hecmas:
## Motivation

Following [#2634](https://github.com/alloy-rs/alloy/pull/2634), I've added a method for being able to inject an ECDSA verification.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/crypto.rs
+++ b/crates/consensus/src/crypto.rs
@@ -53,8 +53,7 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
 ]);
 
 /// Serialized uncompressed public key
-#[derive(Debug, Clone)]
-pub struct UncompressedPublicKey(pub [u8; 65]);
+pub type UncompressedPublicKey = [u8; 65];
 
 /// Crypto backend module for pluggable cryptographic implementations.
 #[cfg(feature = "crypto-backend")]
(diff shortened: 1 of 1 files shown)


kevaundray:
This makes sense to me! Would be useful to clean up https://github.com/paradigmxyz/reth/blob/0b5f79e8c9226ee66c96eb35d2a5f4408e8c48ac/crates/stateless/src/recover_block.rs#L79

crates/consensus/src/crypto.rs
```diff
@@ -52,6 +52,10 @@ pub const SECP256K1N_HALF: U256 = U256::from_be_bytes([
     0x5D, 0x57, 0x6E, 0x73, 0x57, 0xA4, 0x50, 0x1D, 0xDF, 0xE9, 0x2F, 0x46, 0x68, 0x1B, 0x20, 0xA0,
 ]);
 
+/// Serialized uncompressed public key
+#[derive(Debug, Clone)]
+pub struct UncompressedPublicKey(pub [u8; 65]);
```
mattsse: do we even need this type?

rn the called would need to copy the 64bytes into this for

pubkey: &UncompressedPublicKey,

kevaundray: I think we can make it a type alias or use [u8;65] -- we needed it in reth-stateless for serialization reasons though it should be easy to convert in that crate

hecmas: Agree with @kevaundray, I think a type alias would be the cleanest solution as for now.

PR #3604, opened 2026-01-26
refactor(provider): remove redundant clone in wallet filler test

Aleksandr1732:
Eliminates unnecessary `clone()` call in the `ingest_pk_signer` test by computing the expected address before moving the private key.


onbjerg:
this is just a test so doesn't matter much

PR #3609, opened 2026-01-28
fix(provider): remove inverted condition in convert_to_hashes

aso20455:
The condition `if block.transactions().is_empty()` was inverted - it only converted to hashes when there were NO transactions, which is pointless. When the block actually had transactions, they were never converted.

Removed the unnecessary check since `BlockTransactions::convert_to_hashes()` already handles this internally by checking `if !self.is_hashes()`.


mattsse (commented):
this isnt equivalent

PR #3610, opened 2026-01-28
chore(provider): remove unused TransactionRequest import from batch tests

conomist:
Remove an unused TransactionRequest import from the batch layer tests in alloy-provider


onbjerg:
this is not unused.

PR #3618, opened 2026-01-29
fix: remove duplicate log in subscription deserialization error

snazzysam933:
Was logging both debug and error for the same failure - merged into single error log with the value context.


onbjerg:
this is ok, otherwise the default log is too noisy

PR #3620, opened 2026-01-30
chore: added empty bal check and slot num fn to payload

Rimeeeeee:


crates/consensus/src/block/header.rs
```diff
@@ -785,12 +786,17 @@ pub trait BlockHeader {
...
-    /// Checks if the header is considered empty - has no transactions, no ommers or withdrawals
+    /// Checks if the header is considered empty - has no transactions, no ommers or withdrawals or
+    /// bal
     fn is_empty(&self) -> bool {
         let txs_and_ommers_empty = self.transactions_root() == EMPTY_ROOT_HASH
             && self.ommers_hash() == EMPTY_OMMER_ROOT_HASH;
-        self.withdrawals_root().map_or(txs_and_ommers_empty, |withdrawals_root| {
-            txs_and_ommers_empty && withdrawals_root == EMPTY_ROOT_HASH
+
+        let bal_empty =
```
Soubhik-10, line 795: Ah..this caused a lot of empty blocks where there weren't any txs

Soubhik-10 (approved):

mattsse (approved):

PR #3621, opened 2026-01-30
Optimize key parsing to avoid unnecessary allocations

conomist:
summary: Create key variants lazily in extract_value to avoid unused Cow allocations while preserving parsing behavior.


DaniPopes:
its the same

PR #3625, opened 2026-01-31
perf(transport-ws): remove unnecessary clone in WsConnect::connect

0xcharry:
Remove unnecessary heap allocation in WebSocket connection establishment


mattsse (commented):
nope

PR #3626, opened 2026-01-31
perf: optimize hex serialization for Index and BlockNumberOrTag

splinter012:
Replace manual `format!` string allocations with `alloy_serde::quantity::serialize` for hex number serialization in `Index` and `BlockNumberOrTag` types.


mattsse (changes requested):
can you add some additional tests

PR #3627, opened 2026-01-31
avoid closure for default retry backoff

Perekup1:
Context: Retry layer uses a constant default backoff when no hint is provided.
Change: Replaced unwrap_or_else with unwrap_or since the fallback is a cheap const fn.
Why: Removes an unnecessary closure and keeps the path leaner without behavior change.


mattsse (commented):
no

PR #3628, opened 2026-02-01
refactor(pubsub): remove redundant clone in reconnect subscription loop

Aleksandr1732:
Eliminates an unnecessary `SerializedRequest` clone during subscription reconnection in `PubSubService::reconnect()`.


mattsse:
Closing - marginal value change.

PR #3631, opened 2026-02-01
fix(rpc-types-eth): use hex serialization for SyncStatusMetadata block fields

snazzysam933:
Use `alloy_serde::quantity` for `starting_block`, `current_block`, and `highest_block` fields in `SyncStatusMetadata` to serialize as hex strings per EIP-1474.


snazzysam933:
ci flaky?

PR #3632, opened 2026-02-01
fix: reduce test flakiness for node-bindings and network tests

mattsse:
## Summary

Reduces CI flakiness by:

1. **Handling transport errors in `test_pending_block_deser`** - The test now skips on any transport error (502, timeouts, connection failures) rather than just "no response" errors.

2. **Running `alloy-node-bindings` tests serially** - Uses nextest's `threads-required = "num-cpus"` to prevent resource contention between heavy e2e tests.

3. **Adding retries for node-bindings tests** - Exponential backoff with 3 retries for transient failures.

Fixes flaky CI failures like: https://github.com/alloy-rs/alloy/actions/runs/21563960416/job/62143557052

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -5,3 +5,11 @@ slow-timeout = { period = "30s", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = "test(can_launch_reth_custom_ports)"
 retries = { backoff = "exponential", count = 5, delay = "3s", jitter = true }
+
+[test-groups.node-bindings]
+max-threads = 1
+
+[[profile.default.overrides]]
+filter = "package(alloy-node-bindings)"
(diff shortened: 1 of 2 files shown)


.config/nextest.toml
```diff
@@ -5,3 +5,8 @@ slow-timeout = { period = "30s", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = "test(can_launch_reth_custom_ports)"
 retries = { backoff = "exponential", count = 5, delay = "3s", jitter = true }
+
+[[profile.default.overrides]]
+filter = "package(alloy-node-bindings)"
+threads-required = "num-cpus"
```
DaniPopes: This is wrong, i think it's max-threads = 1 for sequential tests

PR #3633, opened 2026-02-02
fix(ci, signer-aws): increase recursion limit to fix `error: queries overflow the depth limit!`

zerosnacks:
## Motivation

Fixes CI issue:

```
error: queries overflow the depth limit!
  |
  = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`alloy_signer_aws`)
  = note: query depth increased by 130 when computing layout of `{async block@crates/signer-aws/src/signer.rs:120:5: 120:23}`
```

This is breaking downstream CI's as `aws-*` is pinned to major. I have not fully looked into why this is happening, I can't find reports on the respective repo.

To repro: 

```
cargo update
cargo +nightly clippy --all-features
```

## Solution

Bumps to `#![recursion_limit = "256"]`

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -5,3 +5,11 @@ slow-timeout = { period = "30s", terminate-after = 4 }
 [[profile.default.overrides]]
 filter = "test(can_launch_reth_custom_ports)"
 retries = { backoff = "exponential", count = 5, delay = "3s", jitter = true }
+
+[test-groups.node-bindings]
+max-threads = 1
+
+[[profile.default.overrides]]
+filter = "package(alloy-node-bindings)"
+test-group = "node-bindings"
+retries = { backoff = "exponential", count = 3, delay = "3s", jitter = true }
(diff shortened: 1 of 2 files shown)


onbjerg (approved):

crates/signer-aws/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 )]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![recursion_limit = "256"]
```
onbjerg, line 10: let's add a note on why this is here?

zerosnacks: Good call, added https://github.com/alloy-rs/alloy/pull/3633/commits/4d789b5807a6bae63d3f476b5e3f55b9509a5ee7

PR #3640, opened 2026-02-03
perf(provider): use get_header instead of get_block in get_block_number_by_id

snazzysam933:
We only need the block number here, no reason to fetch the entire block with all its transactions. Now that get_header exists, let's use it.


snazzysam933:
Updated to use fallback: try get_header first, fall back to get_block if not supported. Turns out `eth_getHeaderByHash` isn't part of the standard JSON-RPC spec, so Anvil/Hardhat don't support it. This way we get the perf win on nodes that support it without breaking compatibility.

PR #3600, opened 2026-01-26
refactor(network): unify EIP-4844 tx builder trait

mablr:
## Motivation

Close #3592, consequently resolve #3591

## Solution

Finally, contrary to my initial design proposal, I believe it’s more straightforward to implement this directly as a breaking change, considering Alloy v2.0 is coming soon.

- Renamed methods for setting and getting blob sidecars to specify EIP-4844 and EIP-7594 format variants.
- Merged the TransactionBuilder7594 trait into TransactionBuilder4844.
- Updated tests and transaction request implementations to reflect the new method names and structure.

## PR Checklist

- [ ] Added Tests (updated existing ones)
- [x] Added Documentation
- [x] Breaking changes

Issue #3592, opened 2026-01-26
refactor(provider): Unify `TransactionBuilder4844` and `TransactionBuilder7594` Traits


mablr:
## Problem

`TransactionBuilder4844` and `TransactionBuilder7594` create conflicts when used in the same code due to duplicate method definitions (`max_fee_per_blob_gas()`/`set_max_fee_per_blob_gas()`)

## Proposal

Create unified `TransactionBuilderSidecar` trait:

```rust
pub trait TransactionBuilderSidecar: Default + Sized + Send + Sync + 'static {
    fn max_fee_per_blob_gas(&self) -> Option<u128>;
    fn set_max_fee_per_blob_gas(&mut self, max_fee_per_blob_gas: u128);
    fn with_max_fee_per_blob_gas(mut self, val: u128) -> Self { ... }

    fn blob_sidecar(&self) -> Option<&BlobTransactionSidecarVariant>;
    fn set_blob_sidecar(&mut self, sidecar: BlobTransactionSidecarVariant);
    fn with_blob_sidecar(mut self, sidecar: BlobTransactionSidecarVariant) -> Self { ... }

    // Convenience methods
    fn blob_sidecar_4844(&self) -> Option<&BlobTransactionSidecar>;
    fn set_blob_sidecar_4844(&mut self, sidecar: BlobTransactionSidecar);
    fn with_blob_sidecar_4844(mut self, sidecar: BlobTransactionSidecar) -> Self { ... }

    fn blob_sidecar_7594(&self) -> Option<&BlobTransactionSidecarEip7594>;
    fn set_blob_sidecar_7594(&mut self, sidecar: BlobTransactionSidecarEip7594);
    fn with_blob_sidecar_7594(mut self, sidecar: BlobTransactionSidecarEip7594) -> Self { ... }
}
```

Keep old traits as stubs delegating to new trait (blanket impl):

```rust
#[deprecated]
impl<T: TransactionBuilderSidecar> TransactionBuilder4844 for T {
    fn max_fee_per_blob_gas(&self) -> Option<u128> {
        TransactionBuilderSidecar::max_fee_per_blob_gas(self)
    }
    // ... delegates all methods
}
```

Same for `TransactionBuilder7594`.

## Benefits

- Unified trait w/ explicit methods eliminate name collisions
- Non-breaking change for the moment (Fully backward compatible since keep old traits)
- Elegant way to fix #3591

dipanshuhappy:
Happy to work on this

[...]

Issue #3591, opened 2026-01-26
[Bug] `BlobGasFiller` doesn't support eip7594 sidecar format


mablr:
### Component

network, json-rpc, provider, pubsub

### What version of Alloy are you on?

1.5.2

### Operating System

None

## Description                                                                                                                                                          

`BlobGasFiller` only fills `max_fee_per_blob_gas` for eip4844 transactions with eip4844 sidecar format, eip7594 sidecar is kipped.

## Root Cause
There is only this trait bound:
https://github.com/alloy-rs/alloy/blob/615f4c6d9ef06b4e908b789313d5d22500089906/crates/provider/src/fillers/gas.rs#L283

The presence of eip7594 sidecar is not checked
https://github.com/alloy-rs/alloy/blob/615f4c6d9ef06b4e908b789313d5d22500089906/crates/provider/src/fillers/gas.rs#L288-L294

This problem is similar to the one fixed here https://github.com/foundry-rs/foundry/pull/13218. I'll propose shortly a more elegant fix here.

PR #3641, opened 2026-02-04
fix(provider): BlobGasFiller now handles EIP-7594 sidecars

neowangreal:
BlobGasFiller only checked `blob_sidecar()` (EIP-4844) and completely ignored `blob_sidecar_7594()`, so blob gas estimation was skipped for EIP-7594 txs. Added the check and trait bound.

### DIFF
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -12,7 +12,7 @@ use crate::{
 };
 use alloy_eips::eip4844::BLOB_TX_MIN_BLOB_GASPRICE;
 use alloy_json_rpc::RpcError;
-use alloy_network::{Network, TransactionBuilder, TransactionBuilder4844};
+use alloy_network::{Network, TransactionBuilder, TransactionBuilder4844, TransactionBuilder7594};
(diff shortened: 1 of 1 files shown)


mattsse (commented):
wdyt @mablr

mablr (commented):
@neowangreal, you’re completely right about this issue. I encountered the same problem a few days ago while working on Foundry refactoring task, see #3591.

In parallel, I proposed the unification of the 4844/7594 sidecar tx builder traits in issue #3592, and the corresponding PR #3600. Finally, I realized that by naming the new unified trait `TransactionBuilder4844` (as the old one), we elegantly solve the missing `TransactionBuilder7594` trait bound everywhere, alloy+foundry (for sure😎), and maybe reth (haven't checked so far). This is because (new `TransactionBuilder4844`) <=> (old `TransactionBuilder4844` + `TransactionBuilder7594`).

So, this quick fix works right now, but it will become pointless once #3600 is shipped in alloy 2.0.

mablr (commented):
As #3600 was merged, the problem is now fixed.

Thank you anyway @neowangreal 🙌

PR #3644, opened 2026-02-04
fix(rpc-types-eth): fix tautological test assertions in TransactionRequest tests

Aleksandr1732:
Fixes three test assertions in TransactionRequest::build_consensus_tx_works that were accidentally comparing shadowed variables to themselves, making them tautological (always true). Also refactors TransactionInputKind::from_str for better readability.


mattsse:
Closing this — the tautological assertion fixes are valid observations, but the impact is low (test-only, and the assertions still pass structurally even if the guards are no-ops). The `FromStr` refactor is a style-only change that reduces readability of the idiomatic `match` version.

Thanks for the contribution though!

PR #3649, opened 2026-02-05
feat: add reth traits to alloy for op-alloy dependency removal

theochap:
## Summary

- Add 10 trait definitions from reth core crates into alloy so that op-alloy can implement them natively without depending on reth
- `alloy-consensus`: `InMemorySize`, `SerdeBincodeCompat`, `SignedTransaction` traits + impls for Ethereum types
- `alloy-rpc-types-engine`: `PayloadAttributes`, `ExecutionPayload` traits + impls for alloy's own structs
- `alloy-network`: `FromConsensusTx`, `TryFromTransactionResponse`, `TryFromReceiptResponse`, `SignableTxRequest`, `TryIntoSimTx`, `RpcTypes` traits + Ethereum impls

## Motivation

This is part of a multi-repo effort to remove `op-alloy-*` dependencies from core (non-optimism) reth crates. By moving trait definitions into alloy, op-alloy can implement them natively for OP types without needing to depend on reth, and reth core crates can drop their op-alloy dependencies entirely.

## Test plan

- [ ] `cargo check --workspace --all-features`
- [ ] `cargo nextest run --workspace`
- [ ] Verify traits are correctly re-exported from each crate's `lib.rs`

### DIFF
--- a/crates/consensus/src/serde_bincode_compat_traits.rs
+++ b/crates/consensus/src/serde_bincode_compat_traits.rs
@@ -1,0 +1,276 @@
+//! Bincode compatibility support traits.
+//!
+//! This module provides traits and implementations to work around bincode's limitations
+//! with optional serde fields. The bincode crate requires all fields to be present during
+//! serialization, which conflicts with types that have `#[serde(skip_serializing_if)]`
+//! attributes for RPC compatibility.
+//!
+//! # Overview
+//!
+//! The main trait is [`SerdeBincodeCompat`], which provides a conversion mechanism between
+//! types and their bincode-compatible representations. There are two main ways to implement
+//! this trait:
+//!
+//! 1. **Using RLP encoding** - Implement [`RlpBincode`] for types that already support RLP
+//! 2. **Custom implementation** - Define a custom representation type
+
+use alloc::vec::Vec;
+use alloy_primitives::Bytes;
+use core::fmt::Debug;
+use serde::{de::DeserializeOwned, Serialize};
+
+/// Trait for types that can be serialized and deserialized using bincode.
+///
+/// This trait provides a workaround for bincode's incompatibility with optional
+/// serde fields. It ensures all fields are serialized, making the type bincode-compatible.
+///
+/// # Implementation
+///
+/// The easiest way to implement this trait is using [`RlpBincode`] for RLP-encodable types:
+///
+/// ```rust,ignore
+/// impl RlpBincode for MyType {}
+/// // SerdeBincodeCompat is automatically implemented
+/// ```
+///
+/// The recommended way to add bincode compatible serialization is via the
+/// [`serde_with`] crate and the `serde_as` macro.
+pub trait SerdeBincodeCompat: Sized + 'static {
+    /// Serde representation of the type for bincode serialization.
+    ///
+    /// This type defines the bincode compatible serde format for the type.
+    type BincodeRepr<'a>: Debug + Serialize + DeserializeOwned;
+
+    /// Convert this type into its bincode representation
+    fn as_repr(&self) -> Self::BincodeRepr<'_>;
+
+    /// Convert from the bincode representation
+    fn from_repr(repr: Self::BincodeRepr<'_>) -> Self;
+}
+
+/// Type alias for the [`SerdeBincodeCompat::BincodeRepr`] associated type.
+pub type BincodeReprFor<'a, T> = <T as SerdeBincodeCompat>::BincodeRepr<'a>;
+
+/// A helper trait for using RLP-encoding for providing bincode-compatible serialization.
+///
+/// By implementing this trait, [`SerdeBincodeCompat`] will be automatically implemented for the
+/// type and RLP encoding will be used for serialization and deserialization for bincode
+/// compatibility.
+pub trait RlpBincode: alloy_rlp::Encodable + alloy_rlp::Decodable {}
+
+impl<T: RlpBincode + 'static> SerdeBincodeCompat for T {
+    type BincodeRepr<'a> = Bytes;
+
+    fn as_repr(&self) -> Self::BincodeRepr<'_> {
+        let mut buf = Vec::new();
+        self.encode(&mut buf);
+        buf.into()
+    }
+
+    fn from_repr(repr: Self::BincodeRepr<'_>) -> Self {
+        Self::decode(&mut repr.as_ref()).expect("Failed to decode bincode rlp representation")
+    }
+}
+
+// --- Implementations for alloy-consensus types ---
+
+impl SerdeBincodeCompat for crate::Header {
+    type BincodeRepr<'a> = crate::serde_bincode_compat::Header<'a>;
+
+    fn as_repr(&self) -> Self::BincodeRepr<'_> {
+        self.into()
+    }
+
+    fn from_repr(repr: Self::BincodeRepr<'_>) -> Self {
+        repr.into()
+    }
+}
+
+impl SerdeBincodeCompat for crate::EthereumTxEnvelope<crate::TxEip4844> {
+    type BincodeRepr<'a> =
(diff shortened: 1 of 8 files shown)


mattsse (commented):
wdyt @onbjerg ?

theochap:
Btw I implemented some new type wrappers for the DB traits in op-reth directly (Compact and such). Here are all the traits that could be cleanly pulled out of op-reth

crates/consensus/src/transaction/signed_tx.rs
```diff
@@ -0,0 +1,148 @@
...
+/// # Recovery Methods
+///
+/// This trait provides two types of recovery methods:
+/// - Standard methods (e.g., `try_recover`) - enforce EIP-2 low-s signature requirement
+/// - Unchecked methods (e.g., `try_recover_unchecked`) - skip EIP-2 validation for pre-EIP-2
+///   transactions
+///
+/// Use unchecked methods only when dealing with historical pre-EIP-2 transactions.
+#[auto_impl::auto_impl(&, Arc)]
+pub trait SignedTransaction:
```
onbjerg, line 30: we already have `Signed<T>`, I'm not sure why this is different in reth

crates/consensus/src/serde_bincode_compat_traits.rs
```diff
@@ -0,0 +1,276 @@
...
+/// The easiest way to implement this trait is using [`RlpBincode`] for RLP-encodable types:
+///
+/// ```rust,ignore
+/// impl RlpBincode for MyType {}
+/// // SerdeBincodeCompat is automatically implemented
+/// ```
+///
+/// The recommended way to add bincode compatible serialization is via the
+/// [`serde_with`] crate and the `serde_as` macro.
+pub trait SerdeBincodeCompat: Sized + 'static {
```
onbjerg, line 38: I don't think this belongs in alloy tbh

crates/network/src/convert.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Returns `Ok(Self)` on successful conversion, or `Err(Self::Error)` if the conversion fails.
+    fn from_receipt_response(receipt_response: N::ReceiptResponse) -> Result<Self, Self::Error>
+    where
+        Self: Sized;
+}
+
+/// Converts `self` into `T`.
+///
+/// Should create a fake transaction for simulation using [`TransactionRequest`].
+pub trait TryIntoSimTx<T>
```
onbjerg, line 127: these probably belong in alloy-evm?

crates/network/src/lib.rs
```diff
@@ -36,6 +42,51 @@ pub use alloy_network_primitives::{
     self as primitives, BlockResponse, ReceiptResponse, TransactionResponse,
 };
 
+/// RPC types used by the `eth_` RPC API.
+///
+/// This is a subset of [`Network`] trait with only RPC response types kept.
+pub trait RpcTypes: Send + Sync + Clone + Unpin + Debug + 'static {
```
onbjerg: these already exist in `Network`, no need to duplicate them.

crates/network/src/lib.rs
```diff
@@ -36,6 +42,51 @@ pub use alloy_network_primitives::{
...
+
+/// Adapter for network specific header response.
+pub type RpcHeader<T> = <T as RpcTypes>::Header;
+
+/// Adapter for network specific block type.
+pub type RpcBlock<T> =
+    alloy_rpc_types_eth::Block<RpcTransaction<T>, RpcHeader<T>>;
+
+/// Adapter for network specific transaction request.
+pub type RpcTxReq<T> = <T as RpcTypes>::TransactionRequest;
```
onbjerg: these should be over `Network`

crates/rpc-types-engine/src/lib.rs
```diff
@@ -34,6 +34,16 @@ pub use jwt::*;
...
 
+#[cfg(feature = "serde")]
+pub mod traits;
+#[cfg(feature = "serde")]
+pub use traits::{
+    ExecutionPayload as ExecutionPayloadTrait, PayloadAttributes as PayloadAttributesTrait,
+};
+
+// Re-export trait names without the `Trait` suffix inside the traits module for direct use.
+// Users can also import from `alloy_rpc_types_engine::traits::{PayloadAttributes, ExecutionPayload}`.
```
onbjerg: unsure what this comment is doing here?

onbjerg (changes requested):
generally supportive, but it seems like a lot of these are because reth is not fully using alloy and instead uses a bunch of wrapper traits/duplicated functionality

crates/consensus/src/serde_bincode_compat_traits.rs
theochap: Where do you think it would be better to integrate?

crates/consensus/src/transaction/signed_tx.rs
theochap: `Signed<T>` looks pretty different from `SignedTransaction`. One is a trait, the other is a struct. I think reth wants to make it possible to generically override transactions with custom types that don't come from `alloy`

crates/network/src/convert.rs
theochap: Ack, will move it there

theochap: Just to confirm, "these" mean all the traits in this file?

crates/network/src/lib.rs
```diff
@@ -36,6 +42,24 @@ pub use alloy_network_primitives::{
...
+pub type RpcReceiptEnvelope<T> = <T as Network>::ReceiptEnvelope;
+
+/// Adapter for network specific receipt response.
+pub type RpcReceiptResponse<T> = <T as Network>::ReceiptResponse;
+
+/// Adapter for network specific header response.
+pub type RpcHeader<T> = <T as Network>::Header;
+
+/// Adapter for network specific block type.
+pub type RpcBlock<T> = alloy_rpc_types_eth::Block<RpcTransaction<T>, RpcHeader<T>>;
```
klkvr, line 58: not sure if those aliases belong here

that one is also not exactly correct because from alloy pov there's a `BlockResponse` generic

crates/network/src/convert.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Converts a network receipt response to a primitive receipt type.
+    ///
+    /// # Returns
+    ///
+    /// Returns `Ok(Self)` on successful conversion, or `Err(Self::Error)` if the conversion fails.
+    fn from_receipt_response(receipt_response: N::ReceiptResponse) -> Result<Self, Self::Error>
+    where
+        Self: Sized;
+}
```
klkvr, line 122: those traits are now removed https://github.com/paradigmxyz/reth/issues/22320

crates/network/src/convert.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// An additional context, usually [`TransactionInfo`] in a wrapper that carries some
+    /// implementation specific extra information.
+    type TxInfo;
+    /// An associated RPC conversion error.
+    type Err: error::Error;
+
+    /// Performs the conversion consuming `tx` with `signer` and `tx_info`. See [`FromConsensusTx`]
+    /// for details.
+    fn from_consensus_tx(tx: T, signer: Address, tx_info: Self::TxInfo) -> Result<Self, Self::Err>;
+}
```
klkvr, line 38: not sure if this belongs to alloy, also note that you don't necessarily need those to be implemented for op-alloy types but could instead have converter types on op-reth side https://github.com/ethereum-optimism/optimism/blob/aadf0fcf078255fa973232f2f16bc06a8eeb514f/rust/op-reth/crates/rpc/src/eth/receipt.rs#L33

mattsse (commented):
we can mark this as completed

PR #3652, opened 2026-02-06
perf(serde): Remove unnecessary clone in OtherFields::deserialize_as

marukai67:
Removes redundant .clone() call in OtherFields::deserialize_as method, eliminating unnecessary allocation and memory copy.


mattsse (approved):
we can keep thi

PR #3654, opened 2026-02-07
fix(contract): check receipt status before contract address in deploy

snazzysam933:
When a contract constructor reverts, `deploy()` and `deploy_sync()` return the misleading error 
"**missing `contractAddress` from deployment transaction receipt**"
because the code only checks `contract_address()` without first checking `receipt.status()`.                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                       
adds a `status()` check before the `contract_address()` check, and introduces a new `Error::DeploymentReverted` variant. Now when the constructor reverts, users see "**deployment transaction reverted**" instead of being pointed toward a missing address problem that doesn't reflect the actual failure.

### DIFF
--- a/crates/contract/src/call.rs
+++ b/crates/contract/src/call.rs
@@ -574,7 +574,7 @@ impl<P: Provider<N>, D: CallDecoder, N: Network> CallBuilder<P, D, N> {
         let pending_tx = self.send().await?;
         let receipt = pending_tx.get_receipt().await?;
         if !receipt.status() {
-            return Err(Error::DeploymentReverted);
+            return Err(Error::ContractNotDeployed);
         }
         receipt.contract_address().ok_or(Error::ContractNotDeployed)
     }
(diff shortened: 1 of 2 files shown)


crates/contract/src/call.rs
```diff
@@ -600,6 +603,9 @@ impl<P: Provider<N>, D: CallDecoder, N: Network> CallBuilder<P, D, N> {
             return Err(Error::NotADeploymentTransaction);
         }
         let receipt = self.send_sync().await?;
+        if !receipt.status() {
+            return Err(Error::DeploymentReverted);
+        }
```
mattsse, line 608: can we reuse the existing ContractNotDeployed variant here?

snazzysam933: Good point, updated to reuse `ContractNotDeployed` with an updated error message.

PR #3658, opened 2026-02-07
feat(consensus-any): add serde aliases for cross-client compatibility in AnyHeader

marukai67:
Adds #[serde(alias)] attributes to AnyHeader fields to support alternative field naming conventions used by different Ethereum clients and networks.


mattsse:
Closing — these serde aliases aren't needed at this time. Thank you.

PR #3659, opened 2026-02-07
feat(rpc-types-eth): add `gasLimit` serde alias for TransactionRequest.gas field

Aleksandr1732:
Adds serde alias `gasLimit` for the `gas` field in `TransactionRequest` to improve compatibility with different JSON-RPC implementations.


mattsse:
Closing — the `gas` field name is per the Ethereum JSON-RPC spec. Adding a `gasLimit` alias without citing a specific RPC implementation that actually sends it is speculative and adds unnecessary surface area. If there's a concrete compatibility issue with a known implementation, please reopen with details.

PR #3660, opened 2026-02-07
feat(transport-ws): make keepalive interval configurable

gakonst:
## Summary

Add `with_keepalive_interval` to `WsConnect`, replacing the hardcoded 10s `KEEPALIVE` constant.

## Motivation

reth-bench persistence subscriptions disconnect during long MDBX fsyncs (29s+). The server can't respond to pings while stalled by I/O, so alloy's 10s keepalive kills the connection with "WS server missed a pong". Making this configurable lets consumers like reth-bench set a longer interval.

## Changes

- Add `keepalive_interval: Duration` field to `WsConnect` (default 10s, backwards-compatible)
- Add `with_keepalive_interval(Duration)` builder method
- Thread the interval through `WsBackend` and use it in `spawn()` instead of the hardcoded constant
- Update WASM backend to match the new `WsBackend` struct shape

## Testing

```
cargo check -p alloy-transport-ws
cargo clippy -p alloy-transport-ws -- -D warnings
cargo +nightly fmt --all --check
```

All clean.

### DIFF
--- a/crates/transport-ws/src/wasm.rs
+++ b/crates/transport-ws/src/wasm.rs
@@ -20,12 +20,20 @@ pub struct WsConnect {
     /// The interval between retries.
     /// Default is 3 seconds.
     retry_interval: Duration,
+    /// The interval between keepalive pings.
+    /// Default is 10 seconds.
+    keepalive_interval: Duration,
 }
 
 impl WsConnect {
     /// Creates a new websocket connection configuration.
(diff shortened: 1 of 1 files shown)


gakonst:
Wait why are we fsyncing for 29 seconds lol

DaniPopes:
on startup

PR #3663, opened 2026-02-08
Remove dead Clone impl for RootProviderInner

conomist:
remove the unused `Clone` implementation for `RootProviderInner`
avoid cloning a non-clonable `OnceLock<HeartbeatHandle>`


onbjerg (changes requested):
this is breaking. the `OnceLock` is also cloneable, otherwise this would not compile.

PR #3665, opened 2026-02-09
Add visit_string support for Id deserialization

splinter012:
Implemented visit_string in Deserialize visitor for Id. Allows deserialization from owned String values


mattsse:
I believe this clashes with the hex desered

PR #3674, opened 2026-02-10
ci: use install-action for zepter

snazzysam933:
Switch zepter installation from `cargo install` to `taiki-e/install-action` to match how we install everything else in CI. Saves a couple minutes of compiling from source on every run.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -275,9 +275,11 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+      - uses: taiki-e/install-action@1e67dedb5e3c590e1c9d9272ace46ef689da250d # v2
+        with:
+          tool: zepter
       - name: run zepter
         run: |
-          cargo install zepter -f --locked
           zepter --version
           time zepter run check
 


.github/workflows/ci.yml
```diff
@@ -275,9 +275,11 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+      - uses: taiki-e/install-action@1e67dedb5e3c590e1c9d9272ace46ef689da250d # v2
+        with:
+          tool: zepter
```
zerosnacks, line 280: It is not listed here: https://github.com/taiki-e/install-action/blob/main/TOOLS.md

Which makes me assume this would be installed w/ cargo-bininstall 

https://github.com/ggwpez/zepter?tab=readme-ov-file#install

Which is available

snazzysam933: Right, it goes through cargo-binstall, but zepter does publish precompiled binaries on GitHub Releases (linux/mac/windows) so binstall will just download the binary directly instead of compiling from source. Should go from ~1-3min to a few seconds.

zerosnacks: Yep I'm aware, I posted this mostly for the other reviewers as context - thanks for your PR

PR #3675, opened 2026-02-10
perf(tx-macros): Remove unnecessary clone in Serialize for typed transactions

splinter012:
Optimizes serialization of typed transactions by eliminating an unnecessary full enum clone. Instead of cloning the entire transaction envelope before conversion, the macro now generates code that accepts references and only clones the specific variant during conversion.

### DIFF
--- a/crates/tx-macros/src/expand.rs
+++ b/crates/tx-macros/src/expand.rs
@@ -820,7 +820,8 @@ impl Expander {
             quote! { #alloy_consensus::private::alloy_serde::reject_if_some }.to_string();
 
         let typed_names = self.variants.typed.iter().map(|v| &v.name).collect::<Vec<_>>();
-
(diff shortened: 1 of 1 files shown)


crates/tx-macros/src/expand.rs
```diff
@@ -919,13 +920,26 @@ impl Expander {
...
                 }
 
+                impl #impl_generics From<&#typed_name #ty_generics> for #tagged_enum_name #ty_generics
+                where
+                    #(#variant_types: Clone,)*
+                {
+                    fn from(value: &#typed_name #ty_generics) -> Self {
+                        match value {
+                            #(
+                                #typed_name::#typed_names(tx) => Self::#typed_names(tx.clone()),
```
mattsse, line 930: this also clones

PR #3676, opened 2026-02-11
docs(transport-ws): align WASM keepalive docs with actual behavior

0xcharry:
Updated `with_keepalive_interval` docs to clarify that the WASM backend currently does not send websocket ping frames. Noted that the value is stored for API parity with native transports and reserved for future keepalive support. Removed misleading keepalive claims from the `spawn()` loop comment.


mattsse:
Closing — these documentation changes don't add meaningful clarity over the existing comments. Thank you.

PR #3678, opened 2026-02-11
ci: pin aws-smithy crates for MSRV 1.88 compatibility

neowangreal:
Recent aws-smithy-* releases bumped their MSRV to 1.91, which breaks the 1.88 CI builds since we don't commit Cargo.lock. Pin them to the last known compatible versions before the MSRV build step.


neowangreal:
OMG finally!

PR #3682, opened 2026-02-12
feat(consensus): add `author` serde alias for Header beneficiary field

0xxFloki:
Adds author as a serde deserialization alias for the beneficiary field in consensus::Header, improving compatibility with Ethereum clients that use author instead of miner in their JSON-RPC responses.


mattsse:
Closing — this change isn't needed at this time. Thank you.

PR #3683, opened 2026-02-13
feat(consensus): add `EthereumReceipt` type ported from reth

stevencartavia:
## Motivation

Port reth's `EthereumReceipt` into `alloy-consensus` as a step towards unifying receipt types between alloy and reth.

## Solution

Add `crates/consensus/src/receipt/receipt2.rs` containing:

- **`TxTy` trait** — bounds alias for the tx type generic (`Debug + Copy + Eq + Send + Sync + InMemorySize + Typed2718 + TryFrom<u8> + Decodable + 'static`)
- **`EthereumReceipt<T, L>` struct** — generic over tx type and log type, with `tx_type`, `success`, `cumulative_gas_used`, `logs` fields
- **Trait impls**: `TxReceipt`, `Typed2718`, `IsTyped2718`, `Eip2718EncodableReceipt`, `Eip2718DecodableReceipt`, `RlpEncodableReceipt`, `RlpDecodableReceipt`, `InMemorySize`
- **`From` conversions**: `ReceiptEnvelope → EthereumReceipt`, `EthereumReceipt → Receipt`, `EthereumReceipt<TxType> → ReceiptEnvelope`
- **`serde_bincode_compat`** module for bincode serialization support

Wire it up via `mod.rs` and re-export `EthereumReceipt` and `TxTy` from the crate root.

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipt2.rs
+++ b/crates/consensus/src/receipt/receipt2.rs
@@ -386,6 +386,7 @@ pub(crate) mod serde_bincode_compat {
     mod tests {
         use crate::TxType;
         use arbitrary::Arbitrary;
+        use bincode::config;
         use rand::Rng;
         use serde_with::serde_as;
 
@@ -402,14 +403,15 @@ pub(crate) mod serde_bincode_compat {
             }
 
             let mut bytes = [0u8; 1024];
-            rand::rng().fill(bytes.as_mut_slice());
+            rand::thread_rng().fill(bytes.as_mut_slice());
             let data = Data {
                 receipt: EthereumReceipt::arbitrary(&mut arbitrary::Unstructured::new(&bytes))
                     .unwrap(),
             };
 
-            let encoded = bincode::serialize(&data).unwrap();
-            let decoded: Data = bincode::deserialize(&encoded).unwrap();
+            let encoded = bincode::serde::encode_to_vec(&data, config::legacy()).unwrap();
+            let (decoded, _): (Data, _) =
+                bincode::serde::decode_from_slice(&encoded, config::legacy()).unwrap();
             assert_eq!(decoded, data);
         }
     }


crates/consensus/src/receipt/mod.rs
```diff
@@ -27,6 +27,11 @@ pub trait TxReceipt: Clone + fmt::Debug + PartialEq + Eq + Send + Sync {
     /// The associated log type.
     type Log;
 
+    /// Returns the EIP-2718 transaction type.
+    fn tx_type(&self) -> u8 {
+        0
+    }
+
```
onbjerg: we should not provide a default here

crates/consensus/src/receipt/envelope.rs
```diff
@@ -188,6 +188,16 @@ where
...
 
+    fn tx_type(&self) -> u8 {
+        match self {
+            Self::Legacy(_) => LEGACY_TX_TYPE_ID,
+            Self::Eip2930(_) => EIP2930_TX_TYPE_ID,
+            Self::Eip1559(_) => EIP1559_TX_TYPE_ID,
+            Self::Eip4844(_) => EIP4844_TX_TYPE_ID,
+            Self::Eip7702(_) => EIP7702_TX_TYPE_ID,
+        }
+    }
```
mattsse: This is basically:

https://github.com/alloy-rs/alloy/blob/db11d05eb8b045c2c90c0cadb2dd3ab36ec605f3/crates/eips/src/eip2718.rs#L344-L344

can we instead add this to the TxReceipt, all types should already satisfy this

crates/consensus/src/receipt/receipts.rs
```diff
@@ -78,6 +78,10 @@ where
 {
     type Log = T;
 
+    fn tx_type(&self) -> u8 {
+        0
+    }
```
mattsse: this feels wrong

mattsse (changes requested):
if we need this logic then we should be able to selectively enforce the Typed trait but always requiring a txtype introduces issues for

https://github.com/alloy-rs/alloy/blob/db11d05eb8b045c2c90c0cadb2dd3ab36ec605f3/crates/consensus/src/receipt/receipts.rs#L21-L21

stevencartavia:
Updated based on review, `tx_type()` is no longer on the `TxReceipt` trait. Instead, `Receipt` now has a `tx_type: u8` field exposed via `Typed2718`. `TransactionReceipt::tx_type()` bounds on `Typed2718` instead of `TxReceipt`.

crates/consensus/src/receipt/envelope.rs
```diff
@@ -58,12 +58,14 @@ impl<T> ReceiptEnvelope<T> {
...
-            TxType::Eip2930 => Self::Eip2930(receipt.into()),
-            TxType::Eip1559 => Self::Eip1559(receipt.into()),
-            TxType::Eip4844 => Self::Eip4844(receipt.into()),
-            TxType::Eip7702 => Self::Eip7702(receipt.into()),
+            TxType::Legacy => Self::Legacy(receipt),
+            TxType::Eip2930 => Self::Eip2930(receipt),
+            TxType::Eip1559 => Self::Eip1559(receipt),
+            TxType::Eip4844 => Self::Eip4844(receipt),
+            TxType::Eip7702 => Self::Eip7702(receipt),
         }
```
mattsse: this is a bit redundant imo

PR #3693, opened 2026-02-14
refactor(provider): deduplicate call conversion in MulticallBuilder

marukai67:
Extracts repeated call conversion logic in MulticallBuilder into a single private helper method, eliminating code duplication across six internal builder methods.

### DIFF
--- a/crates/provider/src/provider/multicall/mod.rs
+++ b/crates/provider/src/provider/multicall/mod.rs
@@ -258,6 +258,14 @@ where
     P: Provider<N>,
     N: Network,
 {
+    #[inline]
+    fn legacy_calls(&self) -> Vec<Call> {
+        self.calls
+            .iter()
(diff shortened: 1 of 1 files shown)


mattsse (approved):

DaniPopes (approved):

crates/provider/src/provider/multicall/mod.rs
```diff
@@ -258,6 +258,14 @@ where
     P: Provider<N>,
     N: Network,
 {
+    #[inline]
+    fn legacy_calls(&self) -> Vec<Call> {
```
DaniPopes, line 262: please move to the bottom of the impl block and remove the inline attribute

mattsse:
Closing — this refactoring wasn't requested and the existing code is clear as-is. Thank you.

PR #3706, opened 2026-02-17
perf(rpc-types-trace): optimize input truncation in Otterscan serialization

splinter012:
Replace `chars().take().collect()` with in-place `truncate()` when serializing Otterscan transaction inputs to eliminate unnecessary allocations.


mattsse:
Closing — this is an unsolicited micro-optimization in a non-hot path that doesn't provide meaningful benefit. Thank you for the contribution.

PR #3710, opened 2026-02-17
feat(consensus): add `EthereumReceipt::map_logs`

stevencartavia:
Add `pub fn map_logs` on `impl<T, L> EthereumReceipt<T, L>`, following the same pattern as [`Receipt::map_logs`](https://github.com/alloy-rs/alloy/blob/54b1f6645d45a1877cea90a5e0d3850341e31168/crates/consensus/src/receipt/receipts.rs#L38-L41).

### DIFF
--- a/crates/consensus/src/receipt/receipt2.rs
+++ b/crates/consensus/src/receipt/receipt2.rs
@@ -73,7 +73,12 @@ impl<T, L> EthereumReceipt<T, L> {
     /// Returns the receipt with the new log type.
     pub fn map_logs<U>(self, f: impl FnMut(L) -> U) -> EthereumReceipt<T, U> {
         let Self { tx_type, success, cumulative_gas_used, logs } = self;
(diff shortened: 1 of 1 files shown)


crates/rpc-types-eth/src/log.rs
```diff
@@ -152,6 +152,21 @@ impl Log<LogData> {
...
+/// number, transaction hash, transaction index, and log index.
+pub fn into_rpc_receipt<T>(
+    receipt: EthereumReceipt<T>,
+    next_log_index: usize,
+    meta: TransactionMeta,
+) -> EthereumReceipt<T, Log> {
+    let EthereumReceipt { tx_type, success, cumulative_gas_used, logs } = receipt;
+    let logs = Log::collect_for_receipt(next_log_index, meta, logs);
+    EthereumReceipt { tx_type, success, cumulative_gas_used, logs }
+}
```
mattsse: wait, isnt this just map receipt?

I actually dont like a standalone fn for this if this can also be solved with a simple map

PR #3711, opened 2026-02-17
fix(signer-ledger): bounds check APDU response in get_address

vanityKoa:
`get_address` indexes into the APDU response without checking bounds, which panics if the device returns a malformed/short response. Use `.first()`/`.get()` and return `ShortResponse` instead, matching what `version()` already does.

Before:
  result[0]              // panics on empty response
  result[offset]         // panics if pub_key_len > actual data
  result[offset+1..end]  // panics if address data is truncated

After:
  result.first().ok_or(ShortResponse { .. })?
  result.get(offset).ok_or(ShortResponse { .. })?
  result.get(offset+1..end).ok_or(ShortResponse { .. })?


mattsse (commented):
i think we can expect this to work properly

PR #3715, opened 2026-02-18
fix: eliminate duplicate length calculations in receipt RLP encoding

Galoretka:
The `rlp_encode_with_bloom` and `rlp_encoded_length_with_bloom` methods were calling `eip2718_encoded_length_with_bloom` multiple times for the same bloom value, causing redundant computations.
Cached the result of `eip2718_encoded_length_with_bloom` in a local variable before using it for both the Header construction and the subsequent encoding call.


mattsse:
Closing — the only actual duplicate call is in test code, and the production change is purely cosmetic variable extraction. Not worth the churn. Thanks for the interest though.

PR #3719, opened 2026-02-18
fix(consensus): use unchecked recovery in recover_unchecked_with_buf default impl

vanityKoa:
The default implementation of `recover_unchecked_with_buf` falls back to `recover_signer()` instead of `recover_signer_unchecked()`. This means the "unchecked" variant still applies EIP-2 low-S validation, which silently rejects valid pre-EIP-2 historical signatures for any type relying on the default impl.

One-line fix: call `recover_signer_unchecked()` to match the documented behavior and the pattern used by all other unchecked methods in the trait.


mattsse (approved):
LGTM — correct bug fix. The default impl of `recover_unchecked_with_buf` was incorrectly calling checked `recover_signer()` instead of `recover_signer_unchecked()`, contradicting its name, docs, and caller expectations.

PR #3721, opened 2026-02-19
fix(rpc-types): off-by-one in TransactionConditional has_exceeded checks

vanityKoa:
`has_exceeded_block_number` and `has_exceeded_timestamp` used `>=` while `matches_block_number` and `matches_timestamp` used `>` for the max boundary, so at `block_number == max` a tx would both "match" and be "exceeded". Changed to `>` to keep inclusive range semantics consistent across both sets of helpers.

Before this fix, with `block_number_max = 1000`:
- `matches_block_number(1000)` → true (tx is valid)
- `has_exceeded_block_number(1000)` → true (tx is expired)

Any caller that short-circuits on `has_exceeded` would incorrectly reject a tx at its exact max block.

### DIFF
--- a/crates/rpc-types-eth/src/erc4337.rs
+++ b/crates/rpc-types-eth/src/erc4337.rs
@@ -75,17 +75,17 @@ impl TransactionConditional {
         self.has_exceeded_block_number(block.number) || self.has_exceeded_timestamp(block.timestamp)
     }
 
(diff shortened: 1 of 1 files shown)


mattsse (approved):
Correct fix — the `>=` was inconsistent with `matches_block_number` which uses `>` for the same max boundary.

PR #3722, opened 2026-02-20
feat: `Provider::watch_blocks_from` and `Provider::watch_canonical_blocks_from`

0xForerunner:
- Adds watch_blocks_from and watch_canonical_blocks_from provider APIs for historical backfill + live tailing using polling.
- Streams yield request futures so callers can control concurrency via buffering and reuse transport retry layers.
- Includes unit tests for progression, recovery after errors, buffering, provider drop semantics, fixed block tags, and cursor-advance guard behavior.

We can now get a historical, concurrent, and reorg aware block stream with live tailing just like this:

```rust
    provider
        .watch_blocks_from(start_block)
        .poll_interval(Duration::from_millis(500))
        .hashes()
        .block_tag(BlockNumberOrTag::Latest)
        .canonical()
        .max_reorg_depth(16)
        .rpc_concurrency(rpc_concurrency)
        .into_stream()
        // Concurrently process canonical events.
        .for_each_concurrent(process_concurrency, |result| async move {
            match result {
                Ok(CanonicalEvent::Added(block)) => {
                    let number = block.header().number();
                    let txs = block.transactions().len();
                    println!("processed block {number} ({txs} txs)");
                }
                Ok(CanonicalEvent::Removed(block)) => {
                    let number = block.header().number();
                    println!("reorged block {number}");
                }
                Err(err) => eprintln!("failed to fetch block: {err}"),
            }
        })
        .await;
```

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes


0xForerunner:
I've updated this PR to remove `watch_logs_from`, because I think without reorg handling it's not useful. Sticking with `watch_blocks_from` and `watch_canonical_blocks_from`.

0xForerunner:
@DaniPopes @mattsse Would love a review on this when you've got a minute!

crates/provider/src/provider/watch_blocks_from.rs
```diff
@@ -0,0 +1,528 @@
...
+use tokio::time::sleep;
+
+const DEFAULT_POLL_INTERVAL: Duration = Duration::from_secs(1);
+
+/// A builder for streaming blocks from a historical block and continuing indefinitely.
+#[derive(Debug, Clone)]
+#[must_use = "this builder does nothing unless you call `.into_stream`"]
+pub struct WatchBlocksFrom<N: Network> {
+    /// A weak reference to the provider's RPC client, used to make requests in the stream.
+    pub client: WeakClient,
```
mattsse: can we make this private as well

crates/provider/src/provider/watch_blocks_from.rs
```diff
@@ -0,0 +1,528 @@
...
+    /// Other errors are surfaced to the caller. Configure retries on the underlying client
+    /// transport (for example with `RetryBackoffLayer`) for transport-level retry behavior.
+    ///
+    /// This can be buffered by the caller, for example with
+    /// [`StreamExt::buffered`](futures::StreamExt::buffered).
+    pub fn into_stream(
+        self,
+    ) -> impl Stream<
+        Item = Pin<Box<dyn Future<Output = TransportResult<N::BlockResponse>> + Send + 'static>>,
+    > + Unpin
```
mattsse: I want named types for this, because eventually we always need extra logic, and this isnt really easy to maintain

crates/provider/src/provider/watch_blocks_from.rs
```diff
@@ -0,0 +1,528 @@
...
+           + 'static {
+        let Self { client, start_block, poll_interval, block_tag, kind, _phantom } = self;
+
+        type Fut<T> = Pin<Box<dyn Future<Output = TransportResult<T>> + Send + 'static>>;
+
+        let stream = stream! {
```
mattsse: I'd much prefer if we have a dedicated stream impl for this

mattsse (changes requested):
this feature makes sense and I wanted something like this in the past.

because I believe we eventually want some additional logic on this, I really want this to be an actual stream type because maitnaining ~100 of stream! code is very difficult

crates/provider/src/provider/watch_blocks_from.rs
0xForerunner: Works for me! Are you okay with `futures::stream::unfold`, or do you prefer a direct `impl Stream`

0xForerunner: Maybe you could let me know what you'd like the function signature to look like for WatchBlocksFrom::into_stream, so I can make sure I make the right types here.

crates/provider/src/provider/watch_blocks_from.rs
0xForerunner: Do you want a dedicated type here for the Stream, the future, or both?

0xForerunner:
@mattsse I've switched to using manual `impl Stream` state machines. I am still using boxed Futures. If you would like me to switch over to named and manually impled futures lemme know! We can just name the one Future that's in the api, or if you want to get rid of all the boxed futures I can implement a state machine for each of these.

0xForerunner:
@mattsse I've also created concrete types for the futures here, and gotten rid of the Boxing. Feel free to revert the last two commits if you preferred it before. Should be good to go for another review, Thanks!

0xForerunner:
@mattsse @klkvr bumping!

0xForerunner:
@mattsse bumping again!

mattsse (approved):
Looks good to me.

One non-blocking note: `watch_canonical_blocks_from` does not emit a same-height tip replacement immediately. If block `n` is replaced while `eth_blockNumber` still reports `n`, the stream waits because the cursor has already advanced to `n + 1`; the reorg is then reconciled once `n + 1` arrives on top of the replacement. That behavior is fine for this API as long as delayed tip reorg reporting is acceptable/documented.

PR #3723, opened 2026-02-20
feat(network): generic `TransactionResponse` to `TransactionRequest` conversion

mablr:
## Motivation

It would be useful in Foundry to get new generic `TransactionResquest` from `TransactionResponse`:

## Solution

Rebased on #3731
- Fixed `impl From<Transaction> for TransactionRequest`to use `into_recovered` in order to properly recover from field in the output `TransactionRequest`
- Added `impl From<AnyRpcTransaction> for WithOtherFields<TransactionRequest>`
- Add Clone on `TxEnvelope`
- It's breaking but Ethereum's, Tempo's, Op's impls already have it

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/provider/src/builder.rs
+++ b/crates/provider/src/builder.rs
@@ -823,7 +823,7 @@ mod tests {
     #[test]
     fn network_replaces_fillers() {
         // Add an extra filler before swapping, it should be dropped.
-        let builder = ProviderBuilder::new().filler(GasFiller).network::<AnyNetwork>();
+        let builder = ProviderBuilder::new().filler(GasFiller::default()).network::<AnyNetwork>();
 
         let _: ProviderBuilder<
             Identity,


mattsse (commented):
this makes sense and the clone should be fine,

I just want to point out that we have:

https://github.com/alloy-rs/alloy/blob/e951250284633f3a89abf843961c0c062c3c8596/crates/rpc-types-eth/src/transaction/request.rs#L153-L153

and

https://github.com/alloy-rs/alloy/blob/e951250284633f3a89abf843961c0c062c3c8596/crates/rpc-types-eth/src/transaction/request.rs#L195-L195

so maybe for this use cases we want a helper fn on the rpc tx response?

mattsse:
@mablr I think we could even enforce this on the type txresponse level that this must be from/into

mattsse:
we also have 

https://github.com/alloy-rs/alloy/blob/cab69be237e582cc23b5d63456789ac9d1c7ca7d/crates/rpc-types-eth/src/transaction/mod.rs#L200-L211

which is a bit suboptimal because this does recovery again

mablr (commented):
@mattsse Just rebased this on your #3731, to add direct `From<Self::TransactionResponse>` bound to `Network::TransactionRequest`.

- Fixed `impl From<Transaction> for TransactionRequest`to use `into_recovered` in order to properly recover `from` field in the output `TransactionRequest`
- Added `impl From<AnyRpcTransaction> for WithOtherFields<TransactionRequest>`

cc @onbjerg

PR #3724, opened 2026-02-20
feat(network): signature flow generic over `Network`

mablr:
## Motivation

Make signature flow generic over `Network`, this is necessary in Foundry.

## Solution
- generic `NetworkWallet<N>` impl for `EthereumWallet` given :
  - `N::TxEnvelope: From<Signed<N::UnsignedTx>>`
  - `N::UnsignedTx: SignableTransaction<Signature>`
- `SignableTransaction<Signature>` impl for `AnyTypedTransaction` to make it compatible with `impl NetworkWallet<N>` (fallback to Ethereum)
-  `SignableTransaction<Signature>` impl for  `TypedTransaction` was already there 👌

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       # Only run tests on latest stable and above
       - name: Install cargo-nextest
         if: ${{ matrix.rust != '1.91' }} # MSRV
-        uses: taiki-e/install-action@cfdb446e391c69574ebc316dfb7d7849ec12b940 # v2
+        uses: taiki-e/install-action@d6e286fa45544157a02d45a43742857ebbc25d12 # v2
         with:
           tool: nextest
       - name: build
@@ -102,7 +102,7 @@ jobs:
         with:
           toolchain: stable
           target: wasm32-unknown-unknown
-      - uses: taiki-e/install-action@cfdb446e391c69574ebc316dfb7d7849ec12b940 # v2
+      - uses: taiki-e/install-action@d6e286fa45544157a02d45a43742857ebbc25d12 # v2
         with:
           tool: cargo-hack
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
@@ -138,7 +138,7 @@ jobs:
         with:
           toolchain: stable
           target: wasm32-wasip1
-      - uses: taiki-e/install-action@cfdb446e391c69574ebc316dfb7d7849ec12b940 # v2
+      - uses: taiki-e/install-action@d6e286fa45544157a02d45a43742857ebbc25d12 # v2
         with:
           tool: cargo-hack
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
@@ -168,7 +168,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@e97e2d8cc328f1b50210efc529dca0028893a2d9 # master
         with:
           toolchain: stable
-      - uses: taiki-e/install-action@cfdb446e391c69574ebc316dfb7d7849ec12b940 # v2
+      - uses: taiki-e/install-action@d6e286fa45544157a02d45a43742857ebbc25d12 # v2
         with:
           tool: cargo-hack
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
@@ -191,7 +191,7 @@ jobs:
         with:
           toolchain: stable
           targets: riscv32imac-unknown-none-elf
-      - uses: taiki-e/install-action@cfdb446e391c69574ebc316dfb7d7849ec12b940 # v2
+      - uses: taiki-e/install-action@d6e286fa45544157a02d45a43742857ebbc25d12 # v2
         with:
           tool: cargo-hack
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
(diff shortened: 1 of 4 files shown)


crates/network/src/lib.rs
```diff
@@ -64,9 +64,12 @@ pub trait Network: Debug + Clone + Copy + Sized + Send + Sync + 'static {
     #[doc(alias = "TransactionEnvelope")]
     type TxEnvelope: Eip2718Envelope + Transaction + Debug;
 
+    /// The network signature type.
+    type Signature;
+
     /// An enum over the various transaction types.
     #[doc(alias = "UnsignedTransaction")]
-    type UnsignedTx: From<Self::TxEnvelope>;
+    type UnsignedTx: From<Self::TxEnvelope> + SignableTransaction<Self::Signature>;
```
mablr: Configurable `Signature` is maybe overkill, but offers potentially full support for Tempo's signature type.

mablr:
> * transactions that are not signable (e.g. optimism's deposit transaction)

`SignableTransaction` trait seems to let you free to do no-op for deposit tx, at least (as far as i understand) this is how it's implemented in op-alloy
https://github.com/alloy-rs/op-alloy/blob/a93a08a021e80db8133f5f7fdccbb5639bc68d45/crates/consensus/src/transaction/typed.rs#L272-L273

> * network with different signature types (e.g. tempo has some txs that can be signed with 1 of 3 signature types, as well as some transactions that can only be signed with 1 signature type)

This is the reason why I added `Network::Signature` type, so that it can be to network's specific one, i.e. `TempoSignature`, but I'm maybe wrong on this...

mablr (commented):
@onbjerg If you prefer less invasive change here, I can re-target this PR.

I actually just need this trait bound for its `encoded_for_signing()` method in Foundry.

So I can :
- revert `Signature` type on `Network` trait
- set simply`Network::UnsignedTx: SignableTransaction<alloy_primitives::Signature>`

**Edit**: I realize we would also need generic `impl NetworkWallet<N> for EthereumWallet`, so this makes `SignableTransaction<alloy_primitives::Signature>` really crucial - I think.

In the following, we could remove `UnsignedTx`/`TxEnvelope` type bindings:
https://github.com/alloy-rs/alloy/blob/c7b548a568266d49e7c638be617899ebb9a66648/crates/network/src/ethereum/wallet.rs#L120-L123

onbjerg:
this might work, wdyt @mattsse? i'm not sure about the silent behavior in AnyTypedTransaction but that type is cursed anyway

crates/network/src/lib.rs
```diff
@@ -62,11 +63,11 @@ pub trait Network: Debug + Clone + Copy + Sized + Send + Sync + 'static {
 
     /// The network transaction envelope type.
     #[doc(alias = "TransactionEnvelope")]
-    type TxEnvelope: Eip2718Envelope + Transaction + Debug;
+    type TxEnvelope: Eip2718Envelope + Transaction + Debug + From<Signed<Self::UnsignedTx>>;
 
     /// An enum over the various transaction types.
     #[doc(alias = "UnsignedTransaction")]
-    type UnsignedTx: From<Self::TxEnvelope>;
+    type UnsignedTx: From<Self::TxEnvelope> + SignableTransaction<Signature>;
```
mattsse: imo it's a bad idea to enshrine the signature type that will just cause problems eventually

mablr: > will just cause problems eventually

@mattsse Are you saying that there are networks that don't support alloy_primitive::Signature at all for any tx type variant of their envelope? In such case, yes SignableTransaction<Signature> may be problematic.

I see two main cases:
- Signable txs that need custom signature, like Tempo’s AA
  - if this custom signature also support primitive one (tempo case), it’s straightforward
  - otherwise we need generic signature type on Network
- Non-signable, like deposit, signature on typed tx get simply dropped during into envelope conversion.

mablr: The actual problem of Alloy is that the generic `impl NetworkWallet<N> for EthereumWallet` enforces concrete Ethereum `TypedTransaction`/`TxEnvelope` types. As the goal is to relax that, let's put the bounds here:

```rust
impl<N: Network> NetworkWallet<N> for EthereumWallet 
where
    N::TxEnvelope: From<Signed<N::UnsignedTx>>, // OR TryFrom<Signed<N::UnsignedTx>>
    N::UnsignedTx: SignableTransaction<Signature>
{
```

So that the PR becomes non-breaking, as all Networks can have now their own `NetworkWallet<CUSTOMNETWORK> for EthereumWallet` impl if they don't match the bounds, masking the generic (as it used to be).

- These bounds on `TxEnvelope`/`UnsignedTx` will now be enforced outside alloy.
- `TryFrom<Signed<N::UnsignedTx>>` could be beneficial to return error when attempting to a non-signable tx (eg: deposit)

crates/network/src/ethereum/wallet.rs
```diff
@@ -117,9 +120,10 @@ impl EthereumWallet {
     }
 }
 
-impl<N> NetworkWallet<N> for EthereumWallet
+impl<N: Network> NetworkWallet<N> for EthereumWallet
 where
-    N: Network<UnsignedTx = TypedTransaction, TxEnvelope = TxEnvelope>,
+    N::TxEnvelope: From<Signed<N::UnsignedTx>>,
+    N::UnsignedTx: SignableTransaction<Signature>,
```
mablr, line 126: This is not breaking anymore for `Network` trait.

But flexible enough to handle signing through `EthereumWallet` for generic tx requests and typed transactions.

In op-alloy `From<Signed<N::UnsignedTx>>` is already implemented, lmk if you want instead `TryFrom<Signed<N::UnsignedTx>>`, to trigger error on  Deposit tx. (see how it looks currently: https://github.com/alloy-rs/op-alloy/compare/main...mablr:op-alloy:chore/pre_bump_alloy_3724)

PR #3729, opened 2026-02-21
fix(provider): reap_timeouts not firing for txs already in waiting_confs

vanityKoa:
When a tx with `with_timeout()` gets included in a block but still needs more confirmations, it moves from `unconfirmed` to `waiting_confs`. The `reap_at` entry stays, but when the timeout fires, `reap_timeouts` only checks `unconfirmed` — the remove returns None and the timeout is silently swallowed.

Example: 
user sends a tx with `required_confirmations = 20` and `timeout = 60s`. The tx lands in a block after 5s and moves to `waiting_confs`. At 60s the timeout fires, but `unconfirmed.remove()` finds nothing, so the user never gets a `Timeout` error and hangs forever if confirmations stall.

Now `reap_timeouts` falls back to scanning `waiting_confs` when the tx isn't in `unconfirmed`.


onbjerg:

PR #3730, opened 2026-02-21
feat(network-primitives): add blob_versioned_hashes to TransactionBuilder4844

figtracer:
## Summary

Adds `blob_versioned_hashes` getter/setter/builder methods to the `TransactionBuilder4844` trait.

`TransactionRequest` already stores `blob_versioned_hashes` as a field and uses it extensively
for EIP-4844 type detection, validation, and tx building. However, the `TransactionBuilder4844`
trait only exposes `max_fee_per_blob_gas` and `blob_sidecar` — there is no way for generic code
(`T: TransactionBuilder4844`) to read or write blob versioned hashes.

This is needed by downstream consumers (e.g. Foundry) that have functions generic over
`T: TransactionBuilder4844` and need to populate EVM `TxEnv.blob_hashes` from the transaction
request.

## Changes

- Add `blob_versioned_hashes(&self) -> Option<&[B256]>` (default: `None`)
- Add `set_blob_versioned_hashes(&mut self, Vec<B256>)` (default: no-op)
- Add `with_blob_versioned_hashes(self, Vec<B256>) -> Self` (builder pattern)
- Add `WithOtherFields<T>` delegation for the new methods
- Implement on `TransactionRequest` backed by the existing `blob_versioned_hashes` field

All defaults are no-ops so existing trait implementations are unaffected.


mablr (commented):
For more context: this is not anymore on the Foundry Cast refactor hotpath. See https://github.com/foundry-rs/foundry/pull/13512.

But I still think it's a nice a "nice-to-have" feat, to add **not before** Alloy 2.0 and its `TransactionBuilder4844` unification #3600

figtracer:
> For more context: this is not anymore on the Foundry Cast refactor hotpath. See [foundry-rs/foundry#13512](https://github.com/foundry-rs/foundry/pull/13512).
> 
> But I still think it's a nice a "nice-to-have" feat, to add **not before** Alloy 2.0 and its `TransactionBuilder4844` unification #3600

will note this down since we might have more cases after release

PR #3731, opened 2026-02-22
feat: impl From<Recovered<T>> for TransactionRequest

mattsse:
Implements `From<Recovered<T>>` for `TransactionRequest` where `T: Transaction`, delegating to the existing `from_recovered_transaction` method.

ref #3723

cc @mablr


mablr (approved):
Lgtm!

PR #3732, opened 2026-02-22
fix(txpool): eliminate redundant allocations in `TxpoolInspectSummary` deserialization

sashass1315:
The previous implementation used `split().collect::<Vec<_>>()` four times during parsing, creating unnecessary heap allocations for each deserialization.

Replace `split().collect()` pattern with zero-allocation `split_once()` and `strip_suffix()` methods, matching the approach already used elsewhere in the codebase (e.g., provider/src/ext/mev/with_auth.rs).


mattsse (commented):
we can keep as is

PR #3747, opened 2026-02-25
feat: Add support for chunked queries

### LINKED ISSUE #3744: [Feature] Add support for chunked queries
### Component

rpc

https://github.com/alloy-rs/alloy/blob/5896ca918f1890e75edab36ef98661ddcd9c1801/crates/contract/src/event.rs#L52-L57

@stevencartavia similar to how you implemented this on foundry

### Additional context

_No response_

stevencartavia:
## Motivation
closes #3744

## Solution
Adds `query_chunked` and `query_raw_chunked` methods to `Event` in `crates/contract/src/event.rs`. Both try the full block range optimistically first, and if that fails, split the range into `chunk_size`-block windows queried concurrently (up to 5 in parallel). If an individual chunk fails, it falls back to block-by-block for that chunk. Similar to the approach implemented in foundry-rs/foundry#12692.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/contract/src/event.rs
+++ b/crates/contract/src/event.rs
@@ -740,7 +740,7 @@ mod tests {
     }
 
     /// Verifies that the chunked query algorithm collects the correct logs and preserves
-    /// block ordering when events are spread across a large range requiring many chunks.
+    /// block ordering when events are spread across a range requiring multiple chunks.
     #[tokio::test]
     async fn chunked_query_collects_and_orders_logs() {
         use alloy_provider::ext::AnvilApi;
@@ -756,19 +756,19 @@ mod tests {
 
         let contract = MyContract::deploy(&provider).await.unwrap();
 
-        // Emit three events spread across a 10k-block range: start, middle, and end.
-        // This stresses chunk boundary handling and result ordering after the merge.
+        // Emit three events spread across a 100-block range: start, middle, and end.
+        // This exercises chunk boundary handling and result ordering after the merge.
         contract.doEmit().send().await.unwrap().get_receipt().await.unwrap();
-        provider.anvil_mine(Some(4998), None).await.unwrap();
+        provider.anvil_mine(Some(48), None).await.unwrap();
         contract.doEmit().send().await.unwrap().get_receipt().await.unwrap();
-        provider.anvil_mine(Some(4998), None).await.unwrap();
+        provider.anvil_mine(Some(48), None).await.unwrap();
         contract.doEmit().send().await.unwrap().get_receipt().await.unwrap();
 
-        // Use chunk_size=7 to force ~1429 chunks, exercising the concurrent dispatch and merge.
+        // Use chunk_size=7 to force ~15 chunks, exercising the concurrent dispatch and merge.
         // get_logs_chunked() bypasses the optimistic full-range attempt so the chunking code
         // is always exercised. query_raw() serves as the reference via the optimistic path.
         let event =
-            contract.MyEvent_filter().from_block(0u64).to_block(10_000u64).chunked().chunk_size(7);
+            contract.MyEvent_filter().from_block(0u64).to_block(100u64).chunked().chunk_size(7);
 
         let chunked = event.get_logs_chunked().await.unwrap();
         let reference = event.query_raw().await.unwrap();


crates/contract/src/event.rs
```diff
@@ -161,6 +161,107 @@ impl<P: Provider<N>, E: SolEvent, N: Network> Event<P, E, N> {
...
 }
 
+impl<P: Provider<N> + Clone, E: SolEvent, N: Network> Event<P, E, N> {
+    /// Queries the blockchain for the selected filter using chunked requests to handle large
+    /// block ranges, and returns a vector of matching event logs.
+    ///
+    /// First tries the full range optimistically. If that fails, splits the range into chunks
+    /// of `chunk_size` blocks and queries them concurrently (up to 5 in parallel).
+    /// If an individual chunk fails, falls back to block-by-block queries for that chunk.
+    pub async fn query_chunked(&self, chunk_size: u64) -> Result<Vec<(E, Log)>, Error> {
```
mattsse: can we design this a bi differently, by introducing a future type.

then we can do `.chunked().concurrent(...)` etc.

then we can also easily add stream support for this

crates/contract/src/event.rs
```diff
@@ -218,6 +319,13 @@ impl<E: SolEvent> EventPoller<E> {
     }
 }
 
+fn extract_block_range(filter: &Filter) -> (Option<u64>, Option<u64>) {
+    let FilterBlockOption::Range { from_block, to_block } = &filter.block_option else {
+        return (None, None);
+    };
+    (from_block.and_then(|b| b.as_number()), to_block.and_then(|b| b.as_number()))
+}
```
mattsse: this should be a native, fn of filter, maybe we added this laready?

crates/contract/src/event.rs
```diff
@@ -168,6 +194,141 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+                                    let single_filter =
+                                        chunk_filter.clone().from_block(block).to_block(block);
+                                    fallback_logs.extend(provider.get_logs(&single_filter).await?);
+                                }
+                                Ok((start_block, fallback_logs))
+                            }
+                        }
+                    }
+                })
+                .buffered(self.max_concurrent)
```
mattsse: since you sort by `start_block` afterwards anyway, `.buffer_unordered()` would be a better fit here — `.buffered()` preserves arrival order which causes head-of-line blocking if an earlier chunk is slow while later ones already finished. no need to pay that cost when we re-sort regardless.

crates/contract/src/event.rs
```diff
@@ -168,6 +194,141 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+            return Ok(logs);
+        }
+        self.get_logs_chunked().await
+    }
+
+    /// Divides the block range into chunks and queries them concurrently, falling back to
+    /// single-block queries for any chunk that fails.
+    async fn get_logs_chunked(&self) -> TransportResult<Vec<Log>> {
+        let (from, to) = self.filter.extract_block_range();
+        let (Some(from), Some(to)) = (from, to) else {
```
mattsse: this will hard-fail when users set `to_block(Latest)` or similar tags. the optimistic path in `query_raw` happily accepts those, but if it fails for a transient reason (timeout, rate limit), the fallback lands here and errors with "chunked queries require numeric from_block and to_block" — which is pretty confusing.

worth resolving tags to concrete block numbers via `provider.get_block_number()` before chunking.

crates/contract/src/event.rs
```diff
@@ -168,6 +194,141 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+        let all_results: Vec<TransportResult<(u64, Vec<Log>)>> =
+            futures::stream::iter(chunk_ranges)
+                .map(|(start_block, end_block)| {
+                    let chunk_filter =
+                        self.filter.clone().from_block(start_block).to_block(end_block);
+                    let provider = self.provider.clone();
+                    async move {
+                        match provider.get_logs(&chunk_filter).await {
+                            Ok(logs) => Ok((start_block, logs)),
+                            Err(_) => {
```
mattsse: might want to at least `tracing::debug!` the error here before falling back. right now if something unexpected goes wrong (auth issue, malformed request, etc.) the error gets silently swallowed and the fallback kicks in for no good reason — makes debugging pretty painful.

crates/contract/src/event.rs
```diff
@@ -168,6 +194,163 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+            },
+        };
+        let to = match to_block {
+            None => self.provider.get_block_number().await?,
+            Some(tag) => match tag.as_number() {
+                Some(n) => n,
+                None if tag == BlockNumberOrTag::Earliest => 0,
+                None => self.provider.get_block_number().await?,
+            },
+        };
```
mattsse: can we make this a private helper fn, then this fn is easier to read

crates/contract/src/event.rs
```diff
@@ -168,6 +194,163 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+/// A chunked event query builder.
+///
+/// Created via [`Event::chunked`]. Configure chunk size and concurrency, then call
+/// [`query`](ChunkedEvent::query) or [`query_raw`](ChunkedEvent::query_raw).
+///
+/// Attempts the full block range optimistically first. If that fails, the range is split into
+/// `chunk_size`-block windows queried concurrently (bounded by `max_concurrent`). If an
+/// individual chunk still fails, each block in that chunk is queried individually.
+#[must_use = "ChunkedEvent does nothing unless you call `query` or `query_raw`"]
+pub struct ChunkedEvent<P, E, N = Ethereum> {
```
mattsse, line 207: I think this type can also implement future, which then just does `Box::pin(self.query())`

we can also derive clone or impl clone

crates/contract/src/event.rs
```diff
@@ -168,6 +194,163 @@ impl<P: Clone, E, N> Event<&P, E, N> {
...
+                        match provider.get_logs(&chunk_filter).await {
+                            Ok(logs) => Ok((start_block, logs)),
+                            Err(err) => {
+                                tracing::debug!(
+                                    %err,
+                                    start_block,
+                                    end_block,
+                                    "chunk query failed, falling back to single-block queries"
+                                );
+                                // Fallback: query each block individually
```
mattsse: let's try to refactor this section to a helper fn, because then we can easily add stream support for this as well if we have 
```
fn chunk_stream() {
futures::stream::iter(chunk_ranges)
                .map(|(start_block, end_block)| {
....
}
```

PR #3749, opened 2026-02-26
chore(node-bindings): remove Clique consensus support

Mapogundal:
Clique was removed in geth >=1.14 and only the setter was deprecated, leaving is_clique(), clique_address() and the entire spawn logic as live dead code, this removes all of it along with the related NodeError variants and the ignored test.


mattsse (commented):
doesnt hurt to keep

PR #3750, opened 2026-02-26
fix: skip serializing None fields in Forking

Mapogundal:
`Forking` is used directly in `anvil_reset`, None fields were serialized as `null` instead of being omitted.


mattsse (commented):
can keep this as is

PR #3753, opened 2026-02-27
chore: ignore test_fill_transaction on Anvil

Ocheretovich:
Anvil does not implement eth_fillTransaction, so this test fails when run against Anvil. Mark it as ignored to avoid
spurious failures.


onbjerg (changes requested):
anvil does support eth_fillTransaction.

PR #3344, opened 2025-12-13
refactor(network): extract `NetworkTransactionBuilder` from `TransactionBuilder`

lean-apple:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/3331.

## Solution

Split `TransactionBuilder` into 2 traits, one new,  `NetworkTransactionBuilder`, with methods depending on network, the other with the other methods.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes -> Integration in [op-alloy](https://github.com/alloy-rs/op-alloy/pull/618), [reth]( https://github.com/paradigmxyz/reth/pull/20356).

PR #3757, opened 2026-02-27
refactor(network): extract dyn-compatible/network-based methods from `TransactionBuilder`

mablr:
## Motivation

Provide dyn-compatible transaction builder.

## Solution

- Rebased https://github.com/alloy-rs/alloy/pull/3344
- Introduced `DynTransactionBuilder` trait
- Smol trick: using concrete types in `TransactionBuilderDyn`, extend in Sized `TransactionBuilder`, e.g `TransactionBuilderDyn::set_input(&mut self, input: Bytes)` / `TransactionBuilder::set_input<T: Into<Bytes>>(&mut self, input: T)`

## PR Checklist

- [ ] Added Tests (no new functionality, all existing pass)
- [x] Added Documentation
- [x] Breaking changes

### DIFF
--- a/crates/network/src/transaction/builder.rs
+++ b/crates/network/src/transaction/builder.rs
@@ -69,7 +69,7 @@ impl<N: Network> TransactionBuilderError<N> {
 /// object-safe. For generic wrapper setters (e.g., `set_input<T: Into<Bytes>>`), use
 /// [`TransactionBuilder`].
 #[doc(alias = "TxBuilderDyn")]
-pub trait DynTransactionBuilder: Send + Sync {
+pub trait DynTransactionBuilder: Send + Sync + 'static {
     /// Get the chain ID for the transaction.
     fn chain_id(&self) -> Option<ChainId>;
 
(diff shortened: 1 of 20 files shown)


crates/network/src/transaction/builder.rs
```diff
@@ -57,28 +57,25 @@ impl<N: Network> TransactionBuilderError<N> {
...
-#[doc(alias = "TxBuilder")]
-pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'static {
+/// - **Getters** for all transaction fields (`chain_id()`, `nonce()`, `input()`, etc.)
+/// - **Setters** with concretized `Bytes` parameters (no generic type parameters)
+///
+/// All setter methods take concrete types rather than generics, enabling the trait to be
+/// object-safe. For generic wrapper setters (e.g., `set_input<T: Into<Bytes>>`), use
+/// [`TransactionBuilder`].
+#[doc(alias = "TxBuilderDyn")]
+pub trait TransactionBuilderDyn: Send + Sync {
```
figtracer: nice!

 i would only change to `DynTransactionBuilder` to keep consistency w the naming
https://github.com/alloy-rs/alloy/blob/5c5bc48cc2dedd29c94f762873e2879889e62cbc/crates/contract/src/call.rs#L28

mablr: okayy let's do that

crates/network/src/transaction/builder.rs
```diff
@@ -321,30 +396,6 @@ pub trait TransactionBuilder<N: Network>: Default + Sized + Send + Sync + 'stati
         self
     }
```
mablr, line 391: This is only method that require `NetworkTransactionBuilder` trait to be `Sized`.

mablr (commented):
~It would be ideal if we'd provide a minimal dyn-compatible e2e transaction builder trait~

crates/network/src/transaction/builder.rs
figtracer: the 4-way separation makes this very ugly
NetworkTransactionBuilder<N>: DynNetworkTransactionBuilder<N> + TransactionBuilder

do we really need another tier?

maybe we should add good docs for this too

mablr: Mb this is actually not possible, since `build` method takes `self` not `&self`, so this is`Sized`

PR #3758, opened 2026-02-27
fix(rpc-types-trace): use BTreeMap in MuxConfig/MuxFrame for deterministic serialization

vanityKoa:
MuxConfig and MuxFrame use HashMap with a random-seeded hasher (foldhash RandomState), which produces non-deterministic JSON key ordering across process runs.

Before (key order varies per run):
{"prestateTracer":{"diffMode":true},"4byteTracer":null,"callTracer":{"onlyTopCall":true,"withLog":true}}

After (keys always sorted):
{"4byteTracer":null,"callTracer":{"onlyTopCall":true,"withLog":true},"prestateTracer":{"diffMode":true}}

also fixes a latent flaky test (test_serialize_mux_tracer_config) that compares serialized output against a fixed string assuming alphabetical key order.


mattsse:
Closing this as not planned for now.

PR #3764, opened 2026-03-01
fix(node-bindings): implicitly enable IPC when setting ipc_path

neowangreal:
Setting `ipc_path()` without calling `enable_ipc()` results in both `--ipcdisable` and `--ipcpath` being passed to the node command, which is contradictory and silently ignores the user's IPC path. Now `ipc_path()` also sets `ipc_enabled = true` since there's no reason to set a path without wanting IPC. Affects both geth and reth.


DaniPopes (approved):

neowangreal:
CI failed seems not related?

PR #3769, opened 2026-03-03
refactor(rpc-types-eth): centralize SimulateError codes

Mapogundal:
Replace hardcoded SimulateError code values with named constants.


mattsse (approved):
fair enough

PR #3771, opened 2026-03-03
feat(provider): add default timeout for PendingTransactionConfig

### LINKED ISSUE #3104: [Feature] `PendingTransactionConfig` should have a default timeout
### Component

provider, pubsub

This is somewhat opinionated of course but having no timeout set can lead to the call hanging indefinitely which i think is not what most users would expect.

To me it seems more reasonable to set a high default timeout and make setting `timeout = None` an explicit user choice.

related code:
https://github.com/alloy-rs/alloy/blob/c1987556cdcd0e99a1ce9c6512a8eea800df34fa/crates/provider/src/heart.rs#L290

### Additional context

_No response_

TanayK07:
## Summary

- Adds `DEFAULT_PENDING_TX_TIMEOUT` constant (5 minutes) as the default timeout for `PendingTransactionConfig`
- Changes `PendingTransactionConfig::new()` to use `Some(DEFAULT_PENDING_TX_TIMEOUT)` instead of `None`
- Users can still disable the timeout with `.with_timeout(None)`

Previously, `PendingTransactionConfig` defaulted to `timeout: None`, meaning pending transaction watchers could hang indefinitely if a transaction was never confirmed. This sets a reasonable default while preserving the ability to opt out.

Closes #3104

## Test plan

- [x] `cargo build -p alloy-provider` passes
- [x] `cargo clippy -p alloy-provider` clean (no warnings)
- [x] `cargo fmt --check` passes
- [ ] CI tests


TanayK07:
hi @mattsse could you review this once when you get the chance

PR #3776, opened 2026-03-05
fix(provider): SendTransaction Failed on BSC due to Legacy TxType)

### LINKED ISSUE #3775: [Bug] SendTransaction Failed on BSC due to Legacy TxType
### Component

provider, pubsub

1.6.1

### Operating System

macOS (Apple Silicon)

### Describe the bug

In `provider/trait.rs`, function `estimate_eip1559_fees_with`, there is a condition to check if the latest block even has a base fee/supports EIP1559, it works for all evm chains which support eip1559 but fails on **Binance Smart Chain**.  For BSC, although it does include `baseFeePerGas` in block headers and its RPC likely succeeds on EIP-1559 fee estimation calls, but it doesn't actually process EIP-1559 transaction types. If you call RPC something like "eth_getBlockByNumber", it returns a header that includes `baseFeePerGas` with "0x0" instead of None. So the condition check might fail with BSC, a small fix that additionally checks whether `baseFeePerGas` is non-zero might work well.

bournezjc:
## Motivation

fix #3775 

## Solution

The fix adds `.filter(|&base_fee| base_fee != 0)`,  so that when BSC (or any chain) returns baseFeePerGas: "0x0" in the block header, it's treated the same as None — resulting in an UnsupportedFeature("eip1559") error instead of silently proceeding with a zero base fee.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -305,6 +305,7 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
                     .header()
                     .as_ref()
                     .base_fee_per_gas()
+                    .filter(|&base_fee| base_fee != 0)
                     .ok_or(RpcError::UnsupportedFeature("eip1559"))?
                     .into()
             }
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -2543,6 +2543,19 @@ mod tests {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


onbjerg:
Thanks for the contribution. Closing this because a zero base fee does not imply that a chain lacks EIP-1559 support. Applying that heuristic globally would misclassify legitimate zero-base-fee EIP-1559 chains. The underlying BSC compatibility issue should be addressed with a chain-specific solution or explicit configuration.

PR #3777, opened 2026-03-05
refactor(rpc-types-eth): add explicit SimulatePayload block builders

Shresth79:
Introduce add_block/add_blocks on SimulatePayload while keeping deprecated wrappers for compatibility, and add tests covering both builder paths.


mattsse (commented):
whats the point of this

PR #3787, opened 2026-03-08
feat(consensus): add `MaybeImpersonatedTransaction<T>`

stevencartavia:
Upstreams `MaybeImpersonatedTransaction` from foundry's `anvil-core` crate.
https://github.com/foundry-rs/foundry/blob/master/crates/anvil/core/src/eth/transaction/mod.rs#L22-L26

A generic wrapper around any transaction type `T` that optionally overrides the sender address.

- Foundry stops owning a concrete type and depends on a generic one from alloy
- Any simulation node built on alloy gets impersonation support for free with their own tx envelope type

### DIFF
--- a/crates/consensus/src/transaction/impersonated.rs
+++ b/crates/consensus/src/transaction/impersonated.rs
@@ -1,4 +1,5 @@
 use crate::crypto::RecoveryError;
+use alloc::vec::Vec;
 use alloy_eips::{eip2718::Encodable2718, Typed2718};
 use alloy_primitives::{keccak256, Address, B256};
 use alloy_rlp::{bytes::BufMut, Decodable, Encodable};


stevencartavia:
> this doesnt add a lot but afaik this is anvil only so unsure what use case we have for this in the consensus crate?

anvil-core currently defines this type hardcoded to `FoundryTxEnvelope`. If it lives here, that becomes `MaybeImpersonatedTransaction<N::TxEnvelope>` and `TypedBlockInfo<N: Network>` works for any chain out of the box.

mattsse:
that change we can do on foundry as well?
we dont need to have the type here

stevencartavia:
not needed 🫡

PR #3789, opened 2026-03-09
fix(pubsub): use unused id for eth_unsubscribe instead of hardcoded

neowangreal:
The hardcoded `Id::Number(1)` in `service_unsubscribe` can collide with an actual in-flight request, causing its response to be fulfilled with the unsubscribe result (a bool). This is a known issue (there was a TODO for it).

Fix: add `RequestManager::unused_id()` that hands out an id not present in the current in-flights map.


mattsse (commented):
not needed

PR #3794, opened 2026-03-11
fix(network): fail fast on unknown AnyNetwork signing and 2718 encoding

Shresth79:
Make unknown AnyNetwork signing and 2718 encoding fail consistently, and update the rough-edges docs to match the enforced panic contract.


Shresth79:
Hi @mattsse, if you have some time, please take a look at this PR. Thanks for your time!

PR #3795, opened 2026-03-11
fix(flashblocks): correct receipts type in FlashBlock metadata

0xSojol:
## Motivation

The current `FlashBlock` type defines the `receipts` field as:

    `BTreeMap<B256, BTreeMap<String, Receipt>>`

However, the actual FlashBlocks response from the Base FlashBlocks endpoint returns receipts in the following format:
```
    "receipts": {
        "0x...tx_hash": {
            "cumulativeGasUsed": "...",
            "logs": [...],
            "status": "0x1"
        }
    }
```
There is no transaction-type wrapper (e.g. `"Legacy"` or `"Eip1559"`) around the
`Receipt`. Because of this mismatch, deserialization fails with errors such as:

    invalid type: string "...", expected struct Receipt

when attempting to parse real FlashBlocks data.

## Solution

Update the `receipts` field type in `Metadata` to match the actual FlashBlocks schema:

    `BTreeMap<B256, Receipt>`

The FlashBlock roundtrip test JSON was also updated to remove the extra
`Legacy` / `Eip1559` wrapper fields so that it reflects the real response
structure returned by the FlashBlocks API.

This change fixes deserialization when consuming live FlashBlocks data.

### DIFF
--- a/crates/eips/src/eip4844/mod.rs
+++ b/crates/eips/src/eip4844/mod.rs
@@ -128,6 +128,7 @@ where
 
 #[cfg(all(not(debug_assertions), feature = "serde"))]
 #[inline(always)]
+/// Helper function to deserialize boxed blobs from a serde deserializer.
 pub fn deserialize_blobs<'de, D>(deserializer: D) -> Result<alloc::vec::Vec<Blob>, D::Error>
 where
     D: serde::de::Deserializer<'de>,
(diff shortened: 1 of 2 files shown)


mattsse (approved):
makes sense, although, unsure if this is even inuse

https://github.com/alloy-rs/op-alloy/blob/a93a08a021e80db8133f5f7fdccbb5639bc68d45/crates/rpc-types-engine/src/flashblock/metadata.rs#L20

PR #3818, opened 2026-03-18
fix(json-rpc): propagate key parse errors in PubSubItem deserializer

neowangreal:
The `visit_map` implementation used `while let Ok(Some(key))` to iterate map keys, which causes `Err` from `next_key()` to silently break the loop instead of propagating the error. This could lead to misleading downstream errors (e.g., "missing field `params`") instead of the actual parse error.

inconsistent with all `next_value()?` calls in the same function, and with the idiomatic serde visitor pattern used elsewhere in the repo.


mattsse (approved):

neowangreal:
Rebase @mattsse

onbjerg (approved):

PR #3825, opened 2026-03-22
fix(transport): clarify fallback misconfiguration errors

AaronAjose:
## Motivation
                                                                                                                                   
  This is a follow-up to #3806.                                                                                                    
                                                                                                                                   
  `FallbackService` now returns a clear error when no transports are available, but a different misconfiguration still surfaces through the same path: callers can construct `FallbackService::new(..., 0)` even though `FallbackLayer::with_active_transport_count` already requires a non-zero count.
                                                                                                                                   
  When `active_transport_count` is `0`, `top_transports()` truncates the configured transports to an empty list and the request path reports a misleading "no transports available" error. That makes the actual configuration issue harder to diagnose.         
                                                                                                                                   
  ## Solution                                                                                                                      
                                                                                                                                   
  Add a shared fallback configuration check before entering the parallel and sequential request paths.                             
                                                                                                                                   
  This change returns a dedicated error when `active_transport_count == 0`, keeps a separate error for an empty transport list, and
  adds regression tests for both misconfiguration cases.
                                                                                                                                   
  ## PR Checklist                                                                                                                  

  - [x] Added Tests                                                                                                                
  - [ ] Added Documentation                                                                                                        
  - [ ] Breaking changes

### DIFF
--- a/crates/transport/src/layers/fallback.rs
+++ b/crates/transport/src/layers/fallback.rs
@@ -20,6 +20,9 @@ const STABILITY_WEIGHT: f64 = 0.7;
 const LATENCY_WEIGHT: f64 = 0.3;
 const DEFAULT_SAMPLE_COUNT: usize = 10;
 const DEFAULT_ACTIVE_TRANSPORT_COUNT: usize = 3;
+const ZERO_ACTIVE_TRANSPORT_COUNT_ERROR: &str =
+    "Fallback service active transport count must be greater than zero";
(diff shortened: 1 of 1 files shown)


mattsse:
Closing this in favor of the current intended behavior. We previously merged #3806 to treat an empty candidate set as "No transports available for fallback service", and an earlier attempt to special-case `active_transport_count = 0` was explicitly rejected in #2960 with the note that setting it to 0 is valid. The remaining inconsistency between `FallbackLayer::with_active_transport_count` and `FallbackService::new` seems better handled as a separate API/design discussion if we want to revisit it.

PR #3835, opened 2026-03-25
feat: add MPP payment layer

stevencartavia:
Adds MPP 402 payment handling to alloy's HTTP and WS transports.

- `MppLayer<P>` + `MppService<S,P>` — tower middleware that intercepts 402 responses, pays, and retries
- `MppReqwestClient<P>` — reqwest transport with 402 interception
- `MppWsConnect<P>` — wraps `WsConnect`, handles 402 during WS upgrade


mattsse (commented):
@onbjerg do we want this here or should all of this live in the mpp crate?

imo having this here will be a pita to maintain because we kinda have a cyclical dep via tempo-alloy here

PR #3849, opened 2026-03-27
feat(alloy-provider): add methods for creating custom gas estimator for `GasFiller`

StackOverflowExcept1on:
## Motivation

Our case is that we want to send transactions in next 3 blocks from head block.

```rust
ProviderBuilder::default()
    .with_eip1559_estimator(Eip1559Estimator::new(move |base_fee_per_gas, rewards| {
        utils::eip1559_default_estimator(base_fee_per_gas, rewards)
            .scaled_by_pct(eip1559_fee_increase_percentage)
    }))
    // ...
```

@mattsse

## Solution

Same approach as in #3447 and #3489. I haven't implemented same for `prepare_legacy` yet, but I think current implementation is sufficient if you, for example, want to use more aggressive approach to gas calculation.

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [x] Breaking changes
  ```diff
  -.filler(GasFiller)
  +.filler(GasFiller::default())
  ```

### DIFF
--- a/crates/provider/src/fillers/gas.rs
+++ b/crates/provider/src/fillers/gas.rs
@@ -7,7 +7,7 @@ use std::{
 use crate::{
     fillers::{FillerControlFlow, TxFiller},
     provider::SendableTx,
-    utils::Eip1559Estimation,
+    utils::{Eip1559Estimation, Eip1559Estimator},
     Provider,
 };
 use alloy_eips::eip4844::BLOB_TX_MIN_BLOB_GASPRICE;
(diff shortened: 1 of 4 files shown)


crates/provider/src/fillers/gas.rs
```diff
@@ -70,8 +70,11 @@ pub enum GasFillable {
 /// # Ok(())
 /// # }
 /// ```
-#[derive(Clone, Copy, Debug, Default)]
-pub struct GasFiller;
```
mattsse, line 74: sigh, this type should have been marked with non_exhaustive...

mattsse (commented):
this is technically a breaking change
so i first need to check how problematic this is and maybe delay until 2.0

crates/provider/src/fillers/gas.rs
StackOverflowExcept1on: done in 5b0c9789a4d25ce5030a19c9f5346a64695f1512! we might also consider that we need `legacy_estimator` and `eip1559_estimator` here as well. perhaps we should just rename it to `eip1559_estimator`? wdyt?

StackOverflowExcept1on:
@mattsse also, what do you think about the consistency of imports?
```rust
use alloy::providers::{utils::Eip1559Estimator, fillers::BlobGasEstimator};
```

mattsse (approved):

PR #3880, opened 2026-04-11
feat: add rlp `Decodable`/`Encodable` to consensus receipt

lean-apple:
## Motivation

Er(a)e type use a slim receipt, we can try to use `alloy_consensus::Receipt`, and not `ReceiptwithEnvelope` in era test when doing roundtrip tests

## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/receipts.rs
+++ b/crates/consensus/src/receipt/receipts.rs
@@ -12,7 +12,7 @@ use alloy_rlp::{BufMut, Decodable, Encodable, Header};
 use core::fmt;
 
 /// Receipt containing result of transaction execution.
-#[derive(Clone, Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


crates/consensus/src/receipt/receipts.rs
klkvr, line 32: hmm do you mind linking which scheme uses such encoding? it doesn't include the txtype even, nor bloom

mattsse: I believe this is related to https://github.com/paradigmxyz/reth/pull/23459

lean-apple: We dont' need this anymore as we introduce a slim receipt directly in https://github.com/paradigmxyz/reth/pull/24718, see https://github.com/paradigmxyz/reth/pull/24718#discussion_r3331428278

lean-apple:
Not needed anymore

PR #3887, opened 2026-04-14
fix(node-bindings): drain stdout for Anvil and Reth when not kept

DivooOliver:
- Drain stdout for Anvil and Reth when not kept, preventing pipe buffer from blocking the node.
  - Clear drain buffer each iteration to avoid unbounded memory growth (also fixed in Geth).


onbjerg (changes requested):
Could you provide a reproducer for Anvil or Reth on current main? When stdout isn’t kept, the reader is dropped and the pipe is closed, so the claimed buffer-filling issue isn’t clear. The Geth buffer fix is already on main, and the new reader loops would spin after EOF. Closing for now; happy to revisit with a reproducer.

PR #3889, opened 2026-04-15
fix(provider): update `test_anvil_set_time` for corrected `evm_setTime`

stevencartavia:
Anvil's `evm_setTime` was fixed in foundry-rs/foundry#14292 to return the offset in seconds directly instead of wrapping it through `Duration::from_millis().as_secs()` (which divided by 1000).


zerosnacks (approved):

figtracer (approved):

zerosnacks:
Merging to unblock CI, @mattsse please revert if you disagree

PR #3890, opened 2026-04-15
fix(provider): default create_access_list to pending

Shresth79:
## Motivation

`StorageSlotFinder::find_slot` collects candidate slots with `eth_createAccessList` and then validates them with `eth_call`. The former defaulted to `latest` while the latter defaulted to `pending`, so plain providers could compare two different state contexts and miss the correct slot.

## Solution

Default `Provider::create_access_list` to the pending block so it matches `eth_call` while still allowing explicit `.latest()` / `.block_id(...)` overrides. Clarify the provider docs, add a focused unit test for the new default, and document that `StorageSlotFinder` follows the provider's default block context.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/provider/trait.rs
+++ b/crates/provider/src/provider/trait.rs
@@ -179,7 +179,7 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
     /// #    my_overrides: StateOverride
     /// # ) -> Result<(), Box<dyn std::error::Error>> {
     /// # let tx = alloy_rpc_types_eth::transaction::TransactionRequest::default();
-    /// // Execute a call on the latest block, with no state overrides
+    /// // Execute a call on the pending block, with no state overrides
     /// let output = provider.call(tx).await?;
     /// # Ok(())
     /// # }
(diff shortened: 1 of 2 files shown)


Shresth79:
The new create_access_list regression test passes locally; the remaining failure is test_anvil_set_time, which looks unrelated to this PR.

crates/provider/src/provider/trait.rs
```diff
@@ -241,11 +241,15 @@ pub trait Provider<N: Network = Ethereum>: Send + Sync {
...
+    ///
+    /// This function returns [`RpcWithBlock`] which can be used to execute the
+    /// request or set a [`BlockId`]. If no block ID is provided, the access
+    /// list will be generated on the pending block with the current state.
     fn create_access_list<'a>(
         &self,
         request: &'a N::TransactionRequest,
     ) -> RpcWithBlock<&'a N::TransactionRequest, AccessListResult> {
-        self.client().request("eth_createAccessList", request).into()
+        RpcWithBlock::new_rpc(self.client().request("eth_createAccessList", request)).pending()
```
mattsse, line 252: since this is for accesslist, maybe we should keep this unchanged, but I think pending is actually more correct ehre

Shresth79: @mattsse Thanks for the review! I was mainly trying to keep `create_access_list` in line with `call`, since
  `StorageSlotFinder` uses both and the differing defaults can cause context drift. If you’d rather keep
  the change narrower, I can limit it to the `StorageSlotFinder` call site.

PR #3898, opened 2026-04-16
chore(BAL): added from_block_slow_optional_bal

Rimeeeeee:


crates/rpc-types-engine/src/payload.rs
```diff
@@ -2153,6 +2153,34 @@ impl ExecutionPayload {
...
+    /// Converts [`alloy_consensus::Block`] to [`ExecutionPayload`] and also returns the
+    /// [`ExecutionPayloadSidecar`] extracted from the block along with optional block access list.
+    ///
+    /// This preserves the full RLP-encoded block access list for Amsterdam/V4 payloads.
+    ///
+    /// See also [`ExecutionPayloadV3::from_block_unchecked`].
+    /// See also [`ExecutionPayloadSidecar::from_block`].
+    ///
+    /// Note: This re-calculates the block hash.
+    pub fn from_block_slow_with_optional_bal<T, H>(
```
mattsse: I have a better idea for this that is perhaps more future proof, lets introduce a helper struct

```
#[non_exhaustive]
struct PayloadExtras {
  pub bal: Option<Bytes>
}
impl From<Option<..>> for PaylodExtras

```

and then we make this arg `impl Into<PayloadExtras>`

so if we eventually need to add any additonal fields down the line, we dont need another fn

mattsse (changes requested):

mattsse (approved):

PR #3899, opened 2026-04-16
fix(anvil): add quantity serialization to `MineOptions::blocks` field

DivooOliver:
The `blocks` field in `MineOptions::Options` was missing the hex serialization attribute, causing Anvil to ignore it. Added `#[serde(default, with = "alloy_serde::quantity::opt")]` to match the expected RPC format. Re-enabled `test_evm_mine_with_configuration`, `test_anvil_mine_detailed_with_configuration`, and rewrote `test_anvil_reset` to verify actual reset behavior.

### DIFF
--- a/crates/provider/src/ext/anvil.rs
+++ b/crates/provider/src/ext/anvil.rs
@@ -517,7 +517,6 @@ mod tests {
     use alloy_rpc_types_eth::TransactionRequest;
     use alloy_sol_types::{sol, SolCall};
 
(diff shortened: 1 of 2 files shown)


mattsse:
Closing this PR because the separate provider-side test re-enable does not seem necessary here; this looks symmetrical with the existing Anvil coverage.

DivooOliver:
> Closing this PR because the separate provider-side test re-enable does not seem necessary here; this looks symmetrical with the existing Anvil coverage.

oh fair enough, thanks for the hint @mattsse

PR #3900, opened 2026-04-17
fix(node-bindings): align Reth::default with Reth::new

Shresth79:
## Motivation

`Reth` currently derives `Default`, but `Reth::new()` sets a different set of defaults, including the default ports, discovery, and a randomized instance. That makes `Reth::default()` behave differently from the constructor used throughout the docs and local call sites.

## Solution

Implement `Default` manually for `Reth` by delegating to `Reth::new()`, and add a unit test that locks in the shared defaults.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/reth.rs
+++ b/crates/node-bindings/src/nodes/reth.rs
@@ -153,7 +153,7 @@ impl Drop for RethInstance {
 ///
 /// drop(reth); // this will kill the instance
 /// ```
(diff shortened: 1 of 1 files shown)


zerosnacks (approved):
Makes sense 👍

mattsse (approved):

PR #3905, opened 2026-04-17
fix(rpc-types-anvil): add quantity serialization to `MineOptions::blocks` field

DivooOliver:
The `blocks` field in `MineOptions::Options` was missing the hex quantity serde attribute that `timestamp` already has, so it was serialized as a raw integer instead of a hex quantity string, which is inconsistent with the JSON-RPC quantity convention used by `evm_mine` / `anvil_mine_detailed`.

Adds `#[serde(default, with = "alloy_serde::quantity::opt")]` to match `timestamp`, plus a test case verifying hex-encoded `blocks` deserialize correctly.


mattsse (commented):
we can keep this as is

PR #3910, opened 2026-04-20
fix(provider): skip batching eth_call with block overrides

Shresth79:
## Motivation

Batched `eth_call` requests currently only screen out block ids and state overrides.
When a caller supplies non-empty `block_overrides`, the batching path drops that
context and silently executes a different call shape than the direct RPC path.

## Solution

Reject batched `eth_call` requests that carry non-empty block overrides so they
fall back to the normal request path, and add a targeted regression test for the
compatibility gate.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
Looks correct. The batching guard now rejects non-empty block overrides so these calls fall back to the direct eth_call path, and the added regression test covers the gate.

PR #3912, opened 2026-04-22
fix(node-bindings): avoid panic if graceful kill fails

Elena343-ai:
## Motivation

`GracefulShutdown::shutdown` currently panics if the final `kill()` call fails, and that helper is used from the `Drop` paths for the node instances. A kill failure during drop should stay best-effort so cleanup does not escalate into a process abort while unwinding.

## Solution

Return early when the child has already exited, and treat the final `kill()` as best-effort by logging the error instead of panicking. Add a focused unit test for the already-exited path and keep the existing force-kill reap coverage.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg (changes requested):
Avoiding the panic makes sense, but we still call `wait()` if `kill()` fails. If the child is still running, shutdown can hang indefinitely. Could you handle that case without an unconditional wait and add a test for it?

PR #3915, opened 2026-04-23
refactor(transport-ws): make TLS opts selectable through features

### LINKED ISSUE #3914: [Feature] feature flags for chosing TLS provider
### Component

transports

As far as I can see #3712 caused `aws-lc-rs` to always be in the dependency graph. This dependency is unfortunately extremely slow to compile compared to the alternative `ring` or `native-tls`.
It would be great if a feature flag could be introduced that gives the user of the libraries finer control over which TLS provider gets used for HTTPS support.

### Additional context

_No response_

lean-apple:
## Motivation

Closes https://github.com/alloy-rs/alloy/issues/3914

## Solution

  - `transport-ws`: split the hardcoded `aws-lc-rs` rustls provider into selectable backends (`aws-lc-rs` default, `ring`, `native-tls`), only install a default crypto provider if none is set.
  - Bubble `ws-ring` / `ws-native-tls` variants up through `rpc-client` → `provider` → `alloy`; `ws` keeps `aws-lc-rs` as the default for backwards compatibility

## PR Checklist

- [ ] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (approved):
Reviewed the feature cascade changes across transport-ws, rpc-client, provider, and alloy. This looks correct; the feature propagation and public cfg surface line up with the new TLS backend variants.

PR #3920, opened 2026-04-24
fix(node-bindings): use `create_dir_all` for datadir creation

DivooOliver:
Switched from `create_dir` to `create_dir_all` in geth.rs and reth.rs. Users can now specify nested paths like `/tmp/parent/child/data` without hitting errors when parent directories don't exist.


mattsse (approved):
Reviewed the diff in crates/node-bindings; create_dir_all is the right behavior for nested datadir paths. Local cargo check -p alloy-node-bindings passes.

PR #3925, opened 2026-04-27
perf(serde): avoid heap allocation in JsonStorageKey Display

DivooOliver:
Replace `alloc::format!("{num:#x}").fmt(f)` with `write!(f, "{num:#x}")` in `JsonStorageKey::fmt`. The old code created a temporary `String` on every call, wasting allocations on hot paths like `eth_getStorageAt`. Same pattern fixed in cdcc69c4 for `BlockNumberOrTag` and `Index`.


mattsse (approved):
Reviewed locally: no issues found. The change preserves JsonStorageKey number formatting while avoiding the temporary allocation.

PR #3926, opened 2026-04-27
perf(txpool): use `split_once` instead of `collect::<Vec>` in deserialization

DivooOliver:
Replace 4 temporary `Vec<&str>` allocations with `split_once()` in `TxpoolInspectSummary` deserialization. Each call to `value.split(...).collect()` followed by index access can be replaced with a single `split_once()` that returns the two parts directly without heap allocation.


mattsse (approved):
Looks good after formatting.

PR #3927, opened 2026-04-28
feat: add `alloy-transport-mpp`

stevencartavia:
Adds MPP WebSocket transport as a new `PubSubConnect` implementation. Drop-in replacement for `WsConnect` that handles MPP challenge/credential/receipt handshakes and wraps JSON-RPC frames in MPP message envelopes.

### DIFF
--- a/crates/pubsub/src/service.rs
+++ b/crates/pubsub/src/service.rs
@@ -4,7 +4,7 @@ use crate::{
     managers::{InFlight, RequestManager, SubscriptionManager},
     PubSubConnect, PubSubFrontend, RawSubscription,
 };
-use alloy_json_rpc::{Id, PubSubItem, Request, Response, ResponsePayload, SubId};
+use alloy_json_rpc::{Id, PubSubItem, Request, Response, ResponsePayload, RpcError, SubId};
 use alloy_primitives::B256;
 use alloy_transport::{
     utils::{to_json_raw_value, Spawnable},
@@ -202,6 +202,10 @@ impl<T: PubSubConnect> PubSubService<T> {
             match self.reconnect().await {
                 Ok(()) => break Ok(()),
                 Err(e) => {
+                    if matches!(&e, RpcError::Transport(k) if k.is_non_retryable()) {
+                        error!("Reconnect aborted (non-retryable), shutting down: {e}");
+                        break Err(e);
+                    }
                     retry_count += 1;
                     if retry_count >= max_retries {
                         error!("Reconnect failed after {max_retries} attempts, shutting down: {e}");
@@ -320,6 +324,25 @@ mod tests {
         }
     }
 
+    /// Returns a non-retryable error and counts `try_reconnect` calls.
+    #[derive(Clone, Debug, Default)]
+    struct NonRetryableConnect(Arc<std::sync::atomic::AtomicUsize>);
+
+    impl PubSubConnect for NonRetryableConnect {
+        fn is_local(&self) -> bool {
+            true
+        }
+
+        async fn connect(&self) -> TransportResult<ConnectionHandle> {
+            Err(TransportErrorKind::non_retryable_str("non-retryable test failure"))
+        }
+
+        async fn try_reconnect(&self) -> TransportResult<ConnectionHandle> {
+            self.0.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+            Err(TransportErrorKind::non_retryable_str("non-retryable test failure"))
+        }
+    }
+
     #[test]
     fn reconnect_retry_interval_uses_capped_exponential_backoff() {
         let base = Duration::from_secs(1);
@@ -385,5 +408,37 @@ mod tests {
                 .expect("request should be dispatched after reconnect")
                 .expect("new backend should receive the request");
         assert_eq!(dispatched.get(), expected);
+    }
+
+    #[tokio::test]
+    async fn non_retryable_reconnect_error_short_circuits_retry_loop() {
+        let (dead_handle, dead_interface) = ConnectionHandle::new();
+        let ConnectionInterface { from_frontend, to_frontend, error, shutdown } = dead_interface;
+        drop(from_frontend);
+        let _keep_dead_backend_alive = (to_frontend, error, shutdown);
+
+        let connector = NonRetryableConnect::default();
+        let counter = connector.0.clone();
+        let (tx, reqs) = mpsc::unbounded_channel();
+        let service = PubSubService {
+            handle: dead_handle,
+            connector,
+            reqs,
+            subs: SubscriptionManager::default(),
+            in_flights: RequestManager::default(),
+        };
+        service.spawn();
+
+        let req = Request::new("eth_blockNumber", Id::Number(1), ()).serialize().unwrap();
+        let (in_flight, rx) = InFlight::new(req, 16);
+        tx.send(PubSubInstruction::Request(in_flight)).unwrap();
+
+        timeout(Duration::from_secs(1), rx)
+            .await
+            .expect("non-retryable reconnect should resolve promptly")
+            .expect_err("request should fail when backend is gone and reconnect aborts");
+
+        // Exactly one attempt, not `max_retries`.
+        assert_eq!(counter.load(std::sync::atomic::Ordering::SeqCst), 1);
     }
 }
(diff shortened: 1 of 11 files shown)


gakonst:
We should make this a default probably for both HTTP and WS? So that it just works?

stevencartavia:
I think that for WS we could add a handshake timeout so if the server never sends a challenge, it falls through to regular WS. And for HTTP wrap the transport to handle 402s with a payment credential and retry

gakonst:
Yep - and some automagical discovery for keys that ties nicely in how we store things in foundry / tempo wallet?

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,781 @@
...
+            interface,
+            self.payment_provider.clone(),
+            self.voucher_provider.clone(),
+            self.keepalive_interval,
+            self.receipt_tx.clone(),
+            self.events_tx.clone(),
+        )
+        .spawn_task();
+
+        Ok(handle.with_max_retries(self.max_retries).with_retry_interval(self.retry_interval))
```
grandizzy, line 361: is it OK to use same retry for both network and MPP failures?

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,781 @@
...
+///
+/// `MppWsConnect` is a drop-in [`PubSubConnect`] that speaks the
+/// [MPP WS wire protocol](https://github.com/tempoxyz/mpp-rs). Outbound
+/// JSON-RPC frames are wrapped as `{"type":"message","data":<payload>}`;
+/// server-issued `challenge` / `needVoucher` frames are settled by the
+/// supplied [`PaymentProvider`] / [`VoucherProvider`].
+///
+/// On reconnect, [`PaymentProvider::pay`] is called again with the server's
+/// new challenge — providers should handle repeated calls cheaply.
+#[derive(Clone, Debug)]
```
grandizzy: nit - debug can leak user sensitive info, let's impl it with redacting sensitive fields

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,781 @@
...
+    // Off-loop tasks: spawning `pay`/`next_voucher` keeps the keepalive arm
+    // responsive while signing/broadcasting takes place.
+    let mut pending_pay: Option<JoinHandle<Result<PaymentCredential, MppError>>> = None;
+    let mut pending_voucher: Option<JoinHandle<Result<PaymentCredential, MppError>>> = None;
+
+    loop {
+        let payment_in_flight = pending_pay.is_some() || pending_voucher.is_some();
+        let frontend_open = handshake_complete && !payment_in_flight;
+
+        tokio::select! {
```
grandizzy, line 408: I think we need arm for handshake timeout here to avoid hanging the client

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,781 @@
...
+                    return Err(());
+                }
+            };
+            debug!(method = %parsed.method, intent = %parsed.intent, "MPP challenge received");
+            let _ = events_tx.send(MppEvent::Challenge(parsed.clone()));
+
+            // Spawn `pay()` so the main loop stays responsive (keepalive,
+            // socket reads) while the provider signs/broadcasts.
+            let provider = payment_provider.clone();
+            *pending_pay = Some(tokio::spawn(async move { provider.pay(&parsed).await }));
```
grandizzy, line 731: should we check if provider supports given method and pay before spawn here?

grandizzy (commented):
looks good overall, ty! left couple of comments, please check

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,952 @@
...
+/// `MppWsConnect` is a drop-in [`PubSubConnect`] that speaks the
+/// [MPP WS wire protocol](https://github.com/tempoxyz/mpp-rs). Outbound
+/// JSON-RPC frames are wrapped as `{"type":"message","data":<payload>}`;
+/// server-issued `challenge` / `needVoucher` frames are settled by the
+/// supplied [`PaymentProvider`] / [`VoucherProvider`].
+///
+/// On reconnect, [`PaymentProvider::pay`] is called again with the server's
+/// new challenge — providers should handle repeated calls cheaply.
+///
+/// Only socket-level failures are retried; deterministic MPP failures are
```
grandizzy, line 157: is this statement true or will it use max retries regardless MPP failures?

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,952 @@
...
+        let payment_in_flight = pending_pay.is_some() || pending_voucher.is_some();
+        let frontend_open = handshake_complete && !payment_in_flight;
+
+        tokio::select! {
+            biased;
+
+            // 1. Outbound JSON-RPC from the alloy frontend → MPP `message` frame.
+            //    Held while a credential exchange is in flight, so RPC never
+            //    races a challenge.
+            inst = interface.recv_from_frontend(), if frontend_open => {
```
grandizzy: what happens in case of payment in progress and shutdown initiated meanwhile? will pay still continue, should we somehow signal this and break?

crates/transport-mpp/src/ws.rs
```diff
@@ -0,0 +1,952 @@
...
+async fn send_credential(
+    socket: &mut WsStream,
+    cred: &PaymentCredential,
+    events_tx: &broadcast::Sender<MppEvent>,
+    is_voucher: bool,
+) -> Result<(), ()> {
+    let auth = match format_authorization(cred) {
+        Ok(s) => s,
+        Err(err) => {
+            error!(?err, "failed to format MPP credential");
```
grandizzy, line 615: probably we should b detailed here and err with a reason that would result in a `TerminationReason::Fatal` instead transient and retry?

grandizzy (commented):
couple of more questions, pls check. ty!

crates/transport-mpp/src/ws.rs
stevencartavia: added `TransportErrorKind::NonRetryable` and made `connect()` return it on fatal MPP failures. The pubsub reconnect loop now short-circuits on non-retryable errors instead of consuming the full retry budget.

crates/transport-mpp/src/ws.rs
stevencartavia: `recv_from_frontend` is now polled unconditionally and RPCs are buffered into a `VecDeque`, so a shutdown is observed immediately even while `pay()` is in flight, and the pending task is aborted via its `JoinHandle`.

crates/transport-mpp/src/ws.rs
stevencartavia: `send_credential` now returns `TerminationReason` directly: `format_authorization` failures map to `Fatal` (deterministic, won't retry), while socket send errors stay `Transient` since a reconnect gets a fresh challenge.

grandizzy (approved):
lgtm!

crates/transport-mpp/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+url.workspace = true
+
+# non-WASM only
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
+http.workspace = true
+tokio = { workspace = true, features = ["sync", "rt", "time", "macros"] }
+tokio-tungstenite = { workspace = true }
+rustls = { version = "0.23", default-features = false, optional = true }
+
+mpp = { git = "https://github.com/tempoxyz/mpp-rs", rev = "1bdea74189ccc9502a170e788789891b001abc63", default-features = false, features = ["ws"] }
```
mattsse, line 48: we cant do this, then we cannot publish this

crates/rpc-types-trace/src/geth/mux.rs
mattsse, line 1: should remain unchanged

crates/pubsub/src/service.rs
mattsse, line 1: why do we need these changes for this feature?

mattsse (changes requested):
why do we want this in here and not in mpp repo

this makes it tricky to maintain because now we have a cyclic repo dep:

https://github.com/tempoxyz/mpp-rs/blob/11fe0dd25cccd067d1e260b92307e96db0930985/Cargo.toml#L82-L85

can we not add this as crate there?

stevencartavia:
added in https://github.com/tempoxyz/mpp-rs/pull/243

PR #3929, opened 2026-04-28
perf(json-rpc): preallocate capacity in ResponsePacket deserialization

DivooOliver:
Use `Vec::with_capacity(seq.size_hint().unwrap_or(0))` instead of `Vec::new()` when deserializing batch responses. For large batches (100+ responses), this avoids 6-7 reallocation cycles during the while-push loop.

### DIFF
--- a/crates/json-rpc/src/packet.rs
+++ b/crates/json-rpc/src/packet.rs
@@ -212,7 +212,7 @@ where
             where
                 A: SeqAccess<'de>,
             {
-                let mut responses = Vec::new();
+                let mut responses = Vec::with_capacity(seq.size_hint().unwrap_or(0));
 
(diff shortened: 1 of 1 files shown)


mattsse:
Thanks for the PR. I do not think we should take this as-is: the normal Alloy HTTP response path deserializes with `serde_json::from_slice`, whose streaming `SeqAccess` does not provide a size hint, so this does not appear to improve the hot path described here. For deserializers that do provide a hint, reserving it directly is less defensive than the standard serde `Vec<T>` deserializer, which caps preallocation.

Closing for now, but happy to reconsider with a benchmark showing an actual win on Alloy response deserialization path and an implementation that avoids unbounded preallocation.

PR #3931, opened 2026-04-28
fix(rpc-types-trace): default missing/null CallOutput.output to empty bytes

### LINKED ISSUE #3930: [Bug] CallOutput fails to deserialize when output field is missing or null
### Component

rpc

`2.0.0` (also reproduces on `main`)

### Operating System

macOS (Apple Silicon)

### Describe the bug

`alloy_rpc_types_trace::parity::CallOutput.output` is a required `Bytes`, so deserialization fails when the field is missing or `null`. This breaks `trace_filter` parsing for chains that return `suicide`/SELFDESTRUCT traces with only `gasUsed` in `result` (seen on Sonic).

Example trace (real, Sonic block 66762070):

```json
{
  "result": { "gasUsed": "0x0" },
  "type": "suicide",
  "...": "..."
}
```

Error:

```
data did not match any variant of untagged enum TraceOutput
```

Minimal repro:

```rust
let json = r#"{ "gasUsed": "0x0" }"#;
let _: alloy_rpc_types_trace::parity::CallOutput =
    serde_json::from_str(json).unwrap();
```

The same bug existed for `TraceResults.output` and was fixed in #1102. The same one-liner fixes `CallOutput`:

```rust
pub struct CallOutput {
    pub gas_used: u64,
    #[serde(default, deserialize_with = "alloy_serde::null_as_default")]
    pub output: Bytes,
}
```

Found via graphprotocol/graph-node#6489.

incrypto32:
## Motivation

`CallOutput.output` is a required `Bytes`, so deserialization fails when the field is missing or `null`. This breaks `trace_filter` parsing for chains that return `suicide`/SELFDESTRUCT traces with only `gasUsed` in `result` (seen on Sonic).

Example trace (real, Sonic block 66762070):

```json
{
  "result": { "gasUsed": "0x0" },
  "type": "suicide",
  "...": "..."
}
```

Error:

```
data did not match any variant of untagged enum TraceOutput
```

The same bug existed for `TraceResults.output` and was fixed in #1102. The same fix was never applied to `CallOutput.output`.

Closes #3930. Found via graphprotocol/graph-node#6489.

## Solution

Mirror the #1102 fix on `CallOutput.output` by adding `#[serde(default, deserialize_with = "alloy_serde::null_as_default")]`. Handles both missing field and `"output": null`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-trace/src/parity.rs
+++ b/crates/rpc-types-trace/src/parity.rs
@@ -464,13 +464,16 @@ pub struct CreateOutput {
 }
 
 /// Represents the output of a trace.
+///
+/// Variant order is significant: `Create` is tried first because `CallOutput`
+/// only requires `gasUsed`, so a create result would otherwise match `Call`.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(untagged)]
 pub enum TraceOutput {
-    /// Output of a regular call transaction.
-    Call(CallOutput),
     /// Output of a CREATE transaction.
     Create(CreateOutput),
+    /// Output of a regular call transaction.
+    Call(CallOutput),
 }
 
 impl TraceOutput {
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
This change fixes the missing/null call output case, but it also introduces a regression in `TraceOutput` deserialization because the untagged enum now matches `CallOutput` too broadly. Please disambiguate the call/create output variants before merging.

`CallOutput.output` now defaults, leaving `CallOutput` with only `gasUsed` as a required field. Since `TraceOutput` is an untagged enum and tries `CallOutput` before `CreateOutput`, valid create results like `{ "address": ..., "code": ..., "gasUsed": ... }` now deserialize as `TraceOutput::Call`, dropping the created contract address/code and changing roundtrip serialization. Reordering `Create` before `Call`, or otherwise explicitly disambiguating these result shapes, should avoid the regression while still accepting missing/null call output.

mattsse (approved):
The ambiguity from my earlier review is addressed by trying Create before Call, with regression coverage. Focused alloy-rpc-types-trace tests pass locally.

PR #3938, opened 2026-04-29
fix(rpc-client): validate response ids for single requests

Elena343-ai:
## Motivation

The single-request path currently accepts the first `ResponsePacket::Single`
payload it receives and drops the response id before deserialization. That means
the client never verifies that the response it is about to decode actually
matches the request id it sent.

## Solution

Carry the expected request id through `RpcCall`, reject mismatched single
responses before stripping the payload id, and add a focused regression test for
the mismatch case.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


Elena343-ai:
This looks like flaky CI.

PR #3941, opened 2026-04-30
fix(pubsub): reject zero channel sizes

Elena343-ai:
## Motivation

A zero subscription channel size can panic later when the pubsub or poller path creates a Tokio broadcast channel.

## Solution

Validate configured channel sizes before they reach broadcast channel creation, and document the panic behavior on the public setters.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes


mattsse (commented):
we can keep as is

PR #3942, opened 2026-04-30
fix(ci): stabilize main red tests

decofe:
## Summary
- switch the flaky ENS public resolver tests to `https://ethereum.reth.rs/rpc`
- stop exposing reqwest-backed HTTP constructors on `wasm32-wasip1`
- make reqwest a non-`wasm32-wasip1` dependency in the HTTP/provider/rpc-client crates so the WASI build no longer pulls unsupported Tokio and aws-lc paths

## Why
`main` is currently red in two places:
- `all-features` jobs fail because two ENS tests still hit the plain-HTTP `ethereum.reth.rs` endpoint, which now returns `520`
- `wasm-wasi` fails while building the `alloy` package because default reqwest/rustls paths on `wasm32-wasip1` pull unsupported Tokio and `aws-lc-sys` codepaths

## Validation
- `cargo fmt --check`
- `git diff --check`
- targeted cargo tests were blocked in this sandbox by crates.io `429 Rate limit exceeded` responses while resolving dependencies

### DIFF
--- a/crates/ens/src/lib.rs
+++ b/crates/ens/src/lib.rs
@@ -459,7 +459,7 @@ mod tests {
     #[tokio::test]
     async fn test_pub_resolver_text() {
         let provider =
-            ProviderBuilder::new().connect_http("http://ethereum.reth.rs/rpc".parse().unwrap());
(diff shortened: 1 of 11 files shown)


mattsse (approved):
ty!

PR #3944, opened 2026-04-30
fix(provider): clean up failed impersonated sends

nivanovvv:
## Motivation

`anvil_send_impersonated_transaction_with_config` starts impersonating before it sends the transaction. If the send fails, the configured default is still to stop impersonating, but the previous control flow returned before cleanup.

## Solution

Attempt the stop-impersonation request after a failed send when `stop_impersonate` is enabled, while preserving the original send error. A mock transport regression test verifies that the cleanup request is issued.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/provider/src/ext/anvil.rs
+++ b/crates/provider/src/ext/anvil.rs
@@ -455,13 +455,16 @@ where
             impersonate_future.await?;
         }
 
-        let tx_hash = self.anvil_send_impersonated_transaction(request).await?;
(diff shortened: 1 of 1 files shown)


mattsse (approved):
LGTM. Reviewed head 212ca9948b01fc8b2cdb1bc4a8ffc53d1b90c61e; targeted regression test passed locally.

PR #3953, opened 2026-05-04
fix: deduplicate fork activation checks in genesis config

anim001k:
`ChainConfig` used separate private helpers for block-based and timestamp-based fork activation checks, but both implemented the same comparison logic. Keeping two copies makes the shared semantics less obvious.


mattsse (commented):
Thanks for the cleanup. I agree this is behavior-preserving, but I don't think we need to take this change: the existing block/timestamp helpers keep the fork activation semantics explicit at the call sites, and the deduplication here does not materially improve the API or maintenance burden. Closing this out.

PR #3954, opened 2026-05-04
perf(eips): optimize c-kzg byte conversions

DaniPopes:
ideally the bindings would let you pass outptr so we can write into the vec directly, but for now we just get a box. also refactors so we remove the scope of unsafe transmutes down to a single trait.

### DIFF
--- a/crates/eips/src/eip4844/mod.rs
+++ b/crates/eips/src/eip4844/mod.rs
@@ -233,20 +233,58 @@ impl alloy_rlp::Decodable for HeapBlob {
 /// A commitment/proof serialized as 0x-prefixed hex string
 pub type Bytes48 = FixedBytes<48>;
 
-/// Conversion helpers for c-kzg blobs.
+/// Conversion helpers for c-kzg byte wrappers.
 #[cfg(feature = "kzg")]
-pub trait BlobCkzgExt {
-    /// Returns this blob as a c-kzg blob.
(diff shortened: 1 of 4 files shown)


crates/eips/src/eip4844/mod.rs
```diff
@@ -233,6 +233,213 @@ impl alloy_rlp::Decodable for HeapBlob {
...
+            fn vec_as_ckzg(vec: alloc::vec::Vec<Self>) -> alloc::vec::Vec<Self::Ckzg> {
+                // SAFETY: See `AsCkzg::as_ckzg`.
+                unsafe { core::mem::transmute(vec) }
+            }
+
+            #[inline]
+            fn vec_from_ckzg(vec: alloc::vec::Vec<Self::Ckzg>) -> alloc::vec::Vec<Self> {
+                // SAFETY: See `AsCkzg::as_ckzg`.
+                unsafe { core::mem::transmute(vec) }
+            }
```
mattsse, line 342: is this valid?

iirc there's some nuance wrt transmuting vecs, hence we have

https://github.com/alloy-rs/alloy/blob/96fa83e44e03ed99c1c9da0ff2a03845c73bdbea/crates/eips/src/eip4844/sidecar.rs#L257-L261

DaniPopes: yes

mattsse (approved):

PR #3956, opened 2026-05-06
fix(transport): make retry queue count cancel-safe

nivanovvv:
## Motivation

`RetryBackoffService` increments its queued-request counter when a request future starts, but manual decrement paths only run when the future reaches an explicit return branch. Dropping the future mid-flight can leave the counter inflated.

## Solution

Replace manual decrements with a small drop guard so completion, error returns, and cancellation all release the queued-request count. Add a focused regression test that polls and drops a pending retry future.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
Looks good. The queued-request guard makes the retry accounting cancellation-safe, and the follow-up documents the intent.

PR #3958, opened 2026-05-06
fix(signer-trezor): dispatch EIP-1559 by tx type

figtracer:
## Problem

`alloy-signer-trezor` selects the EIP-1559 Trezor signing API by doing a concrete downcast:

```rust
(tx as &dyn std::any::Any).downcast_ref::<TxEip1559>()
```

This succeeds only for the concrete `alloy_consensus::TxEip1559` value. Any wrapper that implements `SignableTransaction<Signature>` around a `TxEip1559` (e.g. `TypedTransaction::Eip1559` or network-specific signable wrappers used by tools like Foundry) fails the downcast and silently falls through to the legacy signing path — even though `tx.ty() == 0x02` and `tx.encoded_for_signing()` is a valid EIP-1559 type-2 preimage.

When that happens, Trezor signs the EIP-155 legacy preimage. The resulting r/s/v is then attached to a type-`0x02` envelope, so the decoded transaction recovers to a different signer than the address shown on the device.

The Ledger signer is unaffected because it signs `encoded_for_signing()` directly and never branches on the concrete transaction type.

## Fix

Dispatch on `tx.is_eip1559()` (i.e. the EIP-2718 type byte) and pull all EIP-1559-specific fields (max fees, access list) through the `Transaction` trait methods rather than the concrete struct fields. This keeps:

- legacy signing for legacy transactions,
- EIP-1559 signing for both concrete `TxEip1559` and any `SignableTransaction` wrapper around it.

### DIFF
--- a/crates/signer-trezor/src/signer.rs
+++ b/crates/signer-trezor/src/signer.rs
@@ -1,5 +1,5 @@
 use super::types::{DerivationType, TrezorError};
-use alloy_consensus::{SignableTransaction, TxEip1559};
+use alloy_consensus::SignableTransaction;
 use alloy_primitives::{
     hex, normalize_v, Address, ChainId, Signature, SignatureError, TxKind, B256, U256,
 };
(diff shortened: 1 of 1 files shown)


mattsse (approved):
Reviewed the Trezor signer dispatch change at head `11a6dce6cd3053fe807e41179c6332ac2caf4a13`.

The fix correctly routes EIP-1559 transactions by EIP-2718 type instead of concrete `TxEip1559` downcasting, which covers wrapper types without changing the legacy path. I also checked the Trezor client surface: Alloy currently only has legacy and EIP-1559 signing calls wired here, so the scope matches the available signer API.

PR #3960, opened 2026-05-07
fix(signer-ledger): reject invalid derivation paths

Elena343-ai:
## Motivation

`HDPath::Other` accepts caller-provided derivation paths. Invalid path segments currently panic in `LedgerSigner::path_to_bytes` because the `u32` parse uses `unwrap`, which bypasses the crate's normal `Result`-based error handling.

## Solution

Propagate derivation path parse failures as `LedgerError::InvalidDerivationPath` and thread the result through Ledger address and signing helpers. Add a unit test that verifies malformed custom paths return an error instead of panicking.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/signer-ledger/src/signer.rs
+++ b/crates/signer-ledger/src/signer.rs
@@ -96,7 +96,8 @@ impl Signer for LedgerSigner {
 
     #[inline]
     async fn sign_message(&self, message: &[u8]) -> Result<Signature> {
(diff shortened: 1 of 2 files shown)


mattsse (approved):
Reviewed the derivation path parsing change. The panic path is removed, direct call sites now propagate the error, and the focused test passes. I do not have blocking findings.

PR #3961, opened 2026-05-07
fix(eips): avoid fake exponential overflow

mattsse:
Supersedes #3470 with a smaller fix for `fake_exponential` overflow on large blob fee inputs.

The previous implementation stopped the Taylor expansion when `numerator_accum * numerator` overflowed `u128`, which could return a truncated invalid fee even when the final value still fit in `u128`. This keeps the public return types unchanged and uses the already available `alloy_primitives::ruint::aliases::U512` internally so the intermediate series math can complete correctly without adding `num-bigint` or `num-traits` as direct dependencies.

If the mathematical result is outside the representable `u128` range, `fake_exponential` now saturates at `u128::MAX` instead of unwinding. The only retained panic is the existing denominator-zero panic.

The Osaka excess blob gas comparison now widens both sides before multiplying, since a valid `u128` blob fee can still overflow when multiplied by `DATA_GAS_PER_BLOB`.

`calc_excess_blob_gas` remains `const fn`; the fee helpers that require widened arithmetic run at runtime.


satyakwok (commented):
Reviewed this carefully. Posting full findings + suggestions in case any of it helps unblock review.

## Math correctness

Old `fake_exponential` had three u128 overflow sites:
1. `factor * denominator` on init
2. `output += numerator_accum`
3. `denominator * i` in the accumulator update

The `checked_mul(numerator) ... break` in the old loop is the actual bug — it exits early *even when the final result still fits in u128*, returning a truncated value. The `(1, 88, 1)` and pectra `n=299M` test cases added in this PR are direct repros:

- `(1, 88, 1)`: old returns `25661480018157551077491514070807592801`, correct is `165162653699637111792770913913821835905`
- `(1, 299_453_931, 5_007_716)`: old returns `905999095046699812947807`, correct is `93359993185840258978230108`

Both old values fit comfortably in u128, so this isn't a saturation issue — it's truncation from premature loop exit when `checked_mul` overflows mid-Taylor-expansion.

Widening to U512 cleanly removes all three overflow sites. `factor * denominator ≤ 2^256` worst case leaves ample headroom. When the true mathematical result exceeds u128, the final `(output / denominator).try_into().unwrap_or(u128::MAX)` saturates correctly. For divergent inputs where U512 itself would overflow, the `checked_add` / `checked_mul` short-circuits return `u128::MAX` — verified empirically to never produce spurious saturation for inputs where the spec result fits u128.

`next_accum / denominator / i` is equivalent to `next_accum / (denominator * i)` for positive integers (floor-div identity), so no behavioral change in division order.

The Osaka comparison widening (`eip7840.rs`) is also necessary — `DATA_GAS_PER_BLOB * calc_blob_fee()` can exceed u128 even when `calc_blob_fee` itself fits within u128 (`131072 * u128::MAX ≈ 2^145`), so widening both sides of the inequality to U512 is the right fix.

## Independent verification

- **Spec match**: ported the EIP-4844 Python reference implementation, verified all 19 explicit test vectors in this PR match.
- **Rust ↔ Python cross-check (140 random cases)**: generated `(factor, num, denom, expected)` tuples from the Python reference, added a temporary test module on a local copy of this branch calling `fake_exponential` on each — all 140 match. Inputs covered realistic Cancun (`denom=3338477`), Pectra (`denom=5007716`), zero edges (`factor=0`, `numerator=0`), and huge-numerator saturation cases.
- **Adversarial (38 cases)**: `factor` near `u128::MAX`, `num/denom` near the saturation threshold (~88), small-denom edge cases — 0 spurious saturations. Every case where U512 short-circuits to `u128::MAX` corresponds to a true mathematical result that exceeds u128.
- **U512 behavior**: smoke-tested `checked_add` / `checked_mul` / `try_into::<u128>` / floor-div identity on `alloy_primitives::ruint::U512`. All behave as the diff assumes.
- **no_std + wasm32**: `cargo build -p alloy-eips --no-default-features` and `--target wasm32-unknown-unknown` both compile clean.
- **Downstream**:
  - reth callers in `crates/ethereum/evm/{lib.rs,build.rs}`, `crates/rpc/rpc-eth-types/{fee_history.rs,receipt.rs}`, and `crates/consensus/common/src/validation.rs` — all in runtime contexts, no const usage. `const fn` removal is a non-event.
  - revm has its own `fake_exponential` in `crates/context/interface/src/block/blob.rs:83` (u64 signature) and only uses alloy-eips for `BlockId` / `BlockNumberOrTag` types — doesn't depend on alloy-eips for blob math.

## Perf

Microbench (release, 1M iters per row):

| Input | Old (u128) | New (U512) | Notes |
|---|---|---|---|
| `factor=1, n=50M, denom=cancun` | 632 ns | 5760 ns | ~9× |
| `factor=1, n=148M, denom=cancun` | 2717 ns | 17116 ns | ~6× |
| `factor=1, n=299M, denom=pectra` | 1038 ns | 22342 ns | ~22× — *and old returns the wrong answer here* |

Worst realistic case ~22 µs/call. For block-validation cadence (one call per block, a handful per fee-history RPC) this is comfortably below noise floor.

## Suggested changes (non-blocking)

**1. Collapse the two divisions on line 513.** The current code does `numerator_accum = next_accum / denominator / i;`, but the comment on line 509 says `// Equivalent to: accum = (accum * numerator) / (denominator * i)`. Both forms are mathematically identical for positive integers (floor-div identity), but `next_accum / (denominator * i)` matches the documented intent and saves one U512 division per iteration. `denominator * i` stays comfortably within U512 — `denominator ≤ 2^128`, and `i` is bounded by loop termination (realistic max < 10k iters), so `denominator * i ≤ 2^142 ≪ 2^512`.

```rust
// suggested
numerator_accum = next_accum / (denominator * i);
```

**2. Optional: CHANGELOG mention.** `fake_exponential`, `calc_blob_gasprice`, `calc_blob_fee`, and `next_block_excess_blob_gas_osaka` lose `const fn`. I grepped alloy-rs + reth + revm and found no const-context callers, but worth flagging in CHANGELOG for downstream users with const expressions tied to these.

## CI

The `test windows-latest nightly --no-default-features -F reqwest-rustls-tls` failure is unrelated to this PR — `foundry-rs/foundry-toolchain` failed at the "Install Anvil" setup step with `curl: (22) The requested URL returned error: 403` (GitHub API rate-limit on the unauthenticated foundryup fetch). The `build` and `test` steps never ran on that matrix row. All other Windows matrix rows including other nightly variants (`test windows-latest nightly`, `test windows-latest nightly --all-features`) passed. A re-run should be green.

satyakwok (approved):
Reviewed this carefully. Posting full findings + suggestions in case any of it helps unblock review.

## Math correctness

Old `fake_exponential` had three u128 overflow sites:
1. `factor * denominator` on init
2. `output += numerator_accum`
3. `denominator * i` in the accumulator update

The `checked_mul(numerator) ... break` in the old loop is the actual bug — it exits early *even when the final result still fits in u128*, returning a truncated value. The `(1, 88, 1)` and pectra `n=299M` test cases added in this PR are direct repros:

- `(1, 88, 1)`: old returns `25661480018157551077491514070807592801`, correct is `165162653699637111792770913913821835905`
- `(1, 299_453_931, 5_007_716)`: old returns `905999095046699812947807`, correct is `93359993185840258978230108`

Both old values fit comfortably in u128, so this isn't a saturation issue — it's truncation from premature loop exit when `checked_mul` overflows mid-Taylor-expansion.

Widening to U512 cleanly removes all three overflow sites. `factor * denominator ≤ 2^256` worst case leaves ample headroom. When the true mathematical result exceeds u128, the final `(output / denominator).try_into().unwrap_or(u128::MAX)` saturates correctly. For divergent inputs where U512 itself would overflow, the `checked_add` / `checked_mul` short-circuits return `u128::MAX` — verified empirically to never produce spurious saturation for inputs where the spec result fits u128.

`next_accum / denominator / i` is equivalent to `next_accum / (denominator * i)` for positive integers (floor-div identity), so no behavioral change in division order.

The Osaka comparison widening (`eip7840.rs`) is also necessary — `DATA_GAS_PER_BLOB * calc_blob_fee()` can exceed u128 even when `calc_blob_fee` itself fits within u128 (`131072 * u128::MAX ≈ 2^145`), so widening both sides of the inequality to U512 is the right fix.

## Independent verification

- **Spec match**: ported the EIP-4844 Python reference implementation, verified all 19 explicit test vectors in this PR match.
- **Rust ↔ Python cross-check (140 random cases)**: generated `(factor, num, denom, expected)` tuples from the Python reference, added a temporary test module on a local copy of this branch calling `fake_exponential` on each — all 140 match. Inputs covered realistic Cancun (`denom=3338477`), Pectra (`denom=5007716`), zero edges (`factor=0`, `numerator=0`), and huge-numerator saturation cases.
- **Adversarial (38 cases)**: `factor` near `u128::MAX`, `num/denom` near the saturation threshold (~88), small-denom edge cases — 0 spurious saturations. Every case where U512 short-circuits to `u128::MAX` corresponds to a true mathematical result that exceeds u128.
- **U512 behavior**: smoke-tested `checked_add` / `checked_mul` / `try_into::<u128>` / floor-div identity on `alloy_primitives::ruint::U512`. All behave as the diff assumes.
- **no_std + wasm32**: `cargo build -p alloy-eips --no-default-features` and `--target wasm32-unknown-unknown` both compile clean.
- **Downstream**:
  - reth callers in `crates/ethereum/evm/{lib.rs,build.rs}`, `crates/rpc/rpc-eth-types/{fee_history.rs,receipt.rs}`, and `crates/consensus/common/src/validation.rs` — all in runtime contexts, no const usage. `const fn` removal is a non-event.
  - revm has its own `fake_exponential` in `crates/context/interface/src/block/blob.rs:83` (u64 signature) and only uses alloy-eips for `BlockId` / `BlockNumberOrTag` types — doesn't depend on alloy-eips for blob math.

## Perf

Microbench (release, 1M iters per row):

| Input | Old (u128) | New (U512) | Notes |
|---|---|---|---|
| `factor=1, n=50M, denom=cancun` | 632 ns | 5760 ns | ~9× |
| `factor=1, n=148M, denom=cancun` | 2717 ns | 17116 ns | ~6× |
| `factor=1, n=299M, denom=pectra` | 1038 ns | 22342 ns | ~22× — *and old returns the wrong answer here* |

Worst realistic case ~22 µs/call. For block-validation cadence (one call per block, a handful per fee-history RPC) this is comfortably below noise floor.

## Suggested changes (non-blocking)

**1. Collapse the two divisions on line 513.** The current code does `numerator_accum = next_accum / denominator / i;`, but the comment on line 509 says `// Equivalent to: accum = (accum * numerator) / (denominator * i)`. Both forms are mathematically identical for positive integers (floor-div identity), but `next_accum / (denominator * i)` matches the documented intent and saves one U512 division per iteration. `denominator * i` stays comfortably within U512 — `denominator ≤ 2^128`, and `i` is bounded by loop termination (realistic max < 10k iters), so `denominator * i ≤ 2^142 ≪ 2^512`.

```rust
// suggested
numerator_accum = next_accum / (denominator * i);
```

**2. Optional: CHANGELOG mention.** `fake_exponential`, `calc_blob_gasprice`, `calc_blob_fee`, and `next_block_excess_blob_gas_osaka` lose `const fn`. I grepped alloy-rs + reth + revm and found no const-context callers, but worth flagging in CHANGELOG for downstream users with const expressions tied to these.

## CI

The `test windows-latest nightly --no-default-features -F reqwest-rustls-tls` failure is unrelated to this PR — `foundry-rs/foundry-toolchain` failed at the "Install Anvil" setup step with `curl: (22) The requested URL returned error: 403` (GitHub API rate-limit on the unauthenticated foundryup fetch). The `build` and `test` steps never ran on that matrix row. All other Windows matrix rows including other nightly variants (`test windows-latest nightly`, `test windows-latest nightly --all-features`) passed. A re-run should be green.

---

Approving on the math.

crates/eips/src/eip7840.rs
```diff
@@ -140,8 +141,8 @@ impl BlobParams {
             return 0;
         }
 
-        if self.blob_base_cost as u128 * base_fee_per_gas as u128
-            > DATA_GAS_PER_BLOB as u128 * self.calc_blob_fee(excess_blob_gas)
+        if U512::from(self.blob_base_cost) * U512::from(base_fee_per_gas)
```
DaniPopes: u256 is enough here. N*M where both mod X cannot overflow X*2

mattsse: Good catch, switched to U256 in eacd72ed3. Kept U512 only in `fake_exponential` where the intermediates reach ~2^384.

DaniPopes (approved):

PR #3963, opened 2026-05-07
feat(pubsub): typed terminal-error channel

figtracer:
## Motivation

Today, the only way for an `alloy-pubsub` backend (WebSocket, IPC, or any third-party connector) to signal that it has died is `ConnectionInterface::close_with_error()`, which sends `()` over a `oneshot`. The `PubSubService` cannot tell the difference between:

- a transient socket failure (server dropped the TCP connection, network blip) — should reconnect, and
- a deterministic protocol/auth/handshake failure (bad credential, server-issued protocol error, malformed handshake) — guaranteed to fail again on every retry.

As a result, deterministic failures burn the entire `max_retries` budget (10 attempts with capped exponential backoff = up to ~5 minutes of wakeups) before the service finally gives up.

This PR makes the channel **typed** so backends can opt out of the reconnect loop, while remaining a strictly additive change for existing backends.

### DIFF
--- a/crates/pubsub/src/handle.rs
+++ b/crates/pubsub/src/handle.rs
@@ -24,7 +24,7 @@ pub struct ConnectionHandle {
 
     /// Notification from the backend of a terminal error.
     ///
-    /// The carried [`TransportError`] is used by [`crate::PubSubService`] to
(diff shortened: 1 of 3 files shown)


mattsse (approved):
Looks good, thanks.

PR #3966, opened 2026-05-08
feat: add MPP HTTP transport wrapper

stevencartavia:
Adds `MppHttp`, a wrapper around the default reqwest `Http` transport that handles HTTP 402 challenges. On 402 with a supported `WWW-Authenticate: Payment`challenge, it calls the user-supplied `PaymentProvider`, then replays the request once with `Authorization: Payment <credential>`. Any other status, or 402 with no supported challenge, is forwarded unchanged.

https://github.com/alloy-rs/alloy/pull/3927#issuecomment-4331892104
Enabled by default via the mpp feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -173,3 +173,9 @@ serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.20"
 tower-http = "0.6.1"
+
+mpp = { version = "0.10", default-features = false, features = ["client"] }
+
+# Temporary patch: feature-gate mpp's serde_json/preserve_order. Drop once upstream releases the fix.
+[patch.crates-io]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -173,3 +173,9 @@ serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.20"
 tower-http = "0.6.1"
+
+mpp = { version = "0.10", default-features = false, features = ["client"] }
+
+# Temporary patch: feature-gate mpp's serde_json/preserve_order. Drop once upstream releases the fix.
+[patch.crates-io]
+mpp = { git = "https://github.com/stevencartavia/mpp-rs", branch = "fix/optional-preserve-order" }
```
mattsse, line 181: we cant do this

PR #3967, opened 2026-05-08
fix(node-bindings): harden child process cleanup

Elena343-ai:
## Motivation

`GracefulShutdown::shutdown` could panic when the forced kill failed, and startup error paths in anvil, geth, and reth killed children without waiting for them. These paths run during teardown or before an instance is returned, so cleanup should stay best-effort and reap the child process.

## Solution

Make `GracefulShutdown` tolerate already-exited children and soft-log kill failures before waiting. Reuse the helper in node startup failure branches, including missing pipe handles, timeout, fatal output, and read/parse errors, and add focused shutdown tests.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/node.rs
+++ b/crates/node-bindings/src/node.rs
@@ -5,7 +5,7 @@ use std::time::Duration;
 use thiserror::Error;
 
 /// How long we will wait for the node to indicate that it is ready.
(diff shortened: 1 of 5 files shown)


mattsse (commented):
can keep as is

PR #3968, opened 2026-05-08
feat: add MPP WS transport wrapper

stevencartavia:
Adds `MppWsConnect`, a `WsConnect` wrapper that transparently handles MPP WebSocket sessions: probes for a `challenge` frame at handshake, runs the pay-credential flow via a user-supplied `PaymentProvider`, and falls through to plain WebSocket otherwise (latching the decision so reconnects skip the probe). On by default for non-wasm targets.


mattsse (commented):
we can not add this dep (cicle)

PR #3971, opened 2026-05-10
feat(signer-trezor): add skip_passphrase option to TrezorSigner

ifdario:
The trezor signer doesn't allow skipping the passphrase prompt.

### DIFF
--- a/crates/signer-trezor/src/signer.rs
+++ b/crates/signer-trezor/src/signer.rs
@@ -92,13 +92,23 @@ impl TrezorSigner {
         derivation: DerivationType,
         chain_id: Option<ChainId>,
     ) -> Result<Self, TrezorError> {
+        Self::new_with_passphrase(derivation, chain_id, false).await
+    }
+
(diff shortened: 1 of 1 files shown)


crates/signer-trezor/src/signer.rs
```diff
@@ -91,14 +91,24 @@ impl TrezorSigner {
...
+    ) -> Result<Self, TrezorError> {
+        Self::new_with_passphrase(derivation, chain_id, false).await
+    }
+
+    /// Instantiates a new Trezor signer with the option to skip the passphrase prompt.
+    #[instrument(ret)]
+    pub async fn new_with_passphrase(
+        derivation: DerivationType,
+        chain_id: Option<ChainId>,
+        skip_passphrase: bool,
```
mattsse, line 103: isnt this feature deprecated/removed?

mattsse (commented):
not familiar with this setting but looks like this is deprecated?

https://github.com/trezor/trezor-firmware/blob/main/common/protob/messages-management.proto#L66-L69

PR #3976, opened 2026-05-11
fix(consensus-any): saturate baseFeePerGas above u64::MAX on deser (#3741)

### LINKED ISSUE #3741: [Bug] AnyHeader fails to deserialize baseFeePerGas values exceeding u64::MAX
### Component

consensus, eips, genesis

_No response_

### Operating System

macOS (Apple Silicon)

### Describe the bug

`AnyHeader::base_fee_per_gas` is typed as `Option<u64>`, which causes deserialization failures on chains where `baseFeePerGas` exceeds `u64::MAX`.

We encountered this on a live EVM-compatible chain (Stable, chain ID 988) that returns `baseFeePerGas: "0x3635c9adc5dea00000"` (10^21) for a range of historical blocks. Deserializing the RPC response fails with:

```
invalid type: integer `1000000000000000000000`, expected u64
```

### Spec vs implementation

The [Ethereum execution spec](https://github.com/ethereum/execution-specs/blob/9460f9332968de9ce6370aeab3b1cc7af8b1f8f5/src/ethereum/forks/cancun/blocks.py#L196) defines `base_fee_per_gas` as `Uint` (arbitrary-precision), while explicitly choosing `U64` for [`blob_gas_used`](https://github.com/ethereum/execution-specs/blob/9460f9332968de9ce6370aeab3b1cc7af8b1f8f5/src/ethereum/forks/cancun/blocks.py#L210) and [`excess_blob_gas`](https://github.com/ethereum/execution-specs/blob/9460f9332968de9ce6370aeab3b1cc7af8b1f8f5/src/ethereum/forks/cancun/blocks.py#L218). This distinction appears intentional.

In alloy, `AnyHeader` uses `Option<u64>` for [`base_fee_per_gas`](https://github.com/alloy-rs/alloy/blob/v1.6.3/crates/consensus-any/src/block/header.rs#L68), and the [`BlockHeader` trait](https://github.com/alloy-rs/alloy/blob/v1.6.3/crates/consensus/src/block/header.rs#L487) constrains the return type to `Option<u64>`. 

### Suggestion

Change `AnyHeader::base_fee_per_gas` from `Option<u64>` to `Option<U256>` (or `Option<u128>`). This would be consistent with `AnyHeader`'s documented purpose of accounting for ["possible differences in network implementations"](https://github.com/alloy-rs/alloy/blob/v1.6.3/crates/consensus-any/src/block/header.rs#L3-L4). Ideally the `BlockHeader` trait method would also be widened, but even just changing `AnyHeader` would eliminate the deserialization failure.

The same consideration may apply to `gas_limit` and `gas_used`, which the spec also defines as `Uint`.

satyakwok:
Closes #3741.

## Bug

The Ethereum execution spec defines `base_fee_per_gas` as arbitrary-precision (`Uint`), but alloy types `AnyHeader::base_fee_per_gas` as `Option<u64>` for compatibility with Ethereum mainnet, which has never observed a value above `u64::MAX`. Some EVM-compatible chains do — **Stable (chain id 988)** returns `baseFeePerGas = 0x3635c9adc5dea00000` (10²¹) for a range of historical blocks, causing RPC deserialization to fail with:

```
invalid type: integer `1000000000000000000000`, expected u64
```

## Fix

Per maintainer-blessed direction ([@mattsse in the issue thread](https://github.com/alloy-rs/alloy/issues/3741#issuecomment-...)):

> would simply saturating this on deser would be an option?

Keep `Option<u64>` typing and **saturate** above `u64::MAX` on deserialize instead of erroring. New `saturating_base_fee_per_gas` serde adapter:

- **Serialize**: delegates to `alloy_serde::quantity::opt::serialize` — no behavioral change on output
- **Deserialize**: parses through `Option<U256>` then `try_into::<u64>()` with `unwrap_or(u64::MAX)` — saturates oversize, preserves None

```rust
pub fn deserialize<de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
where
    D: Deserializer<de>,
{
    let opt: Option<U256> = Option::deserialize(deserializer)?;
    Ok(opt.map(|v| v.try_into().unwrap_or(u64::MAX)))
}
```

Field-level doc records the rationale + points consumers needing full precision toward a custom header type.

## Tests

Four new sub-tests under `block::header::tests`:

| Test | Input | Expected |
|------|-------|----------|
| `deserializes_base_fee_within_u64` | `0x3b9aca00` (1 gwei) | `Some(1_000_000_000)` |
| `deserializes_base_fee_saturates_above_u64` | `0x3635c9adc5dea00000` (Stable chain 10²¹) | `Some(u64::MAX)` |
| `deserializes_base_fee_exact_u64_max` | `0xffffffffffffffff` | `Some(u64::MAX)` (no truncation) |
| `deserializes_base_fee_absent` | (field omitted) | `None` (pre-London headers) |

```
running 5 tests
test block::header::tests::deserializes_base_fee_absent ... ok
test block::header::tests::deserializes_base_fee_saturates_above_u64 ... ok
test block::header::tests::deserializes_tron_state_root_in_header ... ok
test block::header::tests::deserializes_base_fee_within_u64 ... ok
test block::header::tests::deserializes_base_fee_exact_u64_max ... ok

test result: ok. 5 passed; 0 failed
```

(Pre-existing `deserializes_tron_state_root_in_header` continues to pass.)

## Compatibility

- In-range values: unchanged behavior, same hex round-trip.
- Pre-London (no base fee field): unchanged, still `None`.
- Out-of-u64-range values: previously errored, now saturate. Any consumer that was relying on the deser error to detect oversize values will instead see `u64::MAX` — consumers that need the full-precision value should construct a custom header type with a wider field (now documented at the field level).

## Open questions

- Should I also apply the same saturating treatment to `blob_gas_used` / `excess_blob_gas`? The execution spec lists those as `U64` (not `Uint`), so the spec-aligned answer is no — the strict u64 typing matches intent. Happy to follow up if the team wants those relaxed too.

### DIFF
--- a/crates/consensus-any/Cargo.toml
+++ b/crates/consensus-any/Cargo.toml
@@ -38,6 +38,7 @@ serde = { workspace = true, features = ["derive"], optional = true }
 [dev-dependencies]
 arbitrary = { workspace = true, features = ["derive"] }
 alloy-consensus = { workspace = true, features = ["arbitrary"] }
+bincode = { workspace = true, features = ["serde"] }
 serde_json.workspace = true
 
 [features]
--- a/crates/consensus-any/src/block/header.rs
+++ b/crates/consensus-any/src/block/header.rs
@@ -424,19 +424,23 @@ mod saturating_base_fee_per_gas {
     use alloy_primitives::U256;
     use serde::{Deserialize, Deserializer, Serializer};
 
-    pub fn serialize<S>(value: &Option<u64>, serializer: S) -> Result<S::Ok, S::Error>
+    pub(super) fn serialize<S>(value: &Option<u64>, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
         alloy_serde::quantity::opt::serialize(value, serializer)
     }
 
-    pub fn deserialize<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
+    pub(super) fn deserialize<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
     where
         D: Deserializer<'de>,
     {
-        let opt: Option<U256> = Option::deserialize(deserializer)?;
-        Ok(opt.map(|v| v.try_into().unwrap_or(u64::MAX)))
+        if deserializer.is_human_readable() {
+            let opt: Option<U256> = Option::deserialize(deserializer)?;
+            Ok(opt.map(|v| v.try_into().unwrap_or(u64::MAX)))
+        } else {
+            alloy_serde::quantity::opt::deserialize(deserializer)
+        }
     }
 }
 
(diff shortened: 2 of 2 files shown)


mattsse (approved):
Looks good now. I pushed a follow-up commit to keep the saturating path limited to human-readable serde, preserve the existing binary serde shape, and cover that with a bincode regression test.

satyakwok:
Sharp split — JSON sees untrusted RPC where saturate is the sane fallback, but binary overflow inside our own storage = corruption bug, fail-loud beats silent-clamp. Good call on the bincode regression test pinning that boundary; would've drifted quietly without it.

PR #3979, opened 2026-05-12
fix(node-bindings): reap child after startup error kill

apetro2:
## Motivation
                                                                                                                                                                                                 
  Startup error paths in `anvil`, `geth`, and `reth` were killing child processes and returning immediately, without an explicit `wait` to reap the process.
  While drop-time shutdown may eventually clean up, this leaves inconsistent process-lifecycle handling and can increase short-lived zombie/resource pressure under repeated startup failures.   
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  This PR introduces `GracefulShutdown::kill_and_wait` in `node-bindings` utils and uses it in startup timeout/fatal branches for:                                                               
                                                                                                                                                                                                 
  - `crates/node-bindings/src/nodes/anvil.rs`                                                                                                                                                    
  - `crates/node-bindings/src/nodes/geth.rs`
  - `crates/node-bindings/src/nodes/reth.rs`                                                                                                                                                     
                                                                                                                                                                                                 
  The helper performs a best-effort reap flow (`try_wait` fast-path, then `kill` + `wait`) so error returns happen after explicit process cleanup.                                               
                                                                                                                                                                                                 
  A unix-only unit test was added to validate that `kill_and_wait` reaps the child process:                                                                                                      
  - `kill_and_wait_reaps_child`                                                                                                                                                                  
                                                                                                                                                                                                 
  ## PR Checklist
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/anvil.rs
+++ b/crates/node-bindings/src/nodes/anvil.rs
@@ -459,7 +459,7 @@ impl Anvil {
         let mut wallet = None;
         loop {
             if start + timeout <= Instant::now() {
-                let _ = child.kill();
+                GracefulShutdown::kill_and_wait(&mut child);
                 return Err(NodeError::Timeout);
             }
 
--- a/crates/node-bindings/src/nodes/geth.rs
+++ b/crates/node-bindings/src/nodes/geth.rs
@@ -627,7 +627,7 @@ impl Geth {
 
         loop {
             if start + NODE_STARTUP_TIMEOUT <= Instant::now() {
-                let _ = child.kill();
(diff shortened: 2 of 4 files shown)


onbjerg (changes requested):
Reaping the child makes sense, but `kill_and_wait` ignores a failed `kill()` and still calls `wait()`. If the child stays alive, a startup timeout can turn into an indefinite wait. Could you handle that case without blocking indefinitely and add a test for it?

PR #3986, opened 2026-05-13
fix(node-bindings): apply custom reth genesis on startup

mk0walsk:
## Motivation
                                                                                                                                                                                                 
  `Reth::genesis()` promises that setting a custom genesis will initialize the node with `reth init` and the same `--datadir`, but `try_spawn()` previously did not execute that init flow.      
  This made the API behavior inconsistent with its documentation and could leave custom genesis settings unapplied at runtime, causing confusing and hard-to-diagnose behavior.
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  This PR implements the missing initialization path in `Reth::try_spawn()`:                                                                                                                     
                                                                                                                                                                                                 
  - When `genesis` is set, require `data_dir`.                                                                                                                                                   
  - Write the provided genesis to `<data_dir>/genesis.json`.                                                                                                                                     
  - Run `reth init --datadir <data_dir> --chain <genesis.json>` before starting `reth node`.
  - Pass the same genesis path to node startup (`--chain`) when `chain_or_path` is not explicitly set.                                                                                           
  - Reject invalid mixed configuration (`genesis` + `chain_or_path`) to avoid conflicting chain sources.                                                                                         
                                                                                                                                                                                                 
  Tests were updated to validate real effect:                                                                                                                                                    

  - Integration test now verifies `eth_chainId` matches the custom genesis chain id.
  - Unit test verifies `genesis` and `chain_or_path` cannot be set together.                                                                                                                     

  ## PR Checklist
                                                                                                                                                                                                 
  - [x] Added Tests
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/node-bindings/src/nodes/reth.rs
+++ b/crates/node-bindings/src/nodes/reth.rs
@@ -8,7 +8,7 @@ use alloy_genesis::Genesis;
 use rand::Rng;
 use std::{
     ffi::OsString,
-    fs::create_dir_all,
+    fs::{create_dir_all, File},
     io::{BufRead, BufReader},
     path::PathBuf,
     process::{Child, ChildStdout, Command, Stdio},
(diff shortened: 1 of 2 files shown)


mattsse:
Closing this because reth can load a custom genesis directly via `reth node --chain <path>`; running `reth init` is unnecessary for the node-bindings flow. I’ll follow up with a docs-only PR clarifying the expected behavior.

PR #3989, opened 2026-05-14
fix(network): preserve transaction request extra fields

jschnelder:
## Motivation
                                                                                                                                                                                                 
  `AnyRpcTransaction -> WithOtherFields<TransactionRequest>` was dropping RPC extension fields during conversion.                                                                                
  The previous implementation rebuilt the wrapper with `Self::new(req)`, which discarded `other`.                                                                                                
  This caused loss of provider-specific fields (for example `tempoFeePayer`) and made passthrough behavior unreliable.                                                                           
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  The conversion now uses `tx.into_parts()` and returns `Self { inner: req, other }`, preserving all extra RPC fields while still converting the inner transaction into `TransactionRequest`.    
  A unit test was added (`preserves_other_fields_when_converting_to_transaction_request`) to verify the behavior, and it passes with:                                                                                                                                                                                    
  `cargo test -p alloy-network preserves_other_fields_when_converting_to_transaction_request`                                                                                                    
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/network/src/any/mod.rs
+++ b/crates/network/src/any/mod.rs
@@ -437,8 +437,9 @@ impl From<AnyRpcTransaction> for Recovered<AnyTxEnvelope> {
 
 impl From<AnyRpcTransaction> for WithOtherFields<TransactionRequest> {
     fn from(tx: AnyRpcTransaction) -> Self {
-        let req: TransactionRequest = tx.0.inner.into_recovered().into();
-        Self::new(req)
+        let (inner, other) = tx.into_parts();
+        let req: TransactionRequest = inner.into_recovered().into();
+        Self { inner: req, other }
     }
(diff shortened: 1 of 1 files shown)


mattsse (approved):
Reviewed the conversion path and added the missing unknown-transaction coverage. The updated implementation preserves recovered sender, unknown envelope fields, and outer RPC extension fields.

PR #3992, opened 2026-05-15
fix(provider): timeout pending tx confirmations

Mapogundal:
## Motivation

Pending transaction timeouts were only reaped from the unconfirmed watcher map. Once a transaction moved into the confirmation-waiting map, the configured timeout could fire without notifying the watcher, leaving callers waiting indefinitely if no later block completed the confirmation count.

## Solution

Schedule timeouts for watchers that enter the confirmation-waiting path and reap timed-out watchers from both pending maps. Add a unit test covering the waiting-confirmations timeout branch.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


onbjerg:

PR #3999, opened 2026-05-20
fix(engine): reject higher-version fields in V1/V2 serde

isagi-y22:
## Motivation
                                                                                                                                                                                                 
  `ExecutionPayloadV1` and `ExecutionPayloadV2` deserialization was implicitly accepting higher-version payload fields (for example V3 fields) and silently dropping them.                       
  This causes version-boundary ambiguity and breaks round-trip expectations, and it is inconsistent with the stricter validation already implemented in `ExecutionPayload` enum deserialization. 
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  - Made `ExecutionPayloadV1` deserialization strict by adding `deny_unknown_fields`.                                                                                                            
  - Made `ExecutionPayloadV2` deserialization strict by adding `deny_unknown_fields` to its serde helper.                                                                                        
  - Updated the existing `serde_faulty_roundtrip_payload_input_v2/v3` tests to assert deserialization errors instead of allowing silent downgrade behavior.                                      
                                                                                                                                                                                                 
  This is a targeted validation fix: valid V1/V2 payloads continue to deserialize as before, while mixed-version inputs are now rejected.                                                        
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/rpc-types-engine/src/payload.rs
+++ b/crates/rpc-types-engine/src/payload.rs
@@ -584,7 +584,7 @@ pub struct ExecutionPayloadEnvelopeV6 {
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(feature = "ssz", derive(ssz_derive::Encode, ssz_derive::Decode))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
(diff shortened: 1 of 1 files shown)


mattsse (commented):
we can keep this as is

PR #4000, opened 2026-05-21
feat(consensus): add eth72 pooled transaction encoding

Soubhik-10:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/transaction/pooled.rs
+++ b/crates/consensus/src/transaction/pooled.rs
@@ -3,8 +3,9 @@
 
 use super::EthereumTxEnvelope;
 use crate::{
-    error::ValueError, transaction::RlpEcdsaEncodableTx, Signed, TxEip4844, TxEip4844Variant,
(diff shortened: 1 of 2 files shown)


decofe:
Closing this PR because the additional sparse sidecar codec is not needed for the current Geth/EIP-8070 wire path. Geth represents the elided blob payload using the existing EIP-7594 sidecar structure with an empty blob list (RLP `0xc0`), while retaining commitments and cell proofs; Alloy already encodes/decodes an empty `Vec<Blob>` through the existing codec. The `0x80` sparse encoding and duplicate sidecar methods here therefore should not be introduced as a separate wire format. The remaining work belongs in Reth’s eth-version-gated networking and runtime cell availability/validation/inclusion paths.\n\nReference: https://github.com/ethereum/go-ethereum/blob/d91b71fb3627253537c2144a1d174dab871fec3/core/txpool/blobpool/blobpool.go#L262-L289

PR #4001, opened 2026-05-21
provider: avoid caching pending tx snapshots by hash

gfischer7:
## Motivation
                                                                                                                                                                                                 
  CacheProvider::get_transaction_by_hash cached every Some(tx) response, including pending transactions.                                                                                         
  When the first response is pending (blockHash/blockNumber are null), that snapshot can be served from cache after the transaction is mined, which hides inclusion updates and causes stale state to persist.                                                                                                                                                                              
                                                                                                                                                                                                 
  ## Solution                                                                                                                                                                                    
                                                                                                                                                                                                 
  This PR updates the cache behavior for eth_getTransactionByHash to cache only included transactions.                                                                                           
  Concretely, the response is cached only when tx.block_hash_num().is_some(), so pending snapshots are not persisted and mined-state fields remain observable on later calls. A regression test was added to verify this flow: pending response is not cached, mined response is cached, and subsequent reads hit the mined cache entry.                                                       
                                                                                                                                                                                                 
  ## PR Checklist                                                                                                                                                                                
                                                                                                                                                                                                 
  - [x] Added Tests                                                                                                                                                                              
  - [ ] Added Documentation                                                                                                                                                                      
  - [ ] Breaking changes

### DIFF
--- a/crates/provider/src/layers/cache.rs
+++ b/crates/provider/src/layers/cache.rs
@@ -4,6 +4,7 @@ use crate::{
 use alloy_eips::BlockId;
 use alloy_json_rpc::{RpcError, RpcSend};
 use alloy_network::Network;
+use alloy_network_primitives::TransactionResponse;
 use alloy_primitives::{
     keccak256, Address, Bytes, StorageKey, StorageValue, TxHash, B256, U256, U64,
 };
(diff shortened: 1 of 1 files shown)


mattsse (approved):
LGTM

PR #4006, opened 2026-05-22
fix(transport-http): preserve JSON-RPC errors from non-2xx

Mohiiit:
## Summary

Preserves valid JSON-RPC error responses returned with non-2xx HTTP status codes in the reqwest and hyper HTTP transports.

## Example

Some providers return protocol-level JSON-RPC errors with an HTTP error status. One observed case from Alchemy was an `eth_getFilterChanges` request returning HTTP `400` with this JSON-RPC body:

```json
{
  "jsonrpc": "2.0",
  "id": 1766,
  "error": {
    "code": -32000,
    "message": "filter not found"
  }
}
```

Before this change, the HTTP transports converted that response into a transport `HttpError`, so callers could not inspect the JSON-RPC error payload. For long-running filter pollers, that means existing terminal-error handling for `filter not found` is bypassed.

## Fix

When a non-2xx HTTP response is received, the transports now first try to parse the body as a JSON-RPC response packet. If the body contains a valid JSON-RPC error response, it is returned through the normal response path so existing RPC error handling can classify it. Non-JSON-RPC HTTP failures still return the existing transport `HttpError`.

## Validation

Validated the focused transport crate across default, no-feature, and hyper-only feature sets, including clippy for the touched feature paths.

### DIFF
--- a/crates/transport-http/src/hyper_transport.rs
+++ b/crates/transport-http/src/hyper_transport.rs
@@ -152,6 +152,10 @@ where
         }
 
         if !status.is_success() {
+            if let Some(response) = crate::json_rpc_error_response(body.as_ref()) {
+                return Ok(response);
+            }
+
             return Err(TransportErrorKind::http_error(
                 status.as_u16(),
                 String::from_utf8_lossy(&body).into_owned(),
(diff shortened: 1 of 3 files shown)


mattsse (approved):
Looks reasonable. Preserving the JSON-RPC error payload here gives callers the provider-specific code/message/data, which is more useful than collapsing these responses into a generic HTTP transport error. Plain non-RPC HTTP failures still fall back to HttpError.

Mohiiit:
gm @mattsse, is there any other issue I can pick up?

PR #4013, opened 2026-05-26
fix(provider): avoid watch hang on stale block numbers

esorense:
## Motivation
                                                                                                                    
  PendingTransactionBuilder::watch could hang for a long time on HTTP providers when eth_blockNumber is stale.      
  In this case, heartbeat-based block progression may not advance even though the transaction receipt is already available, so .watch() can fail to resolve in practice.                                                           
                                                                                                                    
  ## Solution                                                                                                       

  Add a receipt-based fallback path to PendingTransactionBuilder::watch for single-confirmation watches (confirmations == 1): while waiting on heartbeat, poll eth_getTransactionReceipt on the provider poll interval and  return once a receipt is observed.                                                                                
  This preserves existing multi-confirmation behavior and aligns watch with the robustness already present in the get_receipt flow.                                                                                                 
  Also add a regression test that simulates stale eth_blockNumber responses and verifies watch() still completes when the receipt becomes available.                                                                               
                                                                                                                    
  ## PR Checklist                                                                                                   
                                                                                                                    
  - [x] Added Tests                                                                                                 
  - [ ] Added Documentation                                                                                         
  - [ ] Breaking changes

### DIFF
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -208,7 +208,37 @@ impl<N: Network> PendingTransactionBuilder<N> {
     /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
     ///   confirmed.
     pub async fn watch(self) -> Result<TxHash, PendingTransactionError> {
-        self.register().await?.await
+        let hash = self.config.tx_hash;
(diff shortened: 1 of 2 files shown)


onbjerg:

PR #4015, opened 2026-05-27
chore: replaced typed tx with consensus tx in build_typed_simulate_tx…

Rimeeeeee:
This is needed for rpc-compat tests.
```
 differs from expected (-- client, ++ test):
{
  -- "error": {
    -- "code": -32602,
    -- "message": "Failed to convert transaction into RPC response: Transaction is not buildable"
  -- },
  "id": 1,
  "jsonrpc": "2.0",
```


mattsse (changes requested):
can we add a few easy test cases for this

mattsse (approved):

PR #4022, opened 2026-06-01
feat: Provider::watch_canonical_logs_from

0xForerunner:
- Adds Provider::watch_logs_from(start_block, filter) — streams per-block BlockLogs futures from a historical height to head, fetching block and one-block log range concurrently with fallback to a block-hash-pinned log query for fork-safe batches.
  - Adds Provider::watch_canonical_logs_from(...) — reorg-aware variant that emits CanonicalEvent::Added / Removed for BlockLogs, reusing retained logs for Removed events.
  
## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [] Breaking changes


mattsse (approved):
LGTM

PR #4024, opened 2026-06-04
feat: add ssz encoding and decoding to Payload Envelopes

Soubhik-10:


crates/rpc-types-engine/src/payload.rs
```diff
@@ -352,6 +390,66 @@ pub struct ExecutionPayloadEnvelopeV4 {
...
+        encoder.finalize();
+    }
+
+    fn ssz_bytes_len(&self) -> usize {
+        self.envelope_inner.execution_payload.ssz_bytes_len()
+            + self.envelope_inner.block_value.ssz_bytes_len()
+            + self.envelope_inner.blobs_bundle.ssz_bytes_len()
+            + self.envelope_inner.should_override_builder.ssz_bytes_len()
+            + self.execution_requests.ssz_bytes_len()
+    }
```
mattsse, line 427: `ssz_bytes_len()` is expected to match `as_ssz_bytes().len()`, but this only sums the nested variable payload bytes and misses the 4-byte offsets written for `execution_payload`, `blobs_bundle`, and `execution_requests`. The roundtrip tests do not catch it because encoding does not call `ssz_bytes_len()`.

```suggestion
    fn ssz_bytes_len(&self) -> usize {
        let fixed_section_len = <ExecutionPayloadV3 as ssz::Encode>::ssz_fixed_len()
            + <U256 as ssz::Encode>::ssz_fixed_len()
            + <BlobsBundleV1 as ssz::Encode>::ssz_fixed_len()
            + <bool as ssz::Encode>::ssz_fixed_len()
            + <Requests as ssz::Encode>::ssz_fixed_len();

        fixed_section_len
            + self.envelope_inner.execution_payload.ssz_bytes_len()
            + self.envelope_inner.blobs_bundle.ssz_bytes_len()
            + self.execution_requests.ssz_bytes_len()
    }
```

mattsse (approved):

PR #4027, opened 2026-06-07
fix(pubsub): drop tokio/time from the wasm build (#4023)

### LINKED ISSUE #4023: [Bug] Can't use `alloy-provider/ws` on wasm because of transitive dependency `tokio-stream` on `alloy-pubsub`
### Component

provider, pubsub

alloy-provider v2.0.5

### Operating System

Linux

### Describe the bug

Including `alloy-provider` with the `ws` feature leads to `tokio` being included with the `time` feature, crashing a wasm application on building a `Runtime`

`cargo tree -i tokio -e features --target wasm32-unknown-unknown`
```
tokio v1.50.0
└── tokio feature "time"
        └── tokio-stream feature "time"
                └── tokio-stream feature "default"
                        ├── alloy-pubsub v2.0.5
                        │     ├── alloy-provider v2.0.5 (*)
                        │     ├── alloy-rpc-client v2.0.5
                        │     └── alloy-transport-ws v2.0.5
                        └── alloy-rpc-client v2.0.5 (*)
```

Image

0xMars42:
## Motivation

`alloy-provider` with the `ws` feature pulls `tokio` with the `time` feature on
`wasm32-unknown-unknown`. `tokio/time` panics at runtime when a tokio `Runtime` is built there, so a
wasm app using `alloy-provider/ws` crashes (#4023).

## Root cause

Two paths enable `tokio/time` in the wasm build:

1. `tokio-stream`'s **default** features include `time` (`time = ["tokio/time"]`), and `alloy-pubsub`
   pulls `tokio-stream`. This is the path shown in the issue's `cargo tree`.
2. `alloy-pubsub`'s `handle.rs` imports `tokio::time::Duration`. The `tokio::time` module path
   requires the `time` feature, even though `Duration` itself is just `std::time::Duration`.

## Fix

1. Set `default-features = false` on the workspace `tokio-stream` (alloy-pubsub only uses its `sync`
   feature).
2. Import `std::time::Duration` instead of `tokio::time::Duration` in `alloy-pubsub`.

`alloy-pubsub` already uses `wasmtimer::tokio::sleep` on wasm and `tokio::time::sleep` off-wasm, and
gets `tokio/time` on native via `alloy-transport`, so this only drops the feature from the wasm
target.

## Verification

- `tokio/time` is no longer pulled into the wasm build:
  `cargo tree -e features -i tokio -p alloy-provider --features ws --target wasm32-unknown-unknown`
  goes from 3 matches to 0 with this change.
- `alloy-provider` with `ws` compiles for `wasm32-unknown-unknown` (the issue's scenario), and the
  repo's `wasm-unknown` CI job (`cargo hack build --features ws --target wasm32-unknown-unknown`,
  same excludes) builds clean locally.
- Native build of `alloy-pubsub` is unaffected (`tokio/time` comes from `alloy-transport`).

Closes #4023.


DaniPopes (approved):

0xMars42:
Updated. The first version was incomplete: alloy-pubsub also imported `tokio::time::Duration` in `handle.rs`, which kept `tokio/time` enabled on wasm. Switched that to `std::time::Duration` as well, so `tokio/time` is now actually dropped from the wasm build.

Verified by reproducing the `wasm-unknown` CI job locally (`cargo hack build --features ws --target wasm32-unknown-unknown` with the same excludes): builds clean. `tokio/time` is no longer in the `alloy-provider` ws feature graph on wasm.

For reference, the `test ... stable|nightly` failures on this PR are pre-existing on `main`, unrelated to this change.

hukasu:
is this fix exclusive to the `ws` feature or also the `ws-ring` feature?

DaniPopes:
It's for both

hukasu:
ty boss

PR #4038, opened 2026-06-13
feat(engine): add REST-SSZ wire types

Soubhik-10:
Implements the REST-SSZ wire containers from ethereum/execution-apis#793, including fork-specific payloads, attributes, bodies, blobs, and list-encoded optionals. Preserves absent validation errors with a separate status tag and bounded UTF-8 error value, and enforces schema bounds when decoding the dedicated containers while reusing Alloy's existing payload representations and legacy codecs.

Organizes the types by endpoint family and shares identical fork shapes and blob conversions. The experimental witness extension remains in Reth.


0xalpharush:
The SSZ tag derive accepts trailing bytes when `PayloadStatusKind` is decoded standalone (for example, `[0, 0]` decodes as `Valid`). I opened a minimal fix and regression test against this PR’s source branch: https://github.com/Rimeeeeee/alloy/pull/7. The full `alloy-rpc-types-engine` test suite passes (136 tests).

mattsse (approved):

PR #4040, opened 2026-06-16
fix(eips): saturate base fee on the increase path

satyakwok:
The increase branch of `calc_next_block_base_fee` adds the per-block delta with `+`, while the decrease branch already uses `saturating_sub`. Since `base_fee` can reach `u64::MAX` (saturated on deserialization, #3741), a block with `gas_used` above target overflows `u64` — panics in debug, wraps in release.

Use `saturating_add` to match the decrease path. Added a regression test (fails before / passes after).


mattsse (commented):
we can keep this as is

PR #4044, opened 2026-06-20
feat(signer-trezor): implement EIP-712 typed data signing

### LINKED ISSUE #4043: feat(signer-trezor): implement EIP-712 typed data signing
## Summary

The Trezor signer does not implement `sign_dynamic_typed_data`, so EIP-712 typed data signing falls back to the default trait implementation which hashes locally and calls `sign_hash`. Since the Trezor signer returns `UnsupportedOperation` for `sign_hash`, EIP-712 signing fails entirely.

Trezor devices (Model T, Model One) support on-device EIP-712 signing via the `EthereumSignTypedData` streaming protocol (message type 464). The device computes the EIP-712 hash internally and displays struct fields on screen for user verification, which is more secure than signing a blind hash.

## Expected behavior

`sign_dynamic_typed_data` should use the Trezor's native EIP-712 streaming protocol to sign typed data on-device, with struct field verification on the device screen.

## Current behavior

Calling `sign_dynamic_typed_data` on a `TrezorSigner` fails because it falls through to `sign_hash`, which is unsupported.

## Dependencies

This requires `trezor-client` to expose `ethereum_sign_typed_data()`, which is proposed in trezor/trezor-firmware#7038.

ifdario:
Closes #4043

## Summary

- Override `sign_dynamic_typed_data` in `TrezorSigner` to use the Trezor's native EIP-712 streaming protocol (`EthereumSignTypedData`, message type 464)
- The device computes the EIP-712 hash on-device and displays struct fields for user verification
- Inject `EIP712Domain` into the types map since alloy's `Resolver` omits it, but the Trezor device needs it to compute the domain separator hash
- Add `Eip712` error variant to `TrezorError` for serialization errors
- Add `alloy-dyn-abi` (with `eip712` feature) and `serde_json` as dependencies

## Dependencies

This depends on `trezor-client` 0.1.6 gaining `ethereum_sign_typed_data()`, proposed in trezor/trezor-firmware#7038. The workspace `[patch.crates-io]` currently points to a fork with this method; it should be updated once `trezor-client` 0.1.6 is published.

## Test plan

- Tested with a Trezor Model T signing EIP-712 multisig payloads (Hyperliquid L1)
- Verified struct fields display correctly on device screen during signing
- Confirmed signature recovery matches the expected signer address

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -173,3 +173,6 @@ serial_test = "3.0"
 similar-asserts = "1.5"
 tempfile = "3.20"
 tower-http = "0.6.1"
+
+[patch.crates-io]
+trezor-client = { git = "https://github.com/infinitefield/trezor-firmware.git", branch = "main" }
--- a/crates/signer-trezor/Cargo.toml
+++ b/crates/signer-trezor/Cargo.toml
@@ -24,16 +24,18 @@ workspace = true
 
 [dependencies]
 alloy-consensus = { workspace = true, features = ["std"] }
+alloy-dyn-abi = { workspace = true, features = ["std", "eip712"] }
 alloy-network.workspace = true
 alloy-primitives.workspace = true
-alloy-signer.workspace = true
+alloy-signer = { workspace = true, features = ["eip712"] }
 
-trezor-client = { version = "=0.1.5", default-features = false, features = [
+trezor-client = { version = "=0.1.6", default-features = false, features = [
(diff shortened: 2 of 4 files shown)


ifdario:
Closing for now. This depends on trezor/trezor-firmware#7038 being merged and trezor-client 0.1.6 published to crates.io. Will resubmit once the dependency is available.

PR #4045, opened 2026-06-21
feat(consensus): serialize block from refs

decofe:
## Summary
- add serde-gated `Block::serialize_ref` for serializing from borrowed header/body parts
- preserve the regular `Block` serde shape without constructing or cloning an owned block
- cover parity with owned block serialization

Prompted by: @klkvr

## Tests
- cargo test -p alloy-consensus --features serde serialize_ref_matches_block_serde


decofe:
Closing in favor of keeping the fix scoped to Tempo for now.

PR #4050, opened 2026-06-27
fix(provider): serialize eth_getBlockReceipts block hash as plain string

### LINKED ISSUE #4049: get_block_receipts serializes BlockId::Hash as EIP-1898 object, rejected by some nodes
### Summary

`Provider::get_block_receipts(BlockId)` serializes a `BlockId::Hash` to the EIP-1898 object form `{"blockHash": "0x.."}` (via the manual `Serialize for BlockId` impl in `alloy-eips`). Some clients implement `eth_getBlockReceipts` to accept **only** the plain hash string `"0x.."` and reject the object form with `INVALID_PARAMS`.

### Repro

Calling `provider.get_block_receipts(BlockId::from(block_hash))` sends:

```json
{"jsonrpc":"2.0","method":"eth_getBlockReceipts","params":[{"blockHash":"0x.."}],"id":1}
```

A node such as taraxa-node rejects this with:

```
INVALID_PARAMS: Invalid method parameters ... Type is not convertible to string
```

The plain-string form, accepted by every Geth-family client I've tested, is:

```json
{"jsonrpc":"2.0","method":"eth_getBlockReceipts","params":["0x.."],"id":1}
```

### Question

Would alloy consider serializing a bare block-hash param for `eth_getBlockReceipts` (i.e. a `BlockId::Hash` with `require_canonical = None`) as a plain hash string rather than the EIP-1898 object, for broader node compatibility? Or is the position that such nodes are non-conformant and should accept the EIP-1898 object?

### Context

Downstream report and workaround: https://github.com/graphprotocol/graph-node/issues/5835 (graph-node currently sends the object form, which silently degrades affected indexers to per-transaction receipt fetching).

### Version

`alloy` 2.0.5

cargopete:
Closes #4049.

## Motivation

`Provider::get_block_receipts(BlockId)` serializes a `BlockId::Hash` to the
EIP-1898 object form `{"blockHash":"0x.."}`. Some clients (e.g. taraxa-node)
implement `eth_getBlockReceipts` to accept **only** the plain hash string
`"0x.."` and reject the object with `INVALID_PARAMS`. The plain-string form is
accepted by every Geth-family client tested.

This can't be fixed by changing `BlockId`'s `Serialize`: for genuine EIP-1898
methods (`eth_call`, `eth_getProof`) a bare block-hash string is *invalid* and
the `{blockHash}` object is required. The fix has to be local to
`eth_getBlockReceipts`.

## Solution

Introduce `BlockReceiptsParams`, a thin wrapper over `BlockId` used only as the
`eth_getBlockReceipts` parameter. It serializes:

- `BlockId::Hash` with `require_canonical == None` → plain hash string `"0x.."`
- `BlockId::Hash` with `require_canonical == Some(_)` → standard `{blockHash, requireCanonical}` object (so the flag is preserved)
- `BlockId::Number` (number/tag) → unchanged (already a string)

All four `get_block_receipts` impls (trait default, `CacheLayer`, `Filler`,
`ErasedProvider`) route through it. The cache layer still tracks the raw
`BlockId` via `with_block_id` for tag detection, so caching behaviour is
unchanged.

## API note

This changes the parameter type in the public return signature from
`ProviderCall<(BlockId,), _>` to `ProviderCall<(BlockReceiptsParams,), _>`. The
method signature (`block: BlockId`) and the awaited result are unchanged, so
typical callers are unaffected, but it is technically a breaking change to that
generic. Happy to reshape this however you prefer — e.g. a different type name,
location (`alloy-eips`), or approach — if you'd rather take it another way.

## Tests

- New unit tests for `BlockReceiptsParams` serialization (hash → string,
  hash + `requireCanonical` → object, number → quantity, tag → string).
- Existing `gets_block_receipts` and cache `test_block_receipts` tests pass
  unchanged.

Downstream context: graphprotocol/graph-node#5835.

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -49,10 +49,10 @@ use tracing::error;
 #[cfg(feature = "pubsub")]
 use crate::GetSubscription;
 use crate::{
-    provider::SendableTx, EthCall, EthCallMany, EthGetBlock, FilterPollerBuilder, Identity,
-    PendingTransaction, PendingTransactionBuilder, PendingTransactionConfig,
-    PendingTransactionError, Provider, ProviderCall, ProviderLayer, RootProvider, RpcWithBlock,
-    SendableTxErr,
+    provider::SendableTx, BlockReceiptsParams, EthCall, EthCallMany, EthGetBlock,
+    FilterPollerBuilder, Identity, PendingTransaction, PendingTransactionBuilder,
+    PendingTransactionConfig, PendingTransactionError, Provider, ProviderCall, ProviderLayer,
+    RootProvider, RpcWithBlock, SendableTxErr,
 };
 use alloy_json_rpc::RpcError;
 use alloy_network::{AnyNetwork, Ethereum, Network};
@@ -536,7 +536,7 @@ where
     fn get_block_receipts(
         &self,
         block: BlockId,
-    ) -> ProviderCall<(BlockId,), Option<Vec<N::ReceiptResponse>>> {
+    ) -> ProviderCall<(BlockReceiptsParams,), Option<Vec<N::ReceiptResponse>>> {
         self.inner.get_block_receipts(block)
     }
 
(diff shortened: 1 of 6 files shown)


mattsse:
Closing as superseded by #4052.

PR #4057, opened 2026-07-03
perf(json-rpc): skip re-parse when returning owned Box<RawValue>

mfw78:
`try_deserialize_ok` calls `serde_json::from_str::<Box<RawValue>>` even when `J` is already `Box<RawValue>`, so the raw passthrough path (`Provider::raw_request`, `RpcClient::request`/batch with `Box<RawValue>`) copies and re-validates a payload it already owns on every response.

When `J` and `T` are both `Box<RawValue>`, return the owned payload directly behind a `TypeId` guard, matching the `transmute_copy` specialization in `IntoBoxTransport::into_box_transport`. The generic deserialize path is unchanged.

This adds a `'static` bound to the (public) `try_deserialize_ok`, needed for `TypeId`. Both in-tree callers pass `Box<RawValue>`; only external callers passing a borrowed `J` would be affected.

### DIFF
--- a/crates/json-rpc/src/result.rs
+++ b/crates/json-rpc/src/result.rs
@@ -1,6 +1,6 @@
 use crate::{Response, ResponsePayload, RpcError, RpcRecv};
 use serde_json::value::RawValue;
-use std::borrow::Borrow;
+use std::{any::TypeId, borrow::Borrow};
 
 /// The result of a JSON-RPC request.
 ///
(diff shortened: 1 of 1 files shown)


mattsse (approved):
this seems okay, we only use it for `raw_request_dyn` and there we already enforce 'static owned I believe so the additional 'static is okay here.

wdyt @DaniPopes

mfw78:
The failing `docs` job is an upstream rustc regression, not this PR. The 2026-07-02 nightly ICEs in `rustc_metadata` (`get_item_attrs` decoder assertion) when rustdoc inlines a `pub extern crate` re-export, which here is `prost::alloc`. It is tracked in rust-lang/rust#158686 and triggered by rust-lang/rust#158569. It reproduces on `main` without this change (the base commit ICEs identically), and every other check passes. The job should clear on a nightly that carries the fix, or on a re-run once it lands.

DaniPopes (approved):

PR #4058, opened 2026-07-03
fix(consensus): reject string withdrawals in block RLP

0xalpharush:
## Motivation

`BlockBody` and the internal block RLP helper accepted an explicitly present RLP
string as the optional trailing withdrawals field. That allowed non-canonical
bytes such as `c3 c0 c0 80` to decode as the same body as the canonical omitted
withdrawals form.

## Solution

- Add regressions for omitted withdrawals, present empty-list withdrawals, and
  present string withdrawals.
- Use `#[rlp(trailing(no_gaps))]` for `BlockBody` and the block RLP helper types
  so a present withdrawals field must decode as a withdrawals list.

## Tests

- `cargo test -p alloy-consensus block_body_rejects_present_string_withdrawals --lib`
- `cargo test -p alloy-consensus block_decoders_reject_present_string_withdrawals --lib`

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,7 +80,7 @@ alloy-json-abi = { version = "1.6.0", default-features = false }
 alloy-primitives = { version = "1.6.0", default-features = false }
 alloy-sol-types = { version = "1.6.0", default-features = false }
 
-alloy-rlp = { version = "0.3.9", default-features = false }
+alloy-rlp = { version = "0.3.14", default-features = false }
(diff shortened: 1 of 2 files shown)


DaniPopes (approved):

mattsse (approved):
updated the min rlp dep, because no_gaps wasnt present in 0.3.9

PR #4061, opened 2026-07-04
fix(eips): forward eip7928 borsh feature

ArshLabs:
## Motivation

`alloy-eip7928` exposes a `borsh` feature, but the aggregate `alloy-eips` crate did not forward it from its own `borsh` feature. This meant EIP-7928 types re-exported through `alloy-eips` did not get borsh impls when users enabled `alloy-eips/borsh`.

## Solution

Forward `alloy-eip7928/borsh` from the aggregate `borsh` feature and add a compile assertion covering `Bal`.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

PR #4062, opened 2026-07-04
fix(eips): forward eip7928 borsh feature

DaniPopes:
Supersedes #4061.

This keeps the same feature forwarding fix but drops the extra test module from the original PR, so the change is limited to forwarding alloy-eips's borsh feature to alloy-eip7928.


mattsse (approved):

DaniPopes:
Merged in #4063

PR #4073, opened 2026-07-10
feat(network): add AnyRpcBlock header conversion

mattsse:
## Summary

- add `AnyRpcBlock::into_header_with_other`
- preserve outer block fields when consuming an RPC block into its header
- cover the conversion with a regression test for an additional `timestampMillis` field

## Motivation

RPC implementations such as [Foundry's header endpoints](https://github.com/foundry-rs/foundry/pull/15681) need to retain network-specific block fields when returning a header. Providing this conversion on `AnyRpcBlock` centralizes that behavior and avoids repeating the wrapper destructuring at each call site.

### DIFF
--- a/crates/network/src/any/mod.rs
+++ b/crates/network/src/any/mod.rs
@@ -102,6 +102,12 @@ impl AnyRpcBlock {
         self.0.into_inner()
     }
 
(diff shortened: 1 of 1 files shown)


mablr (commented):
so fast 😎, makes sense

PR #4075, opened 2026-07-10
fix(provider): skip cache for requireCanonical block hash queries

### LINKED ISSUE #4069: [Bug] CacheLayer returns stale success for requireCanonical block hashes after a reorg
### Component

provider, pubsub

Alloy v2.1.1 (bce5f50d63bb7d6d9a3b5b455a9a4508bf9dcdc4)

### Operating System

Windows

### Describe the bug

`CacheLayer` caches EIP-1898 `BlockId::hash_canonical(hash)` responses. After that hash is reorged out, an uncached provider returns an RPC error, but the cached provider returns the prior successful value without consulting the backend.

Reproduction against a live Anvil node:

1. Send a transfer of `456` wei and obtain its block hash.
2. Query `get_balance(recipient).block_id(BlockId::hash_canonical(block_hash))` through `CacheLayer`; it returns and caches `456`.
3. Reorg that block with `anvil_reorg`.
4. Repeat the query. The uncached provider errors; the cached provider succeeds with `456`.

EIP-1898 specifies that a `requireCanonical: true` query should raise an RPC error when the block is no longer canonical: https://eips.ethereum.org/EIPS/eip-1898

Expected: a cached call must not satisfy `requireCanonical: true` after the hash becomes non-canonical.

Actual: the cache converts the backend error into stale successful state.

The same cache behavior also returns stale state for numeric block selectors after a reorg.

Osraka:
Fixes #4069

## Summary
- Treat `BlockId::hash_canonical(...)` requests as non-cacheable in `CacheLayer`.
- Keep the existing cache behavior for plain block hashes and numeric block ids.
- Add a regression test that confirms a second `requireCanonical: true` request observes the backend error instead of returning a stale cached success.

## Why
`requireCanonical: true` is a validity check against the current canonical chain. A previously successful response can become invalid after a reorg, so serving it from cache can hide the backend error required by EIP-1898.

## Tests
- `cargo test -p alloy-provider test_hash_canonical_block_ids_do_not_use_cache -- --nocapture`
- `cargo test -p alloy-provider test_get_transaction_by_hash -- --nocapture`
- `cargo test -p alloy-provider test_get_raw_transaction_by_hash -- --nocapture`
- `cargo test -p alloy-provider test_get_transaction_receipt -- --nocapture`
- `cargo +nightly fmt --check`
- `cargo fmt --check`

Note: I also tried `test_get_balance` and `test_get_transaction_count`, but this local environment does not have the `anvil` binary available, so those fail at process spawn before exercising this change.


mattsse (approved):
Reviewed locally against the current head; the provider cache test module and formatting checks pass.

Issue #4082, opened 2026-07-15
[Bug] pubsub: duplicate subscribes leak server-side subscriptions; subscription lifecycle never unsubscribes upstream


cyborg42:
**Component**: provider, pubsub
**Version**: 2.1.1 (relevant code is identical on `main` @ `ffaf1da70ef3bc2f85b792257b67fce7b49999ca`)
**Operating System**: macOS (Apple Silicon)

## Summary

`alloy-pubsub`'s subscription management leaks server-side subscriptions in several independent ways. On any long-lived WS connection the server-side subscription count grows monotonically until the provider's per-connection subscription quota is exhausted (we hit this in production against a provider with a 1024-subscriptions-per-connection limit; the quota is just the observable symptom — the leaks affect any long-lived pubsub connection).

There is one deterministic resource-leak bug (1 below), plus several related lifecycle defects found while auditing the same code paths (2–6). The complete implementation is now consolidated in #4084 so the wire-level fixes and receiver lifecycle semantics can be reviewed and accepted as one PR. Item 3 currently follows Option A; it can be adjusted within that same PR if maintainers prefer Option B.

## 1. Duplicate subscribes create orphan server subscriptions (deterministic leak)

Dedup happens *after* the wire request. Every `eth_subscribe` call is dispatched to the server first; only when the response arrives does `SubscriptionManager::upsert` (`crates/pubsub/src/managers/sub.rs`) check whether a subscription with the same params already exists — and if it does, it **overwrites** `local_to_server` with the new server id. The previous server id is never unsubscribed and no local index can reach it anymore.

Consequences:

- Two sequential identical `subscribe_logs(...)` calls → server has 2 active subscriptions, Alloy has 1 entry (last server id only).
- N concurrent identical calls → up to N−1 orphans.
- Orphan notifications are silently dropped locally (`local_id_for` misses), but the server keeps computing and pushing them.
- A later explicit `unsubscribe(local_id)` only cleans up the last server id; orphans are permanently unreachable.
- Orphans are *not* re-subscribed after reconnect (the entry is shared), but see item 4 for what reconnect does leak.

### Minimal reproducer (sketch)

Run a local `jsonrpsee` WS server that implements `eth_subscribe`/`eth_unsubscribe` and records active subscription ids, then:

```rust
let provider = ProviderBuilder::new().connect_ws(ws_url).await?;
let filter = Filter::new().address(addr);

let sub1 = provider.subscribe_logs(&filter).await?;
[...]

PR #4083, opened 2026-07-16
fix(pubsub): deduplicate subscriptions and track cleanup

cyborg42:
## Summary

- derive subscription local IDs from the full RPC method plus exact serialized params and single-flight matching requests before wire dispatch
- preserve every caller's JSON-RPC response ID while sharing one upstream subscription, and compensate cancelled or duplicate successful responses without overwriting a live server ID
- track unsubscribe requests with reserved, unique IDs and add `unsubscribe_and_wait` without changing the existing enqueue-only `unsubscribe` behavior
- carry channel capacity and custom cleanup method as per-request metadata, preserving first-creator-wins semantics for matching local IDs
- make reconnect routing epoch-aware, retain effective channel capacity, avoid replaying fully cancelled starts, and settle old cleanup requests on connection close

## Motivation

The previous `SubscriptionManager::upsert` deduplicated only after every subscribe request had already reached the server. Matching sequential or concurrent requests therefore created unreachable server-side subscriptions. Subscription identity also omitted the RPC method, and the fixed numeric unsubscribe request ID could hijack an unrelated in-flight response.

This change fixes those deterministic lifecycle bugs without changing receiver-Drop retention semantics; that behavior remains deferred to PR B pending the decision in #4082.

## Identity and route retention

The service computes one fixed-size `B256` local ID from an unambiguous encoding of the RPC method, params presence, and exact serialized params. Service maps and response routes store only that local ID; large params remain only in the serialized request retained for replay.

Subscribe response routes remain until their connection closes so a duplicate or late successful response can still be associated with its cleanup method and compensated. This trades a small amount of connection-lifetime route state for stronger orphan protection; Active reuse does not create additional routes.

## Custom cleanup metadata

The existing helpers now carry the protocol methods verified against their implementations:

- `admin_peerEvents` -> `admin_unsubscribe` (geth's namespace unsubscribe routing)
- `reth_subscribeChainNotifications` -> `reth_unsubscribeChainNotifications`
- `reth_subscribePersistedBlock` -> `reth_unsubscribePersistedBlock`

Generic custom subscriptions do not guess a cleanup method and do not fall back to `eth_unsubscribe`. If none is configured, the entry is marked as connection-owned and cleanup completes only when that connection closes.

References:

- https://github.com/ethereum/go-ethereum/blob/06b23b4293950d8c08b624b90f310d1e918048e8/rpc/json.go
- https://github.com/ethereum/go-ethereum/blob/06b23b4293950d8c08b624b90f310d1e918048e8/rpc/handler.go
- https://github.com/paradigmxyz/reth/blob/4ed6acaffdf0823cfef0f512fc7319b332710c3d/crates/rpc/rpc-api/src/reth.rs

## Compatibility

- `unsubscribe` remains enqueue-only; `unsubscribe_and_wait` is additive.
- Receiver-Drop retention and automatic cleanup are intentionally unchanged.
- Local aliases now hash the method plus exact params, so aliases produced by the old params-only scheme change.
- Unknown custom namespaces require an explicit cleanup method; there is no unsafe naming fallback.

## Validation

- `cargo +1.94.1 test -p alloy-json-rpc -p alloy-pubsub`
- `cargo +1.94.1 test -p alloy-rpc-client --features pubsub,ws`
- `cargo +1.94.1 test -p alloy-provider --features pubsub,ws --lib -- --skip provider::r#trait::tests::test_fill_transaction`
- `cargo +1.94.1 test -p alloy-provider --features pubsub,ws --test it`
- `cargo +1.94.1 test -p alloy-provider --all-features --doc`
- `cargo +1.94.1 clippy -p alloy-pubsub -p alloy-rpc-client -p alloy-provider --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo +1.94.1 doc -p alloy-pubsub -p alloy-rpc-client -p alloy-provider --all-features --no-deps`

The one skipped provider unit test uses a local test node that returns JSON-RPC `-32601 Method not found` for its transaction method; the remaining provider run passed 130 tests (4 ignored), and all 5 WS integration tests passed.

Ref #4082

### DIFF
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -394,6 +394,31 @@ impl SerializedRequest {
         }
     }
 
+    /// Clone this request with a different JSON-RPC request ID.
+    ///
+    /// The serialized params are preserved verbatim, including the distinction between omitted
+    /// params, `null`, and an empty collection.
+    pub fn with_id(&self, id: Id) -> serde_json::Result<Self> {
+        #[derive(Serialize)]
+        struct RequestWithId<'a> {
+            method: &'a str,
+            #[serde(skip_serializing_if = "Option::is_none")]
+            params: Option<&'a RawValue>,
+            id: &'a Id,
+            jsonrpc: &'static str,
+        }
+
+        let request = serde_json::value::to_raw_value(&RequestWithId {
+            method: self.method(),
+            params: self.params_with_presence(),
+            id: &id,
+            jsonrpc: "2.0",
+        })?;
+        let mut meta = self.meta.clone();
+        meta.id = id;
+        Ok(Self { meta, request })
+    }
+
     /// Mark the request as a non-standard subscription (i.e. not
     /// `eth_subscribe`)
     pub const fn set_is_subscription(&mut self) {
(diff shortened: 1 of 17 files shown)


cyborg42:
Closing in favor of #4084, which now contains the complete wire-level and receiver-lifecycle fix as a single PR. #4082 and #4084 have been updated accordingly.

PR #4084, opened 2026-07-16
fix(pubsub): prevent subscription lifecycle leaks

### LINKED ISSUE #4082: [Bug] pubsub: duplicate subscribes leak server-side subscriptions; subscription lifecycle never unsubscribes upstream

cyborg42:
## Summary

This is the complete implementation for #4082 and supersedes #4083. It intentionally combines the wire-level correctness fixes and receiver lifecycle semantics into one review and merge unit.

- deduplicate matching subscriptions before wire dispatch while preserving every caller's JSON-RPC response ID;
- derive local subscription identity from the RPC method, params presence, and exact serialized params;
- compensate unexpected, duplicate, late, and cancelled successful subscribe responses so they cannot leave orphan server subscriptions;
- track cleanup requests with collision-safe service IDs and add `unsubscribe_and_wait` without changing existing `unsubscribe` enqueue semantics;
- carry channel capacity and verified custom cleanup methods as per-request metadata;
- model typed receiver ownership explicitly, clean receiver-scoped subscriptions after the last receiver is dropped, and reconnect only entries that remain locally owned.

## Retention semantics

This draft implements **Option A** proposed in #4082:

- typed `Provider::subscribe_*` builders default to `WhileReceivers`;
- the low-level two-phase/raw path keeps the compatible `UntilExplicitUnsubscribe` default;
- both paths can override the policy per request.

A successfully delivered `UntilExplicitUnsubscribe` waiter commits the shared persistent hold, independent of waiter arrival order. A later `get_subscription(local_id)` does not upgrade receiver-scoped retention.

## Wire lifecycle correctness

Subscription creation is single-flighted before dispatch. Matching sequential or concurrent callers share one upstream subscription, while response fanout restores each caller's original request ID. Cancelled waiters and duplicate or late server responses are compensated without replacing a live server ID.

Cleanup requests use unique service-owned string IDs and their boolean or RPC-error result is tracked. The existing `unsubscribe` API remains enqueue-only; `unsubscribe_and_wait` is additive.

Reconnect routing is epoch-aware, preserves channel capacity, skips fully cancelled starts, restores only owned entries, and compensates a late resubscribe response if ownership disappears.

## Receiver ownership

Typed subscription ownership is delivered through an ephemeral one-shot receiver ticket in the same service turn as the subscribe response. The manager's hidden initial receiver is removed; legacy delivery is represented by explicit manual claims and a persistent hold.

Receiver-scoped subscriptions are checked at notification, acquire, reconnect, and periodic 30-second sweep checkpoints. A zero-receiver generation cannot be reopened by `get_subscription` or an identical subscribe request.

## Identity and route retention

The service computes one fixed-size `B256` local ID from an unambiguous encoding of the RPC method, params presence, and exact serialized params. Service maps and response routes store only that local ID; large params remain only in the serialized request retained for replay.

Subscribe response routes remain until their connection closes so duplicate or late successful responses can still be associated with their cleanup method and compensated. This trades a small amount of connection-lifetime route state for stronger orphan protection; active reuse does not create additional routes.

## Custom cleanup metadata

The existing helpers carry protocol methods verified against their implementations:

- `eth_subscribe` -> `eth_unsubscribe`
- `admin_peerEvents` -> `admin_unsubscribe`
- `reth_subscribeChainNotifications` -> `reth_unsubscribeChainNotifications`
- `reth_subscribePersistedBlock` -> `reth_unsubscribePersistedBlock`

Generic custom subscriptions do not guess a cleanup method and do not fall back to `eth_unsubscribe`. If none is configured, the entry is connection-owned and cleanup completes only when that connection closes.

References:

- https://github.com/ethereum/go-ethereum/blob/06b23b4293950d8c08b624b90f310d1e918048e8/rpc/json.go
- https://github.com/ethereum/go-ethereum/blob/06b23b4293950d8c08b624b90f310d1e918048e8/rpc/handler.go
- https://github.com/paradigmxyz/reth/blob/4ed6acaffdf0823cfef0f512fc7319b332710c3d/crates/rpc/rpc-api/src/reth.rs

## Compatibility

- Existing low-level/manual subscriptions remain persistent by default and have no claim deadline.
- Typed builders adopt Option A's receiver-scoped default.
- Force unsubscribe still closes every local receiver for the shared entry.
- Local aliases now include the method and exact params, so aliases produced by the old params-only scheme change.
- Unknown custom namespaces require explicit cleanup metadata; there is no naming fallback.
- Existing `unsubscribe` behavior is unchanged.

## Validation

- `cargo +1.94.1 test -p alloy-json-rpc -p alloy-pubsub`
- `cargo +1.94.1 test -p alloy-rpc-client --features pubsub,ws`
- `cargo +1.94.1 test -p alloy-provider --features pubsub,ws --lib -- --skip provider::r#trait::tests::test_fill_transaction`
- `cargo +1.94.1 test -p alloy-provider --features pubsub,ws --test it`
- `cargo +1.94.1 test -p alloy-provider --all-features --doc`
- `cargo +1.94.1 clippy -p alloy-pubsub -p alloy-rpc-client -p alloy-provider --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' cargo +1.94.1 doc -p alloy-pubsub -p alloy-rpc-client -p alloy-provider --all-features --no-deps`
- `cargo +1.94.1 check -p alloy-pubsub --target wasm32-unknown-unknown`

The PR conversation also records the downstream Base node 1,024-subscription cap reproduction and validates same-key sharing plus complete slot reclamation after the 30-second sweep.

Supersedes #4083.

Fixes #4082.

### DIFF
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -394,6 +394,31 @@ impl SerializedRequest {
         }
     }
 
+    /// Clone this request with a different JSON-RPC request ID.
+    ///
+    /// The serialized params are preserved verbatim, including the distinction between omitted
+    /// params, `null`, and an empty collection.
+    pub fn with_id(&self, id: Id) -> serde_json::Result<Self> {
+        #[derive(Serialize)]
+        struct RequestWithId<'a> {
+            method: &'a str,
+            #[serde(skip_serializing_if = "Option::is_none")]
+            params: Option<&'a RawValue>,
+            id: &'a Id,
+            jsonrpc: &'static str,
+        }
+
+        let request = serde_json::value::to_raw_value(&RequestWithId {
+            method: self.method(),
+            params: self.params_with_presence(),
+            id: &id,
+            jsonrpc: "2.0",
+        })?;
+        let mut meta = self.meta.clone();
+        meta.id = id;
+        Ok(Self { meta, request })
+    }
+
     /// Mark the request as a non-standard subscription (i.e. not
     /// `eth_subscribe`)
     pub const fn set_is_subscription(&mut self) {
@@ -441,6 +466,40 @@ impl SerializedRequest {
         req.params
     }
 
+    /// Get the exact serialized params when the `params` field is present.
+    ///
+    /// Unlike [`Self::params`], this distinguishes an omitted `params` field from an explicitly
+    /// serialized `null` value.
+    pub fn params_with_presence(&self) -> Option<&RawValue> {
+        #[derive(Default)]
+        enum ParamsField<'a> {
+            #[default]
+            Missing,
+            Present(&'a RawValue),
+        }
+
+        impl<'de: 'a, 'a> Deserialize<'de> for ParamsField<'a> {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                <&'de RawValue>::deserialize(deserializer).map(Self::Present)
+            }
+        }
+
+        #[derive(Deserialize)]
+        struct Req<'a> {
+            #[serde(borrow, default)]
+            params: ParamsField<'a>,
+        }
+
+        let req: Req<'_> = serde_json::from_str(self.request.get()).unwrap();
+        match req.params {
+            ParamsField::Missing => None,
+            ParamsField::Present(params) => Some(params),
+        }
+    }
+
     /// Get the hash of the serialized request's params.
     ///
     /// This partially deserializes the request, and should be avoided if
@@ -478,5 +537,21 @@ mod test {
         test_inner(Request::<u64>::new("test", "hello".to_string().into(), 1));
         test_inner(Request::<String>::new("test", Id::None, "test".to_string()));
         test_inner(Request::<Vec<u64>>::new("test", u64::MAX.into(), vec![1, 2, 3]));
+    }
+
+    #[test]
+    fn with_id_preserves_exact_params_presence_and_value() {
+        let omitted = Request::new("test", Id::Number(1), ()).serialize().unwrap();
+        let null =
+            Request::new("test", Id::Number(1), serde_json::Value::Null).serialize().unwrap();
+        let empty = Request::new("test", Id::Number(1), Vec::<u8>::new()).serialize().unwrap();
+
+        let omitted = omitted.with_id(Id::String("next".into())).unwrap();
+        let null = null.with_id(Id::String("next".into())).unwrap();
+        let empty = empty.with_id(Id::String("next".into())).unwrap();
+
+        assert!(omitted.params_with_presence().is_none());
+        assert_eq!(null.params_with_presence().unwrap().get(), "null");
+        assert_eq!(empty.params_with_presence().unwrap().get(), "[]");
     }
 }
--- a/crates/json-rpc/src/request.rs
+++ b/crates/json-rpc/src/request.rs
@@ -403,6 +403,31 @@ impl SerializedRequest {
         }
     }
 
+    /// Clone this request with a different JSON-RPC request ID.
+    ///
+    /// The serialized params are preserved verbatim, including the distinction between omitted
+    /// params, `null`, and an empty collection.
+    pub fn with_id(&self, id: Id) -> serde_json::Result<Self> {
+        #[derive(Serialize)]
+        struct RequestWithId<'a> {
+            method: &'a str,
+            #[serde(skip_serializing_if = "Option::is_none")]
+            params: Option<&'a RawValue>,
+            id: &'a Id,
+            jsonrpc: &'static str,
+        }
+
+        let request = serde_json::value::to_raw_value(&RequestWithId {
+            method: self.method(),
+            params: self.params_with_presence(),
+            id: &id,
+            jsonrpc: "2.0",
+        })?;
+        let mut meta = self.meta.clone();
+        meta.id = id;
+        Ok(Self { meta, request })
+    }
+
     /// Mark the request as a non-standard subscription (i.e. not
     /// `eth_subscribe`)
     pub const fn set_is_subscription(&mut self) {
@@ -450,6 +475,40 @@ impl SerializedRequest {
         req.params
     }
 
+    /// Get the exact serialized params when the `params` field is present.
+    ///
+    /// Unlike [`Self::params`], this distinguishes an omitted `params` field from an explicitly
+    /// serialized `null` value.
+    pub fn params_with_presence(&self) -> Option<&RawValue> {
+        #[derive(Default)]
+        enum ParamsField<'a> {
+            #[default]
+            Missing,
+            Present(&'a RawValue),
+        }
+
+        impl<'de: 'a, 'a> Deserialize<'de> for ParamsField<'a> {
+            fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+            where
+                D: serde::Deserializer<'de>,
+            {
+                <&'de RawValue>::deserialize(deserializer).map(Self::Present)
+            }
+        }
+
+        #[derive(Deserialize)]
+        struct Req<'a> {
+            #[serde(borrow, default)]
+            params: ParamsField<'a>,
+        }
+
+        let req: Req<'_> = serde_json::from_str(self.request.get()).unwrap();
+        match req.params {
+            ParamsField::Missing => None,
+            ParamsField::Present(params) => Some(params),
+        }
+    }
+
     /// Get the hash of the serialized request's params.
     ///
     /// This partially deserializes the request, and should be avoided if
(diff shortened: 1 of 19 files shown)


cyborg42:
### Downstream Base node subscription-cap validation

Tested commit `0dd01276f5b5b7194aa155382e06dd8b53868c0f` from this PR through an Alloy typed-provider integration test against a self-hosted Base node with a hard limit of 1,024 subscriptions per WebSocket connection.

The test used a fresh, independent WS connection:

1. Created and retained 1,025 `Provider::subscribe_logs` subscriptions with the same `Filter`. All succeeded and returned the same Alloy local subscription ID.
2. Created 1,023 subscriptions with unique filters, bringing the total upstream subscriptions to exactly 1,024.
3. The next unique subscription was rejected by the node as expected:
   `error code -32006: Too many subscriptions on the connection, data: "Exceeded max limit of 1024"`
4. Dropped all typed `Subscription` handles and waited 35 seconds, crossing the 30-second subscription sweep interval.
5. Created 1,024 subscriptions with an entirely new generation of unique filters. All succeeded.

Observed test output:

```text
same-key requests=1025 upstream subscriptions=1024 overflow=server returned an error response: error code -32006: Too many subscriptions on the connection, data: "Exceeded max limit of 1024"
reclaimed subscriptions=1024 after waiting 35s
test result: ok. 1 passed; 0 failed; finished in 36.56s
```

The filters used unique addresses with no expected log traffic. This validates both same-key upstream subscription sharing and complete automatic slot reclamation after the last typed receivers are dropped.

cyborg42:
### Baseline comparison: crates.io Alloy 2.1.1

I repeated the same Base node hard-cap probe with an isolated temporary crate pinned to crates.io `alloy = "=2.1.1"`.

On a fresh WebSocket connection:

1. Created 1,024 typed `Provider::subscribe_logs` subscriptions with an identical `Filter`. Every call succeeded and exposed the same Alloy local subscription ID.
2. The 1,025th request with that same identical filter was rejected by the node:
   `error code -32006: Too many subscriptions on the connection, data: "Exceeded max limit of 1024"`
3. Dropped all 1,024 typed `Subscription` handles and waited 35 seconds.
4. The first subscription with a new filter was still rejected with the same `-32006` response.

Observed output:

```text
alloy=2.1.1 same-key successes=1024 node subscriptions=1024 identical-overflow=server returned an error response: error code -32006: Too many subscriptions on the connection, data: "Exceeded max limit of 1024"
REGRESSION_REPRODUCED reclaimed=0 unreclaimed=1024 after 35s error=server returned an error response: error code -32006: Too many subscriptions on the connection, data: "Exceeded max limit of 1024"
```

The full baseline probe was run twice on separate fresh connections and produced the same result both times.

So, in the 2.1.1 baseline, identical typed subscriptions expose the same local ID but still consume one upstream node slot per request, and dropping all typed receivers does not reclaim any of those slots. In contrast, commit `0dd01276f5b5b7194aa155382e06dd8b53868c0f` shares the upstream subscription and reclaimed all 1,024 slots after the 30-second sweep.

PR #4089, opened 2026-07-20
fix(tx-macros): restrict TransactionEnvelope fallback decode to legacy (EIP-2718)

chfast:
**Issue.** The `#[derive(TransactionEnvelope)]`-generated `Decodable2718::fallback_decode` tries *every* transaction variant against un-typed input (no EIP-2718 type byte, i.e. an RLP list) and returns the first that parses. Per EIP-2718 an untagged byte string can only be a legacy transaction, so input that fails the legacy schema must be rejected — but a byte string that is invalid as legacy yet happens to be a valid field list of another type (e.g. EIP-1559) is silently mis-decoded as that type instead of erroring. Found via differential fuzzing across six EVM implementations: only alloy accepts the sample input, all others reject it. (Decoder-conformance bug, not a consensus split — reth independently cross-checks the transactions-root and rejects such a block regardless — but the decoder should be correct on its own; consumers without that cross-check would accept a non-conformant transaction.)

**Fix.** Restrict the untagged/fallback decode path to the legacy variant only — plus flattened variants, which delegate to an inner envelope that may itself carry the legacy variant — mirroring the crate's existing serde untagged-path logic. Includes regression tests for both the mis-decode rejection and the flatten carve-out.

Related: #2585 (introduced the derive), #4059 (fixed a narrower manifestation of the same fallback loop).

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1473,6 +1473,77 @@ mod tests {
         assert_eq!(buf, raw.as_slice());
     }
 
+    #[test]
+    fn untagged_fallback_decode_only_accepts_legacy() {
+        // Per EIP-2718, a byte string with no type-byte prefix (first byte >= 0xc0, i.e. it
+        // is an RLP list) can only ever represent a legacy transaction; if it fails the legacy
+        // schema, decoding must fail outright -- it must never be reinterpreted as some other
+        // (typed) transaction kind.
+        //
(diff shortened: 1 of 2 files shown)


chfast:
Test cases and the fix are AI generated, so feel free to modify/replace them as you wish.

klkvr:
we've addressed the issue a bit differently in #4090 and will be merging the tests you've suggested in https://github.com/alloy-rs/alloy/pull/4091, ty for reporting and fixing 🫡

PR #4097, opened 2026-07-24
feat(transport): expose retry count to backoff hint

### LINKED ISSUE #4066: [Feature] Expose retry attempt count to RetryPolicy::backoff_hint
### Component

transports

 RetryPolicy::backoff_hint only receives the error:

 ` fn backoff_hint(&self, error: &TransportError) -> Option<Duration>;`

  When it returns None, RetryBackoffService falls back to a constant initial_backoff on every attempt — there is no exponential growth:

```
  let backoff_hint = this.policy.backoff_hint(&err);
  let next_backoff = backoff_hint.unwrap_or_else(|| this.initial_backoff());
```

  Without access to the current attempt number, a custom policy cannot implement standard exponential backoff, e.g.:

`  next_backoff = initial_backoff * (1 + backoff_coefficient)^(num_retries - 1)`

  Proposal: pass the attempt count to the policy, e.g.

`  fn backoff_hint(&self, error: &TransportError, num_retries: u32) -> Option<Duration>;`

  (or a context struct to keep it extensible). Alternatively, implement exponential backoff with a max ceiling directly in RetryBackoffService when no hint is
  returned.

### Additional context

Downstream use case: hoprnet/blokli — custom retry policy with configurable initial_backoff, backoff_coefficient, max_backoff, currently unable to apply
  them (hoprnet/hoprnet#7140).

Soubhik-10:
closes #4066


crates/transport/src/layers/retry.rs
```diff
@@ -120,8 +120,11 @@ pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
     /// Whether to retry the request based on the given `error`
     fn should_retry(&self, error: &TransportError) -> bool;
 
-    /// Providers may include the `backoff` in the error response directly
-    fn backoff_hint(&self, error: &TransportError) -> Option<std::time::Duration>;
+    /// Providers may include the `backoff` in the error response directly.
+    ///
+    /// `num_retries` is the one-based number of the retry that is about to be attempted.
+    fn backoff_hint(&self, error: &TransportError, num_retries: u32)
+        -> Option<std::time::Duration>;
```
mattsse, line 127: Changing the signature of this required public trait method breaks every downstream `RetryPolicy` implementation. Please keep the existing method and add a provided count-aware entry point so existing implementations continue to compile. `RetryBackoffService` should call the new method, and `OrRetryPolicyFn` should forward it to its base policy.

```suggestion
    /// Providers may include the `backoff` in the error response directly.
    fn backoff_hint(&self, error: &TransportError) -> Option<std::time::Duration>;

    /// Provides a backoff hint for the retry that is about to be attempted.
    ///
    /// `num_retries` is the one-based retry number.
    fn backoff_hint_with_retry_count(
        &self,
        error: &TransportError,
        num_retries: u32,
    ) -> Option<std::time::Duration> {
        let _ = num_retries;
        self.backoff_hint(error)
    }
```

mattsse (changes requested):
The retry number is useful and the one-based semantics look correct, but changing a required method on this public trait is source-breaking for every downstream `RetryPolicy` implementation. Please preserve the existing method and introduce a defaulted count-aware method for the service to call.

PR #4103, opened 2026-07-28
fix(transport-http): honor Retry-After

mattsse:
## Motivation

The HTTP transports discard response headers before constructing errors. Providers such as Robinhood Chain return Cloudflare-backed 429 responses with a `Retry-After` header, so the retry layer ignores the server's reset window and applies its short fallback backoff instead.

## Solution

Read the `Retry-After` header (delay-seconds form only) in the reqwest and hyper transports and carry it on a new `TransportErrorKind::HttpErrorWithRetryAfter` variant wrapping the existing `HttpError`, which `backoff_hint` and therefore the retry layer now honor. The HTTP-date form is intentionally not parsed to avoid a date-parsing dependency; such values simply fall back to the default backoff.

A JSON-RPC error response in the body still takes precedence and is returned as `ErrorResp` exactly as before; the header is only attached when the body is not a JSON-RPC error. Since a server sending `Retry-After` explicitly requests a retry, the new variant is always considered retryable for error statuses (4xx/5xx); redirects are excluded because their `Retry-After` delays the redirected request rather than a replay of the original URL. Honored hints are clamped to 5 minutes so a misbehaving server cannot stall the retry loop, which also bounds the pre-existing infura-style payload hints, and the compute-budget offset is added with saturating arithmetic to rule out `Duration` overflow. A failed body read on an error response now preserves the captured status and `Retry-After` instead of degrading to a `Custom` error, so e.g. a truncated 429 is still retried. The accessors (`is_http_error`, `as_http_error`) treat the new variant like a plain HTTP error, so existing status/body handling keeps working.

### DIFF
--- a/crates/transport-http/src/reqwest_transport.rs
+++ b/crates/transport-http/src/reqwest_transport.rs
@@ -5,7 +5,7 @@ use alloy_transport::{
     TransportFut, TransportResult,
 };
 use itertools::Itertools;
-use std::task;
+use std::{task, time::Duration};
 use tower::Service;
 use tracing::{debug, debug_span, instrument, trace, Instrument};
 use url::Url;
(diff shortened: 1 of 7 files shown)


crates/transport-http/src/reqwest_transport.rs
```diff
@@ -65,9 +72,10 @@ impl Http<Client> {
                 return Ok(response);
             }
 
-            return Err(TransportErrorKind::http_error(
+            return Err(TransportErrorKind::http_error_with_retry_after(
```
stevencartavia: **Non-blocking:** Should `Retry-After` take precedence over a valid JSON-RPC error body?

Both transports return the parsed JSON-RPC error before constructing `HttpErrorWithRetryAfter`, so the retry layer loses the header and may use its fallback delay instead.

This was reproduced locally with both reqwest and HyperTransport, but the response shape was not observed from the public RPCs tested. Is this precedence intentional?

PR #4107, opened 2026-07-30
fix(consensus): default missing receipt type to legacy

### LINKED ISSUE #4106: [Bug] Receipt deserialization fails when a node omits the type field.
### Component

consensus, eips, genesis, rpc

alloy-consensus v2.2.0  alloy-rpc-types-eth v2.2.0

### Operating System

macOS (Apple Silicon)

### Describe the bug

`ReceiptEnvelope` is deserialized as an internally tagged enum:

```rust
#[cfg_attr(feature = "serde", serde(tag = "type"))]
pub enum ReceiptEnvelope<T = Log> {
```

Serde has no way to default a missing tag, so any receipt JSON without a `type` field fails to deserialize. Several Geth-derived chains never emit the field - their `marshalReceipt` builds the response map without it, and it cannot be added without a node upgrade:

```go
fields := map[string]interface{}{
    "blockHash":         blockHash,
    "blockNumber":       hexutil.Uint64(blockNumber),
    "transactionHash":   tx.Hash(),
    "transactionIndex":  hexutil.Uint64(txIndex),
    "from":              from,
    "to":                tx.To(),
    "gasUsed":           hexutil.Uint64(receipt.GasUsed),
    "cumulativeGasUsed": hexutil.Uint64(receipt.CumulativeGasUsed),
    "contractAddress":   nil,
    "logs":              receipt.Logs,
    "logsBloom":         receipt.Bloom,
}
```

This surfaces in Foundry as a hard failure on `forge script --broadcast`, since the deployment can never be confirmed:

```
Error: Failure on receiving a receipt for 0x13c2e2de4e886256727ac69efd52d066fae21c37d4e458f009e47a033d09e252:
deserialization error: missing field type at line 1 column 949
```

Both `eth_getTransactionReceipt` and `eth_getBlockReceipts` are affected.

A receipt with no type flag predates [EIP-2718] and is unambiguously legacy, so the absence of the field is not actually ambiguous - it can be defaulted to `TxType::Legacy` rather than rejected.

There is direct precedent for this leniency in the same struct. `effectiveGasPrice` is also required by the JSON-RPC spec, and alloy already tolerates its absence (`crates/rpc-types-eth/src/transaction/receipt.rs:97-124`, from #2019):

```rust
// Optional fields for gas prices
// 1. Use effectiveGasPrice if present
// 2. Fallback to use gasPrice
// 3. Default to 0 if neither is present
#[serde(default, alias = "gasPrice", with = "alloy_serde::quantity::opt")]
effective_gas_price: Option<u128>,
```

Unlike `effectiveGasPrice`, whose fallback to `0` is lossy, defaulting a missing `type` to legacy recovers the correct value.

### Reproduction

```rust
let json = r#"{"blockHash":"0x18cf669d61484fafd3fd01ca88c0b6706332f444f90de28d04fe84c6a922c5d2","blockNumber":"0x6abbdf1","contractAddress":"0xd937e6195ddf60b91de26a27c59492bbb48690fa","cumulativeGasUsed":"0xc3b68","from":"0xeff507ead35c808439741f5d1b228918c00a4329","gasUsed":"0xc3b68","logs":[],"logsBloom":"0x0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","status":"0x1","to":null,"transactionHash":"0x13c2e2de4e886256727ac69efd52d066fae21c37d4e458f009e47a033d09e252","transactionIndex":"0x0"}"#;

let receipt: TransactionReceipt = serde_json::from_str(json).unwrap();
// panics: missing field `type`
```

### Proposed fix

Replace the derived `Deserialize` on `ReceiptEnvelope` with a manual impl that reads `type` as an optional quantity and falls back to `TxType::Legacy`. `Serialize` stays derived, so output is unchanged. Explicit and unknown type flags keep their current behavior.

Kept generic over the `Deserializer` (no `serde_json::Value` round-trip) so `no_std` still holds.

[EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718

0xendale:
## Motivation

  `ReceiptEnvelope` is deserialized as an internally tagged enum (`#[serde(tag = "type")]`), and
  serde has no way to default a missing tag, so a receipt JSON object without a `type` field is
  rejected with `missing field type`. Several Geth-derived chains never emit the field — their
  `marshalReceipt` builds the response map without it — which makes every receipt from those chains
  undeserializable, on both `eth_getTransactionReceipt` and `eth_getBlockReceipts`.

  It surfaces downstream as a hard failure. In Foundry, `forge script --broadcast` can never confirm
  a deployment against such a chain:

  ```
  Error: Failure on receiving a receipt for 0x13c2e2de…:
  deserialization error: missing field type at line 1 column 949
  ```
  
  Closes #4106

  ## Solution

  Replace the derived `Deserialize` with a manual one that reads `type` as an optional quantity and
  falls back to `TxType::Legacy` when absent. A receipt with no type flag predates EIP-2718 and is
  unambiguously legacy, so it can be accepted rather than rejected. Explicit and unknown type flags
  behave as before.

  `Serialize` stays derived, so serialized output is byte-identical. The helper is generic over the
  `Deserializer` rather than round-tripping through `serde_json::Value`, so `no_std` support is unaffected.

  This mirrors the existing leniency for `effectiveGasPrice`, which is also required by the spec but
  defaults to 0 when a node omits it (#2019). Unlike that fallback, defaulting a missing `type` to
  legacy recovers the correct value rather than a placeholder.

  ## PR Checklist

  - [x] Added Tests
  - [x] Added Documentation
  - [ ] Breaking changes

### DIFF
--- a/crates/consensus/src/receipt/envelope.rs
+++ b/crates/consensus/src/receipt/envelope.rs
@@ -22,7 +22,7 @@ use core::fmt;
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
 #[derive(Clone, Debug, PartialEq, Eq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
 #[cfg_attr(feature = "serde", serde(tag = "type"))]
 #[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
 #[doc(alias = "TransactionReceiptEnvelope", alias = "TxReceiptEnvelope")]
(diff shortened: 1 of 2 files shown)


mattsse (approved):
The compatibility tradeoff is acceptable here, and the focused receipt deserialization tests plus the no-default-features Serde build pass.

PR #4110, opened 2026-08-04
Migrate ENS resolution to Universal Resolver

yashgo0018:
Part 1 of the ENS and CCIP Read stack.

## What changed

- Route ENS forward, reverse, text, and resolver discovery through the ENS Universal Resolver.
- Add corrected ENSIP-23 bindings and ENSIP-11 multicoin resolution.
- Replace `dns_encode` with a single validated, fallible DNS wire-format encoder and retain the `UNIVERSAL_RESOLVER_ADDRESS` name.
- Remove registry dependence from the resolution helpers.
- Reject direct `get_resolver` use when the Universal Resolver reports an extended/wildcard resolver (avoids wrong direct `addr` for names like `ur.integration-tests.eth`).
- Align DNS label length limit with NameCoder/viem (255) and map UR `ResolverNotFound` to `EnsError::ResolverNotFound`.

## Why

This establishes the [ENSv2-ready onchain resolution path](https://docs.ens.domains/web/ensv2-readiness) independently from the generic CCIP Read transport.

## Validation

- `cargo test -p alloy-ens --all-features`
- `cargo clippy -p alloy-ens --all-features --all-targets`
- `cargo check -p alloy-ens --no-default-features`
- `cargo check -p alloy-ens --no-default-features --features contract`

This is the bottom PR in the stack. Follow-ups:
- Part 2: generic CCIP Read in `alloy-provider`
- Part 3: wire ENS helpers through CCIP Read

Supersedes ens-integrations/alloy-2#2.

### DIFF
--- a/crates/ens/src/lib.rs
+++ b/crates/ens/src/lib.rs
@@ -9,7 +9,7 @@
 //! ENS Name resolving utilities.
 
 mod utils;
-pub use utils::{dns_encode, namehash, reverse_address, try_dns_encode, DnsEncodeError};
+pub use utils::{dns_encode, namehash, reverse_address, DnsEncodeError};
 
 use alloy_primitives::{address, Address};
 use std::str::FromStr;
@@ -223,7 +223,7 @@ mod contract {
 #[cfg(feature = "provider")]
 mod provider {
     use crate::{
-        coin_type, namehash, try_dns_encode, EnsError, EnsMulticoinResolver, EnsResolver,
+        coin_type, dns_encode, namehash, EnsError, EnsMulticoinResolver, EnsResolver,
         EnsResolver::EnsResolverInstance, UniversalResolver, UNIVERSAL_RESOLVER_ADDRESS,
     };
     use alloy_primitives::{Address, Bytes, U256};
@@ -274,7 +274,7 @@ mod provider {
         N: Network,
     {
         async fn get_resolver(&self, name: &str) -> Result<EnsResolverInstance<&P, N>, EnsError> {
-            let dns = try_dns_encode(name)?;
+            let dns = dns_encode(name)?;
 
             let ur = UniversalResolver::new(UNIVERSAL_RESOLVER_ADDRESS, self);
             let info = ur
@@ -288,7 +288,7 @@ mod provider {
 
         async fn resolve_name(&self, name: &str) -> Result<Address, EnsError> {
             let node = namehash(name);
-            let dns = try_dns_encode(name)?;
+            let dns = dns_encode(name)?;
             let calldata = EnsResolver::addrCall { node }.abi_encode();
 
             let ur = UniversalResolver::new(UNIVERSAL_RESOLVER_ADDRESS, self);
@@ -307,7 +307,7 @@ mod provider {
             coin_type: u64,
         ) -> Result<Bytes, EnsError> {
             let node = namehash(name);
-            let dns = try_dns_encode(name)?;
+            let dns = dns_encode(name)?;
             let calldata = EnsMulticoinResolver::addrCall {
                 node,
                 coin_type: U256::from(coin_type),
@@ -337,7 +337,7 @@ mod provider {
 
         async fn lookup_txt(&self, name: &str, key: &str) -> Result<String, EnsError> {
             let node = namehash(name);
-            let dns = try_dns_encode(name)?;
+            let dns = dns_encode(name)?;
             let calldata = EnsResolver::textCall { node, key: key.to_string() }.abi_encode();
 
             let ur = UniversalResolver::new(UNIVERSAL_RESOLVER_ADDRESS, self);
--- a/crates/ens/src/lib.rs
+++ b/crates/ens/src/lib.rs
@@ -202,6 +202,16 @@ mod contract {
         /// No resolver found for the given name.
         #[error("ENS resolver not found for name {0:?}")]
         ResolverNotFound(String),
+        /// Direct resolver calls are unsafe for this name; use Universal Resolver helpers.
+        ///
+        /// Returned when the name resolves through an ENSIP-10 extended resolver or a
+        /// parent/wildcard resolver. Calling `addr`/`text`/`name` on the raw resolver
+        /// instance can return incorrect data — use `resolve_name`, `lookup_txt`, or
+        /// related helpers instead.
+        #[error(
+            "ENS name {0:?} requires Universal Resolver resolution (extended or wildcard resolver)"
+        )]
+        RequiresUniversalResolver(String),
         /// Failed to perform a reverse lookup.
         #[error("Failed to lookup ENS name from an address: {0}")]
         Lookup(alloy_contract::Error),
@@ -237,6 +247,12 @@ mod provider {
         /// Returns the resolver contract instance for the given ENS name.
         ///
         /// Determines the resolver address via the Universal Resolver.
+        ///
+        /// Returns [`EnsError::RequiresUniversalResolver`] when the name uses an
+        /// ENSIP-10 extended resolver or a parent/wildcard resolver. In those cases
+        /// direct `addr`/`text`/`name` calls on the resolver are not equivalent to
+        /// Universal Resolver resolution — use [`Self::resolve_name`],
+        /// [`Self::lookup_txt`], or the other helpers instead.
         async fn get_resolver(&self, name: &str) -> Result<EnsResolverInstance<&P, N>, EnsError>;
 
         /// Performs a forward lookup of an ENS name to an Ethereum address.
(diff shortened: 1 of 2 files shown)


crates/ens/src/utils.rs
```diff
@@ -0,0 +1,159 @@
...
+
+/// DNS wire-format encodes an ENS name for use with the Universal Resolver's `resolve()`.
+///
+/// Each label is prefixed with its byte length and the sequence is terminated with `\x00`.
+/// For example, `"foo.eth"` encodes to `[3, 'f', 'o', 'o', 3, 'e', 't', 'h', 0]`.
+///
+/// Returns an error if any label is empty or exceeds 255 bytes. This matches ENS
+/// [`NameCoder`](https://github.com/ensdomains/ens-contracts/blob/staging/contracts/utils/NameCoder.sol),
+/// which uses a full length byte (not the RFC 1035 63-octet DNS message limit).
+pub fn dns_encode(name: &str) -> Result<Vec<u8>, DnsEncodeError> {
```
figtracer: `dns_encode` is already public and returns `Vec<u8>`; changing it to `Result` breaks downstream source compatibility. can we keep that wrapper and expose the checked path as `try_dns_encode`, as `92682bd` did?

crates/ens/src/lib.rs
```diff
@@ -92,109 +110,176 @@ impl FromStr for NameOrAddress {
...
+        /// Returns the resolver contract instance for the given ENS name.
+        ///
+        /// Determines the resolver address via the Universal Resolver.
+        ///
+        /// Returns [`EnsError::RequiresUniversalResolver`] when the name uses an
+        /// ENSIP-10 extended resolver or a parent/wildcard resolver. In those cases
+        /// direct `addr`/`text`/`name` calls on the resolver are not equivalent to
+        /// Universal Resolver resolution — use [`Self::resolve_name`],
+        /// [`Self::lookup_txt`], or the other helpers instead.
+        async fn get_resolver(&self, name: &str) -> Result<EnsResolverInstance<&P, N>, EnsError>;
```
figtracer: this replaces `get_resolver(node, error_name)` and removes `get_reverse_registrar`, `EnsRegistry`, `ReverseRegistrar`, `ENS_ADDRESS`, and `ENS_REVERSE_REGISTRAR_DOMAIN`. can we retain or deprecate those public apis and add a separate name-based universal-resolver helper?

crates/ens/src/lib.rs
```diff
@@ -8,21 +8,39 @@
...
+
+    /// Converts an EVM chain ID to its ENSIP-11 coin type.
+    ///
+    /// Chain ID `1` (Ethereum mainnet) returns [`ETH`] (`60`) per SLIP-0044.
+    /// All other chains use `0x80000000 | chain_id`.
+    pub fn evm_chain(chain_id: u32) -> u64 {
+        if chain_id == 1 {
+            return ETH;
+        }
+        0x8000_0000u64 | u64::from(chain_id)
```
mablr: `0x80000000 | chain_id` is not injective for `u32`: IDs `0` and `0x80000000` both map to `0x80000000`, while ENSIP-11’s reverse mask assumes a 31-bit ID. Can this reject IDs with the high bit set?

crates/ens/src/lib.rs
yashgo0018: added those back with deprecated flag

crates/ens/src/utils.rs
yashgo0018: fixed

crates/ens/src/lib.rs
```diff
@@ -109,129 +148,256 @@ mod contract {
...
+            let info = ur
+                .requireResolver(dns.into())
+                .call()
+                .await
+                .map_err(|error| map_ur_error(error, name, EnsError::Resolver))?;
+
+            // Extended and parent/wildcard resolvers must be queried through the
+            // Universal Resolver. Returning a raw instance invites incorrect direct
+            // `addr`/`text` calls (e.g. ur.integration-tests.eth).
+            if info.extended || !info.offset.is_zero() {
```
mablr: This is the headline safety fix of the PR (rejecting extended/wildcard resolvers), but it's only covered by a single live-mainnet integration test (`test_get_resolver_for_name_rejects_extended_resolver`, hits `ur.integration-tests.eth` on a real RPC). If that endpoint is flaky/unavailable or the fixture's resolver config ever changes, this safety property stops being verified with no offline signal. Consider a table-driven unit test constructing `ResolverInfo` values directly (offset=0/extended=false -> accept; extended=true or offset>0 -> reject) to pin the branching logic independent of network state.

crates/ens/src/utils.rs
```diff
@@ -0,0 +1,197 @@
...
+///
+/// ```
+/// use alloy_ens::dns_encode;
+/// assert_eq!(dns_encode("eth"), vec![3, b'e', b't', b'h', 0]);
+/// assert_eq!(
+///     dns_encode("vitalik.eth"),
+///     vec![7, b'v', b'i', b't', b'a', b'l', b'i', b'k', 3, b'e', b't', b'h', 0]
+/// );
+/// ```
+pub fn dns_encode(name: &str) -> Vec<u8> {
```
mablr, line 50: Nit: now that `try_dns_encode` is the recommended path, maybe consider re-adding the prior explicit warning that an oversized label's length prefix silently truncates (producing corrupted wire data).

crates/ens/src/lib.rs
```diff
@@ -109,129 +148,256 @@ mod contract {
...
+            /// Reverts with `OffchainLookup` when CCIP Read (ERC-3668) is required.
+            function resolve(bytes calldata name, bytes calldata data) external view returns (bytes memory result, address resolver);
+
+            /// Like `resolve`, but uses `gateways` for CCIP Read instead of the default.
+            function resolveWithGateways(bytes calldata name, bytes calldata data, string[] memory gateways) public view returns (bytes memory result, address resolver);
+
+            /// Reverse-resolves an address to its primary ENS name.
+            ///
+            /// `lookupAddress` is the raw byte encoding of the address.
+            /// `coinType` specifies the chain (use [`coin_type::ETH`] for Ethereum).
```
figtracer, line 237: this still breaks the existing public binding: base users can access `resolveReturn._0/_1` and `reverseCall.reverseName`, while this head renames or replaces all three. i confirmed the downstream compile fails here. can we preserve the legacy binding shape and add the ensip-23 binding separately?

crates/ens/src/lib.rs
```diff
@@ -109,129 +148,256 @@ mod contract {
...
-        /// Failed to get reverse registrar from the ENS registry.
+        /// Direct resolver calls are unsafe for this name; use Universal Resolver helpers.
+        ///
+        /// Returned when the name resolves through an ENSIP-10 extended resolver or a
+        /// parent/wildcard resolver. Calling `addr`/`text`/`name` on the raw resolver
+        /// instance can return incorrect data — use `resolve_name`, `lookup_txt`, or
+        /// related helpers instead.
+        #[error(
+            "ENS name {0:?} requires Universal Resolver resolution (extended or wildcard resolver)"
+        )]
```
figtracer, line 281: `EnsError` is public and exhaustive, so these added variants break downstream exhaustive matches. i confirmed the same match compiles against base and fails here. can we keep the old variant surface or explicitly treat this as a semver break?

crates/ens/src/lib.rs
yashgo0018: Extracted the accept/reject branch into `direct_resolver_address` and added table-driven unit tests (unextended+offset=0 accepts; extended or offset>0 rejects). The live `ur.integration-tests.eth` test remains.

crates/ens/src/utils.rs
yashgo0018: Re-added an explicit warning on `dns_encode` that oversized labels silently truncate the length prefix and produce corrupted wire data; `try_dns_encode` is the checked path.

crates/ens/src/lib.rs
yashgo0018: Restored the main `UniversalResolver` ABI (`resolveReturn._0/_1`, `reverseCall.reverseName`) and moved ENSIP-23 onto a separate `UniversalResolver23` binding. Provider helpers use `UniversalResolver23` only.

crates/ens/src/lib.rs
yashgo0018: Intentional semver break: the new variants (`RequiresUniversalResolver`, `DnsEncode`, `InvalidResponse`) are the safety/UR failure modes and are not folded into `Resolve`/`ResolverNotFound`. Marked `EnsError` `#[non_exhaustive]` so later additions do not keep breaking matches.

crates/ens/src/lib.rs
yashgo0018: Follow-up: kept a single `UniversalResolver`. Unnamed `resolve` returns remain so `resolveReturn._0/_1` still compile. The 1-arg `reverse(bytes reverseName)` ABI is not on the deployed contract (only `reverse(bytes,uint256)`), so that phantom binding is not preserved.

mattsse:
Thanks for the work on this. Maintainers cannot push to PR branches of organization-owned forks, so the reviewed and rebased version continues in #4183 with your commits preserved. Closing in favor of that PR.

PR #4111, opened 2026-08-04
Add generic CCIP Read support

yashgo0018:
Part 2 of the ENS and CCIP Read stack. Depends on Part 1 (Universal Resolver migration); please review/merge that first.

**Note:** Because intermediate stack branches are not on `alloy-rs/alloy`, this PR is opened against `main` and therefore includes Part 1 commits. Prefer reviewing only the CCIP Read commits (`b8b13e2a` and later on this branch), or wait until Part 1 lands and update this branch.

## What changed

- Add a feature-gated generic CCIP Read client to `alloy-provider`.
- Implement ERC-3668 redirects and callback calls.
- Implement the ENSIP-21 batch gateway protocol.
- Add configurable redirect, batch, concurrency, request, URL, revert-data, and response-size limits.
- Keep CCIP callback `input`/`data` fields in sync (`TransactionInputKind::Both`).
- Enforce `max_response_size` on local ENSIP-21 batch payloads.
- Support custom gateways, native HTTP, browser WASM, and a clear WASI Preview 1 fallback.

## Why

CCIP Read applies to arbitrary `eth_call` flows, so the transport belongs in `alloy-provider` rather than being embedded in ENS-specific code.

## Validation

- Targeted CCIP Read unit tests
- `cargo clippy -p alloy-provider --features ccip-read,reqwest-default-tls --lib --tests`
- `cargo check -p alloy-provider --all-features`
- `cargo check -p alloy-provider --target wasm32-wasip1 --no-default-features --features ccip-read`

This layer does not yet change ENS helpers to invoke CCIP Read.

Supersedes ens-integrations/alloy-2#3.


crates/provider/Cargo.toml
```diff
@@ -105,6 +105,7 @@ tokio-tungstenite.workspace = true
 
 [features]
 default = ["reqwest", "reqwest-default-tls"]
+ccip-read = ["reqwest"]
```
mablr: `ccip-read` forces the provider’s full reqwest transport feature even when callers supply a custom `CcipReadGateway`. Can the protocol/client and default HTTP gateway use separate features so custom gateways do not pull reqwest and an unused transport?

yashgo0018: fixed

crates/ens/src/lib.rs
```diff
@@ -92,109 +110,176 @@ impl FromStr for NameOrAddress {
...
+        /// ENS Universal Resolver (ENSIP-23).
+        ///
+        /// The single entry-point for all ENS resolution. Handles routing to wildcard
+        /// resolvers and CCIP Read (ERC-3668).
+        ///
+        /// Spec: <https://docs.ens.domains/ensip/23>
+        ///
+        /// Note: CCIP Read requires client-side handling of the `OffchainLookup` revert
+        /// (ERC-3668). Alloy does not currently implement this; calls to names that require
+        /// CCIP Read will surface as [`EnsError::Resolve`].
```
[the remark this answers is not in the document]

crates/provider/src/provider/ccip_read.rs
```diff
@@ -0,0 +1,1189 @@
...
+
+#[cfg(all(feature = "ccip-read-http", not(all(target_os = "wasi", target_env = "p1"))))]
+impl Default for HttpCcipReadGateway {
+    fn default() -> Self {
+        #[cfg(not(target_family = "wasm"))]
+        let client = reqwest::Client::builder()
+            .redirect(reqwest::redirect::Policy::custom(|attempt| {
+                if attempt.previous().len() >= 10 {
+                    attempt.stop()
+                } else if attempt.url().scheme() == "https" {
```
mablr, line 216: SSRF: the only validation applied to gateway URLs (initial and redirect) is `scheme == "https"`. `OffchainLookup.urls` originate from the contract being called, which per this PR's own threat model is untrusted, there's no check against private/link-local/metadata-service IPs (e.g. `169.254.169.254`, `127.0.0.0/8`, `10.0.0.0/8`) and no DNS-rebinding protection. A malicious/compromised resolver can direct a server-side consumer to make outbound HTTPS requests into its internal network. Consider an optional URL-validation hook on `CcipReadConfig`/`HttpCcipReadGateway` (resolved at request time to avoid TOCTOU) that rejects non-public IPs by default, or at least document the risk prominently.

crates/provider/src/provider/ccip_read.rs
```diff
@@ -0,0 +1,1189 @@
...
+#[cfg(all(feature = "ccip-read-http", not(all(target_os = "wasi", target_env = "p1"))))]
+struct GatewayResponse {
+    data: Bytes,
+}
+
+#[cfg_attr(target_family = "wasm", async_trait::async_trait(?Send))]
+#[cfg_attr(not(target_family = "wasm"), async_trait::async_trait)]
+#[cfg(all(feature = "ccip-read-http", not(all(target_os = "wasi", target_env = "p1"))))]
+impl CcipReadGateway for HttpCcipReadGateway {
+    async fn request(
```
mablr, line 350: Test coverage gap: every test in `mod tests` exercises the redirect/batch orchestration via a hand-rolled `CcipReadGateway` mock — none exercise this real `HttpCcipReadGateway::request` impl, which is the code that actually parses untrusted HTTP responses (HTTPS-only redirect policy, Content-Length pre-check, streaming size cap, content-type handling, `serde_json::from_slice`). Given the stated concern about malformed-gateway-response cases, this is the part of the module most in need of adversarial tests (oversized body without Content-Length, non-JSON body, truncated JSON, wrong content-type, redirect to non-HTTPS).

crates/provider/src/provider/ccip_read.rs
```diff
@@ -0,0 +1,1189 @@
...
+        request: CcipReadRequest,
+        context: &'a BatchContext<'a>,
+    ) -> CcipFuture<'a, Result<Bytes, CcipReadError>> {
+        Box::pin(async move {
+            if context.config.max_concurrent_requests == 0 {
+                return Err(CcipReadError::InvalidConfig(
+                    "max_concurrent_requests must be greater than zero".into(),
+                ));
+            }
+            let is_batch = request.urls.iter().any(|url| url == BATCH_GATEWAY_SENTINEL);
```
mablr, line 616: If `BATCH_GATEWAY_SENTINEL` appears anywhere in `request.urls`, all other URLs in that same list are silently discarded regardless of position. ERC-3668/ENSIP-21 clients are expected to try URLs in fallback order — a resolver listing `["https://real-gateway/...", "x-batch-gateway:true"]` (real gateway preferred, batch as fallback) would have the real gateway silently skipped. Worth documenting this assumption explicitly or handling mixed lists per ordering.

crates/provider/src/provider/ccip_read.rs
```diff
@@ -0,0 +1,1189 @@
...
+        unreachable!("redirect loop always returns")
+    }
+
+    fn fetch<'a>(
+        &'a self,
+        request: CcipReadRequest,
+        context: &'a BatchContext<'a>,
+    ) -> CcipFuture<'a, Result<Bytes, CcipReadError>> {
+        Box::pin(async move {
+            if context.config.max_concurrent_requests == 0 {
```
mablr, line 611: `max_concurrent_requests == 0` is only validated deep inside `fetch()`, i.e. only surfaces once an `OffchainLookup` is actually hit. A caller who misconfigures this and calls `.call()` on a transaction that doesn't trigger CCIP Read gets no error at all. Consider validating `CcipReadConfig` eagerly at construction time instead.

crates/provider/src/provider/ccip_read.rs
yashgo0018: Documented on the module and `HttpCcipReadGateway`: only `https` is required; private/link-local/metadata IPs are not blocked. Callers that need an allowlist, blocklist, or IP check should inject a custom `CcipReadGateway` (or ENS `*_with_ccip_read`).

crates/provider/src/provider/ccip_read.rs
yashgo0018: Split response classification (`Content-Length` pre-check, size cap, content-type, JSON) into helpers with table tests, plus real `HttpCcipReadGateway::request` coverage for empty / `http://` / invalid URLs. Did not add a local TLS stack.

crates/provider/src/provider/ccip_read.rs
yashgo0018: Documented: if `x-batch-gateway:true` appears anywhere in the URL list, HTTP URLs in that same list are not tried (ENSIP-21 local-batch precedence). Left the algorithm unchanged.

crates/provider/src/provider/ccip_read.rs
yashgo0018: `call_at` now returns `InvalidConfig` when `max_concurrent_requests == 0` before the first `eth_call`. The `fetch()` check remains as a backstop because `CcipReadConfig` fields are public.

mattsse:
Thanks for the work on this. Maintainers cannot push to PR branches of organization-owned forks, so the reviewed and rebased version continues in #4184 with your commits preserved. Closing in favor of that PR.

PR #4112, opened 2026-08-04
Enable CCIP Read for ENS resolution

yashgo0018:
Part 3 of the ENS and CCIP Read stack. Depends on Parts 1 and 2; please review/merge those first.

**Note:** Because intermediate stack branches are not on `alloy-rs/alloy`, this PR is opened against `main` and therefore includes Parts 1–2. Prefer reviewing only the ENS wiring commits (`06d11901` and later on this branch), or wait until the lower PRs land and update this branch.

## What changed

- Make ENS forward, reverse, multicoin, and text helpers use `call_with_ccip_read`.
- Resolve ENS CCIP calls at `BlockId::latest()`.
- Enable provider CCIP Read and its default TLS backend from `alloy-ens`.
- Keep `EnsError::CcipRead` mapping explicit (no blanket `From`) while preserving existing ENS transport-error categories for non-CCIP failures.
- Replace relevant live tests with canonical `ensdomains/resolution-tests` fixtures, including offchain ETH, text, wildcard, Base, reverse, and DNS cases.

## Why

This final layer connects the generic ERC-3668/ENSIP-21 transport to ENS resolution and verifies the complete behavior against maintained integration fixtures.

## Validation

- `cargo test -p alloy-ens --all-features`
- `cargo clippy -p alloy-ens --all-features --all-targets`
- `cargo check -p alloy-ens --no-default-features --features provider`
- `cargo check -p alloy-ens --target wasm32-wasip1 --no-default-features --features provider`

Supersedes ens-integrations/alloy-2#4.


crates/ens/Cargo.toml
```diff
@@ -36,4 +37,11 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 
 [features]
 contract = ["dep:alloy-contract", "dep:thiserror", "dep:alloy-sol-types"]
-provider = ["contract", "dep:alloy-provider", "dep:async-trait"]
+provider = [
+    "contract",
+    "dep:alloy-eips",
+    "dep:alloy-provider",
+    "alloy-provider/ccip-read",
+    "alloy-provider/reqwest-default-tls",
```
figtracer: `provider` now forces `reqwest-default-tls`, so even `alloy --no-default-features --features ens,reqwest-native-tls` activates reqwest default tls backend. can this forward separate tls features instead of overriding the caller selection?

crates/ens/src/lib.rs
```diff
@@ -205,276 +299,287 @@ mod provider {
...
-                .resolve(Bytes::from(dns_name), call_data)
-                .call()
+            let dns = dns_encode(name)?;
+            let calldata = EnsResolver::addrCall { node }.abi_encode();
+
+            let transaction = UniversalResolver::new(UNIVERSAL_RESOLVER_ADDRESS, self)
+                .resolve(dns.into(), calldata.into())
+                .into_transaction_request();
+            let output = self
+                .call_with_ccip_read_at(transaction, BlockId::latest())
```
figtracer: `CcipReadClient` supports custom gateways and config, but this path always creates the default client, so ens users cannot proxy, allowlist, blocklist, or disable contract-controlled urls as erc-3668 recommends. can we accept a client here or add configurable ens variants?

crates/ens/Cargo.toml
yashgo0018: fixed

crates/provider/src/provider/ccip_read.rs
```diff
@@ -0,0 +1,1171 @@
...
+where
+    P: Provider<N>,
+    N: Network,
+{
+    async fn call_with_ccip_read(
+        &self,
+        transaction: N::TransactionRequest,
+    ) -> Result<Bytes, CcipReadError> {
+        CcipReadClient::default().call(self, transaction).await
+    }
```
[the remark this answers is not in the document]
yashgo0018: Fixed with shared_http_ccip_read_client()

crates/ens/src/lib.rs
```diff
@@ -259,258 +470,334 @@ mod provider {
...
+        ) -> Result<Address, EnsError> {
+            let node = namehash(name);
+            let dns = try_dns_encode(name)?;
+            let calldata = EnsResolver::addrCall { node }.abi_encode();
+
+            let transaction = UniversalResolver::new(UNIVERSAL_RESOLVER_ADDRESS, self)
+                .resolve(dns.into(), calldata.into())
+                .into_transaction_request();
+            let output = client
+                .call_at(self, transaction, BlockId::latest())
```
mablr, line 526: `BlockId::latest()` is re-resolved by the node on every RPC call, and a CCIP round trip involves an initial `eth_call`, an off-chain HTTP gateway fetch (real-world latency, possibly retried across multiple gateway URLs), and a follow-up `eth_call`. Chain state backing "latest" can legitimately drift between the initial call (which produces the `OffchainLookup` revert) and the callback call, a behavior change from the pre-PR single non-CCIP `eth_call`. In practice this is mitigated by ERC-3668 resolvers validating their own signed payloads, but that lives entirely in third-party contracts and isn't discussed here. Worth documenting this tradeoff in the doc comment.

crates/ens/src/lib.rs
```diff
@@ -259,258 +470,334 @@ mod provider {
...
-        // Hash the label.
-        let mut label_hasher = Keccak256::new();
-        label_hasher.update(label.as_bytes());
-        label_hasher.finalize_into(&mut buffer[32..]);
-
-        // Hash both the node and the label hash, writing into the node.
-        let mut buffer_hasher = Keccak256::new();
-        buffer_hasher.update(buffer.as_slice());
-        buffer_hasher.finalize_into(&mut buffer[..32]);
+    fn map_ccip_error(
```
mablr, line 646: The only pure unit test for this mapping (`preserves_existing_ens_transport_error_category`) covers just the `Transport` branch. No unit test covers the other `CcipReadError` branches (`Gateway`, `TooManyRedirects`, `SenderMismatch`, etc.) falling through to `EnsError::CcipRead`, or verifies `ResolverNotFound` is still distinguished when wrapped in `CcipReadError::Transport`. All other CCIP-path coverage depends on live mainnet RPC + live third-party gateways. Consider a couple more pure unit tests to lock in the mapping contract independent of network availability.

crates/ens/src/lib.rs
yashgo0018: Documented on the ENS helpers: CCIP round trips use `BlockId::latest` for the initial call and the callback, so chain head can move between them. ERC-3668 signed payloads are contract-side; Alloy does not enforce that.

crates/ens/src/lib.rs
yashgo0018: Added unit tests for `Gateway` / `TooManyRedirects` / `SenderMismatch` → `EnsError::CcipRead`, and for `ResolverNotFound` still distinguished when wrapped in `CcipReadError::Transport`.

mattsse:
Thanks for the work on this. Maintainers cannot push to PR branches of organization-owned forks, so the reviewed and rebased version continues in #4185 with your commits preserved. Closing in favor of that PR.

PR #4114, opened 2026-08-04
fix(rpc): align transaction response types

figtracer:
Aligns `eth_fillTransaction` and `eth_signTransaction` with ethereum/execution-apis#803. Fill now returns only the unsigned filled transaction, while sign returns both encoded bytes and the signed RPC transaction object.

This updates provider return types and preserves `Web3Signer::sign_transaction`'s raw-byte convenience API.

Checks:
- `cargo test -p alloy-rpc-types-eth`
- `cargo check -p alloy-provider --all-features --all-targets`
- `cargo clippy -p alloy-rpc-types-eth -p alloy-provider --all-targets --all-features -- -D warnings`

### DIFF
--- a/crates/provider/src/fillers/mod.rs
+++ b/crates/provider/src/fillers/mod.rs
@@ -61,7 +61,7 @@ use alloy_rpc_types_eth::{
     erc4337::TransactionConditional,
     simulate::{SimulatePayload, SimulatedBlock},
     AccessListResult, EIP1186AccountProofResponse, EthCallResponse, FeeHistory, Filter,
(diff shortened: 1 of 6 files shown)


mattsse (commented):
this is breaking, so perhaps we should try to make this backwards compat by issueing new fns, but I guess if this is incompatible with the server's impl then this isnt useful currently anywy?

figtracer:
> this is breaking, so perhaps we should try to make this backwards compat by issueing new fns, but I guess if this is incompatible with the server's impl then this isnt useful currently anywy?

let's park it for now

PR #4120, opened 2026-08-07
feat(eip7594): reconstruct sidecars from sparse cells

Soubhik-10:
EIP-7594 allows nodes to store and exchange only subsets of a blob’s 128 cells. However, transaction validation, execution APIs, RPC responses, and existing blobpool code still require a complete BlobTransactionSidecar. This change provides the missing bridge: once 64+ cells are available, Alloy reconstructs the original blob and all cell proofs using c-kzg, verifies them against the transaction commitments, and returns the normal complete sidecar type.

### DIFF
--- a/crates/eips/src/eip7594/sidecar.rs
+++ b/crates/eips/src/eip7594/sidecar.rs
@@ -3,7 +3,7 @@ use crate::{
         Blob, BlobAndProofV2, BlobTransactionSidecar, Bytes48, BYTES_PER_BLOB,
         BYTES_PER_COMMITMENT, BYTES_PER_PROOF,
     },
-    eip7594::{CELLS_PER_EXT_BLOB, EIP_7594_WRAPPER_VERSION, FIELD_ELEMENTS_PER_EXT_BLOB},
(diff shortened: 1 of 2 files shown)


crates/eips/src/eip7594/sidecar.rs
```diff
@@ -559,7 +594,205 @@ impl BlobTransactionSidecarEip7594 {
...
+    /// The cells must be laid out in blob-major order, with exactly
+    /// [`CELLS_PER_EXT_BLOB`] cells for each blob. The supplied proofs must match the recovered
+    /// proofs. This uses the default KZG settings.
+    #[cfg(feature = "kzg")]
+    pub fn try_from_cells(
+        cells: Vec<crate::eip7594::Cell>,
+        commitments: Vec<Bytes48>,
+        cell_proofs: Vec<Bytes48>,
+    ) -> Result<Self, c_kzg::Error> {
+        use crate::eip4844::env_settings::EnvKzgSettings;
```
Soubhik-10:
```suggestion
```

mattsse (approved):

PR #4124, opened 2026-08-08
fix(signer): reject BIP32 indices that already set the harden bit

SashaMIT:
## Summary
- Ledger `path_to_bytes` and Trezor `convert_path` parsed BIP32 segments with `parse::<u32>()` then optionally applied `0x80000000`.
- Raw index `2147483648` (no `'`) already equals the harden bit, so `Other(\"m/2147483648\")` silently selected the same child as `m/0'`.
- Reject `index >= 0x80000000` before applying the harden marker. Trezor path conversion now returns `Result` instead of unwrapping parse errors.

## Test plan
- [x] `cargo test -p alloy-signer-ledger --lib rejects_raw_hardened`
- [x] `cargo test -p alloy-signer-trezor --lib rejects_raw_hardened`
- [x] `cargo test -p alloy-signer-ledger --lib invalid_derivation_path`

### DIFF
--- a/crates/signer-ledger/src/signer.rs
+++ b/crates/signer-ledger/src/signer.rs
@@ -387,10 +387,17 @@ impl LedgerSigner {
         let mut bytes = vec![depth as u8];
         for derivation_index in elements {
             let hardened = derivation_index.contains('\'');
-            let mut index = derivation_index.replace('\'', "").parse::<u32>()?;
-            if hardened {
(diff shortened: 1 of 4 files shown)


SashaMIT:
Nightly rustfmt was red on the two path helpers. Reformatted those parse sites so cargo fmt --check is clean. Behavior unchanged.

PR #4122, opened 2026-08-07
feat: send_raw_transaction_sync_with_timeout

kirsanium:
## Motivation

Closes #4041.
Increases conformance to [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966) by introducing method `send_raw_transaction_sync_with_timeout` on `Provider` trait which allows usage of `eth_sendRawTransactionSync` RPC method with optional timeout parameter.

## Solution
Introduces a method `Provider::send_raw_transaction_sync_with_timeout`. A cleaner but unfortunately breaking change would be to introduce a new parameter to `Provider::send_raw_transaction_sync`.

The tests are added but they work only with Anvil nightly (https://github.com/foundry-rs/foundry/commit/283dabaf5cb71090fe33b3393b24b117d16d26ae).

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

PR #4125, opened 2026-08-09
feat(provider): add raw transaction sync future

mattsse:
## Summary

- replace the opaque `async fn Provider::send_raw_transaction_sync` return with a named `SendRawTransactionSync` future
- add a client-side `.timeout(Duration) -> Timeout<Self>`
- add `.server_timeout(Option<Duration>)`, serialized as EIP-7966's optional millisecond parameter
- validate server durations that cannot be represented as a positive `u64` millisecond value

## Motivation

This is an alternative to #4122 that follows the composable future pattern introduced in #4119. Existing call sites can continue to await `send_raw_transaction_sync` directly, while callers that need timeout control can independently bound the local wait and the RPC server's inclusion wait.

The client timeout only stops waiting for the response; the server timeout is sent to the node and surfaces through the normal RPC result. Custom `Provider` implementations that override this method will need to adopt the named return type.

### DIFF
--- a/crates/provider/src/provider/send_raw_transaction_sync.rs
+++ b/crates/provider/src/provider/send_raw_transaction_sync.rs
@@ -1,0 +1,170 @@
+use alloy_network::Network;
+use alloy_rpc_client::{ClientRef, RpcCall};
+use alloy_transport::TransportResult;
+use serde::ser::{Error as _, SerializeSeq};
+use std::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+};
+
+#[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
+use tokio::time::{timeout as timeout_future, Timeout};
+
+#[cfg(all(target_family = "wasm", target_os = "unknown"))]
+use wasmtimer::tokio::{timeout as timeout_future, Timeout};
+
+/// Parameters for an `eth_sendRawTransactionSync` request.
+#[derive(Clone, Debug)]
+struct SendRawTransactionSyncParams {
+    encoded_tx: String,
+    server_timeout: Option<Duration>,
+}
+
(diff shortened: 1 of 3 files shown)


crates/provider/src/provider/send_raw_transaction_sync.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// Awaiting the returned future produces a timeout result around the existing transport result,
+    /// so the two error cases can be handled separately.
+    ///
+    /// # Panics
+    ///
+    /// On Tokio-backed targets, including WASI, the returned future panics when polled if there
+    /// is no current Tokio timer, for example when polled outside of a Tokio runtime.
+    pub fn timeout(self, duration: Duration) -> Timeout<Self> {
+        timeout_future(duration, self)
+    }
```
onbjerg, line 81: this is not entirely correct. this is a client side timeout, the timeout that was requested is the one in the eip: https://eips.ethereum.org/EIPS/eip-7966#example-request-no-timeout

so we'd want something like `EthGetBlock`: https://github.com/alloy-rs/alloy/blob/c2201618345c2e4a75b9ef911ab689aff61bf2f6/crates/provider/src/provider/get_block.rs#L61-L154

onbjerg (changes requested):

crates/provider/src/provider/send_raw_transaction_sync.rs
mattsse: we have both, see 

https://github.com/alloy-rs/alloy/pull/4125/changes#diff-25df010ec5afa518727f6d791ef7600e61058e023bb1c7b1ae41d1cf3f4df7f9R93

imo the server side timeout is a bit stupid anyway

onbjerg: ah right

PR #4147, opened 2026-08-16
perf(transport-ipc): frame JSON incrementally instead of re-parsing the buffer

### LINKED ISSUE #4136: [Bug] ReadJsonStream is O(n²) on chunked IPC frames
## Summary

`ReadJsonStream` re-runs `serde_json::Deserializer::from_slice` over the **entire buffer** after every socket read. `serde_json` cannot resume, so a frame that arrives in K chunks is parsed K times from byte zero. `poll_read_buf` also never `reserve`s, so once the initial 4 KiB `BytesMut` is full, reads shrink to ~64 B and the restart count explodes.

Large responses (`eth_getLogs`, multi-kilobyte `eth_simulateV1` results) are quadratic in their size. Small frames that arrive in one read are unaffected.

Fix: incremental framing so each byte is examined once and each complete frame is deserialized once, plus `reserve` before `poll_read_buf`. Public API unchanged. Proposed PR: #4147 (numbers live there).

## Component

- `alloy-transport-ipc` — `ReadJsonStream::poll_next` (`crates/transport-ipc/src/lib.rs`)

## Version

Reproduced on alloy `main` (and 2.2.0 / 2.4.0 — same `ReadJsonStream` loop). Colocated mainnet Reth v2.4.1.

## What the current code does

1. `ReadJsonStream::poll_next` (on `main`) does `Deserializer::from_slice(buf)` on every wake, then `poll_read_buf` without `reserve`.
2. After 4 KiB, `BytesMut::chunk_mut` yields ~64 B, so a large frame is re-parsed on every tiny read.

## Requested outcome

#4147: resumable frame scan + one `serde_json` pass per complete value + `reserve(4 KiB)` before each read. Small complete frames keep the serde-first path so they do not regress.

## Related (not duplicates)

- #957 / #958 — large IPC responses / buffer capacity
- #69, #1914 — `ReadJsonStream` internals / reuse
- #4147 — the fix

pata-eth:
## Motivation

Fixes #4136.

`ReadJsonStream` re-runs `serde_json::Deserializer::from_slice` over the **entire buffer** after every socket read. `serde_json` cannot resume, so a frame that arrives in K chunks is parsed K times from byte zero. Reads are bounded by the socket buffer rather than by the frame, so K grows with the response and the total parsing work is quadratic in its size.

Large responses (`eth_getLogs`, multi-kilobyte `eth_simulateV1` results) pay for this. Small frames that arrive in one read are unaffected.

## Solution

Public API unchanged. `ReadJsonStream` now:

- tries `serde_json` first when the buffer looks complete, so a small frame that arrives in one read costs the same as before
- on `is_eof` / `is_data`, flips onto a resumable `FrameScanner` (`memchr` over string interiors and `{ } [ ] "`) so each byte of a large frame is examined once, no matter how the socket chunks it
- reserves before each read so `poll_read_buf` always gets a full page to fill instead of whatever slack is left over

Each complete frame is deserialized exactly once.

One behaviour change worth calling out: a frame that is well-formed JSON but does not deserialize into the stream's item type used to leave its bytes in the buffer and re-fail on every subsequent read, so the reader spun until the peer closed the socket. It now ends the stream, and the backend reports the error to in-flight requests instead of hanging.

## Measurements

Colocated mainnet Reth v2.4.1, 32 cores. Interleaved medians, 5 reps, unless noted.

### Large frames — `eth_getLogs` Transfer

Span 16, 32 requests @ concurrency 8:

| transport | rps | p50 |
| --- | --- | --- |
| IPC `main` | 57 | 136 ms |
| IPC this PR | **108** | **65 ms** |

Span 16 @ concurrency 16 (same tip, earlier single-shot A/B): IPC `main` ~78 rps → this PR **~154**.

Span 40, 64 requests @ concurrency 16:

| transport | rps | p50 |
| --- | --- | --- |
| IPC `main` | 12 | 1298 ms |
| IPC this PR | **56** | **252 ms** |

### Small frames — no regression

`eth_simulateV1`, 1 call, 1 account / 1 slot, `returnFullTransactions: false`:

| shape | IPC `main` rps | IPC this PR rps |
| --- | --- | --- |
| batch 10, 13 in flight | 7918 | **7924** |
| batch 150, 2 in flight | 382 | **394** |

`eth_blockNumber` @ concurrency 16 is in the noise (same order, ~65–67k rps).

## Correctness

`FrameScanner` was differential-fuzzed against `serde_json`'s `byte_offset` over 50k generated documents — nested objects and arrays, strings carrying `"`, `\`, braces, brackets, newlines, non-ASCII and 4-byte code points — each scanned both whole-buffer and re-scanned across random chunk boundaries. No divergences, and every run inspected each byte once.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [ ] Breaking changes

### DIFF
--- a/crates/transport-ipc/Cargo.toml
+++ b/crates/transport-ipc/Cargo.toml
@@ -37,6 +37,7 @@ serde.workspace = true
 
 bytes = "1.5.0"
 interprocess = { version = "2", features = ["tokio"] }
+memchr = "2.7"
 tempfile = { workspace = true, optional = true }
 
 [dev-dependencies]
--- a/crates/transport-ipc/Cargo.toml
+++ b/crates/transport-ipc/Cargo.toml
@@ -41,10 +41,6 @@ memchr = "2.7"
 tempfile = { workspace = true, optional = true }
 
 [dev-dependencies]
-alloy-json-rpc.workspace = true
-alloy-pubsub.workspace = true
-interprocess = { version = "2", features = ["tokio"] }
-serde_json.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["macros", "rt-multi-thread", "io-util", "time"] }
 tokio-test.workspace = true
--- a/crates/transport-ipc/Cargo.toml
+++ b/crates/transport-ipc/Cargo.toml
@@ -40,6 +40,12 @@ interprocess = { version = "2", features = ["tokio"] }
 tempfile = { workspace = true, optional = true }
 
 [dev-dependencies]
+alloy-json-rpc.workspace = true
+alloy-pubsub.workspace = true
+interprocess = { version = "2", features = ["tokio"] }
+serde_json.workspace = true
+tempfile.workspace = true
+tokio = { workspace = true, features = ["macros", "rt-multi-thread", "io-util", "time"] }
 tokio-test.workspace = true
 
 [features]
--- a/crates/transport-ipc/README.md
+++ b/crates/transport-ipc/README.md
@@ -5,10 +5,11 @@ IPC transport implementation.
 The backend keeps a long-lived connection (unix socket or Windows named
 pipe) so subscriptions can share a mux with request/response traffic.
 
-JSON-RPC values are framed incrementally: a small state machine finds the
-end of each top-level object or array as bytes arrive, and each complete
-frame is deserialized once. Partial responses are not re-parsed from the
-start of the buffer on every socket read.
+JSON-RPC values that arrive complete in one read are deserialized with
+`serde_json` directly. A partial value flips the reader onto an incremental
+framer (`memchr` over strings and structural bytes) so a large response is
+scanned once, not re-parsed from the start of the buffer on every socket
+read.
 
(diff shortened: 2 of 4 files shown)


mattsse (approved):
Reviewed the incremental JSON-RPC framing changes and exercised the transport-ipc tests locally. Looks good.

kayabaNerve:
<details>
<summary> Relevant self-shill, feel free to delete if too off-topic </summary>
<br>
While `memchr` may be fine, as a heads up, I'm incredibly proud of my work on `core-json` which can be used as a streaming JSON deserializer: https://github.com/core-json/core-json
<br>
<br>
It's well tested and would be usable to definitively know while the item ends. It also has explicitly-bounded memory, so it's resilient to DoSs, and is competitive with serde-json on performance. It is published to crates.io, though I was working on an `async` variant on a branch which isn't yet published and may be preferable here.

<br>
<br>
If it is of potential interest, I'd be happy to discuss it more.
</details>

PR #4153, opened 2026-08-17
fix(rpc-types-eth): preserve specified topic positions in Filter

0xOmarA:
## Motivation

Say that we wish to construct a `Filter` to represent the following topic constraints:

- The event signature is `some_signature`.
- Topic 1 belongs to `set A`
- Topic 2 is anything
- There is at least 3 topics in the log.

The traditional way of constructing such a topics filter in standard JSON would be as follows: `[some_signature, set_a, []]` which would apply all the constraints above and also apply the constraint that there must be at least 3 topics.

This is unrepresentable through `Filter` at the moment since we store topics as `[BTreeSet<B256>; 4]` where an explicit empty set (used as a wild card) is indistinguishable from an empty set which was created as part of the array’s initialization.

The outcome of this is the loss of the "there is at least 3 topics in the log" constraint from the serialized filter arriving at geth (or any other node) thus the filter matches logs with at least 2 topics.

For context, we depend on alloy heavily in [parity's eth-rpc](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/revive/rpc) for some of the logic it provides especially around log filters.

## The Solution

The first part of the solution is that we need to untangle two concepts: `FilterSet::default()` was used to mean both "the user didn’t specify any constraints for this topic position" and also "this topic is unconstrained" which are two concepts that seem the same but are different (as shown in the motivation above).

To do this I changed the `topics` field type on `Filter` from `[Topic; 4]` to `[Option<Topic>; 4]` where:

- `Option::None` means that the user didn’t specify a constraint on this topic position.
- `Option::Some` means that the user specified a constraint on this topic position (even if the constraint is "unconstrained")

This means that a wildcard is now representable in one of two ways:

- The normal `FilterSet::default()` which we always had.
- A `None` in the topics array which is then followed by a `Some` somewhere in the array.

## PR Checklist

- [x] Added Tests
- [x] Added Documentation
- [x] Breaking changes
  - Sadly the changes here are breaking for any client was constructing `Filter` through its fields rather than using the builder pattern implemented on the type since the type of `topics` was changed. Otherwise, I tried my best to preserve the signature of all other functions across the codebase (e.g., `Filter::topic1`'s signature and other similar methods have the same signature)

### DIFF
--- a/crates/rpc-types-eth/src/filter.rs
+++ b/crates/rpc-types-eth/src/filter.rs
@@ -421,9 +421,12 @@ pub struct Filter {
     pub address: FilterSet<Address>,
     /// Topic filters, with at most four positions.
     ///
-    /// Values within a position use OR semantics; populated positions use AND semantics. An empty
-    /// position is a wildcard and serializes as `null` when followed by a populated position.
-    pub topics: [Topic; 4],
+    /// Values within a position use OR semantics; populated positions use AND semantics. A [`None`]
+    /// position is unspecified: it matches anything and is truncated from the serialized array when
+    /// no later position is specified. An explicitly specified empty set is a wildcard that matches
+    /// any value but still requires the log to have a topic at that position; it serializes as
+    /// `null`.
+    pub topics: [Option<Topic>; 4],
 }
 
 impl Filter {
@@ -605,28 +608,28 @@ impl Filter {
     /// Sets event_signature(topic0) (the event name for non-anonymous events)
     #[must_use]
     pub fn event_signature<T: Into<Topic>>(mut self, topic: T) -> Self {
-        self.topics[0] = topic.into();
+        self.topics[0] = Some(topic.into());
         self
     }
 
     /// Sets the 1st indexed topic
     #[must_use]
     pub fn topic1<T: Into<Topic>>(mut self, topic: T) -> Self {
-        self.topics[1] = topic.into();
+        self.topics[1] = Some(topic.into());
         self
     }
 
     /// Sets the 2nd indexed topic
     #[must_use]
     pub fn topic2<T: Into<Topic>>(mut self, topic: T) -> Self {
-        self.topics[2] = topic.into();
+        self.topics[2] = Some(topic.into());
         self
     }
 
     /// Sets the 3rd indexed topic
     #[must_use]
     pub fn topic3<T: Into<Topic>>(mut self, topic: T) -> Self {
-        self.topics[3] = topic.into();
+        self.topics[3] = Some(topic.into());
         self
     }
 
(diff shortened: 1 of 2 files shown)


crates/rpc-types-eth/src/filter.rs
```diff
@@ -421,9 +421,12 @@ pub struct Filter {
...
     ///
-    /// Values within a position use OR semantics; populated positions use AND semantics. An empty
-    /// position is a wildcard and serializes as `null` when followed by a populated position.
-    pub topics: [Topic; 4],
+    /// Values within a position use OR semantics; populated positions use AND semantics. A [`None`]
+    /// position is unspecified: it matches anything and is truncated from the serialized array when
+    /// no later position is specified. An explicitly specified empty set is a wildcard that matches
+    /// any value but still requires the log to have a topic at that position; it serializes as
+    /// `null`.
+    pub topics: [Option<Topic>; 4],
```
0xOmarA, line 429: This is the main change which allows us to represent "the user didn't specify any constraints for this topic position" vs "the user has specified it and they wish to keep it unconstrained"

crates/rpc-types-eth/src/filter.rs
```diff
@@ -1518,31 +1522,7 @@ impl FilteredParams {
...
-                // Check that `log_topic` is included in `filter_topic`
-                Both(filter_topic, log_topic) => {
-                    if !filter_topic.matches(log_topic) {
-                        return false;
-                    }
-                }
-            }
-        }
-        true
+        self.filter.as_ref().is_none_or(|filter| filter.matches_topics(log_topics))
```
0xOmarA, line 1525: Would appreciate another pair of eyes on this. I found that this method's logic was almost identical to `matches_topics` so I decided to just forward the call to `matches_topics` (and still preserve the return `true` if there's no filter behavior).

If there's any context I'm missing on why we needed a very similar second implementation here then please let me know.

0xOmarA (commented):
I will be using the comments just to highlight some bits of the code which I think need highlighting.

crates/rpc-types-eth/src/filter.rs
```diff
@@ -678,9 +681,9 @@ impl Filter {
     /// Returns `true` if the filter matches the given topics.
     pub fn matches_topics(&self, topics: &[B256]) -> bool {
         self.topics.iter().zip_longest(topics.iter()).all(|topic| match topic {
-            Both(filter, log) => filter.matches(log),
-            Left(filter) => filter.is_empty(),
-            Right(_) => false,
+            Both(Some(filter), log) => filter.matches(log),
+            Left(None) | Both(None, _) | Right(_) => true,
+            Left(Some(_)) => false,
```
0xOmarA, line 686: The "There is at least 3 topics in the log" constraint from my PR description is applied here. If there's ever a specified topic filter at this position but there's no associated log then the filter does not match.

PR #4155, opened 2026-08-21
fix(consensus): bound sealed block RLP decoding

0xalpharush:
## Motivation

`Block::decode_sealed` decoded the outer RLP header but allowed its child decoders to read from the complete caller buffer. A malformed block could therefore consume fields beyond the payload length declared by the outer header.

This violates the outer RLP boundary and can make decoding depend on bytes belonging to a following value.

## Solution

Restrict child decoding to the payload declared by the outer header and require the bounded payload to be consumed exactly.

The regression tests cover:

- fields extending past the outer RLP boundary;
- a hostile child decoder that expands its input cursor;
- valid sealed-block round trips.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## Specification

The [Ethereum RLP specification](https://ethereum.org/developers/docs/data-structures-and-encoding/rlp/) defines a list payload as the concatenated encodings of its items with an exact declared payload length. [go-ethereum's `Stream.List` and `ListEnd`](https://github.com/ethereum/go-ethereum/blob/02b73d4ea7181464175e0a6cbecc0a3a2655a562/rlp/decode.go#L793-L830) enforce the same boundary by tracking the remaining bytes of every enclosing list and requiring the decoder to finish at the list's end.

### DIFF
--- a/crates/consensus/src/block/mod.rs
+++ b/crates/consensus/src/block/mod.rs
@@ -399,21 +399,31 @@ mod block_rlp {
         /// This is more efficient than decoding the block and then sealing it, as the header
         /// hash is computed from the raw RLP bytes without re-encoding.
         pub fn decode_sealed(buf: &mut &[u8]) -> alloy_rlp::Result<Sealed<Self>> {
-            // Decode the outer block list header
-            let block_rlp_head = alloy_rlp::Header::decode(buf)?;
-            if !block_rlp_head.list {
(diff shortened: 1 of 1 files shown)


mattsse (approved):
lgtm pending @klkvr

klkvr (approved):

PR #4156, opened 2026-08-21
fix(eips): reject unwrapped typed network encoding

0xalpharush:
## Motivation

EIP-2718 typed transactions have distinct canonical encodings:

- direct encoding: `type || payload`;
- network encoding: an RLP string containing `type || payload`.

`network_decode` also accepted the unwrapped direct form. Re-encoding the resulting transaction as a network value added the missing wrapper, violating decode/encode canonicality.

## Solution

Require typed values passed to `network_decode` to have the network RLP-string wrapper. Direct typed transaction decoding remains supported through `decode_2718`.

A regression test verifies that unwrapped typed transactions are rejected while valid wrapped network transactions continue to round trip.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## Specification

The [Ethereum Wire Protocol transaction encoding](https://github.com/ethereum/devp2p/blob/77264dbc983a06aab74a79aec85d9e18361df03e/caps/eth.md#L174-L198) specifies EIP-2718 typed transactions as RLP byte arrays whose contents are `tx-type || tx-data`. go-ethereum likewise distinguishes [`Transaction.DecodeRLP`](https://github.com/ethereum/go-ethereum/blob/02b73d4ea7181464175e0a6cbecc0a3a2655a562/core/types/transaction.go#L142-L176), which reads the network RLP wrapper, from [`UnmarshalBinary`](https://github.com/ethereum/go-ethereum/blob/02b73d4ea7181464175e0a6cbecc0a3a2655a562/core/types/transaction.go#L178-L198), which reads the bare EIP-2718 encoding.

### DIFF
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1449,6 +1449,17 @@ mod tests {
     }
 
     #[test]
+    fn network_decode_rejects_unwrapped_typed_transaction() {
+        let tx = TxEnvelope::Eip1559(TxEip1559::default().into_signed(Signature::test_signature()));
+        let encoded = tx.encoded_2718();
+        assert_eq!(encoded.first().copied(), Some(TxType::Eip1559 as u8));
+
+        let mut input = encoded.as_slice();
+        assert!(TxEnvelope::network_decode(&mut input).is_err());
+        assert_eq!(input, encoded.as_slice());
+    }
+
+    #[test]
     fn decode_encode_known_rpc_transaction() {
         // test data pulled from hive test that sends blob transactions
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm pending @klkvr

decofe:
Closing in favor of documenting the existing compatibility behavior in https://github.com/alloy-rs/alloy/pull/4217. Requiring the RLP string wrapper would reject unwrapped typed envelopes that network_decode currently accepts, which is a behavioral breaking change for existing callers. We are retaining that permissive behavior; #4217 documents it and explicitly records the breaking-change constraint.

PR #4158, opened 2026-08-21
fix(consensus): validate legacy signing payload RLP

0xalpharush:
## Motivation

Unsigned EIP-155 signing payloads use the same nine-field RLP shape as signed legacy transactions, with the final fields required to be `chain_id, 0, 0`.

The legacy signing-payload decoder did not fully enforce this structure: it did not require a list-bounded payload and decoded and discarded arbitrary values in the two sentinel fields.

This allowed malformed inputs to decode successfully and normalize when re-encoded.

## Solution

Decode strictly within the declared RLP-list payload, require both EIP-155 sentinel fields to be zero, reject extra payload fields, and preserve bytes following the decoded value in the caller's cursor.

Regression tests cover malformed sentinels, incorrect headers and payload boundaries, trailing fields, and valid signing payloads.

## PR Checklist

- [x] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes

## Specification

[EIP-155](https://eips.ethereum.org/EIPS/eip-155#specification) defines the signing payload as the nine-element RLP tuple `(nonce, gasprice, startgas, to, value, data, chainid, 0, 0)`. go-ethereum constructs the same tuple, including both zero sentinels, in [`LegacyTx.sigHash`](https://github.com/ethereum/go-ethereum/blob/02b73d4ea7181464175e0a6cbecc0a3a2655a562/core/types/tx_legacy.go#L127-L138).

### DIFF
--- a/crates/consensus/src/transaction/legacy.rs
+++ b/crates/consensus/src/transaction/legacy.rs
@@ -368,30 +368,56 @@ impl Encodable for TxLegacy {
 
 impl Decodable for TxLegacy {
     fn decode(data: &mut &[u8]) -> Result<Self> {
-        let header = Header::decode(data)?;
-        let remaining_len = data.len();
+        let mut payload = Header::decode_bytes(data, true)?;
+        let mut transaction = Self::rlp_decode_fields(&mut payload)?;
 
-        let transaction_payload_len = header.payload_length;
-
-        if transaction_payload_len > remaining_len {
-            return Err(alloy_rlp::Error::InputTooShort);
+        // If the payload has more fields, they must be the EIP-155 chain ID and zero sentinels.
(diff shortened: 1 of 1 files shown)


DaniPopes (approved):

mattsse (approved):
lgtm pending @klkvr

decofe:
Closing in favor of documenting the existing compatibility behavior in https://github.com/alloy-rs/alloy/pull/4217. Requiring a list header and zero EIP-155 sentinel fields would reject inputs that TxLegacy::decode currently accepts, which is a behavioral breaking change for existing callers. We are retaining that permissive behavior; #4217 documents the accepted forms, the signed-RLP caveat, and the breaking-change constraint.

Issue #3881, opened 2026-04-12
[Bug] .watch() can hang when RPC block num is stale


lsankar4033:
### Component

provider, pubsub

### What version of Alloy are you on?

1.7.3

### Operating System

macOS (Apple Silicon)

### Describe the bug

I've noticed that watch() on an http provider can hang for long times || indefinitely when the underlying RPC nodes have stale block nums (i.e. returned by `eth_blockNumber`). 

This isn't that uncommon with node providers like Alchemy due to the way they seem to do routing. 

The issue is that the way `watch_pending_transactions()` works with an http provider seems to be to get the latest block number, then iterate (via `NewBlocks`) through recent blocks, checking for existing of the tx being watched in each. As polling/processing time gets close to block time, this can hang for a long time, sometimes even indefinitely!

I've moved away from using `watch()` in favor of just locally looping `eth_getTransactionReciept`, but it might be worth doing the dumb thing of that and/or documenting on `watch()` that this inconsistency exists.

PR #4162, opened 2026-08-25
fix(provider): recheck receipts of unconfirmed transactions in heartb…

0xSoftBoi:
## Summary
Fixes heartbeat to periodically recheck receipts of unconfirmed transactions, preventing indefinite hangs when a transaction confirmation is missed by the initial watch.
## Problem
When a transaction confirmation arrives out-of-order or is missed by the initial listener, the heartbeat never validates the receipt. This causes watchers to hang indefinitely, blocking all subsequent block processing.
## Solution
- Added `RecheckResult` struct to track recheck attempts and receipts
- Implemented `spawn_recheck()` to asynchronously validate unconfirmed txs one at a time
- Implemented `apply_recheck()` to reconcile rechecked receipts against pending watchers
- Rechecks only occur while watchers exist; spawned tasks do not block block processing
## Testing
Added 4 regression tests that hang without the fix and pass with it:
- `hang_on_missed_confirmation` — confirms hang when confirmation is missed
- `hang_on_out_of_order_confirmation` — confirms hang when confirmation arrives out-of-order
- `recheck_finds_missed_confirmation` — confirms fix recovers from missed confirmation
- `recheck_finds_out_of_order_confirmation` — confirms fix handles out-of-order arrivals

### DIFF
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -6,7 +6,7 @@ use alloy_json_rpc::RpcError;
 use alloy_network::{primitives::ReceiptResponse, BlockResponse, Network};
 use alloy_primitives::{
     map::{B256HashMap, B256HashSet},
-    TxHash, B256, U64,
+    TxHash, B256,
 };
 use alloy_rpc_client::WeakClient;
 use alloy_transport::{utils::Spawnable, TransportError};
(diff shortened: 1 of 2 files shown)


0xSoftBoi:
Pushed a follow-up: the first version of this PR still hung in the exact case #3881 describes.

The recheck asked the server for the chain height via `eth_blockNumber` — but in the reported failure that call is the very thing that is stale. If the reported height never catches up, `confirmed_at <= height` can never become true, so the watcher goes into `waiting_confs` and never resolves. The receipt was found; the height was what blocked it.

A receipt in block `n` already proves the chain is at least `n` blocks long, so the receipts can supply the height themselves. The `eth_blockNumber` request is now gone entirely — one fewer RPC call per recheck cycle, and the block stream already polls it.

To show the gap concretely, I added a regression test that mines the transaction's block while leaving the reported block number pinned at `0` — a server that is stale and never recovers, rather than one that catches up:

```
watch_resolves_when_block_number_never_catches_up
```

It fails on the previous commit of this branch:

```
panicked at crates/provider/src/heart.rs:1005:
watcher hung despite the transaction being confirmed: Elapsed(())
```

and passes after the change. The four existing stale-RPC tests still pass — they let the reported height catch up, which is why they did not catch this.

Two smaller fixes in the same commit:

- the next recheck is scheduled when the previous result **lands** rather than when it is spawned, so a recheck slower than `poll_interval` no longer triggers the next one back-to-back;
- the recheck task now always replies, so the loop cannot get stuck with `recheck_running` still set if the client has been dropped.

`RecheckResult` collapses to a plain vector now that it only carries the mined transactions.

230 `alloy-provider` lib tests pass, fmt and clippy clean. The `heart::` tests also drop from 6.80s to 0.97s, since the recheck no longer waits on the stale height response.

onbjerg:

PR #4167, opened 2026-08-26
fix(consensus): reject 0x00-tagged legacy transactions in typed_decode

### LINKED ISSUE #4165: [Bug] 0x00-tagged legacy transaction is accepted by decode_2718 and re-encodes untagged

```rust

fn main() {
    let tx = TxLegacy {
        chain_id: None,
        nonce: 0,
        gas_price: 10,
        gas_limit: 21_000,
        value: U256::ZERO,
        input: Default::default(),
    };

            Ok(env) => {
                let re = env.encoded_2718();
                println!(
                );
            }
        }
    }
    for ty in [0x05u8, 0x7f] {
        let mut b = vec![ty];
    }
}
```

```
```

Found by evmone-fuzz.

mayanksharma-eth:
## Summary

Per EIP-2718, `0x00` is not an assigned `TransactionType` — legacy transactions are the untagged form. `TxType::try_from(0)` returns `TxType::Legacy` (since `Legacy` is declared `#[envelope(ty = 0)]`), so the macro-generated `Decodable2718::typed_decode` accepted an EIP-2718 envelope whose type byte is a literal `0x00` and decoded the body as a legacy transaction — the same as `0x05`/`0x7f` should be, but currently aren't, rejected.

Because `EthereumTxEnvelope::Legacy` has no way to record that a tag byte was present on the wire, `encode_2718` always re-emits the untagged form, so a `0x00`-tagged input does not round-trip through `decode_2718`/`encode_2718`. geth, erigon, nethermind, and besu all reject the type byte at decode.

- Adds a guard to `Signed<T>::typed_decode` (`crates/consensus/src/signed.rs`) rejecting `ty == 0`, mirroring the guard already present in `fallback_decode` (added in #4090 for the mirror-case bug: untagged legacy incorrectly accepting a non-legacy RLP list). This fixes the issue for every envelope built via `#[derive(TransactionEnvelope)]`, matching the existing hand-written `ReceiptEnvelope::typed_decode`, which already rejects an explicit `0x00` type byte the same way.
- Updates the (previously inaccurate) doc comment on `EthereumTxEnvelope` to describe the corrected behavior.
- Adds regression tests: a top-level tagged/untagged round-trip test in `crates/consensus/src/transaction/envelope.rs`, and a flatten-composition test in `crates/consensus/src/transaction/mod.rs` mirroring the existing `flattened_variant_still_decodes_untagged_legacy` coverage from #4091.

Fixes #4165

## Test plan

- [x] `cargo test -p alloy-consensus --all-features` — 131 passed (0 failed), 22 doctests passed
- [x] `cargo test -p alloy-eips --all-features` — all passed
- [x] `cargo test -p alloy-network -p alloy-network-primitives -p alloy-consensus-any -p alloy-rpc-types -p alloy-rpc-types-eth --all-features` — all passed
- [x] `cargo clippy -p alloy-consensus --all-features -- -D warnings` — clean
- [x] `cargo fmt --check -p alloy-consensus` — clean
- [x] Verified the issue's repro script directly: `0x00`, `0x05`, and `0x7f`-tagged legacy transactions are now all consistently rejected with `UnexpectedType`, while genuinely untagged legacy transactions still decode and round-trip correctly.

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -543,6 +543,10 @@ where
     T: RlpEcdsaDecodableTx + Typed2718 + Send + Sync,
 {
     fn typed_decode(ty: u8, buf: &mut &[u8]) -> Eip2718Result<Self> {
+        if ty == 0 {
+            return Err(Eip2718Error::UnexpectedType(0));
+        }
+
         let decoded = T::rlp_decode_signed(buf)?;
 
         if decoded.ty() != ty {
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -14,11 +14,9 @@ use alloy_primitives::{Bytes, Signature, B256};
 ///
 /// # Note:
 ///
-/// This enum distinguishes between tagged and untagged legacy transactions, as
-/// the in-protocol merkle tree may commit to EITHER 0-prefixed or raw.
-/// Therefore we must ensure that encoding returns the precise byte-array that
-/// was decoded, preserving the presence or absence of the `TransactionType`
-/// flag.
+/// Represents an untagged legacy transaction, or a typed EIP-2718 variant. Binary decoding
+/// rejects a literal `0x00` type prefix; legacy transactions use the untagged fallback
+/// encoding.
 ///
 /// [EIP-2718]: https://eips.ethereum.org/EIPS/eip-2718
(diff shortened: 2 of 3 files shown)


mattsse:
Thanks for the PR. Superseded by #4197, which applies the guard in `Signed<T>::typed_decode` and also covers the receipt decoders (`EthereumReceipt`, `AnyReceiptEnvelope`) and `decode_unsigned`, with round-trip tests on every decode path.

PR #4168, opened 2026-08-26
fix(consensus): reject 0x00 type byte in EIP-2718 typed decoding

### LINKED ISSUE #4165: [Bug] 0x00-tagged legacy transaction is accepted by decode_2718 and re-encodes untagged

Ishant5436:
### Problem
Per EIP-2718, legacy transactions are strictly untagged RLP lists and `0x00` is not an assigned `TransactionType`. Previously, `EthereumTypedTransaction::decode_unsigned` and macro-derived `Decodable2718::typed_decode` accepted `0x00`-prefixed envelopes, decoding them as legacy transactions. When re-encoded via `encode_2718`, the untagged format was emitted, breaking round-trip serialization integrity and causing divergence during differential fuzzing (#4165).

### Solution
- Explicitly reject `ty == 0` in macro-generated `Decodable2718::typed_decode` and `Signed<T>::typed_decode` so that `0x00` cannot be decoded as a typed envelope.
- Removed the `0x00` match arm from `EthereumTypedTransaction::decode_unsigned` so `0x00` falls through to `Err(Eip2718Error::UnexpectedType(0x00))`.
- Added unit test `test_decode_2718_rejects_tagged_0x00_legacy` in `crates/consensus/src/transaction/envelope.rs`.

Fixes #4165

### DIFF
--- a/crates/consensus/src/signed.rs
+++ b/crates/consensus/src/signed.rs
@@ -543,6 +543,10 @@ where
     T: RlpEcdsaDecodableTx + Typed2718 + Send + Sync,
 {
     fn typed_decode(ty: u8, buf: &mut &[u8]) -> Eip2718Result<Self> {
+        if ty == 0 {
+            return Err(Eip2718Error::UnexpectedType(0));
+        }
+
         let decoded = T::rlp_decode_signed(buf)?;
 
         if decoded.ty() != ty {
--- a/crates/consensus/src/transaction/envelope.rs
+++ b/crates/consensus/src/transaction/envelope.rs
@@ -1085,6 +1085,33 @@ mod tests {
     }
 
     #[test]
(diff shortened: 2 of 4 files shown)


onbjerg:
dupe https://github.com/alloy-rs/alloy/pull/4167

PR #4171, opened 2026-08-28
feat(serde): lenient JSON-RPC param deserializers

mattsse:
Every JSON-RPC server built on alloy types has to be lenient about how integers arrive in a request's `params`. The same block number shows up as a JSON number (`100`), as a quantity string (`"0x64"`) or as a decimal string (`"100"`), and a single positional parameter arrives either bare (`100`) or wrapped in the one-element sequence the spec actually prescribes (`[100]`, `["0x64"]`). Nothing in `alloy-serde` covers this today: `quantity` handles the three scalar shapes for primitive integers but has no notion of the positional-params sequence, and there is no single generic entry point that also works for `U256`.

The evidence that this belongs upstream is that Foundry maintains its own copy in `crates/common/src/serde_helpers.rs` (https://github.com/foundry-rs/foundry/blob/master/crates/common/src/serde_helpers.rs), where anvil's `EthRequest` uses it on roughly thirty variants, and the number-or-hex half of it is duplicated a second time inside Foundry in `crates/config/src/utils.rs`.

This adds a `json_rpc` module following the layout of `quantity`: `json_rpc::deserialize` for the scalar shapes, `json_rpc::opt::deserialize` for the optional version, `json_rpc::seq::deserialize` for a value that may be wrapped in a one-element sequence, and `json_rpc::seq::opt::deserialize`, which additionally accepts `null`, `[]` and `[null]` as `None`. The helpers are generic over `u8` through `u128` and over `Uint` targets such as `U256`, so a single implementation serves both the `u64` and `U256` call sites that Foundry currently duplicates.

Deserialization delegates to the target type's own ruint-backed implementation rather than reimplementing hex parsing, which keeps error messages consistent with the rest of the crate and means overflow and malformed input are always errors rather than silent truncation: `"0x10000000000000000"` and `"18446744073709551616"` both fail for a `u64` target, as does `256` for a `u8` one. Sequences longer than one element are rejected with the observed length. Everything is deserialize-only, since serialization already has a canonical home in `quantity`, and the module stays within the crate's `no_std` plus `alloc` constraints.

The port is a behavioral superset of Foundry's helpers with one deliberate difference: Foundry's `deserialize_u64_seq_opt` rejects a bare value, accepting only sequences, while `seq::opt` here accepts a bare value the same way `seq` does. The asymmetry did not seem worth preserving in a general-purpose helper.

This PR was written with AI assistance (Claude Code); the design, review and verification are mine.

### DIFF
--- a/crates/serde/src/json_rpc.rs
+++ b/crates/serde/src/json_rpc.rs
@@ -35,7 +35,7 @@ use serde::Deserializer;
 /// # use serde::Deserialize;
 /// #[derive(Deserialize)]
 /// struct Params {
-///     #[serde(deserialize_with = "alloy_serde::lenient::deserialize")]
+///     #[serde(deserialize_with = "alloy_serde::json_rpc::deserialize")]
 ///     value: U256,
 /// }
 ///
@@ -53,7 +53,7 @@ where
 
 /// Serde functions for leniently deserializing optional numbers.
 ///
-/// See [`lenient`](crate::lenient) for more information.
+/// See [`json_rpc`](crate::json_rpc) for more information.
 pub mod opt {
     use super::private::{Lenient, LenientNumber};
     use serde::{Deserialize, Deserializer};
@@ -65,7 +65,7 @@ pub mod opt {
     /// # use serde::Deserialize;
     /// #[derive(Deserialize)]
     /// struct Params {
-    ///     #[serde(default, deserialize_with = "alloy_serde::lenient::opt::deserialize")]
+    ///     #[serde(default, deserialize_with = "alloy_serde::json_rpc::opt::deserialize")]
     ///     value: Option<u64>,
     /// }
     ///
@@ -88,7 +88,7 @@ pub mod opt {
 /// Serde functions for leniently deserializing a single number that may be wrapped in a
 /// one-element sequence, which is the JSON-RPC positional parameter shape.
 ///
-/// See [`lenient`](crate::lenient) for more information.
+/// See [`json_rpc`](crate::json_rpc) for more information.
 pub mod seq {
     use super::private::{LenientNumber, ValueOrSeq};
     use serde::Deserializer;
@@ -101,7 +101,7 @@ pub mod seq {
     /// ```
     /// # use serde::Deserialize;
     /// #[derive(Deserialize)]
-    /// struct Params(#[serde(deserialize_with = "alloy_serde::lenient::seq::deserialize")] u64);
+    /// struct Params(#[serde(deserialize_with = "alloy_serde::json_rpc::seq::deserialize")] u64);
     ///
     /// for raw in ["100", r#""0x64""#, "[100]", r#"["0x64"]"#] {
     ///     assert_eq!(serde_json::from_str::<Params>(raw).unwrap().0, 100);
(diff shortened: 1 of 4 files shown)


crates/serde/src/json_rpc.rs
```diff
@@ -0,0 +1,480 @@
...
+//! | [`deserialize`] | yes | yes | yes | no | no | no |
+//! | [`opt::deserialize`] | yes | yes | yes | no | yes | no |
+//! | [`seq::deserialize`] | yes | yes | yes | yes | no | no |
+//! | [`seq::opt::deserialize`] | yes | yes | yes | yes | yes | yes |
+//!
+//! Sequences of more than one element are rejected, as is any value that does not fit the target
+//! type: overflow and malformed input are always errors, never silent truncation.
+//!
+//! The helpers support `u8` through `u128` and [`Uint`](alloy_primitives::Uint) types such as
+//! [`U256`](alloy_primitives::U256). They are deserializers only; to serialize, use the
```
mattsse, line 21: `serde_json` only exposes integer tokens above `u64::MAX` through `deserialize_u128`, while the ruint visitor and both sequence entry points use `deserialize_any`. As a result, a bare `18446744073709551616` is rejected for `u128`/`U256`, even though the same value as a string works. Since this generic path cannot preserve both exact large integer tokens and string forms, could we document the boundary?

```suggestion
//! [`U256`](alloy_primitives::U256). Bare JSON number tokens are limited to values representable
//! as `u64`; larger values must use a decimal or hexadecimal string. They are deserializers only;
//! to serialize, use the
```

crates/serde/src/json_rpc.rs
```diff
@@ -0,0 +1,480 @@
...
+//!
+//! | Function | `100` | `"0x64"` | `"100"` | `[100]` | `null` | `[]`, `[null]` |
+//! |---|---|---|---|---|---|---|
+//! | [`deserialize`] | yes | yes | yes | no | no | no |
+//! | [`opt::deserialize`] | yes | yes | yes | no | yes | no |
+//! | [`seq::deserialize`] | yes | yes | yes | yes | no | no |
+//! | [`seq::opt::deserialize`] | yes | yes | yes | yes | yes | yes |
+//!
+//! Sequences of more than one element are rejected, as is any value that does not fit the target
+//! type: overflow and malformed input are always errors, never silent truncation.
```
mattsse, line 18: String parsing intentionally delegates to ruint's `Uint::from_str`, so it mirrors that parser's broader grammar rather than accepting only decimal and `0x` forms: `0b`/`0o` prefixes and `_` separators are accepted, and empty digit sequences such as `""`, `"_"`, or `"0x"` resolve to zero. Could we document that inherited behavior explicitly here? This does not need to diverge from ruint, but server users should be able to see the exact leniency contract.

DaniPopes:
maybe name it mod json_rpc?

mattsse:
Renamed to `alloy_serde::json_rpc` in 50409ea, so the helpers now read as `json_rpc::deserialize`, `json_rpc::opt`, `json_rpc::seq` and `json_rpc::seq::opt`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

DaniPopes:
thanks boss

onbjerg (approved):

PR #4172, opened 2026-08-28
feat(json-rpc): add RpcError::error_code

mattsse:
Some providers return JSON-RPC errors inside non-2xx HTTP responses. A missing-method error served as HTTP 403 becomes a transport error, so callers using `as_error_resp()` cannot inspect its JSON-RPC code.

Add `RpcError::error_code()` to return the code from either a regular error response or a transport error implementing the new `ErrorResponseCode` trait. The trait lives in `alloy-json-rpc`; `alloy-transport` implements it by parsing HTTP error bodies. Both nested error responses and bare error objects are supported, including bodies with trailing diagnostics. Codes use `i64` throughout, matching `ErrorPayload::code` and preserving application codes outside the `i32` range.

This supersedes #4169, which exposed extraction as a free function. Consumers such as Foundry can use `error.error_code()` instead of maintaining their own extraction helpers.

This PR was prepared with AI assistance.

### DIFF
--- a/crates/json-rpc/src/error.rs
+++ b/crates/json-rpc/src/error.rs
@@ -2,15 +2,20 @@ use crate::{ErrorPayload, RpcRecv};
 use alloy_primitives::B256;
 use serde_json::value::RawValue;
 
-/// A transport error kind that may carry the raw body of an error response.
+/// A transport error that may carry a JSON-RPC error in its raw response body.
 ///
 /// [`RpcError`] is generic over its transport error, which is defined by downstream transport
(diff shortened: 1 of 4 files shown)


onbjerg (changes requested):
Could you use `i64` throughout, matching `ErrorPayload::code`? The HTTP path currently narrows codes to `i32`, so the same code can be returned for a normal RPC error but disappear when it comes in an HTTP error body. JSON-RPC doesn't impose an `i32` limit on application error codes.

onbjerg (approved):

PR #4179, opened 2026-08-31
feat(rpc-types-engine): add amsterdam payload errors

Soubhik-10:
## PR Checklist

- [ ] Added Tests
- [ ] Added Documentation
- [ ] Breaking changes


mattsse (approved):
Reviewed the Amsterdam payload error additions and checked current reth main for downstream compatibility. Reth does not exhaustively match alloy_rpc_types_engine::PayloadError, so these variants do not introduce a reth compile break. The rpc-types-engine all-feature check and all 99 crate tests pass locally.

PR #4182, opened 2026-09-03
feat(genesis): support feature-gated account extensions

decofe:
Adds optional raw account payloads to GenesisAccount and RPC AccountInfo behind `account-ext`, preserving them through genesis root computation and RPC responses. Empty payloads are omitted from JSON. Depends on alloy-rs/trie#156.

### DIFF
--- a/crates/genesis/src/lib.rs
+++ b/crates/genesis/src/lib.rs
@@ -315,6 +315,7 @@ impl From<GenesisAccount> for TrieAccount {
             balance: account.balance,
             storage_root,
             code_hash,
+            extension: (),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


crates/genesis/src/lib.rs
```diff
@@ -315,6 +315,7 @@ impl From<GenesisAccount> for TrieAccount {
             balance: account.balance,
             storage_root,
             code_hash,
+            extension: (),
```
mediocregopher: GenesisAccount should properly support extensions in the genesis data, probably just as an RLP-encoded blob

mediocregopher: GenesisAccount now has an optional ThinArc-backed extension containing trailing RLP items, represented as hex in genesis JSON. The account-ext feature propagates it into the trie account and genesis state root; a regression checks that it changes the root and preserves the shared allocation.

mattsse (approved):

PR #4191, opened 2026-09-07
fix(provider): recover watch() when heartbeat skips blocks

### LINKED ISSUE #3881: [Bug] .watch() can hang when RPC block num is stale

### LINKED ISSUE #4181: PendingTransactionBuilder::watch can hang when the HTTP poller skips the mined block
## Describe the bug

`PendingTransactionBuilder::watch()` can wait indefinitely when an HTTP provider's block poll interval is longer than the chain's block time. If the heartbeat observes block N and then N+2, it never examines the transaction hashes in N+1. A transaction mined in N+1 therefore remains in the heartbeat's `unconfirmed` map even though `eth_getTransactionReceipt` already returns its receipt.

This is reproducible with `alloy-provider` v2.4.1 and is causing intermittent 30-minute CI timeouts on Tempo's 100ms local dev chain.

## Evidence

- [Failing Tempo CI job](https://github.com/tempoxyz/tempo/actions/runs/33450870922/job/99680319351): 1,435/1,436 tests completed, then the final test remained pending until the job's 30-minute timeout.
- [The hanging call site](https://github.com/tempoxyz/tempo/blob/782ca7cc53b6c749801effef01ae34ee7b645869/crates/node/tests/it/storage_credits.rs#L1418-L1425) uses `.watch().await` before fetching the receipt manually.
- [A subsequent run of the same shard passed](https://github.com/tempoxyz/tempo/actions/runs/33505361441/job/99848100928) in 5m32s, confirming the failure is intermittent.
- Tempo previously mitigated the same race at individual call sites by reducing the Alloy polling interval to 10ms ([example 1](https://github.com/tempoxyz/tempo/blob/782ca7cc53b6c749801effef01ae34ee7b645869/crates/node/tests/it/storage_credits.rs#L116-L122), [example 2](https://github.com/tempoxyz/tempo/blob/782ca7cc53b6c749801effef01ae34ee7b645869/crates/node/tests/it/storage_credits.rs#L1126-L1132)).

## Reproducer

The existing `test_send_tx` provider test can be adapted to make the block producer outrun the HTTP heartbeat:

```rust
#[tokio::test]
async fn watch_recovers_when_http_poller_skips_mined_block() {
    let provider = ProviderBuilder::new().connect_anvil_with_wallet_and_config(|anvil| {
        anvil.block_time_f64(0.1)
    });
    provider.client().set_poll_interval(Duration::from_secs(1));

    let tx = TransactionRequest {
        value: Some(U256::from(100)),
        ..Default::default()
    };

    let pending = provider.send_transaction(tx).await.unwrap();
    tokio::time::timeout(Duration::from_secs(5), pending.watch())
        .await
        .expect("watch hung after the mined block was skipped")
        .unwrap();
}
```

Run this repeatedly if needed; the bug is scheduling-sensitive. Increasing the poll interval or adding a short delay after submission makes the missed-block window wider.

## Root cause

In v2.4.1, [`watch()` only registers with and awaits the heartbeat](https://github.com/alloy-rs/alloy/blob/e1a217f10c059b87c0e741c36190c0940f5c8b9e/crates/provider/src/heart.rs#L203-L212). The heartbeat [checks transaction hashes only in blocks delivered by its stream](https://github.com/alloy-rs/alloy/blob/e1a217f10c059b87c0e741c36190c0940f5c8b9e/crates/provider/src/heart.rs#L625-L660). Its lookbehind contains only observed blocks, so detecting a height gap does not recover a transaction from a skipped block. Because the default pending-transaction timeout is `None`, the watcher can remain pending forever.

[`get_receipt()` already documents and mitigates this race](https://github.com/alloy-rs/alloy/blob/e1a217f10c059b87c0e741c36190c0940f5c8b9e/crates/provider/src/heart.rs#L225-L253) by polling `eth_getTransactionReceipt` independently ([#614](https://github.com/alloy-rs/alloy/pull/614)). The mitigation is not used by `watch()`.

## Recommended fix

Make `watch()` recover when the heartbeat skips heights. Two possible approaches:

1. Poll `eth_getTransactionReceipt` for each unconfirmed transaction, as `get_receipt()` already does; or
2. When a height gap is detected, fetch and process the missing blocks before advancing the heartbeat.

The first option reuses the existing mitigation and is likely the smaller change. Please also add a regression test with an Anvil block interval shorter than the HTTP poll interval. Lowering the client poll interval alone is not sufficient because scheduler or RPC latency can still produce a gap.

## Version

- `alloy-provider`: 2.4.1
- Transport: HTTP
- Runtime: Tokio

Aditya-9-6:
## Motivation

When using an HTTP provider, `PendingTransactionBuilder::watch()` can hang indefinitely if the block poll interval or RPC latency causes the heartbeat to skip past the block the transaction was mined in (or when RPC nodes return stale block numbers).

As documented in #4181 and #3881:
- The heartbeat only checks transactions in blocks delivered by its stream (`crates/provider/src/heart.rs#L625-L660`).
- If a block height gap occurs (e.g. block N then N+2), transactions mined in block N+1 remain in the unconfirmed map forever because the default timeout is `None`.
- `get_receipt()` already documents and mitigates this exact race condition by periodically polling `eth_getTransactionReceipt` on interval. However, `watch()` previously only awaited `self.register().await?.await`, lacking this recovery mechanism.

Fixes #4181
Fixes #3881

## Solution

Mirrors the fallback recovery pattern already used in `get_receipt()`:
- For 1 confirmation (`required_confirmations <= 1`), set up a periodic ticker based on `self.provider.client().poll_interval()`.
- In each tick, check `self.provider.get_transaction_receipt(hash)`. If a receipt is present, the transaction has already been mined in a block skipped by the stream, so return `Ok(hash)` immediately rather than hanging indefinitely.

## PR Checklist

- [x] Added Tests / verified against existing provider test suite
- [x] Documentation integrity preserved
- [x] `cargo clippy -p alloy-provider -- -D warnings` passes cleanly
- [x] `cargo fmt --check -p alloy-provider` passes cleanly

### DIFF
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -208,7 +208,38 @@ impl<N: Network> PendingTransactionBuilder<N> {
     /// - [`get_receipt`](Self::get_receipt) for fetching the receipt after the transaction has been
     ///   confirmed.
     pub async fn watch(self) -> Result<TxHash, PendingTransactionError> {
-        self.register().await?.await
+        let hash = self.config.tx_hash;
+        let required_confirmations = self.config.required_confirmations;
+        let mut pending_tx = self.provider.watch_pending_transaction(self.config).await?;
+
+        // Fallback polling for single confirmation to prevent hangs when the heartbeat
+        // misses or skips the block the tx was mined in (e.g. HTTP poller intervals).
+        let mut interval = if required_confirmations > 1 {
+            None
+        } else {
+            Some(interval(self.provider.client().poll_interval()))
+        };
+
+        loop {
+            let tick_fut = if let Some(interval) = interval.as_mut() {
+                interval.tick().map(|_| ()).left_future()
+            } else {
+                pending::<()>().right_future()
+            };
+
+            select! {
+                _ = tick_fut => {},
+                res = &mut pending_tx => {
+                    return res;
+                }
+            }
+
+            // If the transaction has already been mined in a block skipped by the heartbeat,
+            // return its hash.
+            if let Ok(Some(_)) = self.provider.get_transaction_receipt(hash).await {
+                return Ok(hash);
+            }
+        }
     }
 
     /// Waits for the transaction to confirm with the given number of confirmations, and


onbjerg:

PR #4194, opened 2026-09-08
fix(provider): recover missed transactions in heartbeat

### LINKED ISSUE #3408: [Bug] `watch_pending_transaction` will never pause heartbeat if transaction is missed due to reorg when calling `get_receipt`
### Component

provider, pubsub

1.1.1

### Operating System

Linux

### Describe the bug

Here's an example where the block was missed on arbitrum:

```
2025-12-22T23:56:36.636352Z DEBUG alloy_provider::blocks: yielding block number=413480103
2025-12-22T23:56:36.636359Z DEBUG alloy_provider::heart: handling block block_height=413480103
2025-12-22T23:56:36.947322Z DEBUG alloy_provider::heart: watching tx=0xa957513e3bae1cd9187e65c00bfa33e15d9b3ebb99f96c4c735cd51aaf4e6c75
2025-12-22T23:56:36.947360Z DEBUG alloy_provider::heart: updating heartbeat pause state paused=false
2025-12-22T23:56:38.502948Z DEBUG alloy_provider::blocks: fetching block number=413480111
2025-12-22T23:56:38.522697Z DEBUG alloy_provider::blocks: fetching block number=413480112
2025-12-22T23:56:38.541916Z DEBUG alloy_provider::blocks: yielding block number=413480111
2025-12-22T23:56:38.541927Z DEBUG alloy_provider::heart: handling block block_height=413480111
2025-12-22T23:56:38.541938Z DEBUG alloy_provider::heart: reorg/unpause detected block_height=413480111 last_height=413480103
2025-12-22T23:56:38.541986Z DEBUG alloy_provider::blocks: yielding block number=413480112
2025-12-22T23:56:38.541994Z DEBUG alloy_provider::heart: handling block block_height=413480112
```

The transaction I was waiting for a receipt for here was in `413480106` and it causes `watch_pending_transaction` to poll forever since the transaction is missed.

 
https://github.com/alloy-rs/alloy/blob/174b64362a61010f77f0ada7a4e47f2cd88a357c/crates/provider/src/heart.rs#L228

### LINKED ISSUE #3881: [Bug] .watch() can hang when RPC block num is stale

### LINKED ISSUE #3916: [Bug] PendingTransactionBuilder::get_receipt() can time out for >1 confirmations even when the receipt already exists

### LINKED ISSUE #4181: PendingTransactionBuilder::watch can hang when the HTTP poller skips the mined block

mattsse:
Pending transaction watches can wait forever when the heartbeat never sees the mined block. `NewBlocks` backfills ordinary height gaps, but starts or resumes near the tip: if a transaction lands in block 1 before the stream resumes at blocks 19–20, the heartbeat leaves it unconfirmed even though its receipt is available. The existing single-confirmation `get_receipt()` workaround could return successfully while leaving that watcher alive.

Move receipt recovery into the heartbeat so `register()`, `watch()`, and `get_receipt()` share it. Recheck each pending hash on the client polling interval with bounded concurrency, independently of block processing and timeout handling. Single-confirmation watchers resolve from the receipt without waiting for the head RPC; multiple-confirmation watchers also require the appropriate chain height. Successful recovery removes the watcher, allowing the heartbeat to pause, and replaces the separate `get_receipt()` workaround.

Keep duplicate registrations and their deadlines independent, enforce timeouts after inclusion as well as before it, and prune cancelled watchers. Clear obsolete inclusion heights on detected reorgs and make the pause notification wait safe when recovery pauses the heartbeat again before the block stream resumes. Public APIs and configuration remain unchanged.

Fixes #4181
Fixes #3881
Fixes #3916
Fixes #3408

Supersedes the following PRs:

- Closes #4191, closes #4013 — receipt fallback for `.watch()` when blocks are missed or the RPC head is stale.
- Closes #4162 — heartbeat-level receipt rechecking.
- Closes #4065, closes #3996 — receipt recovery while waiting for multiple confirmations.
- Closes #3992, closes #3729 — enforcing timeouts after inclusion while waiting for confirmations.
- Closes #3551 — the large-gap workaround for #3408; this PR recovers missed receipts instead of introducing an automatic timeout for large gaps.

#3771 (default timeout policy) and #2689 (mempool-drop detection) remain separate work and are not superseded by this PR.

### DIFF
--- a/crates/provider/src/blocks.rs
+++ b/crates/provider/src/blocks.rs
@@ -49,12 +49,18 @@ impl Paused {
     /// Returns `true` if the method actually waited for the paused state to become unpaused,
     /// or `false` if it was already unpaused when called.
     async fn wait(&self) -> bool {
-        if !self.is_paused() {
-            return false;
+        let mut waited = false;
+        loop {
+            // Create the notification before checking the flag so notify_waiters cannot
+            // fall between the check and registration. Recheck after waking: receipt
+            // recovery may already have paused the heartbeat again.
+            let notified = self.notify.notified();
+            if !self.is_paused() {
+                return waited;
+            }
+            notified.await;
+            waited = true;
         }
-        self.notify.notified().await;
-        debug_assert!(!self.is_paused());
-        true
     }
 }
 
@@ -309,5 +315,25 @@ mod tests {
         for (i, block) in blocks.iter().enumerate() {
             assert_eq!(block.header.number, first + i as u64);
         }
+    }
+}
+
+#[cfg(test)]
+mod pause_tests {
+    use super::*;
+    use futures::FutureExt;
+
+    #[tokio::test]
+    async fn pause_again_before_stream_resumes() {
+        let paused = Paused::default();
+        paused.set_paused(true);
+        let mut wait = Box::pin(paused.wait());
+        assert!((&mut wait).now_or_never().is_none());
+        paused.set_paused(false);
+        paused.set_paused(true);
+        assert!((&mut wait).now_or_never().is_none());
+        paused.set_paused(false);
+        assert!(wait.await);
+        assert!(!paused.wait().await);
     }
 }
--- a/crates/provider/src/heart.rs
+++ b/crates/provider/src/heart.rs
@@ -3,13 +3,17 @@
 use crate::{blocks::Paused, Provider, RootProvider};
 use alloy_consensus::BlockHeader;
 use alloy_json_rpc::RpcError;
-use alloy_network::{BlockResponse, Network};
+use alloy_network::{BlockResponse, Network, ReceiptResponse};
 use alloy_primitives::{
     map::{B256HashMap, B256HashSet},
-    TxHash, B256,
+    TxHash, U64,
 };
+use alloy_rpc_client::WeakClient;
 use alloy_transport::{utils::Spawnable, TransportError};
-use futures::{future::pending, stream::StreamExt, FutureExt, Stream};
+use futures::{
+    stream::{FusedStream, StreamExt},
+    FutureExt, Stream,
+};
 use std::{
     collections::{BTreeMap, VecDeque},
     fmt,
@@ -25,14 +29,19 @@ use tokio::{
 #[cfg(all(target_family = "wasm", target_os = "unknown"))]
 use wasmtimer::{
     std::Instant,
-    tokio::{interval, sleep_until},
+    tokio::{sleep_until, timeout},
 };
 
 #[cfg(not(all(target_family = "wasm", target_os = "unknown")))]
 use {
     std::time::Instant,
-    tokio::time::{interval, sleep_until},
+    tokio::time::{sleep_until, timeout},
 };
+
+#[cfg(not(target_family = "wasm"))]
+use futures::stream::BoxStream;
+#[cfg(target_family = "wasm")]
+use futures::stream::LocalBoxStream as BoxStream;
 
 /// Errors which may occur when watching a pending transaction.
 #[derive(Debug, thiserror::Error)]
@@ -224,48 +233,8 @@ impl<N: Network> PendingTransactionBuilder<N> {
     /// - [`watch`](Self::watch) for watching the transaction without fetching the receipt.
     pub async fn get_receipt(self) -> Result<N::ReceiptResponse, PendingTransactionError> {
         let hash = self.config.tx_hash;
-        let required_confirmations = self.config.required_confirmations;
-        let mut pending_tx = self.provider.watch_pending_transaction(self.config).await?;
-
-        // FIXME: this is a hotfix to prevent a race condition where the heartbeat would miss the
-        // block the tx was mined in. Only apply this for single confirmation to respect the
-        // confirmation setting.
-        let mut interval = if required_confirmations > 1 {
-            None
-        } else {
-            Some(interval(self.provider.client().poll_interval()))
-        };
-
-        loop {
-            let mut confirmed = false;
-
-            // If more than 1 block confirmations is specified then we can rely on the regular
-            // watch_pending_transaction and dont need this workaround for the above mentioned race
-            // condition
-            let tick_fut = if let Some(interval) = interval.as_mut() {
-                interval.tick().map(|_| ()).left_future()
-            } else {
-                pending::<()>().right_future()
-            };
-
-            select! {
-                _ = tick_fut => {},
-                res = &mut pending_tx => {
-                    let _ = res?;
-                    confirmed = true;
-                }
-            }
-
-            // try to fetch the receipt
-            let receipt = self.provider.get_transaction_receipt(hash).await?;
-            if let Some(receipt) = receipt {
-                return Ok(receipt);
-            }
-
-            if confirmed {
-                return Err(RpcError::NullResp.into());
-            }
-        }
+        self.provider.watch_pending_transaction(self.config).await?.await?;
+        self.provider.get_transaction_receipt(hash).await?.ok_or_else(|| RpcError::NullResp.into())
     }
 }
 
@@ -379,6 +348,8 @@ struct TxWatcher {
     /// Invariant: any confirmed transaction in `Heart` has this value set.
     received_at_block: Option<u64>,
     tx: oneshot::Sender<Result<(), WatchTxError>>,
+    /// Each registration has its own deadline, including duplicate transaction hashes.
+    deadline: Option<Instant>,
 }
 
 impl TxWatcher {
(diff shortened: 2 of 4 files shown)


mattsse:
Once this PR is merged, we can close the following superseded PRs:

- #4191 and #4013 — receipt fallback for `.watch()` when blocks are missed or the RPC head is stale.
- #4162 — heartbeat-level receipt rechecking.
- #4065 and #3996 — receipt recovery while waiting for multiple confirmations.
- #3992 and #3729 — enforcing timeouts after inclusion while waiting for confirmations.
- #3551 — the large-gap workaround for #3408; this PR recovers missed receipts instead of introducing an automatic timeout for large gaps.

#3771 (default timeout policy) and #2689 (mempool-drop detection) remain separate work and are not superseded by this PR.

Issue #852, opened 2024-06-07
[Bug] ProviderBuilder with Hyper and Https


DaveWK:
### Component

provider, pubsub

### What version of Alloy are you on?

0.1.0

### Operating System

Linux

### Describe the bug

it seems that https does not work with the hyper feature at the moment. I see:
`Err(Transport(Custom(hyper_util::client::legacy::Error(Connect, \"invalid URL, scheme is not http\"))))`

when attempting to pass an https url to ProviderBuilder. This appears to be due to a nuisance with how hyper separates out the tls implementation by default.

My preference would be that alloy-rs enables a tls implementation for hyper by default to abstract this implementation detail, but would also be appreciative for some documentation on how to specify and enable the tls for the alloy-rs's hyper feature.

HrikB:
I also get this issue when using reqwest. Http urls work fine but errors out on using https

DaniPopes:
This has been enabled by default for some time, at least 0.2.0

PR #4195, opened 2026-09-09
fix(trace): decode regularGasUsed and explain gas

mattsse:
Accept both `regularGasUsed` (the execution-apis proposal) and `executionGasUsed` when decoding state-gas, call, and opcode traces. Keep serialization as `executionGasUsed` for compatibility. This prevents proposal-shaped responses from failing to decode or silently losing the optional execution-gas breakdown.

Clarify EIP-8037 gas accounting across transaction traces, opcode logs, call frames, receipts, block headers, transaction limits, and estimation APIs. Explain regular/execution gas versus net state gas, reservoir spillover, state refills versus ordinary refunds, transaction-versus-block totals, calldata-floor caveats, and inactive-network/optional-field behavior. Link each relevant surface to the EIP and add a tracing guide directing users to receipt gas for charges and node estimation for required limits.

The matching Foundry guide is in foundry-rs/book#2076, with cheatcode documentation in foundry-rs/foundry#16772 and foundry-rs/forge-std#910.

AI assistance: Codex authored the documentation, decoding aliases, regression test, and PR text.

### DIFF
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -55,10 +55,17 @@ pub struct Header {
     /// zero; formally Hi.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub number: BlockNumber,
-    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
+    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
+    /// the accumulated regular (execution) and state gas dimensions, not their sum.
+    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob
+    /// gas.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub gas_limit: u64,
-    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
+    /// Gas used for block capacity and base-fee accounting.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
+    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
+    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
+    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub gas_used: u64,
     /// A scalar value equal to the reasonable output of Unix’s time() at this block’s inception;
@@ -647,10 +654,16 @@ pub trait BlockHeader {
     /// Retrieves the block number
     fn number(&self) -> BlockNumber;
 
-    /// Retrieves the gas limit of the block
+    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
+    /// the accumulated regular (execution) and state gas dimensions, not their sum.
+    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob gas.
     fn gas_limit(&self) -> u64;
 
-    /// Retrieves the gas used by the block
+    /// Gas used for block capacity and base-fee accounting.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
+    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
+    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
+    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
     fn gas_used(&self) -> u64;
 
     /// Retrieves the timestamp of the block
--- a/crates/consensus/src/block/header.rs
+++ b/crates/consensus/src/block/header.rs
@@ -55,17 +55,16 @@ pub struct Header {
     /// zero; formally Hi.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub number: BlockNumber,
-    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
-    /// the accumulated regular (execution) and state gas dimensions, not their sum.
-    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob
-    /// gas.
+    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it bounds each of the block's two
+    /// gas dimensions (execution and state) separately, not their sum.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub gas_limit: u64,
-    /// Gas used for block capacity and base-fee accounting.
-    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
-    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
-    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
-    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
+    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this is the larger of the block's
+    /// execution-gas and state-gas totals, counted before refunds ([EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)),
+    /// so it need not equal the last receipt's `cumulativeGasUsed`. State totals remain net of
+    /// state refills.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub gas_used: u64,
     /// A scalar value equal to the reasonable output of Unix’s time() at this block’s inception;
@@ -654,16 +653,16 @@ pub trait BlockHeader {
     /// Retrieves the block number
     fn number(&self) -> BlockNumber;
 
-    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
-    /// the accumulated regular (execution) and state gas dimensions, not their sum.
-    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob gas.
+    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it bounds each of the block's two
+    /// gas dimensions (execution and state) separately, not their sum.
     fn gas_limit(&self) -> u64;
 
-    /// Gas used for block capacity and base-fee accounting.
-    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
-    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
-    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
-    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
+    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this is the larger of the block's
+    /// execution-gas and state-gas totals, counted before refunds ([EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)),
+    /// so it need not equal the last receipt's `cumulativeGasUsed`. State totals remain net of
+    /// state refills.
     fn gas_used(&self) -> u64;
 
     /// Retrieves the timestamp of the block
--- a/crates/consensus/src/receipt/mod.rs
+++ b/crates/consensus/src/receipt/mod.rs
@@ -89,7 +89,12 @@ pub trait TxReceipt: Clone + fmt::Debug + PartialEq + Eq + Send + Sync {
         ReceiptWithBloom { logs_bloom, receipt: self }
     }
 
-    /// Returns the cumulative gas used in the block after this transaction was executed.
+    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
+    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
+    /// active, both regular (execution) and net state gas, after ordinary refunds and the calldata
+    /// floor. Excludes blob gas. This is not just this transaction's gas or the EIP-8037 block
+    /// header's `gasUsed`, which uses the maximum of the block's two dimensions instead of
+    /// their sum.
     fn cumulative_gas_used(&self) -> u64;
 
     /// Returns the logs emitted by this transaction.
--- a/crates/consensus/src/receipt/mod.rs
+++ b/crates/consensus/src/receipt/mod.rs
@@ -89,12 +89,11 @@ pub trait TxReceipt: Clone + fmt::Debug + PartialEq + Eq + Send + Sync {
         ReceiptWithBloom { logs_bloom, receipt: self }
     }
 
-    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
-    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
-    /// active, both regular (execution) and net state gas, after ordinary refunds and the calldata
-    /// floor. Excludes blob gas. This is not just this transaction's gas or the EIP-8037 block
-    /// header's `gasUsed`, which uses the maximum of the block's two dimensions instead of
-    /// their sum.
+    /// Returns the cumulative gas used in the block up to and including this transaction.
+    ///
+    /// This is charged gas, after refunds and the calldata floor. Under
+    /// [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it includes both execution and state gas, so
+    /// the last receipt's value need not equal the block header's `gasUsed`.
     fn cumulative_gas_used(&self) -> u64;
 
     /// Returns the logs emitted by this transaction.
--- a/crates/consensus/src/receipt/receipt2.rs
+++ b/crates/consensus/src/receipt/receipt2.rs
@@ -61,7 +61,7 @@ pub struct EthereumReceipt<T = TxType, L = Log> {
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity", rename = "status"))]
     pub success: bool,
     /// Cumulative gas used in the block up to and including this transaction.
-    /// See [`TxReceipt::cumulative_gas_used`](crate::TxReceipt::cumulative_gas_used).
+    /// See [`TxReceipt::cumulative_gas_used`].
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub cumulative_gas_used: u64,
     /// Log send from contracts.
--- a/crates/consensus/src/receipt/receipt2.rs
+++ b/crates/consensus/src/receipt/receipt2.rs
@@ -60,7 +60,12 @@ pub struct EthereumReceipt<T = TxType, L = Log> {
     /// This is the `statusCode`
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity", rename = "status"))]
     pub success: bool,
-    /// Gas used
+    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
+    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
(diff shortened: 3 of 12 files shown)


mattsse (commented):
The `state_gas` module doc is the right canonical place for the two-dimensional model and reads well; the `estimate_gas` note (alloy forwards, does not add state gas) and the "`None` means not reported" lines on the trace fields are exactly what users will need. Two kinds of suggestions below.

First, `alias = "regularGasUsed"` on the three execution-gas fields: the execution-apis proposal spells the field that way in every revision, alloy chose `executionGasUsed` in #4113, and the module doc currently states that difference backwards. The alias makes decoding work against either kind of node and lets the doc say something true.

Second, the hot types (`Header`, both receipt structs, `ReceiptResponse`, `TransactionRequest`, `Transaction::gas_limit`) now carry five-line EIP paragraphs that every hover shows and that repeat each other. The suggestions bring each back to its canonical one-line definition plus a single EIP-8037 sentence, with the full story staying in the module doc.

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -11,26 +11,52 @@ use serde::{Deserialize, Serialize};
...
-    /// See [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) and the
-    /// [execution-apis call tracer proposal](https://github.com/ethereum/execution-apis/pull/852).
+    /// Includes intrinsic gas and the execution-dimension calldata floor; before ordinary refunds
+    /// under EIP-7778. This is not receipt gas. See the
+    /// [gas accounting guide](crate::geth::state_gas) and
+    /// [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).
+    ///
+    /// A transaction summary on the top-level frame when supported; not a per-child breakdown.
+    /// `None` means not reported (for example, a nested frame, older fork, or unsupported node).
     #[serde(default, rename = "executionGasUsed", skip_serializing_if = "Option::is_none")]
```
mattsse: The execution-apis proposal spells this field `regularGasUsed` (both commits of ethereum/execution-apis#852, and it never uses `executionGasUsed`); `executionGasUsed` came in with #4113 and is what revm-inspectors emits. A node implementing the proposal as written decodes to `None` here. Accepting both on input costs nothing and keeps the wire name for serialization:

```suggestion
    #[serde(
        default,
        rename = "executionGasUsed",
        alias = "regularGasUsed",
        skip_serializing_if = "Option::is_none"
    )]
```

crates/rpc-types-trace/src/geth/mod.rs
```diff
@@ -74,30 +74,44 @@ pub struct ChainBlockTraceResult {
...
-    /// can require downstream struct literals to use `..Default::default()`.
+    /// Includes intrinsic gas and the execution-dimension calldata floor; before ordinary refunds
+    /// under EIP-7778. This is not receipt gas. See the
+    /// [gas accounting guide](crate::geth::state_gas) and
+    /// [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).
+    ///
+    /// Optional because older forks and nodes can omit this field; `None` does not mean zero.
     #[serde(
         default,
         with = "alloy_serde::quantity::opt",
```
mattsse, line 91: Same alias as on `CallFrame`, so `DefaultFrame` decodes both spellings:

```suggestion
        with = "alloy_serde::quantity::opt",
        alias = "regularGasUsed",
```

crates/rpc-types-trace/src/geth/state_gas.rs
```diff
@@ -1,24 +1,95 @@
...
     pub gas_used: u64,
-    /// Gross execution-dimension gas used by the transaction.
+    /// Regular (execution) gas contribution to block accounting, excluding EIP-8037 state gas.
+    ///
+    /// Includes intrinsic gas and the execution-dimension calldata floor. With
+    /// [EIP-7778](https://eips.ethereum.org/EIPS/eip-7778), this is before ordinary transaction refunds.
+    /// Without EIP-8037, state creation remains part of the ordinary gas schedule. A floor can
+    /// make this value larger than the execution work actually consumed; use `gas_used` for
+    /// receipt gas.
     #[serde(with = "alloy_serde::quantity")]
```
mattsse: Same alias for the `stateGasTracer` response:

```suggestion
    #[serde(with = "alloy_serde::quantity", alias = "regularGasUsed")]
```

crates/rpc-types-trace/src/geth/state_gas.rs
```diff
@@ -1,24 +1,95 @@
...
+//! activation. Before activation, state creation uses the ordinary gas schedule and the separate
+//! state dimension is zero. A node may omit optional breakdown fields or reject `stateGasTracer`
+//! if it does not implement the tracer. `None` means not reported, not zero.
+//!
+//! The tracing wire format is still described by an
+//! [execution-apis proposal](https://github.com/ethereum/execution-apis/pull/852); support and
+//! historical responses depend on the node. Alloy uses the wire name `executionGasUsed`; older
+//! proposal/node revisions may use `regularGasUsed`, which these fields do not alias. Match the
+//! node's response schema to the Alloy version. These types decode responses; they do not enable
+//! EIPs.
```
mattsse: This has the direction backwards: `regularGasUsed` is the *current* proposal spelling (both commits of #852, 2 and 6 July), not an older revision, and `executionGasUsed` is alloy's own choice from #4113 (11 August). With the aliases suggested on the three fields the paragraph can say what is true and stop hedging:

```suggestion
//! The tracing wire format is still described by an
//! [execution-apis proposal](https://github.com/ethereum/execution-apis/pull/852), which names the
//! execution dimension `regularGasUsed`. Alloy serializes `executionGasUsed` and accepts both names
//! when decoding. Support and historical responses depend on the node; these types decode
//! responses, they do not enable EIPs.
```

crates/consensus/src/block/header.rs
```diff
@@ -55,10 +55,17 @@ pub struct Header {
     /// zero; formally Hi.
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub number: BlockNumber,
-    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
+    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
+    /// the accumulated regular (execution) and state gas dimensions, not their sum.
+    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob
+    /// gas.
```
mattsse: Every consumer of `Header` hovers this. Keep the canonical one-line definition (the "formally Hl" anchor was dropped) and add one EIP-8037 sentence:

```suggestion
    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it bounds each of the block's two
    /// gas dimensions (execution and state) separately, not their sum.
```

crates/consensus/src/block/header.rs
```diff
@@ -55,10 +55,17 @@ pub struct Header {
...
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub gas_limit: u64,
-    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
+    /// Gas used for block capacity and base-fee accounting.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
+    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
+    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
+    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
```
mattsse: "EIP-7778 keeps ordinary refunds in block usage" is a confusing paraphrase of an EIP whose rule is that refunds are *excluded* from block accounting, i.e. block gas is counted before refunds. Restore "formally Hg" and say it directly:

```suggestion
    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this is the larger of the block's
    /// execution-gas and state-gas totals, counted before refunds ([EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)),
    /// so it need not equal the last receipt's `cumulativeGasUsed`.
```

crates/consensus/src/block/header.rs
```diff
@@ -647,10 +654,16 @@ pub trait BlockHeader {
     /// Retrieves the block number
     fn number(&self) -> BlockNumber;
 
-    /// Retrieves the gas limit of the block
+    /// Block gas limit; under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), bounds each of
+    /// the accumulated regular (execution) and state gas dimensions, not their sum.
+    /// Without EIP-8037 this is the ordinary single-dimensional block gas limit. Excludes blob gas.
```
mattsse: Mirror of the struct field:

```suggestion
    /// A scalar value equal to the current limit of gas expenditure per block; formally Hl.
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it bounds each of the block's two
    /// gas dimensions (execution and state) separately, not their sum.
```

crates/consensus/src/block/header.rs
```diff
@@ -647,10 +654,16 @@ pub trait BlockHeader {
...
     fn gas_limit(&self) -> u64;
 
-    /// Retrieves the gas used by the block
+    /// Gas used for block capacity and base-fee accounting.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), the maximum of the accumulated
+    /// regular (execution) and state gas dimensions, not their sum or cumulative receipt gas.
+    /// EIP-7778 keeps ordinary refunds in block usage; state gas is still net of state refills.
+    /// Without these EIPs, the network's ordinary block gas rules apply. Excludes blob gas.
```
mattsse: Mirror of the struct field:

```suggestion
    /// A scalar value equal to the total gas used in transactions in this block; formally Hg.
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this is the larger of the block's
    /// execution-gas and state-gas totals, counted before refunds ([EIP-7778](https://eips.ethereum.org/EIPS/eip-7778)),
    /// so it need not equal the last receipt's `cumulativeGasUsed`.
```

crates/consensus/src/receipt/mod.rs
```diff
@@ -89,7 +89,12 @@ pub trait TxReceipt: Clone + fmt::Debug + PartialEq + Eq + Send + Sync {
         ReceiptWithBloom { logs_bloom, receipt: self }
     }
 
-    /// Returns the cumulative gas used in the block after this transaction was executed.
+    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
+    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
+    /// active, both regular (execution) and net state gas, after ordinary refunds and the calldata
+    /// floor. Excludes blob gas. This is not just this transaction's gas or the EIP-8037 block
+    /// header's `gasUsed`, which uses the maximum of the block's two dimensions instead of
+    /// their sum.
```
mattsse: The same six lines are copied to four places (this trait, both receipt structs, `ReceiptResponse`). Keep the full explanation once, here, and point the others at it:

```suggestion
    /// Returns the cumulative gas used in the block up to and including this transaction.
    ///
    /// This is charged gas, after refunds and the calldata floor. Under
    /// [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) it includes both execution and state gas, so
    /// the last receipt's value need not equal the block header's `gasUsed`.
```

crates/consensus/src/receipt/receipt2.rs
```diff
@@ -60,7 +60,12 @@ pub struct EthereumReceipt<T = TxType, L = Log> {
     /// This is the `statusCode`
     #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity", rename = "status"))]
     pub success: bool,
-    /// Gas used
+    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
+    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
+    /// active, both regular (execution) and net state gas, after ordinary refunds and the calldata
+    /// floor. Excludes blob gas. This is not just this transaction's gas or the EIP-8037 block
+    /// header's `gasUsed`, which uses the maximum of the block's two dimensions instead of
+    /// their sum.
```
mattsse:
```suggestion
    /// Cumulative gas used in the block up to and including this transaction.
    /// See [`TxReceipt::cumulative_gas_used`](crate::TxReceipt::cumulative_gas_used).
```

crates/consensus/src/receipt/receipts.rs
```diff
@@ -24,7 +24,12 @@ pub struct Receipt<T = Log> {
     /// This is the `statusCode`
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub status: Eip658Value,
-    /// Gas used
+    /// Cumulative charged receipt gas through this transaction, including this transaction's gas.
+    /// Includes intrinsic costs and, where [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is
+    /// active, both regular (execution) and net state gas, after ordinary refunds and the calldata
+    /// floor. Excludes blob gas. This is not just this transaction's gas or the EIP-8037 block
+    /// header's `gasUsed`, which uses the maximum of the block's two dimensions instead of
+    /// their sum.
```
mattsse:
```suggestion
    /// Cumulative gas used in the block up to and including this transaction.
    /// See [`TxReceipt::cumulative_gas_used`](crate::TxReceipt::cumulative_gas_used).
```

crates/network-primitives/src/traits.rs
```diff
@@ -56,13 +56,17 @@ pub trait ReceiptResponse {
     /// Index within the block.
     fn transaction_index(&self) -> Option<u64>;
 
-    /// Gas used by this transaction alone.
+    /// Total gas charged for this transaction alone, after refunds and the applicable calldata
+    /// floor. Includes intrinsic costs and both regular (execution) and net state gas where
+    /// [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) is active. Without it, state creation costs
+    /// use the ordinary gas schedule. Excludes blob gas. This is not a gas-limit estimate.
```
mattsse:
```suggestion
    /// Gas used by this transaction alone, as charged (after refunds and the calldata floor).
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this includes both execution and state gas.
```

crates/network-primitives/src/traits.rs
```diff
@@ -83,7 +87,10 @@ pub trait ReceiptResponse {
     /// Address of the receiver, or `None` for contract creation.
     fn to(&self) -> Option<Address>;
 
-    /// Returns the gas used in the block up to and including this transaction.
+    /// Returns cumulative receipt gas through this transaction, including this transaction's gas.
+    /// Under EIP-8037 this sums charged regular and state gas. It is not the block header's
+    /// `gasUsed`, which uses the larger of the block's two gas dimensions, with its own
+    /// refund/floor rules.
```
mattsse:
```suggestion
    /// Returns the gas used in the block up to and including this transaction.
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this is the sum of charged execution and
    /// state gas, so it need not match the block header's `gasUsed`.
```

crates/consensus/src/transaction/mod.rs
```diff
@@ -91,7 +91,11 @@ pub trait Transaction: Typed2718 + fmt::Debug + any::Any + Send + Sync + 'static
     /// Get `nonce`.
     fn nonce(&self) -> u64;
 
-    /// Get `gas_limit`.
+    /// Combined transaction gas limit, including intrinsic costs.
+    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), this funds regular (execution) and
+    /// state gas through the reservoir model; the regular gas cap still applies independently.
+    /// Without EIP-8037, state creation uses the ordinary gas schedule. Excludes blob gas.
+    /// Gas charged after refunds or state refills need not be a sufficient limit for execution.
```
mattsse: "Gas charged after refunds or state refills need not be a sufficient limit for execution" on the getter of a *signed* transaction's gas limit is advice the reader cannot act on here. The estimate guidance belongs on `TransactionRequest::gas` and `Provider::estimate_gas`, where it already is.

```suggestion
    /// Get `gas_limit`.
    ///
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this single limit funds both execution
    /// and state gas via the reservoir model.
```

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -74,7 +74,14 @@ pub struct TransactionRequest {
...
     pub max_fee_per_blob_gas: Option<u128>,
-    /// The gas limit for the transaction.
+    /// The combined gas limit for the transaction, including intrinsic costs.
+    ///
+    /// On networks with [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), this funds both regular
+    /// (execution) gas and state gas. The protocol splits it into regular gas and a state
+    /// reservoir; callers still supply one limit. The regular gas cap applies independently.
+    /// Without EIP-8037, state creation uses the ordinary gas schedule with no separate reservoir.
+    /// Use `eth_estimateGas` with the target network's rules. Receipt gas or net trace consumption
+    /// may be too low because gas can be required before a later refund or state refill.
```
mattsse: Keeps the two facts a caller needs (one limit funds both dimensions; charged gas can be too low as a limit) and drops the restatements:

```suggestion
    /// The gas limit for the transaction, including intrinsic costs.
    ///
    /// Under [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037) this single limit funds both execution
    /// and state gas; the protocol splits it into regular gas and a state reservoir. A receipt's
    /// `gasUsed` or a trace's net consumption can be too low as a limit, because gas may be needed
    /// before a later refund or state refill. Use `eth_estimateGas` against the target network.
```

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -212,7 +219,9 @@ impl TransactionRequest {
         self
     }
 
-    /// Sets the gas limit for the transaction.
+    /// Sets the combined transaction gas limit, including intrinsic, regular, and EIP-8037 state
+    /// gas. See [`Self::gas`] for the reservoir model and why charged gas is not a sufficient
+    /// estimate.
```
mattsse:
```suggestion
    /// Sets the gas limit for the transaction. See [`Self::gas`].
```

crates/rpc-types-eth/src/transaction/request.rs
```diff
@@ -1465,7 +1474,15 @@ pub(super) mod serde_bincode_compat {
...
-        /// The gas limit for the transaction.
+        /// The combined gas limit for the transaction, including intrinsic costs.
+        ///
+        /// On networks with [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037), this funds both regular
+        /// (execution) gas and state gas. The protocol splits it into regular gas and a state
+        /// reservoir; callers still supply one limit. The regular gas cap applies
+        /// independently. Without EIP-8037, state creation uses the ordinary gas schedule
+        /// with no separate reservoir. Use `eth_estimateGas` with the target network's
+        /// rules. Receipt gas or net trace consumption may be too low because gas can be
+        /// required before a later refund or state refill.
```
mattsse: This is the bincode-compat mirror struct; nobody reads its docs, and duplicating the paragraph here means two places to keep in sync.

```suggestion
        /// The gas limit for the transaction.
```

crates/rpc-types-trace/src/geth/call.rs
```diff
@@ -11,26 +11,52 @@ use serde::{Deserialize, Serialize};
...
+    /// saving; use the transaction's receipt gas for the final charge. `None` means not
+    /// reported. This transaction summary is reported on the top-level frame when supported.
     #[serde(default, rename = "gasRefund", skip_serializing_if = "Option::is_none")]
     pub gas_refund: Option<U256>,
-    /// State gas refunded within the transaction.
+    /// Optional node-reported EIP-8037 state gas refill amount.
+    /// Distinct from the ordinary capped refund in `gas_refund`; refills undo state creation
+    /// charges during execution. The top-level `state_gas_used` is already net, so do not
+    /// subtract this again. This is a client extension: availability and frame scope depend on
+    /// the node. `None` is not zero. See [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).
```
mattsse: Two corrections: the proposal does define this field (as MAY, top-level frame), so "client extension" undersells it, and the refill/refund vocabulary mapping is never stated anywhere in the four PRs even though the wire name says `Refund` and the docs say "refill":

```suggestion
    /// Total EIP-8037 state gas refilled within the transaction, which the proposal calls
    /// "source-based state-gas refunds". Already netted out of `state_gas_used`; do not subtract it
    /// again. Optional in the proposal and only meaningful on the top-level frame; `None` means not
    /// reported. See [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037).
```

mattsse:
Added `regularGasUsed` decoding aliases to all three execution-gas fields while preserving `executionGasUsed` serialization, and corrected the proposal naming and `stateGasRefund` descriptions. The regression test covers both names, the `GethTrace` wrapper, and integer/quantity opcode summaries. Shortened the common header, receipt, and transaction docs while retaining the state-refill distinction in block accounting.

AI assistance: Codex authored this update and reply.

PR #4197, opened 2026-09-11
fix(consensus): reject 0x00-tagged legacy transactions and receipts

### LINKED ISSUE #4165: [Bug] 0x00-tagged legacy transaction is accepted by decode_2718 and re-encodes untagged

mattsse:
Per EIP-2718 legacy transactions and receipts are untagged, and `0x00` is not an assigned type. `TxType::try_from(0)` resolves to `Legacy`, so the derived `typed_decode` accepted an envelope with a literal `0x00` type byte and decoded the body as a legacy transaction. `EthereumTxEnvelope::Legacy` cannot record that a tag was present, so `encode_2718` emitted the untagged form and the input did not round-trip. geth, erigon, nethermind and besu all reject the type byte at decode. The same applied to `EthereumReceipt` and `AnyReceiptEnvelope`.

This is the tagged counterpart of #4090, which restricted `Signed<T>::fallback_decode` to legacy transactions. The guard now lives in `Signed<T>::typed_decode`, so every envelope built with `#[derive(TransactionEnvelope)]`, including flattened variants, rejects `0x00` with `Eip2718Error::UnexpectedType(0)` on the `decode_2718`, `network_decode` and `Decodable` paths. `EthereumTypedTransaction::decode_unsigned` drops its `0x00` arm for the same reason, since `encode_for_signing` never emits it.

For receipts, `EthereumReceipt::typed_decode_with_bloom` rejects `ty == 0`, and `rlp_decode_with_bloom` rejects a legacy type behind a string header. That path reads the type flag as an RLP integer, so it was reachable via `0x80`, while a raw `0x00` already failed as a non-canonical integer. `AnyReceiptEnvelope::typed_decode` gets the same guard and `fallback_decode` constructs the legacy variant directly. `ReceiptEnvelope` already rejected the tag and only gains a regression test. The `EthereumTxEnvelope` docs claiming the enum distinguishes tagged and untagged legacy transactions were inaccurate and now use the wording of `ReceiptEnvelope`.

Tests cover the untagged happy path and the `0x00`-tagged unhappy path on every decode entry point (`decode_2718`, `decode_2718_exact`, `network_decode`, `Decodable`) for `TxEnvelope`, `Signed<TxLegacy>`, a flattened custom envelope, `TypedTransaction::decode_unsigned`, `ReceiptWithBloom<EthereumReceipt>`, `ReceiptEnvelope` and `AnyReceiptEnvelope`, plus typed round-trips to guard against over-correction.

Supersedes #4167 and #4188.

Fixes #4165

### DIFF
--- a/crates/consensus-any/src/receipt/envelope.rs
+++ b/crates/consensus-any/src/receipt/envelope.rs
@@ -1,7 +1,7 @@
 use alloc::vec::Vec;
 use alloy_consensus::{Eip658Value, Receipt, ReceiptWithBloom, TxReceipt};
 use alloy_eips::{
-    eip2718::{Decodable2718, Eip2718Result, Encodable2718},
+    eip2718::{Decodable2718, Eip2718Error, Eip2718Result, Encodable2718},
     Typed2718,
 };
 use alloy_primitives::{bytes::BufMut, Bloom, Log};
@@ -10,8 +10,8 @@ use core::fmt;
 
 /// Receipt envelope, as defined in [EIP-2718].
 ///
-/// Represents legacy and typed EIP-2718 receipts. Type ID 0 is encoded as untagged legacy; this
-/// type does not preserve a literal `0x00` prefix.
+/// Represents untagged legacy receipts and typed EIP-2718 variants. Binary decoding rejects a
+/// literal `0x00` type prefix; type ID 0 uses the untagged legacy encoding.
 ///
 /// Transaction receipt payloads are specified in their respective EIPs.
 ///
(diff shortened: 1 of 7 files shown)


mattsse (commented):
Reviewed b8e4bdcab13fce2e8ac7f07624223d35b870527e for correctness; no issues found.

The `Signed<T>` guard rejects tagged legacy transactions through both direct and flattened envelopes while preserving untagged decoding. The receipt changes cover both the literal `0x00` prefix and the RLP-zero (`0x80`) path. Removing the unsigned decoder's `0x00` arm is consistent with its signing encoder and EIP-2718's untagged legacy representation.

Ran `cargo nextest run --offline -p alloy-consensus -p alloy-consensus-any --all-features` at this commit: 148 tests passed, 0 skipped, including the new regression tests and existing typed round-trips.

klkvr (approved):

PR #4203, opened 2026-09-16
chore(ci): route package installs through Aegis

decofe:
- Route CI package downloads through pinned Aegis secure-runner with job-scoped OIDC.
- Commit `Cargo.lock`, enforce `--locked`, and select older releases for publication-age checks while retaining approved exclusions.
- Pin tool installers; use Taiki with `zepter@1.88.1` and `cargo-binstall` fallback. Keep the Zepter lockfile-change guard.

Anvil/Geth/Reth downloads are unchanged. Prebuilt tools and cached artifacts are not covered by per-crate download checks.

Local configuration and locked-resolution checks passed; see CI for end-to-end validation.

Prepared with AI assistance.

Prompted by: @grandizzy


.github/workflows/ci.yml
```diff
@@ -41,6 +42,8 @@ jobs:
           - rust: "1.94.1" # MSRV
             flags: "--all-features"
     steps:
+      - name: Secure runner
+        uses: tempoxyz/gh-actions/actions/secure-runner@e080a269a3d37e571ad64e94f72536b48eb9921c # main
```
mablr: On fork PRs without OIDC, this action skips Aegis and succeeds with a warning. Could we document that dependency-changing fork PRs need an enforced maintainer-branch run before merge?

.github/workflows/ci.yml
```diff
@@ -272,22 +316,49 @@ jobs:
...
         with:
           persist-credentials: false
-      - uses: taiki-e/install-action@cca35edeb1d01366c2843b68fc3ca441446d73d3 # v2
+      - name: Install Zepter
+        uses: tempoxyz/gh-actions/vendor/taiki-e/install-action@25cce154e7fb10f99361a166468a6c56b9c31aa3 # main
         with:
-          tool: zepter
+          tool: zepter@1.88.1
+          checksum: true
+          fallback: cargo-binstall
```
mablr, line 331: `checksum: true` verifies the cargo-binstall bootstrap, not the Zepter binary: Zepter has no binstall signing metadata. Could we use a checksummed manifest/direct SHA-256 and `fallback: none`?

.github/workflows/ci.yml
```diff
@@ -60,23 +63,28 @@ jobs:
...
+        uses: tempoxyz/gh-actions/vendor/taiki-e/install-action@25cce154e7fb10f99361a166468a6c56b9c31aa3 # main
         with:
-          tool: nextest
+          tool: nextest@0.9.140
+          checksum: true
+          fallback: none
       - name: build
         if: ${{ matrix.rust == '1.94.1' }} # MSRV
-        run: cargo build --workspace ${{ matrix.flags }}
+        run: cargo build --locked --workspace ${{ matrix.flags }}
```
mablr, line 73: With every job locked, CI no longer covers fresh semver-compatible resolutions that downstream users receive. 

As we have now Cargo.lock file, we certainly need Dependabot.

.github/workflows/ci.yml
```diff
@@ -272,22 +316,49 @@ jobs:
...
         with:
-          tool: zepter
+          tool: zepter@1.88.1
+          checksum: true
+          fallback: cargo-binstall
       - name: run zepter
         run: |
           zepter --version
           time zepter run check
+          git diff --exit-code -- Cargo.lock
```
mablr, line 336: This invariant is in `feature-propagation`, but that job is missing from `ci-success.needs`. Could we add it so the aggregate required check cannot stay green when this fails?

.github/workflows/ci.yml
grandizzy: yeah, this will be enforced in a follow-up

.github/workflows/ci.yml
grandizzy: good catch - I think best approach is to add Zepter to the shared tempoxyz/gh-actions installer manifest and then switch this job to `fallback: none`, will do in a follow-up

grandizzy: we could also install locked zepter and have it run through Aegis but will increase CI time

.github/workflows/ci.yml
grandizzy: https://github.com/alloy-rs/alloy/pull/4203/changes/33b916c6d6896b29d733f2ed16b74a5a40c26d42

.github/workflows/ci.yml
grandizzy: https://github.com/alloy-rs/alloy/pull/4203/changes/33b916c6d6896b29d733f2ed16b74a5a40c26d42

mablr (commented):
SGTM, for the fixes

grandizzy (approved):

